PR #54, opened 2019-12-28
RFC: constructs for defining GitHub Actions

skinny85:
This RFC proposes creating a set of libraries that would allow defining GitHub Actions in all JSII-supported languages.

---------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


eladb:
@skinny85 what's up with this?

skinny85:
> @skinny85 what's up with this?

Swamped ATM with other stuff. But I will get back to it, and incorporate everybody's comments (thanks all!).

PR #91, opened 2020-01-26
RFC: 0095 cognito construct library

nija-at:
https://github.com/aws/aws-cdk-rfcs/issues/95

-----

Ready for feedback. 

[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/nija-at/cognito-l2constructs/text/0095-cognito-construct-library.md)

---


text/cognito-construct-library.md
```diff
@@ -0,0 +1,68 @@
+---
+feature name: cognito-construct-library
+start date: 27/01/2020
+rfc pr:
+related issue: n/a
+---
+
+# Summary
+
+This RFC covers the design of Cognito's CDK construct library coverage.
```
eladb:
```suggestion
This RFC covers the design of Cognito's CDK construct library.
```

text/cognito-construct-library.md
```diff
@@ -0,0 +1,68 @@
...
+by the Cognito module.
+
+# Design Summary
+
+This RFC is structured as a working backwards document. Since the focus of this
+RFC is to propose the API design, the best way to propose this is to write up
+the future user guide of the Cognito module, as if it was complete.
+
+The bulk of the RFC is in the supporting document -
+[working-backwards-readme.md](./cognito-construct-library/working-backwards-readme.md)
```
eladb: What's the benefit of managing this in a separate document?

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
+# Amazon Cognito Construct Library
+
+Amazon Cognito provides authentication, authorization, and user management for your web and mobile apps. Your users can
```
eladb:
```suggestion
[Amazon Cognito](link-to-aws-docs) provides authentication, authorization, and user management for your web and mobile apps. Your users can
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+sign in directly with a user name and password, or through a third party such as Facebook, Amazon, Google or Apple.
+
+The two main components of Amazon Cognito are user pools and identity pools. User pools are user directories that
```
eladb:
```suggestion
The two main components of Amazon Cognito are [user pools](link) and [identity pools](link). User pools are user directories that
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+provide sign-up and sign-in options for your app users. Identity pools enable you to grant your users access to other
+AWS services.
+
+This module is part of the [AWS Cloud Development Kit](https://github.com/aws/aws-cdk) project.
+
+## User Pools
+
+User pools allows creating and managing your own directory of users that can sign up and sign in. They enable easy
```
eladb:
```suggestion
User pools allow creating and managing your own directory of users that can sign up and sign in. They enable easy
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+created outside the CDK App.
+
+User pools can be imported using the `UserPool.fromUserPoolAttributes()` API. They can be imported either using the
+user pool ARN or the user pool id. For example:
+
+```ts
+const stack = new Stack(app, 'my-stack');
+
+UserPool.fromUserPoolAttributes(stack, 'my-imported-user-pool', {
+  userPoolId: 'us-east-1_123412341'
```
eladb: I would expect this to be `UserPool.fromUserPoolId` and `UserPool.fromUserPoolArn` and not through "...Attributes"

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+## User Pools
+
+User pools allows creating and managing your own directory of users that can sign up and sign in. They enable easy
+integration with social identity providers such as Facebook, Google, Amazon, Microsoft Active Directory, etc. through
+SAML.
+
+The CDK APIs provide ways to both create user pools, via the `UserPool` constructor, or import existing ones that were
+created outside the CDK App.
+
+User pools can be imported using the `UserPool.fromUserPoolAttributes()` API. They can be imported either using the
```
eladb: Always start with the simplest getting started (Not sure it makes sense to start from the import case.).

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+
+```ts
+const stack = new Stack(app, 'my-stack');
+
+UserPool.fromUserPoolAttributes(stack, 'my-imported-user-pool', {
+  userPoolId: 'us-east-1_123412341'
+});
+```
+
+> TODO: provider name and provider url return values?
```
eladb: ?

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+
+The following sections describe how user pools can be configured. **Unless explicitly stated, all of these
+configurations are available only when creating new user pools.**
+
+### Sign Up
+
+Users need to either signed up by the app's administrators or can sign themselves up. You can read more about both these
```
eladb:
```suggestion
Users need to either be signed up by the app's administrators or can sign themselves up. You can read more about both these
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+kinds of sign up and how they work
+[here](https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html).
+
+Further, a welcome email and/or SMS can be configured to be sent automatically once a user has signed up. This welcome
+email and/or SMS will carry the temporary password for the user. The user will use this password to log in and reset
+their password. The temporary password is valid only for a limited number of days.
+
+All of these options can be configured under the `signUp` property. The pool can be configured to let users sign
+themselves up by setting the `selfSignUp` property. A welcome email template can be configured by specifying the
+`welcomeEmail` property and a similar `welcomeSms` property for the welcome SMS. The validity of the temporary password
```
eladb: I don't know that we need this level of detail here. It mostly belongs to the reference. The README should give people references to entry points, preferably in the form of code samples. In this case I would just say something like:

You can setup sign up options through the `signUp` property. For example, the following code configures a user pool to send a sign up email with a password that expires after 5 days:

```ts
<code snippet here>
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+
+The user pool can be configured such that either the user's email address, phone number or both should be verifed at the
+time of sign in. Verification is necessary for account recovery, so that there is at least one mode of communication for
+a user to reset their password or MFA token when lost.
+
+When either one or both of these are configured to be verified, a confirmation message and/or email are sent at the
+time of user sign up that they then enter back into the system to verify these attributes and confirm user sign up.
+
+*Note*: If both email and phone number are specified, Cognito will only verify the phone number. To also verify the
+email address, see https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-email-phone-verification.html
```
eladb:
```suggestion
email address, see [Email and Phone Verification](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-email-phone-verification.html) in the Cognito Developer Guide
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+The `verifyContactMethods` attribute allows for this to be configured.
+
+*Defaults*:
+* signUp.selfSignUp: true
+* signUp.tempPasswordValidity: 7 days
+* signUp.welcomeEmail.subject: 'Thanks for signing up'
+* signUp.welcomeEmail.body - 'Hello {username}, Your temporary password is {####}'
+* signUp.welcomeSms.message - 'Your temporary password is {####}'
+* signUp.verifyContactMethods - Email
+
```
eladb: This is redundant, it is available through the reference documentation and will be easier to keep up to date. Feel free to link to the reference docs.

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+    selfSignUp: true,
+    tempPasswordValidity: Duration.days(3),
+    welcomeEmail: {
+      subject: 'Welcome to our awesome app!'
+      body: 'Hello {username}, Thanks for signing up to our awesome app! Your temporary password is {####}'
+    },
+    welcomeSms: {
+      message: 'Your temporary password for our awesome app is {####}'
+    },
+    verifyContactMethods: [ ContactMethods.EMAIL, ContactMethods.PHONE_NUMBER ],
```
eladb: Is there a correlation between the methods you select here and the "welcomeXxx" options? Perhaps if you specify "welcomeXxx" it will imply that you want Xxx to be used to verify the contact as sugar?

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+> Internal Note: Implemented via UserPool-AdminCreateUserConfig, temp password via UserPool-Policies,
+> verifyContactMethods via UserPool-AutoVerifiedAttributes.
+
+### Sign In
+
+These are the various ways a user of your app can sign in. There are 4 options available with the enum `SignInType`:
+
+* USERNAME: Allow signing in using the one time immutable user name that the user chose at the time of sign up.
```
eladb:
```suggestion
* `USERNAME`: Allow signing in using the one time immutable user name that the user chose at the time of sign up.
```

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+### Attributes
+
+These are the set of attributes you want to collect and store with each user in your user pool. Cognito provides a set
+of standard attributes that are available all user pools. Users are allowed to select any of these standard attributes
+to be required. Users will not be able to sign up without specifying the attributes that are marked as required. Besides
+these, additional attributes can be further defined, known as custom attributes.
+
+Custom attributes cannot be marked as required.
+
+[Go here](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html) for more info.
```
eladb: "Go here" is not an ideal link title. Something like "See [User Pool Attributes](link) for more information".

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+
+Additionally, two properties `mutable` and `adminOnly` properties can be set for each custom attribute. The former
+specifies that the property can be modified by the user while the latter specifies that it can only be modified by the
+app's administrator and not by the user (using their access token).
+
+*Defaults*:
+* No standard attributes are marked required.
+* For all custom attributes, mutable is true and adminOnly is false.
+
+Code sample:
```
eladb: Switch things up (like above): start with an annotated code sample and provide some details after it instead of lenghts of text before.

text/cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,300 @@
...
+
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  attributes: {
+    required: [ StandardAttrs.address, StandardAttrs.name ],
+    custom: [
+      stringAttr({ name: 'myappid', minLen: 5, maxLen: 15 }),
```
eladb: I am assuming "name" must be unique, which lends itself to a map instead of an array:

```ts
attributes: {
  myappid: ...
  callingcode: { required: true }
}

File: text/cognito-construct-library/working-backwards-readme.md [hunk 18]
```diff
@@ -0,0 +1,300 @@
...
+Custom attributes cannot be marked as required.
+
+[Go here](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html) for more info.
+
+Standard attributes are available via the `StandardAttrs` enum.
+
+*Note*: By default, the standard attributes 'email' and/or 'phone\_number' will automatically be marked required if
+they are one of the verified contact methods. See [Sign up](#sign-up) for details on verified contact methods.
+
+Custom attributes can be specified via the `stringAttr` and `numberAttr` methods, depending on whether the attribute
```
eladb: Are these free-floating functions (not supported by jsii)?

eladb:
Please follow the RFC README, create a tracking issues, reference it here in the title, etc.

text/cognito-construct-library/working-backwards-readme.md
nija-at: Thanks! Good call. I believe the section around attributes requires another iteration of design.

text/cognito-construct-library.md
nija-at: I can keep the RFC structure in this doc while keeping the readme in the format/language that would feature when it is actually shipped. Mixing them up could complicate how the sections are structured and the voice & tense of the language. This main document would still contain things around motivations, limitations, adoptions, breaking changes, roll out (if applicable), etc., while the working-backwards-readme doc can continue to focus purely on the user experience.

Further, I could break up the one working-backwards-readme I have now into 2 or several if they get too big. Separate docs for user pool and identity pool, for instance.

text/cognito-construct-library/working-backwards-readme.md
nija-at: Ignore all `>` comments for now. It's a note to myself to investigate into something, or an implementation note. I'll probably move these off from here and into the RFC's main document.

text/cognito-construct-library/working-backwards-readme.md
nija-at: I intend to remove these defaults when this document becomes an actual readme and move them as the `@default` annotation in tsdoc. The `*Defaults*` section is the way to highlight that before the actual code is written.

What's the standard for defaults that we let CloudFormation apply - do we document that as a `@default` in the tsdoc?

text/cognito-construct-library/working-backwards-readme.md
nija-at: Changed to reflect this.

text/cognito-construct-library/working-backwards-readme.md
nija-at: I've re-worked the defaults in a way that it's not part of the core doc anymore but is still present as a quoted text (`>`). These will go into the tsdoc as `@default` annotations during implementation.

text/cognito-construct-library/working-backwards-readme.md
nija-at: switched around.

text/cognito-construct-library/working-backwards-readme.md
nija-at: Done.

text/cognito-construct-library/working-backwards-readme.md
nija-at: I've been reading up on this page twice and I still can't say for sure if that is or that is not the case - https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html#allowing-users-to-sign-up-and-confirm-themselves

There are too many different ways to signup and verify, and I can't quite make out. However, I think your suggestion is reasonable. Let's start there and we can always break out a separate property if such is required in the future.

text/0095-cognito-construct-library.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Future Changes / Currently out of scope
+
+The following are currently out of scope for this document. Towards the end of the implementation, separate issues will
+be created for these, that will then be prioritized based on customer requests and +1s.
+
+* User Pool
+  * Import using user pool name
+  * Configure SES email actions.
+  * ClientMetadata and ValidationData properties in User
```
duarten: Do these have a representation at user pool creation time?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+});
+```
+
+### Sign Up
+
+Users need to either be signed up by the app's administrators or can sign themselves up. This can be configured via the
+`selfSignUp` property under the `signUp` group.
+
+A welcome email and/or SMS can be configured to be sent automatically once user has signed up. This welcome email and/or
+SMS will carry the temporary password for the user. The user will use this password to log in and reset their password.
```
duarten: This is only true for administrator sign up, right? Otherwise the user can specify a permanent password attribute.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+Consider the following code sample where a number of these properties are configured -
+
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  signUp: {
+    selfSignUp: true,
+    tempPasswordValidity: Duration.days(3),
+    welcomeEmail: {
```
duarten: I don't think this should be a `welcomeEmail`. It's either an invitation e-mail (for administrator sign up with a temporary password) or a verification message (for self-signup and when attributes require validation). These are the terms used in the cognito console and in cloudformation (InviteMessageTemplate and VerificationMessageTemplate).

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  attributes: {
+    required: [ StandardAttrs.address, StandardAttrs.name ],
+    custom: {
+      'myappid': new StringAttr({ minLen: 5, maxLen: 15 }),
+      'callingcode': new NumberAttr({ min: 1, max: 3 }),
+    },
```
duarten: Why not an array for custom attributes too?

`custom: [new StringAttr('myappid', {minLen: 5, maxLen: 15 })]`

text/0095-cognito-construct-library/working-backwards-readme.md
duarten: It should also be configureable whether to confirm with [code or link](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-verificationmessagetemplate.html#cfn-cognito-userpool-verificationmessagetemplate-defaultemailoption).

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+    postSignUp: postsignupLambda,
+    preAuthentication: preauthLambda,
+    // ...
+  }
+});
+```
+
+Triggers can also be added to the user pool outside of its constructor. This can be done via individual methods such as
+`addPreSgnUpTrigger()`, `addPostSignUpTrigger()`, `addPreAuthentication()`, etc.
+
```
duarten: Maybe explain that the CDK adds permissions to the Lambdas so they can invoked by Cognito?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+> Internal Note: Account Recovery settings are missing from UserPool CloudFormation resource.
+
+### Emails
+
+Cognito will handle sending emails to users in the user pool. The address from which emails are sent can be configured
+on the user pool via the `from` property, and the `replyTo` property to configure the email where replies are sent.
+
+User pools can also be configured to send emails through Amazon SES, however, that is not yet supported via the CDK. Use
+the cfn layer to configure this - https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html.
```
duarten: Why isn't this going to be supported? Seems like it could be a simple switch.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+const awesomePool = UserPool.fromUserPoolId(stack, 'awesome-user-pool', 'us-east-1_oiuR12Abd');
+
+const otherAwesomePool = UserPool.fromUserPoolArn(stack, 'other-awesome-user-pool',
+  'arn:aws:cognito-idp:eu-west-1:123456789012:userpool/us-east-1_mtRyYQ14D');
+```
+
+### Users
+
+Users are added to pools either as part of sign up or via the Cognito APIs directly. In addition to these, it is
+possible to add users to the user pool via the CDK.
```
duarten: Ah, now I understand the comment about the validation data :)

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+  // ...
+  // ...
+});
+
+userpool.createUser(this, 'myuserpool-supportuser', {
+  userName: 'support-user',
+  deliveryMediums: [ ContactMethods.EMAIL, ContactMethods.PHONE ],
+  attributes: [
+    { name: StandardAttrs.EMAIL, value: 'support@awesomeapp.com' },
+    { name: 'callingcode' value: '12' }
```
duarten: Missing comma after 'callingcode'.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+      body: 'Hello {username}, Thanks for signing up to our awesome app! Your temporary password is {####}'
+    },
+    welcomeSms: {
+      message: 'Your temporary password for our awesome app is {####}'
+    },
+  }
+});
+```
+
+At least one of `welcomeEmail` or `welcomeSms` is required if `selfSignUp` is set to `true`.
```
rix0rrr: May be worth making a class with factory functions for this?

```ts
new UserPool(this, 'myuserpool', {
  signUp: SignUp.byEmail()  // Has some nice defaults that can be customized
});
```

text/0095-cognito-construct-library/working-backwards-readme.md
rix0rrr: I also don't see from this example code how you make a custom attribute required or a standard attribute optional. The sample seems to conflate required == standard, custom == optional (although that might be a Cognito limitation that I'm unaware of).

Would expect to see this though:

```ts
new UserPool(this, 'myuserpool', {
  // ...
  requiredAttributes: [
    Attribute.NAME,
    Attribute.customString('myappid', { minLen: 5, maxLen: 15 })
  ],
  optionalAttributes: [
    Attribute.ADDRESS,
    Attribute.customNumber('callingcode', { min: 1, max: 3 })
  ]
});
```

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+    required: [ StandardAttrs.address, StandardAttrs.name ],
+    custom: {
+      'myappid': new StringAttr({ minLen: 5, maxLen: 15 }),
+      'callingcode': new NumberAttr({ min: 1, max: 3 }),
+    },
+  }
+});
+```
+
+*Note* that, custom attributes cannot be marked as required.
```
rix0rrr: Aha. Well I'd still expect to see the previous code with a runtime check or something.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+>   The role assumption will be conditioned on a strict equals on an ExternalId that will be unique to this user pool.
+
+Code sample:
+
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  security: {
+    mfa: {
```
rix0rrr: I think I'd like to encourage a factory function here again, and same for `passwordPolicy`. Or flatten out the properties into the top-level props object.

We have a construct library design guideline to flatten out the properties namespace.

text/0095-cognito-construct-library/working-backwards-readme.md
rix0rrr: Yes if CFN can do it there shouldn't really be a blocker.

Or is it because SES doesn't have appropriate L2s yet?

text/0095-cognito-construct-library/working-backwards-readme.md
rix0rrr: At this point I would have a used a map:

```ts

userpool.createUser(this, 'myuserpool-supportuser', {
  attributes: {
    [StandardAttrs.EMAIL]: 'support@awesomeapp.com',
    callingcode: 12
  }
});

File: text/0095-cognito-construct-library/working-backwards-readme.md [hunk 30]
```diff
@@ -0,0 +1,567 @@
...
+When the `resendInvitation` property is set to true, Cognito will resend the invitation message for a user that already
+exists but has not signed up.
+
+Users can be created for imported user pools by using the `UserPoolUser` construct, as so -
+
+```ts
+const userpool = UserPool.fromUserPoolAttributes(this, 'myuserpool', { ... });
+
+new UserPoolUser(this, userpool, {
+  userPoolId: userpool.userPoolId,
```
rix0rrr: Don't use ID, reference by object:

```ts
new UserPoolUser(this, userpool, {
  userPool: userpool,
});
```

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+### Identity Providers
+
+Identity pools can be configured with external providers for sign in. These can be from facebook, google, twitter as
+well as, "login with Amazon" and Cognito user pools. Use the `identityProviers` property to configure these providers.
+Learn more about external identity providers at [Cognito's
+documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/external-identity-providers.html). The CDK
+documentation of the `identityProviders` property has more details on how to configure an identity pool with each of
+these external providers.
+
+> Internal Note: IdentityPool-SupportedLoginProviders takes a JSON with all of the providers configured.
```
rix0rrr: We will provide factory functions for the common use cases, right?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+});
+```
+
+> These are the defaults that will be part of tsdoc, and not part of the README -
+> * allowUnauthenticated: false
+> * authenticationFlow: `AuthenticationFlow.ENHANCED`
+
+### Identity Providers
+
+Identity pools can be configured with external providers for sign in. These can be from facebook, google, twitter as
```
rix0rrr: Is that the difference between UserPool and IdentityPool? Direct users or federated users?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+
+```ts
+const authenRole = new iam.Role(this, 'cognito-authenticated-role', { ... });
+
+const unAuthenRole = new iam.Role(this, 'cognito-unauthenticated-role', { ... });
+
+new IdentityPool(this, 'my-awesome-id-pool', {
+  // ...
+  // ...
+  authenticatedRole: authenRole,
```
rix0rrr: Does it make sense to autocreate these?

```ts
const pool = new IdentityPool(this, 'my-awesome-id-pool');

bucket.grantRead(pool.unauthenticatedRole);
bucket.grantReadWrite(pool.authenticatedRole);
```

Especially since I believe the `assumedBy` parameter for the roles is going to be quite tricky in practice, no?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,567 @@
...
+facebook or via google.
+
+```ts
+const adminRole = new iam.Role(this, 'admin-role', { ... });
+
+new IdentityPool(this, 'my-awesome-id-pool', {
+  // ...
+  // ...
+  roleMappings: {
+    'facebook': new FacebookRoleMapping({
```
rix0rrr: Can't the key be implicit in the mapping class?

rix0rrr: Also, doesn't it make sense to push these mappings into the Facebook provider?

```ts
new IdentityPool(this, 'myidentitypool', {
  // ...
  identityProviders: [
    IdentityProvider.Facebook('7346241598935555' , {
      roleMappingRules: [
        {
          // if the identifier of the user as returned by facebook is 'boris'
          claim: FacebookRoleMappingClaim.SUB,
          matchType: MatchType.EQUALS,
          value: 'boris',
          role: adminRole
        }
      ],
    })
 ]
});
```

text/0095-cognito-construct-library.md
nija-at: Not at user pool creation AFAICT. 

They are inputs to pre-signup triggers but not stored anywhere in the user pool itself. - https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUser.html#CognitoUserPools-AdminCreateUser-request-ClientMetadata

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: You are correct. I'll update the documentation and the model to correctly reflect that.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: You are right. Changes to this coming in the next update.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: @rix0rrr -

>  The sample seems to conflate required == standard, custom == optional (although that might be a Cognito limitation that I'm unaware of).

That is exactly Cognito's modeling.

Quoting from the [documentation for standard attributes](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#cognito-user-pools-standard-attributes) -- "These attributes are available as optional attributes for all users. To make an attribute required, select the check box next to the attribute. ".

Quoting from the [documentation for custom attributes](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-custom-attributes) -- "cannot be required"

Given this information, is my modeling ok?

nija-at: @duarten - switched this from an array to map based on [another comment](https://github.com/aws/aws-cdk-rfcs/pull/91#discussion_r371007799). I believe the map-style is a bit more ergonomic and human friendly.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: > Or is it because SES doesn't have appropriate L2s yet?

Correct! 

> Seems like it could be a simple switch.

It's not a simple switch. A bunch of configurability [comes along with it](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-emailconfiguration.html#aws-properties-cognito-userpool-emailconfiguration-properties) and without proper class modeling in SES, this will look weird.

At the same time, given the extended nature of Cognito's features, something had to be cut for the first round of implementation and this made the cut. This is not saying we will not support it but just not yet. Users can always avail this configuration via escape hatches, albeit it won't be clean.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: Yes, good call out.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: Not quite. User pool lets you manage a pool of users, while identity pool lets users obtain temporary AWS credentials. Identity pools can use either a 3P identity provider (facebook, google, amazon,...) as well as the user pool as an identity provider.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: Good call. Changed.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: I don't follow this comment. Can you re-state?

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: Whoops, that was when my brain took a break.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: > Does it make sense to autocreate these?

Yes, there should be a default role created for these with the `authenticatedRole` property to specify an existing role. I seem to have missed this and will add that in.

> Especially since I believe the assumedBy parameter for the roles is going to be quite tricky in practice, no?

The doc states -- "The CDK, by default, adds a trust policy with these configured to all roles used by the IdentityPool. To disable this, set the configureRoleTrust to false."

Essentially I'm hoping to call `role.assumeRolePolicy.addStatements(...)` and add the correct trust parameters.

Does this satisfy?

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: I've flattened this mapping out by one level. What do you think about this now? Do you think it still need factory functions?

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: The subsequent paragraphs and code snippet is how I propose to model this JSON. This should be sufficient, no?

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: Nice! I like this.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+
+The CDK will add necessary permission to the lambda functions for it to be invoked by the cognito service.
+
+### Importing User Pools
+
+Any user pool that has been created outside of this stack, can be imported into the CDK app. Importing a user pool
+allows for it to be used in other parts of the CDK app that reference an `IUserPool`. However, imported user pools have
+limited configurability. As a rule of thumb, none of the properties that is are part of the
+[`AWS::Cognito::UserPool`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html)
+CloudFormation resource can be configured.
```
shivlaks: really like that you called out this limitation explicitly

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+
+The `role` property refers to the IAM role whose permissions that the users in this group are able to assume.
+
+When several different groups are defined for the same user pool, they can be ordered to have a precendence. Precedence
+will determine which IAM role's permission will be available for a user when they are part of two different groups. See
+[CreateGroup
+API](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateGroup.html#CognitoUserPools-CreateGroup-request-Precedence)
+for more details.
+
+Use the `takesPrecendenceOver()` API to set the precedence of one group over the other.
```
shivlaks: would it be possible to set the precedence on the user pool itself?
this seems like a really easy way to create a circular graph. Or rather will we detect them to fail faster?

```ts
 groupA.takesPrecedenceOver(groupB);
 ...
 groupB.takesPrecedenceOver(groupA);
```

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+> * precedence - `undefined`
+
+## Identity Pools (Federated Identity)
+
+Identity pools enables creation of unique identities for their users and federate them with identity providers. These
+identity providers can be a Cognito user pool, Facebook, Google, or any other SAML-based identity provider.
+With an identity pool, you can obtain temporary, limited-privilege AWS credentials to access other AWS services.
+
+Using the CDK, a new identity pool can be created as part of the stack using the construct's constructor. You must
+specify the `identityPoolName`. This is the quickest way to create an identity pool using all of the defaults -
```
shivlaks: same thought as earlier comment: are we trying to be deliberate about avoiding CFN generated names and therefore made this a required property?

text/0095-cognito-construct-library.md
```diff
@@ -0,0 +1,77 @@
...
+The Cognito module's API stability is 'experimental'. This is a signal to the users of this module that breaking
+changes to the API are to be expected.
+
+# Unresolved questions
+
+None.
+
+# Future Changes / Currently out of scope
+
+The following are currently out of scope for this document. Towards the end of the implementation, separate issues will
```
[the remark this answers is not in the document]
nija-at: You're correct. I'll drop those words

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+This module is part of the [AWS Cloud Development Kit](https://github.com/aws/aws-cdk) project.
+
+## User Pools
+
+User pools allow creating and managing your own directory of users that can sign up and sign in. They enable easy
+integration with social identity providers such as Facebook, Google, Amazon, Microsoft Active Directory, etc. through
+SAML.
+
+Using the CDK, a new user pool can be created as part of the stack using the construct's constructor. You must specify
+the `userPoolName`, as so -
```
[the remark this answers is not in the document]
nija-at: You are correct. It should be optional as well. I'll correct it.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  selfSignUp: {
+    verificationEmail: {
+      subject: 'Verify your email for our awesome app!',
+      body: 'Hello {username}, Thanks for signing up to our aweomse app! Your verification code is {####}',
+      verificationStyle: EmailVerificationStyle.CODE
```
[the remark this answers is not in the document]
nija-at: I went back and forth on this. I was struggling with - is `style`/`type` sufficiently explanatory and  could there be other styles or types added later that conflict with the naming here? 

I think I'll switch this around per your suggestion.

text/0095-cognito-construct-library/working-backwards-readme.md
nija-at: I think we should detect it and fail faster. I'd like to play around with this after implementation to see how the final outcomes looks and adjust accordingly. 

This is just at a proposal level here, as in, let me know if this sounds like a good, bad or if there's a better idea. Smaller adjustments can be reviewed in the actual implementation PR.

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+  }
+});
+```
+
+*Note* that, custom attributes cannot be marked as required.
+
+*Note*: By default, the standard attributes 'email' and/or 'phone\_number' will automatically be marked required if
+they are one of the verified contact methods. See [Sign up](#sign-up) for details on verified contact methods.
+
+Additionally, two properties `mutable` and `adminOnly` properties can be set for each custom attribute. The former
```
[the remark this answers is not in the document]
nija-at: I can't find much information on this in the documentation, so I've left this here. I'll adjust the final outcome based on what I find at implementation when I play around with it. Sound good?

text/0095-cognito-construct-library/working-backwards-readme.md
```diff
@@ -0,0 +1,600 @@
...
+> These are the defaults that will be part of tsdoc, and not part of the README -
+> SignInType: `USERNAME`.
+
+Code sample:
+
+```ts
+new UserPool(this, 'myuserpool', {
+  // ...
+  // ...
+  signInType: [ SignInType.USERNAME, SignInType.EMAIL ],
```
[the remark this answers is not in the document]
nija-at: can you elaborate?

text/0095-cognito-construct-library/working-backwards-readme.md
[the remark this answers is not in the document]
shivlaks: works for me!

text/0095-cognito-construct-library/working-backwards-readme.md
shivlaks: agreed, just capturing thoughts that come to mind.  failing faster sounds like the right customer experience to me

text/0095-cognito-construct-library/working-backwards-readme.md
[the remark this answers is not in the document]
shivlaks: so the constructor is one way to initialize with supported sign-in types. would `UserPool` have methods to the `UserPool` class -> `addSignInType` `removeSignInType`

I'm not sure how it's currently handled, does it replace the UserPool altogether or is this an attribute that can be updated?

PR #93, opened 2020-01-27
feat(continuous-delivery): assets in environment-agnostic stacks

eladb:
Originally, we planned to disallow the use of assets by environment-agnostic stacks (e.g. stacks that are defined with pseudo ACCOUNT/REGION in `env`). The problem is that today this is the default behavior for `new Stack()`, which obviously can't work. The original plan was to modify the default behavior of `env` (see https://github.com/aws/aws-cdk/pull/4131), and we still want to pursue this, but since this change in default behavior has potential risk involved, we wanted to decouple it from our CI/CD delivery.

Therefore, this change updates the CI/CD and cdk-assets RFCs to support assets in environment-agnostic stacks by allowing the use `${AWS::Region}` and `${AWS::AccountId}` in `assets.json`.

---

### DIFF
--- a/text/0049-continuous-delivery.md
+++ b/text/0049-continuous-delivery.md
@@ -211,7 +211,9 @@ However, we also need to be able to express this environment-agnostic location i
 cdk-bootstrap-hnb659fds-container-assets-${AWS::ACCOUNT}-${AWS::REGION}:7a4f25ddcb08358916501c7536cdb11f33daf856896aa2a2bec3c7a5d735ace4
 ```
 
-When `cdk-assets` needs to publish an asset it will substitute `${AWS::ACCOUNT}` and `${AWS::REGION}` with the current account and region.
+Then `cdk-assets` will simply substitute `${AWS::ACCOUNT}` and `${AWS::REGION}` with the account and region derived from the credentials configured for the CLI.
+
+> NOTE: even when an IAM role from another account is assumed for publishing, `${AWS::ACCOUNT}` and `${AWS::REGION}` always resolve to the CLI configuration and not to the other account.
 
 ## Bootstrapping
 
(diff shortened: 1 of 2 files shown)


text/0049-continuous-delivery.md
```diff
@@ -186,17 +185,33 @@ Users should be able to vend custom asset providers to allow customizing how ass
...
+When a template that contains this expression will be deployed, CloudFormation will resolve this to the correct location within the current account and region.
+
+However, we also need to be able to express this environment-agnostic location in `assets.json`. This can be done either by adding explicit support for `Fn::Join` and the region/account pseudo references in their CloudFormation syntax or by rendering a substitution syntax such as:
+
+```
+cdk-bootstrap-hnb659fds-container-assets-${AWS::ACCOUNT}-${AWS::REGION}:7a4f25ddcb08358916501c7536cdb11f33daf856896aa2a2bec3c7a5d735ace4
+```
 
-Since the specific account and region are required when resolving asset consumption and publishing locations, the current plan is for the default asset store implementation to **fail if assets are used from environment-agnostic stacks**. Again, bear in mind that the current behavior (where the default is environment-agnostic stacks) is going to be changed.
+When `cdk-assets` needs to publish an asset it will substitute `${AWS::ACCOUNT}` and `${AWS::REGION}` with the current account and region.
```
rix0rrr: This needs a little more definition to disambiguate, looking at the CDK assets manifest:

```
        {
          "bucketName": "aws-cdk-files-3333333333EU-us-west-2",
          "objectKey": "a0bae29e7b47044a66819606c65d26a92b1e844f4b3124a5539efc0167a09e57.zip",
          "assumeRoleArn": "arn:aws:iam::3333333333EU:role/aws-cdk-publish-3333333333EU-eu-west-2"
        }
```

Specifically, `assumeRoleArn` can change the value of `${AWS::ACCOUNT}`. So what does the following mean:

```
        {
          "bucketName": "aws-cdk-files-${AWS::ACCOUNT}-us-west-2",
          "assumeRoleArn": "arn:aws:iam::3333333333EU:role/aws-cdk-publish-${AWS::ACCOUNT}-eu-west-2"
        }
```

The `${AWS::ACCOUNT}` in `assumeRoleArn` MUST be the "current" account, because anything else wouldn't make sense (we can't know the account we're assuming into in advance... UNLESS we parse the ARN...?)

However, the  `${AWS::ACCOUNT}`  in the `bucketName` could realistically refer to both the "current" account as well as the ASSUMED account. In fact, the assumed account makes most sense, but is hardest to get by.

Now, this is mostly a theoretical complication. In practical cases, it should be enough to define  `${AWS::ACCOUNT}` to always mean the "current" account, and if you do crazy things in `assumeRoleArn` that's your problem (the CDK framework will never output cross-account ARNs in this way).

But we should nail this down unambiguously.

eladb: OK. Good point, updated.

Issue #79, opened 2019-07-23
CDK v2.0


eladb:
|PR|Champion|
|--|--------|
|#156 |  @eladb       |

## Description

Proposal detailing how we release aws-cdk 2.0 and how we want to handle major versions in the future.

### Things to consider

- [ ] Release process
- [ ] Long-term support for previous major versions
- [ ] Regular clean ups across major versions such as:
  - [ ] Remove all deprecated APIs
  - [ ] Reset all [feature flags](https://github.com/aws/aws-cdk/blob/master/design/feature-flags.md)

#### Required

- [Monocdk](https://github.com/aws/aws-cdk-rfcs/issues/6)
- Reset Feature Flags
- Remove All Deprecations

#### Recommended

- [Easier Identification of Experimental Modules and APIs](https://github.com/aws/aws-cdk-rfcs/issues/116)
- [Public S3 Artifacts](https://github.com/aws/aws-cdk-rfcs/issues/39)
- [Rename ID Parameter](https://github.com/aws/aws-cdk/issues/3203)
- Remove all usage of Type unions

#### Not Recommended

- [Unmangled L2 Logical IDs](https://github.com/aws/aws-cdk/issues/1687)
- [Removal of All Union Types](https://github.com/aws/aws-cdk/issues/5067)
- [Remove Custom Resource Implementation of Fargate Event Target](https://github.com/aws/aws-cdk/issues/3930)
- [Change Logical IDs to Avoid Potential Collisions](https://github.com/aws/aws-cdk/issues/6421)
- [Replace `addToPolicy` with `addTo{Principal,Identity}Policy`](https://github.com/aws/aws-cdk/pull/7598)

#### More Info Needed

- [AliasOptions and VersionOptions should not extend EventInvokeConfigOptions](https://github.com/aws/aws-cdk/issues/6966)
- [Remove Support for Docker Assets with Parameters]()

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

eladb:
Closing this in favor of #118

flemjame-at-amazon:
How can we make suggestions for things to consider for CDK v2.0? Open an issue?

MrArnoldPalmer:
> How can we make suggestions for things to consider for CDK v2.0? Open an issue?

[...]

Issue #36, opened 2019-12-10
Constructs Programming Model


eladb:
|PR|Champion|
|--|--------|
|# |        |

## Description

The AWS CDK demonstrates that leveraging general purpose programming languages
to define complex configuration modules brings value to users. At the heart of
the AWS CDK lies the "constructs programming model" which consists of a simple
pattern that enables composition and synthesis.

Recently, there were a few initiatives to build CDKs for other domains with
similar characteristics. Examples are Kubernetes, Terraform and GitHub Actions.

We wish to enable the creation of these "sister" CDK projects by extracting the
constructs programming model into an independent library, which can be leveraged
for any domain without taking a dependency on the AWS CDK.

In order to enable composition of constructs across domains (e.g. Kubernetes app
defined within an AWS CDK app), we will need to extract the `Construct` base
class from the @aws-cdk/core module and into a separate module and AWS CDK users
will need to take a dependency on this new module.

> Note that this means that all CDK apps and libraries will also need to take a peer dependency on this new module.

Spaces:
- [ ] Terraform
- [ ] Kubernetes
- [ ] F5

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

pgollucci:
I am interested in this. Dare I say it other Clouds?  But no really, Kubernetes

eladb:
We have released this as an independent library called [constructs] and also have an alpha version of https://cdk8s.io

[constructs]: https://github.com/aws/constructs

pgollucci:
This seems done esp with 10.x.  Can be closed?

eladb:
Yes, done.

jjkavalam:
@eladb Where could i find an introduction to the Constructs Programming Model ? I can see that the idea is really effective; but i would like to get some guidance on using it. I am interested in applying it on a domain different from aws/k8s.

The readme at https://github.com/aws/constructs is unfortunately not quite helpful.

Issue #6, opened 2019-12-08
Monolithic Packaging


eladb:
|PR|Champion|
|--|--------|
|#122 | @rix0rrr          |

## Description

The CDK consists of 100+ packages, one for every AWS service and then some, with
complex interdependencies. For example, the *aws-ecs* package depends on *core*,
*aws-iam*, *aws-ecr*, *aws-ec2* and more packages to do its work. In fact, it
depends on 23 other packages.

This means that when a user wishes to use the *aws-ecs* module, their package
manager needs to fetch all 23 dependencies.

This proposal suggests to bundle and release the AWS CDK as a single monolithic
module. This means that the way users consume the CDK will change in a breaking
way.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [x] Resolved

eladb:
Based on a whitepaper by [@rix0rrr](https://github.com/rix0rrr).

The ultimate root cause of the issues we are encountering is a specific feature of the NPM package manager, one that both encourages a package ecosystem that “just works” with minimal maintenance and versioning problems, as well one that significantly simplifies the package manager:

_Packages can appear multiple times in the dependency graph, at different versions._

It simplifies the package manager because the package manager does not have to resolve version conflicts to arrive at a single version that will work across the entire dependency tree, and report a usable error if fails to do so. In the presence of version ranges, this is an open research problem, which NPM conveniently sidesteps.

### The Feature

As an example, a package *tablify* could depend on a package called *leftpad* at version 2.0, while the application using *tablify* could be using *leftpad* at an older and incompatible version 1.0. The following dependency graph is a valid NPM dependency graph:

```
app (1.0)
+-- tablify (1.0)
|   +-- leftpad (2.0)
+-- leftpad (1.0)
```

Loading dependencies is done in NPM by calling  require("leftpad") or import { ... } from "leftpad". If this statement is executed in a source file of *app*, it will load *leftpad@1.0*, and if it is executed in a source file of *tablify* it will load *leftpad@2.0*.

[...]

PR #105, opened 2020-01-28
RFC: #79 CDK v2.0

eladb:
Introduce the motivation and explore the implications of releasing a new major version for the AWS CDK (v2.0).

---

### DIFF
--- a/text/0079-cdk-v2.md
+++ b/text/0079-cdk-v2.md
@@ -1,0 +1,339 @@
+---
+feature name: cdk-v2
+start date: 2020-28-01
+rfc pr: (leave this empty)
+related issue: #79
+---
+
+# Summary
+
+[Version 1.0 of the AWS CDK](https://github.com/aws/aws-cdk/releases/tag/v1.0.0)
+have been released about 6 months ago (July 2019) after about a year in
+developer preview. Since then, we have released over 25 minor versions (and
+about 10 patches) which included hundreds of bugs fixes and features.
+
+Throughout this time, we were able to deliver changes to the CDK without
+introducing (intentional) breaking changes to APIs marked as STABLE (when an
+unintentional break was introduced, we quickly released a patch to revert it).
+In a few cases, we introduced breaking changes to EXPERIMENTAL modules.
+
+This RFC proposes content and explores the implications of releasing a new major
+version for the AWS CDK (v2.0).
+
+# Motivation
+
+In the past few weeks, a few issues have been accumulated that warrant the
+consideration of a major version release:
+
+1. Challenges when managing CDK dependencies
+2. Confusion around the usage of experimental features
+3. Desire to use constructs for other domains
+
+The solutions we are considering to address these issues will have an impact
+on how users consume the AWS CDK, and therefore will require a major version
+bump.
+
+The following sections describe each one of the motivations above.
+
+## Dependency Management
+
+The CDK consists of 100+ packages, one for every AWS service and then some, with
+complex interdependencies. For example, the *aws-ecs* package depends on *core*,
+*aws-iam*, *aws-ecr*, *aws-ec2* and more packages to do its work. In fact, it
+depends on 23 other packages.
+
+This means that when a user wishes to use the *aws-ecs* module, their package
+manager needs to fetch all 23 dependencies.
+
+### Unacceptable experience with peer npm dependencies
+
+Most languages (Java, Python, .NET) only support a single instance of a module
+at runtime. This means that if two modules declare that they take a dependency
+on a third module, the dependency manager will resolve a common version and
+fetch it.
+
+Contrary to most languages, the Node.js ecosystem supports the co-existence of
+multiple versions of a module at runtime (since `node_modules` first searched as
+a relative path). This is problematic for the CDK case since CDK modules often
+expose types from other modules in their API (e.g. an `s3.Bucket` accepts a
+`kms.Key` as an option).
+
+To model this type of dependency, npm allows declaring a dependency as a ["peer
+dependency"](https://nodejs.org/es/blog/npm/peer-dependencies). This means
+consumers are expected to **directly** install all transitive dependencies. In
+the *aws-ecs* case, this means that instead of declaring a dependency only on
+*aws-ecs*, users would have to declare a dependency on all 23 transitive modules
+that *aws-ecs* depend on.
+
+This is an unacceptable user experience with our current tools.
+
+See the [appendix](#appendix-cdk-and-npm-dependencies) written by
+[@rix0rrr](https://github.com/rix0rrr) which describes this problem in detail.
+
+### Poor ergonomics due to too many modules
+
+In addition to the peer npm dependency issue described above, due to the fact
+that CDK is "hyper modular", in all languages, we have poor ergonomics when it
+comes to declaring dependencies. Since users are required to explicitly install
+a module for each service they use, even simple projects end up with dozens of
+direct CDK dependencies.
+
+## Experimental Modules
+
+In some cases, a CDK module can include both stable and experimental constructs.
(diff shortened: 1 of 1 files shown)


text/0079-cdk-v2.md
```diff
@@ -0,0 +1,339 @@
...
+
+Throughout this time, we were able to deliver changes to the CDK without
+introducing (intentional) breaking changes to APIs marked as STABLE (when an
+unintentional break was introduced, we quickly released a patch to revert it).
+In a few cases, we introduced breaking changes to EXPERIMENTAL modules.
+
+This RFC proposes content and explores the implications of releasing a new major
+version for the AWS CDK (v2.0).
+
+# Motivation
```
nija-at: We should include a section in motivation with all the cleanup activities we take advantage of with a new major version. You have a good list here in the description - https://github.com/aws/aws-cdk-rfcs/issues/79.

And I'll add one to the list 'reserved word warnings' - https://github.com/aws/jsii/commit/5389def583160d6e06bb399704ff4cce9c8ef13a

text/0079-cdk-v2.md
```diff
@@ -0,0 +1,339 @@
...
+
+The CDK consists of 100+ packages, one for every AWS service and then some, with
+complex interdependencies. For example, the *aws-ecs* package depends on *core*,
+*aws-iam*, *aws-ecr*, *aws-ec2* and more packages to do its work. In fact, it
+depends on 23 other packages.
+
+This means that when a user wishes to use the *aws-ecs* module, their package
+manager needs to fetch all 23 dependencies.
+
+### Unacceptable experience with peer npm dependencies
```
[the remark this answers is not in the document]
rix0rrr: Changing to a monomodule will require people to change their `import` statements (at least in TS/JS). If it requires source code change it's a breaking change that requires an MV bump.

text/0079-cdk-v2.md
```diff
@@ -0,0 +1,339 @@
...
+Our API stability guarantees and adherence to semantic versioning only applies
+to STABLE APIs and not to EXPERIMENTAL APIs.
+
+This situation is confusing to users. It is not uncommon for users to use
+experimental constructs without being aware and then be surprised by breaking
+changes in minor versions.
+
+The current approach makes it hard for us to specify the stability at the module
+level since the L1s are stable by definition, but all high-level constructs are
+still experimental.
```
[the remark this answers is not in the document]
rix0rrr: This section is just describing a problem with a solution in mind where the solution will be breaking.

In particular, I think the solution is to move classes that have experimental APIs to a separate package, which means that if you want them you need to update your source to point to the new package, and hence a breaking change.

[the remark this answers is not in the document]
rix0rrr: For the record, I think we're going to bring pain onto ourselves if we want to split off experimental APIs into a separate package. Here are some thought exercises:

* What if we want to add an experimental method to an otherwise stable class? Do we copy the complete class?
* What if we want to add an experimental property to an otherwise stable props struct? Do we copy the complete class?
* What if we want to provide a stable API that depends on an experimental interface? (Remember, we can easily do this and there is no reason we shouldn't: the breakage caused by any changes to the experimental API will be detected by our complete in-tree build and corrected by the change maker, leaving to trace of it to downstream consumers) We won't be able to do that as we'll get cyclic dependencies.
* What if we stabilize an experimental class? Do we move it from the experimental package to the stable package, thereby breaking everyone in the process?

The better solution is to vend 2 different flavors of CDK from a single codebase:

* `stable` APIs only
* `stable+experimental` APIs

This will solve all issues mentioned above:

* Add experimental features to stable classes: just will not appear in the `stable` flavor, WILL appear in the `s+e` flavor.
* Depend on experimental APIs from stable: will just work, experimental API is just considered *private* in the `s` flavor.
* Stabilizing a class: `s+e` flavor will continue to work as usual, class now appears in `s` flavor.

text/0079-cdk-v2.md
```diff
@@ -0,0 +1,339 @@
...
+
+To improve **transparency about experimental modules**, we propose to move all
+experimental classes into a separate [sub]module with the `-experimental`
+postfix (e.g. `@aws-cdk/aws-eks-experimental`). This means that the non
+"-experimental" modules will always have stable APIs (including L1s) and will
+give us a way to graduate classes from the experimental module to the stable
+module. Users who use a module with the name "xxx-experimental" will have more
+awareness that this API may break in minor versions.
+
+To **enable the reuse of the constructs programming model** we will extract and
```
rix0rrr: I'm in favor of this. Do be aware that it will be added to the list of `peerDependencies` that has to be resolved by users. That is probably still fine, but keep this in mind.

text/0079-cdk-v2.md
```diff
@@ -0,0 +1,339 @@
...
+We wish to enable the creation of these "sister" CDK projects by extracting the
+constructs programming model into an independent library, which can be leveraged
+for any domain without taking a dependency on the AWS CDK.
+
+In order to enable composition of constructs across domains (e.g. Kubernetes app
+defined within an AWS CDK app), we will need to extract the `Construct` base
+class from the @aws-cdk/core module and into a separate module and AWS CDK users
+will need to take a dependency on this new module.
+
+# Design Summary
```
rix0rrr: I see another problem that I think a package split would address:

*Many people are mixing up L1 and L2 APIs* 

I see a lot of people using L2 classes, and then being disappointed when they can't plug them into L1 classes. This happens a lot in Python (where there's obviously no typechecker to tell you you're doing it wrong), but also happens in other languages where people call integration APIs between L2 classes to plug the result into an L1 (see recent changes to CloudWatch dashboards, mutliple people internally and externally were calling `toJson()` on `GraphWidgets` and plugging the result into a `CfnDashboard`, blissfully unaware that a special method needs to be called there to do a Token-aware JSON stringification).

Apparently our current communication methods of `Foo` and `CfnFoo` classes, plus notes that these are separate classes plus notes in every talk that we do aren't enough to dissuade people from doing this.

I would therefore like to propose we drop the L1-and-L2-in-a-single-package style of bundling, and publish 2 different packages, one with all L1s and one with all L2s.

Note that this needn't imply a `peerDependency` of L2-to-L1 library, for most reasonable use cases(*) a regular `dependency` of L2-to-L1 is sufficient (since by design we don't expose L1 types in the L2 API, and in fact jsii won't allow us to do that).

_(*) The existence of `construct.node.defaultChild` breaks this assumption a little, it's important that the consumer app and construct library agree on the version of the `CfnFoo` class that is accidentally exposed this way. Still, I expect this to be a niche use case that will mostly "just work"_

text/0079-cdk-v2.md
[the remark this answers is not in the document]
eladb: @nija-at added a sentence to clarify why this is a breaking change
@rix0rrr I will take your points into address when I expand the experimental modules section

text/0079-cdk-v2.md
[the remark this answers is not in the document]
eladb: Added a clarification

text/0079-cdk-v2.md
eladb: Added a link to #79, although I am not sure about all these changes. Might be too much.

text/0079-cdk-v2.md
eladb: Yes, totally aware. Will make sure it is called out.

text/0079-cdk-v2.md
eladb: Noted. I will incorporate this into the doc

eladb:
I have split this into three RFCs:

- [Dependencies issues caused by hyper-modularity](https://github.com/aws/aws-cdk-rfcs/issues/6)
- [Constructs programming model](https://github.com/aws/aws-cdk-rfcs/issues/36)
- [Experimentation modules](https://github.com/aws/aws-cdk-rfcs/issues/116)

PR #120, opened 2020-02-07
feat: run markdownlint for PRs

nija-at:
Use Github actions to run markdownlint for PRs. A reasonable set of
defaults are chosen and configured under `.markdownlint.json`.

Keeping the `.markdownlint.json` file in the root of the folder so that
it is picked up by VS Code markdownlint plugin by default.

Note, all changes to the markdown files are purely formatting to
conform with the markdown linter rules.

Motivations:
The main motivations is so that all RFCs meet a reasonable bar in terms
of consistency. The main violator is line width - unusually long lines
are hard to review and comment on in PRs. They also don't play very well
with text editors.

The added benefit is that some parts of markdown are flimsy such that
they need to be formatted in a particular way to be rendered correctly.
This will help users keep to that.

---

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -16,6 +16,6 @@ jobs:
           cd ../
       - name: run linter
         run: |
-          ./markdownlint-cli/node_modules/.bin/markdownlint -i text/0049-continuous-delivery.md text/**/*.md
+          ./markdownlint-cli/node_modules/.bin/markdownlint -i text/0049-continuous-delivery.md -i text/0092-asset-publishing.md text/**/*.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -14,6 +14,7 @@ jobs:
           npm install --prefix tools/linters
       - name: run linter
         run: |
-          ./tools/linters/node_modules/.bin/markdownlint -c ./tools/linters/markdownlint.json ./*.md text/**/*.md
+          cd ./tools/linters
+          npx --no-install markdownlint -c markdownlint.json ../../*.md ../../text/**/*.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -16,6 +16,6 @@ jobs:
           cd ../
       - name: run linter
         run: |
-          ./markdownlint-cli/node_modules/.bin/markdownlint -i text/0049-continuous-delivery.md -i text/0092-asset-publishing.md text/**/*.md
+          ./markdownlint-cli/node_modules/.bin/markdownlint -i text/0049-continuous-delivery.md text/**/*.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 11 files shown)


MrArnoldPalmer:
I'd assume we will continuously be adding various scripts to automate pieces of the workflow going forward.

@eladb @nija-at do you guys think we should just add package.json at the root to start collecting dependencies/scripts for use in actions and locally?

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,21 @@
...
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: |
+          mkdir markdownlint-cli && cd markdownlint-cli
+          npm install markdownlint-cli
+          cd ../
+      - name: run linter
+        run: |
+          ./markdownlint-cli/node_modules/.bin/markdownlint -i text/0049-continuous-delivery.md -i text/0092-asset-publishing.md text/**/*.md
```
eladb: Why do we have explicit docs listed here ?

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,21 @@
...
+  markdownlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: |
+          mkdir markdownlint-cli && cd markdownlint-cli
```
eladb: instead of putting all the logic here, I rather we created another tool under tools with a package.json and just “npm install” it.

nija-at: Nice! I like this pattern. It's slightly better than @MrArnoldPalmer's suggestion of keeping a package.json at the package root, and yet satisfies his comment. I'll change this up.

.github/workflows/pr-lint.yml
nija-at: The number of changes to conform these files were high. I've taken the time to fix these up now.

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: |
+          npm install --prefix tools/linters
+      - name: run linter
+        run: |
+          ./tools/linters/node_modules/.bin/markdownlint -c ./tools/linters/markdownlint.json ./*.md text/**/*.md
```
eladb: use `npx markdownlint` instead of explicitly reaching out to `node_modules/.bin`

README.md
```diff
@@ -72,7 +72,7 @@ existing proposals.
 [1](https://github.com/aws/aws-cdk-rfcs/issues/1)|CDK Watch||[@eladb](https://github.com/eladb)||proposed
 <!--END_TABLE-->
 
-## What does all this mean?!
+## What does all this mean?
```
eladb: why was this changed? I think the intention of the author was to use "?!"

nija-at: It was removed to follow the guideline of not having punctuations at the end of headers, except for question mark.

>  I think the intention of the author was to use "?!"

And what is this intention? I believe the intent for this question is to be rhetorical, which is conveyed with just the question mark, no? Very similar to the next header - "What is an RFC?" - https://github.com/aws/aws-cdk-rfcs/blob/master/README.md#what-does-all-this-mean

.github/workflows/pr-lint.yml
nija-at: Done!

PR #122, opened 2020-02-13
RFC: 0006 monocdk

eladb:
Proposal to distribute the AWS CDK as a single module instead of 150+
modules in order to allow third-party CDK modules to declare their dependency on
the AWS CDK as a peer dependency ([rendered version](https://github.com/aws/aws-cdk-rfcs/blob/benisrae/monolithic-packaging/text/0006-monolothic-packaging.md)).

Related to #6 

---


text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+## Prototype
+
+See [monocdk-experiment](https://www.npmjs.com/package/monocdk-experiment).
+
+## TODO
+
+- [ ] We lose per-module analytics which means we will to move to report
+  analytics at the construct level.
+- [ ] We jsii to support the concept of "submodules". I believe this is easier
```
rix0rrr: What was the problem with supporting namespaces in jsii again?

IIRC, the only REAL issue was that in TypeScript, a class can be both a class and a namespace and that's all fine, but that's hard to translate to other languages (i.e., trouble translating `CfnBlaBucket.EncryptionProperty`).

Why are namespaces hard in general? All languages we're targeting (with the possible exception of Go?) support them.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,334 @@
...
+  than arbitrary namespaces which we decided not to support in jsii.
+- [ ] `constructs`
+- [ ] How do we deal with `cx-api`?
```
rix0rrr: `assert` library too. Do we have to do the `assert` library rewrite+jsii-ification before we can execute on this RFC?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
```
rix0rrr: Another TODO: we need to rewrite all example code out there on the internet and in the docs.

text/0006-monolothic-packaging.md
eladb: There isn't an intrinsic issue with namespaces, just that it was hard to find a good way to represent their full semantics in all languages in a consistent way. For example, in Python we ran into issues related to cyclic dependencies between types defined inside the namespace and outside the namespace.

I would argue that it will be "safer" to add more explicit support for "submodules" in jsii, and then we can decide how to represent them in each language.

@RomainMuller can surely expand on that

text/0006-monolothic-packaging.md
eladb: added (and deleting this comment to avoid clutter if that's ok with you).

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+const p = new Paddable('text');
+tablify([[p]]);
+```
+In this code sample, a *Paddable@1* gets constructed and passed to a function which expects to be able to use it as a *Paddable@2*, and the code explodes. 
+Every individual package here was locally correct, but the combination of package versions didn’t work.
+### TypeScript to the rescue
+The CDK uses TypeScript, and the TypeScript compiler actually protects against this issue. If we actually typed the *tablify* function as
+```ts
+import { Paddable } from 'leftpad';
+function tablify(rows:* *Array<Array<Paddable>>) {
```
RomainMuller: What's the two `* *` here?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,338 @@
...
+Every individual package here was locally correct, but the combination of package versions didn’t work.
+### TypeScript to the rescue
+The CDK uses TypeScript, and the TypeScript compiler actually protects against this issue. If we actually typed the *tablify* function as
+```ts
+import { Paddable } from 'leftpad';
+function tablify(rows:* *Array<Array<Paddable>>) { 
+  ...
+}
+```
+The TypeScript compiler would correctly resolve that type to Array<Array<Paddable@2>>, and refuse to compile the calling of *tablify* with an argument of type Array<Array<Paddable@1>>, because of the detected incompatibility between the types.
```
RomainMuller: I recommend wrapping the stuff with chevrons (`<` and `>`) in back-ticks so they don't risk getting interpreted as custom HTML blocks (which are likely to then be filtered out from the rendered view).

text/0006-monolothic-packaging.md
RomainMuller: The main issue we have faced with namespaces was that certain compositions that are perfectly workable in JavaScript cannot be represented in Python. This is in parts due to how interfaces have no physical existence in JavaScript, whereas we materialize them in Python... So if the cyclic dependency is solely caused by interfaces, the TypeScript compiler can work it out... But Python suddenly cannot get out of the cycle.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+- [Unresolved questions](#unresolved-questions)
+- [Future Possibilities](#future-possibilities)
+- [Appendix](#appendix)
+  - [Peer Dependencies in npm](#peer-dependencies-in-npm)
+
+# Motivation
+
+## 3rd-party Construct Libraries
+
+When a 3rd-party construct library declares it's dependency on the AWS CDK as
```
CaerusKaru:
```suggestion
When a 3rd-party construct library declares its dependency on the AWS CDK as
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+["peer dependencies"](https://nodejs.org/es/blog/npm/peer-dependencies) (see
```
CaerusKaru:
```suggestion
When a 3rd-party construct library declares its dependency on the AWS CDK, it's through
["peer dependencies"](https://nodejs.org/es/blog/npm/peer-dependencies) (see
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+[Appendix](#peer-dependencies-in-npm) for details). 
+
+This means that consumers of a 3rd-party construct library are expected to
+**directly** install all transitive dependencies. 
+
+For example, if someone publishes a 3rd-party construct library `foo` that uses the `@aws-cdk/aws-ecs`, the consumers of this library will not only have to take a dependency on `foo`, but also on `@aws-cdk/aws-ecs` **and all it's transitive dependencies** (i.e. 23 additional dependencies).
```
CaerusKaru:
```suggestion
For example, if someone publishes a 3rd-party construct library `foo` that uses the `@aws-cdk/aws-ecs`, the consumers of this library will not only have to take a dependency on `foo`, but also on `@aws-cdk/aws-ecs` **and all of its transitive dependencies** (i.e. 23 additional dependencies).
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+that CDK is "hyper modular", in all languages, we have poor ergonomics when it
+comes to declaring dependencies. Since users are required to explicitly install
+a module for each service they use, even simple projects end up with dozens of
+direct CDK dependencies.
+
+# Basic Example
+
+The AWS CDK will be shipped as a single module that includes the core types and the entire AWS Construct Library.
+
+This means that a 3rd-party library will declare it's dependency on the CDK via a single module `aws-cdk-lib` (name to be decided):
```
CaerusKaru:
```suggestion
This means that a 3rd-party library will declare its dependency on the CDK via a single module `aws-cdk-lib` (name to be decided):
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+The cx-api module is used to coordinate the protocol between the CDK apps and the CLI. Today, both the CLI and the framework are dynamically linked against this module (it is defined in `dependencies`). Once we ship the CDK as a single monolithic module, we will need to decide how to coordinate the protocol.
+
+The current thinking is to *statically link* the protocol definitions between the CLI and the framework. This means that at build time we will create a copy of the protocol definitions (the `.ts` files) and use it both in the CLI and the framework.
+
+## TODO
+
+- [ ] We lose per-module analytics which means we will to move to report
+  analytics at the construct level.
+- [ ] We jsii to support the concept of "submodules". I believe this is easier
```
CaerusKaru:
```suggestion
- [ ] We need jsii to support the concept of "submodules". I believe this is easier
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+  than arbitrary namespaces which we decided not to support in jsii. One thing to consider for example, is where do submodule README files go.
+- [ ] `constructs`
+- [ ] `@aws-cdk/assert` library. Do we have to do the assert library rewrite as a jsii module (we want to do that anyway!).
+- [ ] Rewrite all example code (see [cdk-rewrite-mono-import](https://github.com/rix0rrr/cdk-rewrite-mono-imports)).
+- [ ] Reference documentation needs to also support submodules/namespaces and use the submodule's README file.
+
+# Drawbacks
+
+- [ ] In JavaScript/TypeScript code will always be fully qualified
```
CaerusKaru: Can we have barrel imports? E.g. `aws-cdk-lib/s3`?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+  (`aws_s3.Bucket` instead of `Bucket`) due to limitations of the `import` tool.
+- [ ] The size of the single module will be large (~50MiB) which can create issues of using the CDK in non-standard environments such as from a Lambda function.
+- [ ] Breaking change
+
+# Rationale and Alternatives
+
+I believe our users think of the AWS CDK as a "standard library" (or a
+"framework") and not as another library that they depend on to build their
+applications (like the AWS SDK for example). When users create CDK libraries and
```
CaerusKaru: The most recent version of the SDK uses modular imports (at least in JavaScript, admittedly in alpha). Does this alter your perception?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+- [ ] In JavaScript/TypeScript code will always be fully qualified
+  (`aws_s3.Bucket` instead of `Bucket`) due to limitations of the `import` tool.
+- [ ] The size of the single module will be large (~50MiB) which can create issues of using the CDK in non-standard environments such as from a Lambda function.
+- [ ] Breaking change
+
+# Rationale and Alternatives
+
+I believe our users think of the AWS CDK as a "standard library" (or a
+"framework") and not as another library that they depend on to build their
```
CaerusKaru: Many frameworks are not monolithic. Angular and React on the web, for instance, are not, and not solely due to size concerns. There is still a distinction between what is considered "essential" and what isn't. How is this a strong rationale for monolithic publishing?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+applications (like the AWS SDK for example). When users create CDK libraries and
+apps they don't think of the AWS CDK as yet another library that they use, they
+think of the AWS CDK as the foundation of their app.
```
CaerusKaru: Surely some parts of the CDK are more foundational than others though? Is it enough of a rationale to justify all of the drawbacks of monolithic publishing?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+We can draw the analogy to other standard libraries like the Node.js SDK, the
```
CaerusKaru: See [AWS SDK JS V3](https://github.com/aws/aws-sdk-js-v3). There's definitely a use case for modular imports, but not necessarily on the scale we have now.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+JDK, the .NET Framework. When users write Node.js libraries or apps they don't "depend" on the Node.js SDK, they actually "peer depend" on it (it is expressed via a special `engines` section in `package.json` but for all intents and purposes it's behaves like a peer dependency).
```
CaerusKaru:
```suggestion
JDK, the .NET Framework. When users write Node.js libraries or apps they don't "depend" on the Node.js SDK, they actually "peer depend" on it (it is expressed via a special `engines` section in `package.json` but for all intents and purposes it behaves like a peer dependency).
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+I would argue that if a vendor publishes a 3rd-party construct library, what they *really* want to say is "I am compatible with CDK >= 1.23.0". Then, the decision about which actual CDK version is being used is left to the app level.
```
CaerusKaru: I think it's an acceptable pattern, at least in the NodeJS world, to have multiple dependencies each with the same versioning. HapiJS, Angular, Yarn, and others come to mind. Users don't seem to have a difficult time managing these dependencies, likely because of the strong use of major-versioning to protect against breaking changes. In this way, the API remains stable, so Angular v7 libs work regardless of minor versions, and NPM will strongly complain when you try to upgrade at the major version level.

Where we run into trouble is the 130+ dependencies, or 30+ for a single package. But there is also a middle ground that has to be considered. Essentially, this does not solely justify a monolithic publishing model. It only justifies a "less" publishing model. Why monolithic and not duolithic as @rix0rrr suggests? Or a number less than ten? Or a standard library with IAM, S3, and similar commonly used utilities, and other packages for everything else? This is not clear why we should favor this approach.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+## Alternatives Considered
+
+### Tooling
+
+TODO: we considered building additional tooling (e.g. `cdk install`) that will make hyper-modular peer dependencies sane.
```
CaerusKaru: Why was this rejected?

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+applications (like the AWS SDK for example). When users create CDK libraries and
+apps they don't think of the AWS CDK as yet another library that they use, they
+think of the AWS CDK as the foundation of their app.
+
+We can draw the analogy to other standard libraries like the Node.js SDK, the
+JDK, the .NET Framework. When users write Node.js libraries or apps they don't "depend" on the Node.js SDK, they actually "peer depend" on it (it is expressed via a special `engines` section in `package.json` but for all intents and purposes it's behaves like a peer dependency).
+
+I would argue that if a vendor publishes a 3rd-party construct library, what they *really* want to say is "I am compatible with CDK >= 1.23.0". Then, the decision about which actual CDK version is being used is left to the app level.
+
+## Alternatives Considered
```
CaerusKaru: There needs to be a section on a model that explores "not 130 packages" but not one package. There is a middle ground that could make sense. In the Angular world for instance (which they also consider to be a platform), we have a core set to make apps "work". The core, the "platform" (just the app environment, e.g. browser, server, etc), and the compiler. CDK operates much the same way, where core = constructs, platform = cfn, and compiler = also core, cx-api, etc. Angular has other packages for commonly used utilities like common (includes template extensions), forms, router, animations, etc. The Angular CLI installs all or most of these for new apps, and then manages them moving forward. There are less than ten in total, and library authors typically peer depend on a subset, again shielded by major version protection.

Updating/keeping dependencies in line is seen as relatively stress-free, in part due to the excellent tooling developed, and also just that the level of churn in the Angular world is relatively low (breaking changes don't typically impact everyone, and are very uncommon). HapiJS works in much the same way, where the core engine is exported under the main package, but utilities (very commonly used utilities) are left for the user to install and manage.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+TODO
+
+# Appendix
+
+## Peer Dependencies in npm
+
+> Based on a whitepaper by [@rix0rrr](https://github.com/rix0rrr).
+
+Most languages (Java, Python, .NET) only support a single instance of a module
```
CaerusKaru: Might want to mention this can become an issue in Python as well unless extra care is taken, since Python does not reliably resolve peer dependencies to an expected version by default.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+multiple versions of a module at runtime (since `node_modules` first searched as
+a relative path). This is problematic for the CDK case since CDK modules often
+expose types from other modules in their API (e.g. an `s3.Bucket` accepts a
+`kms.Key` as an option).
+
+The ultimate root cause of the issues we are encountering is a specific feature of the NPM package manager, one that both encourages a package ecosystem that “just works” with minimal maintenance and versioning problems, as well one that significantly simplifies the package manager:
+
+_Packages can appear multiple times in the dependency graph, at different versions._
+
+It simplifies the package manager because the package manager does not have to resolve version conflicts to arrive at a single version that will work across the entire dependency tree, and report a usable error if fails to do so. In the presence of version ranges, this is an open research problem, which NPM conveniently sidesteps.
```
CaerusKaru:
```suggestion
It simplifies the package manager because the package manager does not have to resolve version conflicts to arrive at a single version that will work across the entire dependency tree, and report a usable error if it fails to do so. In the presence of version ranges, this is an open research problem, which NPM conveniently sidesteps.
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+TODO
+
+# Future Possibilities
+
+TODO
+
+# Appendix
+
+## Peer Dependencies in npm
```
CaerusKaru: This is not a section on peer dependencies, this is a section on dependency management in NPM in general. Peer dependencies do not create this problem, explicit dependencies do.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+
+The ultimate root cause of the issues we are encountering is a specific feature of the NPM package manager, one that both encourages a package ecosystem that “just works” with minimal maintenance and versioning problems, as well one that significantly simplifies the package manager:
+
+_Packages can appear multiple times in the dependency graph, at different versions._
+
+It simplifies the package manager because the package manager does not have to resolve version conflicts to arrive at a single version that will work across the entire dependency tree, and report a usable error if fails to do so. In the presence of version ranges, this is an open research problem, which NPM conveniently sidesteps.
+
+### The Feature
+
+As an example, a package *tablify* could depend on a package called *leftpad* at version 2.0, while the application using *tablify* could be using *leftpad* at an older and incompatible version 1.0. The following dependency graph is a valid NPM dependency graph:
```
CaerusKaru: Again, this does not represent peer dependencies, but transitive dependencies, which doesn't seem in line with the problem statement. If library authors declared peer dependencies only, this would not be an issue for users at all. However, the nature of the number of packages to peer depend on is prohibitive, as mentioned above.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+const p = new Paddable('text');
+tablify([[p]]);
+```
+In this code sample, a *Paddable@1* gets constructed and passed to a function which expects to be able to use it as a *Paddable@2*, and the code explodes. 
+Every individual package here was locally correct, but the combination of package versions didn’t work.
+### TypeScript to the rescue
+The CDK uses TypeScript, and the TypeScript compiler actually protects against this issue. If we actually typed the *tablify* function as
+```ts
+import { Paddable } from 'leftpad';
+function tablify(rows:* *Array<Array<Paddable>>) {
```
CaerusKaru:
```suggestion
function tablify(rows: Array<Array<Paddable>>) { 
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+Every individual package here was locally correct, but the combination of package versions didn’t work.
+### TypeScript to the rescue
+The CDK uses TypeScript, and the TypeScript compiler actually protects against this issue. If we actually typed the *tablify* function as
+```ts
+import { Paddable } from 'leftpad';
+function tablify(rows:* *Array<Array<Paddable>>) { 
+  ...
+}
+```
+The TypeScript compiler would correctly resolve that type to Array<Array<Paddable@2>>, and refuse to compile the calling of *tablify* with an argument of type Array<Array<Paddable@1>>, because of the detected incompatibility between the types.
```
CaerusKaru:
```suggestion
The TypeScript compiler would correctly resolve that type to `Array<Array<Paddable@2>>`, and refuse to compile the calling of *tablify* with an argument of type `Array<Array<Paddable@1>>`, because of the detected incompatibility between the types.
```

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+```
+The lack of a caret makes it a fixed version, indicating they really want *1.12.0* and not “at least *1.12.0*” (which would resolve to *1.13.0* in this situation). However, because of the transitive caret dependency the complete dependency graph would look like this:
+```
+app@1.0.0      -> aws-ecs==1.12.0, core==1.12.0
+aws-ecs@1.12.0 -> core^1.12.0
+=== { core@1.13.0 is available } ===> 
+app
++-- core@1.12.0
++-- aws-ecs@1.12.0
+    +-- core@1.13.0
```
CaerusKaru: This should be a peer dependency, which would not cause NPM to install it transitively.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,343 @@
...
+THE OBVIOUS SOLUTION to this is to get rid of the transitive caret dependency, and make intra-CDK dependencies fixed version (*==1.12.0, ==1.13.0, etc*).. This puts all control over the specific version that gets pulled in in the hands of the user:
+```
+app@1.0.0      -> aws-ecs==1.12.0, core==1.12.0
+aws-ecs@1.12.0 -> core==1.12.0
+=== { core@1.13.0 is available, but not selected }===> 
+app
++-- core@1.12.0
++-- aws-ecs@1.12.0
+```
+This works fine for the CDK, but breaks in the face of 3rd party construct
```
CaerusKaru: Again, library authors should declare peer dependencies to fix this issue. Monolithic publishing is not the only or even the best way to solve this problem.

CaerusKaru:
Another point to consider is that the AWS SDK for Java recommends modular imports. They _explicitly_ recommend against a singular dependency import. [Source](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/setup-project-maven.html)

eladb:
> Another point to consider is that the AWS SDK for Java recommends modular imports. They explicitly recommend against a singular dependency import. Source

There is a note about this in the RFC. The SDKs are designed as runtime library dependencies and the AWS CDK is a framework.

eladb:
Added some additional notes in my previous comments.

text/0006-monolothic-packaging.md
eladb: Yes, the solution is peer dependencies (as mentioned above), but our hyper modularity makes it impractical to apply because consumers of a 3rd party may need to install dozens of transitive dependencies they don’t know or care about and may also be broken when a library adds an additional transitive dependency

CaerusKaru:
> CDK apps and libraries cannot exist without the CDK, which is not true for application code which uses the SDKs to interact with AWS resources, but could, for example, move to a different cloud and still have the same functionality.

Strictly speaking I don't agree with this. While it's true that right now the vast majority of CDK projects only use the CDK, it's entirely possible to mix infrastructure in your project with other things, or to have multiple types of infrastructure altogether. For instance, you could have a static website project with CDK and just swap it out for CloudFormation, Serverless, or any other infrastructure provider (as they do today).

text/0006-monolothic-packaging.md
CaerusKaru: Yes I understand, but the point I'm making is that you haven't sold me on why just one package as opposed to two or three? Surely 130+ is the _wrong_ number, but why is one the _right_ number?

eladb: https://en.wikipedia.org/wiki/Zero_one_infinity_rule

eladb: For real. This is actually the reason

CaerusKaru: This alone does not justify moving to “one” (plus this should be in the doc if this is a primary rationale – not everyone knows about this). My argument is not that we set an arbitrary number. It’s that we continue to allow infinity, but we reset the number of packages we currently have lower. In essence, we didn’t specifically target 130+ packages at the start; we didn’t target a specific number at all. In my proposal we don’t target a number either. We _set_ an initial number as a bound for infinity on the negative end. But we can (and probably will) expand it as AWS expands (towards infinity at this rate, I have no doubt).

eladb: I will add to the doc. Good call.

**Infinity** does not address the main problem we are trying to solve, which is the fact that consumers of 3rd party library will need to directly depend on all the transitive dependencies of that library.

CaerusKaru: Third party libraries will always need to depend on a transitive dependency, whether it’s just one in your case, or a number less than we have now in mine. The critical issue that I think we’re missing here is that we can solve this issue to the satisfaction of developers and users alike without choosing one. I don’t see the justification for saying “let’s go nuclear - just one dependency!” when many users (myself included) would be more than happy to manage a number say less than ten (to begin with).

eladb: Let’s say we have 3 modules (A, B and C). consider the case where a 3rd party module peer-depends on module A. 

Now a JavaScript app consumes said 3rd party (defines it under “dependencies”). This means it has to directly also depend on A since A is a peer and won’t be automatically installed.

A month goes by, and our 3rd party evolves in a non-API-breaking way. It now takes a dependency also on B and published a new minor version.

BOOM!

Our app is broken.


The only way I see a solution to this is if we ever only had a single module.

CaerusKaru: That third party app has now violated semver. Even thought the API is non-breaking, it is by definition a breaking change. We cannot not should not be responsible for bad publishing design by third party libs. That could get exhausting very quickly and is not reasonable justification for this entire proposal (eg “people _could_ mess up, so we should bundle the world together forever”)

CaerusKaru: The inconvenience factor of having an additional one or two peer dependencies once in a while, especially when documented in release notes, is not high, nor is it unexpected in the Node world. Dependencies evolve, and people understand this. It’s not the end of the world. The real problem is when there are too many peer dependencies to manage, even with tooling. But the difference between one dep and five deps, for instance, is negligible. Especially when most of them are installed on app init (eg the most common ones)

eladb: I agree. This is a violation of semver that is caused by how the CDK is packaged. It is unacceptable that 3rd party libraries will need to perform a major version bump every time they add a dependency.

eladb: This is not inconvenience, it's a design flaw in the CDK that basically means a healthy 3rd party ecosystem cannot be created at the moment.

eladb: @CaerusKaru I appreciate the discussion, but I think we've exhausted this specific topic. I will try to incorporate some of this feedback under "alternative considered" in the RFC, but otherwise I am resolving this thread for now.

text/0006-monolothic-packaging.md
eladb: See [prototype](https://github.com/aws/jsii/pull/1297)

text/0006-monolothic-packaging.md
eladb: This is a duplicate of https://github.com/aws/aws-cdk-rfcs/pull/122/files#r385260527. No need to repeat.

text/0006-monolothic-packaging.md
eladb: See [discussion](https://github.com/aws/aws-cdk-rfcs/pull/122/files#r385272327)

text/0006-monolothic-packaging.md
eladb: See [discussion](https://github.com/aws/aws-cdk-rfcs/pull/122/files#r385272327). The main concern is the fact that more than a single module will require 3rd party libraries to bump a new major version every time they add a dependency, or otherwise violate semver.

I will add this to **alternatives considered** with the above reasoning.

text/0006-monolothic-packaging.md
```diff
@@ -144,11 +146,13 @@ The current thinking is to *statically link* the protocol definitions between th
...
 - [ ] Reference documentation needs to also support submodules/namespaces and use the submodule's README file.
 
+
+
 # Drawbacks
 
 - [ ] In JavaScript/TypeScript code will always be fully qualified
   (`aws_s3.Bucket` instead of `Bucket`) due to limitations of the `import` tool.
-- [ ] The size of the single module will be large (~50MiB) which can create issues of using the CDK in non-standard environments such as from a Lambda function.
+- [ ] The size of the single module will be large (~50MiB) which can create issues of using the CDK in non-standard environments such as from a Lambda function. See this https://github.com/aws/aws-cdk-rfcs/issues/39#issuecomment-593092612
```
CaerusKaru: You could add that if needed, we could ship a mini ford version of the MonoCDK, which would cut down the size significantly (see [here](https://bundlephobia.com/result?p=monocdk-experiment@1.26.0) for reference)

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,389 @@
...
+The monolithic module will be organized into submodules that match 1:1 the current module system we have for the AWS CDK. These submodules will be implemented using typescript **namespaced exports** (see [jsii PR](https://github.com/aws/jsii/pull/1297)).
+
+* All `@aws-cdk/core` types will be exported without a namespace (root).
+* Hyphens in the current module names will be converted to underscores (`aws-s3` => `aws_s3`). 
+
+### Should we use the `aws_` prefix?
+
+We still need some way to discern between AWS Construct Library modules such as `aws-s3` and framework modules such as `assets`. If we didn't have the `aws-` prefix, and an AWS service will be introduced with the name of a module we already have (e.g. say we launched an "AWS Assets" service one day), we would have a conflict. 
+
+## Issues with Specific Modules
```
CaerusKaru: Is the intention that the MonoCDK combines everything currently under the `@aws-cdk/aws-` scope in the aws-cdk repo currently, or is it just for L1s and L2s?

If it's the latter, might want to add discussion on edge packages like `@aws-cdk/aws-lambda-ndoe`. And if it's the former, might want to clarify that it includes those packages.

text/0006-monolothic-packaging.md
eladb: fixed

text/0006-monolothic-packaging.md
eladb: Yes, good idea!

eladb: https://github.com/aws/aws-cdk/pull/6996

text/0006-monolothic-packaging.md
eladb: Due to the fact that adding peer dependencies is a breaking change, we basically conclude now that the only path forward is monolithic packaging.

text/0006-monolothic-packaging.md
eladb: added

text/0006-monolothic-packaging.md
eladb: Removed this section.

text/0006-monolothic-packaging.md
eladb: Removed this section.

text/0006-monolothic-packaging.md
eladb: Removed this section.

text/0006-monolothic-packaging.md
eladb: Removed this section.

text/0006-monolothic-packaging.md
eladb: added

text/0006-monolothic-packaging.md
eladb: added

misterjoshua:
A small DX perspective from me: I’d like to advocate for incorporating aws and cdk into the `constructs` package name somehow.

Having _aws_ and _cdk_ in a package name identifies a package as being related to both AWS and CDK. `constructs` has neither string in its name, so when looking at the name of the constructs package without foreknowledge of the relationship, it's not obvious that `constructs` is related at all to `aws-cdk-lib`, much less that `constructs` is the core programming model backing aws-cdk-lib.

For comparison, Eslint requires an eslint-plugin prefix for plugin names. I think it’s clear that a plugin name like eslint-plugin-prettier is an eslint plugin. Babel has incorporated a similar naming scheme. At least from my experience, this consistency helps me when I’m pruning my package.json to see what dependencies are related.

My suggestion would be to pick a package name like `aws-cdk-core` instead of `constructs` for the core model.

eladb:
> A small DX perspective from me: I’d like to advocate for incorporating aws and cdk into the `constructs` package name somehow.
> 
> 
> 
> Having _aws_ and _cdk_ in a package name identifies a package as being related to both AWS and CDK. `constructs` has neither string in its name, so when looking at the name of the constructs package without foreknowledge of the relationship, it's not obvious that `constructs` is related at all to `aws-cdk-lib`, much less that `constructs` is the core programming model backing aws-cdk-lib.
> 
> 
> 
> For comparison, Eslint requires an eslint-plugin prefix for plugin names. I think it’s clear that a plugin name like eslint-plugin-prettier is an eslint plugin. Babel has incorporated a similar naming scheme. At least from my experience, this consistency helps me when I’m pruning my package.json to see what dependencies are related.
> 
> 
> 
> My suggestion would be to pick a package name like `aws-cdk-core` instead of `constructs` for the core model.

We intentionally chose a name that _does not_ include AWS or CDK in order to enable new CDK use cases like [cdk8s](https://github.com/awslabs/cdk8s).

eladb:
> We use a subset of the CDK to create infra on demand for customers as part of a SaaS. The CDK is executed within a Lambda (coded in Java). 
> 
> 
> 
> Am I right in thinking that this change would now require us to bundle *all* of the CDK into our Lambda?
> 
> 
> 
> If so, that’s going to come up against deployment size limits and make our Lambda bigger and slower than it needs to be. I’d really rather this proposal didn’t happen, please.

As discussed over Twitter, we expect the monolithic CDK library not to exceed 20MiB, and lambda deployment size limit is 250MiB. Do you still think this will be a major barrier?

mrgrain:
Not sure what the implications would be, but could this change be optional and modularised packages still be published?

eladb:
> Not sure what the implications would be, but could this change be in addition to the existing setup and modularised packages still be published? 

Can you describe your use case for multiple modules?

As described in the rfc, we don’t see much value in continuing to release the individual modules because this will cause the 3rd party library story to break. 3rd party libraries will have to take a dependency on the monocdk which means that any user who depends on any third party library will have to also depend on the monocdk as well.

There are some really powerful construct libraries in the works (both from within amazon and outside) and we want to make sure that using them is a first-class experience.

mrgrain:
Nah you are right, I missed the point about 3rd party libraries. Was really just wondering why we can't have both. I guess it's interesting to see that some tools, e.g. the aws-sdk are moving towards  multiple modules and CDK is moving back to a monolith.

But I thought a bit more about use cases like including it in a lambda and that's nothing that can't be resolved with a good bundler (that removes out unused code).

eladb:
> e.g. the aws-sdk are moving towards multiple modules and CDK is moving back to a monolith.

This is also somewhat discussed in the RFC. Basically the major difference is that the SDKs are used as libraries and can be hosted in all kinds of apps with a huge range of constraints (e.g. performance, memory, load time, disk, etc). The CDK is used as a "framework" and mostly consumed by "CDK Apps" which are normally executed from build environments, and therefore with much less constraints.

text/0006-monolothic-packaging.md
```diff
@@ -0,0 +1,444 @@
...
+able to resolve the **aws-kms** dependency at runtime.
+
+## Implications of peer dependencies
+
+If we modeled all the CDK dependencies as peers (as they should be), it means,
+for example, that if an app uses the **aws-ecs** module, the app will have to
+explicitly install all the 40 transitive dependencies.
+
+The other **critical** implication of using peer dependencies is that adding a peer
+dependency to a module in in fact **A BREAKING CHANGE**. Any direct or indirect
```
trivikr, line 56: nit
```suggestion
dependency to a module is in fact **A BREAKING CHANGE**. Any direct or indirect
```

mikegwhit:
My understanding of this RFC (reading the referenced #6 ) is that two differing package versions will cause conflicts.  I elaborated in an issue ticket that another use case is when compiling TypeScript against a globally installed module.  I lack TypeScript knack, but somehow I suspect the latter use case is solveable with a type definition file.

Focusing on just the version differences, it seems like the root cause is that NPM will install multiple versions of packages.  I'd first want to know why `peerDependencies` isn't an option.  A remark is that this would force developers into a practice that is non-standard.  It would not be unimaginable to provide a smoke test and friendly console message if there are mismatched versions in a `node_modules` directory.  Such a message could guide the user to installing AWS CDK as a set of `peerDependencies`.  An even more invasive version of this would automagically write to a user's package.json file but this could understandably breach trust for some or worse get flagged for security concerns.

It seems like a much heavier version of this resolution is to implement package management within AWS CDK.  This could look like `cdk i [MODULE]`.  This would circumvent needing to package CDK as a monolithic module.  I am experimenting with ideas along these lines in my own software (more for licensing concerns).  This could manage and contain all CDK modules within the single version of CDK.  This method successfully would abstract CDK submodule versions from the problems introduced by mismatched versions within NPM.  The downside is that you'd need to track any new breaking changes in NPM.  It's a heavier effort, but you result with one single CDK version and you can more precisely control interaction of dependencies.

It seems like the third option proposed here is to simply install everything (which amounts to 20mb).  I would propose concerns relating to wanting to stage a CDK installation perhaps within a CICD environment.  I think someone mentioned this in context of Lambda but it seems more accurate to reference in context of a CICD since a usual setup will install CDK dependencies fresh on each build (even if from a cache in the ideal circumstance?).

Am guessing these thoughts have all been kicked around but that's my view of this problem space.  I've looked at this problem space quite a bit for my own work and landed on option #2 (the heavy handed solution) because I have licensing concerns for my code, i.e. I want to do a key exchange before installation.

PR #136, opened 2020-03-03
fix: markdownlint is not actually catching lints

nija-at:
It seems that Github actions don't support shell globbing, so
`markdownlint` was actually looking for files name `*.md` and `**/*.md`
which obviously don't exist.

Switching to `find` and `xargs` to enumerate the files and pass them to
the `markdownlint` explicitly.

---

Proof that the linter actually works - https://github.com/aws/aws-cdk-rfcs/pull/136/checks?sha=de1c5761088d78ef5a21a3b475fe8cae586d30ae

---

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -14,7 +14,5 @@ jobs:
           npm install --prefix tools/linters
       - name: run linter
         run: |
-          cd ./tools/linters
-          npx --no-install markdownlint -c markdownlint.json ../../*.md ../../text/**/*.md
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          find . -type f -name '*.md' -not -path '*/node_modules/*' -not -path '*/.github/*' |\
+            xargs -I'{}' ./tools/linters/node_modules/.bin/markdownlint -c tools/linters/markdownlint.json '{}'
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-lint.yml
```diff
@@ -14,7 +14,5 @@ jobs:
           npm install --prefix tools/linters
       - name: run linter
         run: |
-          cd ./tools/linters
-          npx --no-install markdownlint -c markdownlint.json ../../*.md ../../text/**/*.md
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          find . -type f -name '*.md' -not -path '*/node_modules/*' -not -path '*/.github/*' |\
+            xargs -I'{}' ./tools/linters/node_modules/.bin/markdownlint -c tools/linters/markdownlint.json '{}'
```
nija-at: I've explicitly chosen not to use `npx` here to make the output cleaner. 

If I have to use `npx` like before, I'll have to change into the directory (`cd`) which means every line of the linter output will be prefixed with `../../`.

nija-at:
Proof that the linter works now - https://github.com/aws/aws-cdk-rfcs/pull/136/checks?sha=de1c5761088d78ef5a21a3b475fe8cae586d30ae

Issue #116, opened 2020-02-04
Easier identification of experimental modules


eladb:
|PR|Champion|
|--|--------|
|# |        |

## Description

The AWS CDK is a large class library which is released through a single version. In order to allow innovation and quick iteration, we decided to leverage the model employed by the node.js project and tag APIs either as **STABLE** or **EXPERIMENTAL** (as described in our [user guide](https://docs.aws.amazon.com/cdk/latest/guide/reference.html)).

Technically, each API element (class, enum, method, property) can be have a stability tag. If it is not tagged, it will inherit it's stability from it's parent, all the way up to the module level.

Practically, the current "high resolution" module causes confusion and oftentimes results in users coding against experimental APIs without their knowledge. This is because most of the construct library's modules include "CFN resources" which are always STABLE, while other constructs can still be EXPERIMENTAL. Users are also not accustomed to reviewing the API documentations for every module they use.

The purpose of this RFC is to improve the experience around consuming experimental modules, so that it will be easier for customers (in the common case) to acknowledge that they are using experimental APIs.

@ccfife proposes a solution where experimental constructs will be available only under the "@aws-cdk/aws-xxx-experimental" module and only move to the main module as they stabilize. The main module initially will only include "CFN resources" (which are stable).

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
- [ ] Resolved

eladb:
For the record, I think we're going to bring pain onto ourselves if we want to split off experimental APIs into a separate package. Here are some thought exercises:

* What if we want to add an experimental method to an otherwise stable class? Do we copy the complete class?
* What if we want to add an experimental property to an otherwise stable props struct? Do we copy the complete class?
* What if we want to provide a stable API that depends on an experimental interface? (Remember, we can easily do this and there is no reason we shouldn't: the breakage caused by any changes to the experimental API will be detected by our complete in-tree build and corrected by the change maker, leaving to trace of it to downstream consumers) We won't be able to do that as we'll get cyclic dependencies.
* What if we stabilize an experimental class? Do we move it from the experimental package to the stable package, thereby breaking everyone in the process?

The better solution is to vend 2 different flavors of CDK from a single codebase:

* `stable` APIs only
* `stable+experimental` APIs

This will solve all issues mentioned above:

[...]

PR #156, opened 2020-05-01
RFC: 79 CDK v2.0

MrArnoldPalmer:
RFC: #79 CDK v2.0 
---

[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/feature/cdk-2.0/text/0079-cdk-2.0.md)
Proposal detailing a new major version of aws-cdk. Lays out the plan for
rolling out the release and continuity for support of current v1
modules.

**Note** Lets talk about issues marked as "More info needed" to decide
if they should be included or not. I didn't quite understand what they
were or why they required incurring breaking changes.

Related to: #79

---


text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+After a period of testing and gathering feedback from users, `aws-cdk-lib` will be promoted to a stable release with the version `2.0.0`. All other modules, excluding those under the `@aws-cdk/` namespace, will also get promoted to v2.0, though will remain backwards compatible with `@aws-cdk/` modules.
+
+All `@aws-cdk/` namespaced libraries will continue to be published with current `1.x.x` versions.
+
+### v1 Deprecation
+
+`@aws-cdk/` namespaced modules will continue to receive non-breaking changes and be released regularly.
+
+When v1 module support is discontinued will ultimately be dictated by the introduction of some feature or change that the team agrees is worth the cost of dropping support and requires us to do so. Alternatively, if the burden of maintaining v1 module support is no longer warranted by the number of users continuing to rely on them.
```
nija-at, line 135: I'm not convinced we're doing the right thing here.

> dictated by the introduction of some feature or change that the team agrees is worth the cost of dropping support and requires us to do so

I'd rather us plan out when this happens, rather than pivot this on some feature/change that may or may not happen in the future.

> if the burden of maintaining v1 module support is no longer warranted by the number of users continuing to rely on them

The way I'm reading this is - "not in any reasonable timeline". We've this happen with releases other in other tools or ecosystem (e.g., it took a long time python2.7 could be deprecated).

We should a take a different approach instead - define and announce a sunset period and/or a maintenance period for 1.0 and stop releases after that. 
We should also consider an upgrade tool for customers to move from 1.0 to 2.0 (similar to the monocdk rewrite?) in all our supported languages.

Another benefit we will get from this is that deprecated code and packages can actually be deleted from our code base, and we'll no longer have to maintain backwards compat.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### [`aws-cdk-lib` Version Unlocking]()
+
+Since `aws-cdk-lib` will have a version number differing from the one used by lerna to version all of the other packages in the repo until the "stable" phase, changes to the align-version and bump scripts will have to be made in the repo to handle this.
+
+#### [Deprecation Aware Builds]()
+
+A strategy for selectively including or excluding apis marked `deprecated` during build time. This can be performed at the JSII level and potentially can be linked to specific version numbers with additional information in the `deprecated` annotation. JSII can warn when APIs marked for deprecation within a specific major version release are being included in the build. For example:
```
nija-at, line 155: What about deprecated packages?

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+An api annotated like so:
+```typescript
+  /**
+  * @deprecated-v1
```
nija-at, line 160: This feels more complicated than needed. 
Why not just maintain only 2 active versions at any time? Before we move to the next major version, all items marked removed from the n-1 version is removed from source code.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+  */
+  function myDeprecatedApi() {}
+```
+
+When inside a module versioned `1.x.x`
+`jsii` - builds normally, no warning
+
+when inside a module versioned `2.x.x` or greater
+`jsii` - build fails with error stating that deprecated APIs are included and points to the corresponding lines of code. It will suggest usage of `--no-deprecations` option to exlcude these APIs.
+`jsii --no-deprecations` - builds without warning, but excludes the above method and others marked deprecated during the v1.0 cycle
```
nija-at, line 170: Sorry, I'm not following everything that is said here. Can you clarify better on the text here? It might read better if you start with the use case and then state the flags to use for that use case.

Are you saying that there is now an option to continue to keep deprecated APIs in the new version?

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+CDK v2.0!
+
+This RFC details the strategy for building and releasing aws-cdk 2.0, in addition to changes to tooling required to support this strategy. For details on the specific features themselves, see the corresponding RFCs.
+
+# Motivation
+
+Since the CDK was announcned as "generally available", the team has tried to limit breaking changes for users. However, a handful of features have emerged that the core team believes will make the CDK significantly simpler to use. These changes require a 2.0 release to signal to users that code changes may be required for adoption.
+
+The main change of note is to consolidate all of the aws service construct libraries into a single package. The motivation for this change is detailed in [the corresponding RFC](https://github.com/aws/aws-cdk-rfcs/issues/6).
```
eladb, line 19:
```suggestion
The main change of note is to consolidate all of the aws service construct libraries into a single package. The motivation for this change is detailed in [the corresponding RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0006-monolothic-packaging.md).
```

eladb, line 19:
```suggestion
The main change of note is to release the CDK as a single package instead of multiple package. The motivation for this change is detailed in [the corresponding RFC](https://github.com/aws/aws-cdk-rfcs/issues/6).
```

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+- [Monolithic Packaging](https://github.com/aws/aws-cdk-rfcs/issues/6) - Combine all aws construct libraries into a single package.
+- [Reset all Feature Flags](https://github.com/aws/aws-cdk-rfcs/issues/55) - Delete all defined feature flags and extraneous logic for the "old" state.
+- Remove Deprecated APIs - Delete all constructs/methods marked with `@deprecated` flag.
+- Separate Experimental Code - Extract experimental APIs into a separate entrypoint.
+
+# Detailed Design
+
+## Release Stages
+
+CDK v2.0 will be released in stages. The stages also dictate if and how the v1 modules continue to be published.
```
eladb, line 59: Maybe provide an overview of the stages before diving into them to give readers the full picture upfront

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+`monocdk-experiment` is renamed to `aws-cdk-lib`. The version of this module now becomes v2.0 and is tagged as a "pre-release" version. This has varying formats for different language targets.
+
+#### NodeJS
+version number: `2.0.0-beta.1`
+deploying: `npm publish --tag beta`
+installing: `npm install aws-cdk-lib@beta`
+package.json:
+```sh
+"dependencies": {
```
eladb, line 88: Worth mentioning that this is for apps.  Library vendors should use `peerDependencies` here, correct?

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+  "aws-cdk-lib": "~2.0.0-beta.1"
+}
+```
+**Note: `~` ranges match all versions with the same major, minor, and patch numbers with a greater prerelease number. For example, the above would match `2.0.0-beta.2`. but not `2.0.1-beta.2`.**
```
eladb, line 92: Will it also match `2.1.1-beta.2`?

eladb: What's our strategy for bumping these numbers? I am wondering if perhaps the last number should reflect the major+minor of the corresponding 1.x?

Can we do something like this:

|v1|maps to v2|
|-|-|
|1.32.0|2.0.0-beta.32.0|
|1.33.2|2.0.0-beta.33.2|

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+```xml
+<dependencies>
+    <dependency>
+        <groupId>software.amazon.awscdk</groupId>
+        <artifactId>lib</artifactId>
+        <version>2.0.0-beta-1</version>
+    </dependency>
+</dependencies>
+```
+
```
eladb, line 106: again, what's the strategy for mapping v1 versions to these and how can users take a "caret" version?

text/0079-cdk-2.0.md
eladb: I tend to agree and also becoming a bit concerned about our ability to actually hide deprecated APIs from v2.0 without really modifying the code. As part of #7708 and #7736 I ran into a few situations where stable APIs extend an experimental API (e.g. `assets.IAsset`). Apart from the fact that I think we should probably prohibit this, it will be impossible to hide `IAsset` even though it's marked `deprecated`. We have quite a few places like that in our code and I don't know how this will actually work.

An additional case is the constructs compatiblity layer, which will also require some actual code changes to get rid of.

I am leaning more and more towards a period of transitioning from a v1.0 code base to a v2.0 code base in which we can actually do some clean ups and leave the old stuff behind some how. This, naturally, will require us to plan better for this transition and be explicit about what we intend to backport. It also means that we may need to continue to release v1.x for a while from a branch.

text/0079-cdk-2.0.md
eladb: Ideally we should omit them from the monocdk, but I think it might be impossible given we have some stable APIs that depend on deprecated/experimental APIs.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+#### [JSII Submodule Support](https://github.com/aws/jsii/issues/1286)
+
+JSII currently doesn't handle namespacing of exports. Support for this will be needed to allow building the `aws-cdk-lib` module in all of the cdk supported target languages. Though the initial work is complete, some open questions remain about the scope of this feature. Namely, in order to make migrating to the new module structure as easy as possible, [namespace renaming]() could potentially allow C#, Python, and Java users to not have to change the structure of their import statements.
+
+#### [`aws-cdk-lib` Version Unlocking]()
+
+Since `aws-cdk-lib` will have a version number differing from the one used by lerna to version all of the other packages in the repo until the "stable" phase, changes to the align-version and bump scripts will have to be made in the repo to handle this.
+
+#### [Deprecation Aware Builds]()
```
eladb, line 153: I wonder if this is worth a separate RFC or at least a deeper dive/prototype.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+In addition, since the v1 modules will continue to be supported for some time, and the `aws-cdk-lib` package will be made of up these modules, the inclusion of any of these candidate features would need to be implemented in a way that didn't affect the v1 modules. How this would be done has not been determined, but may be affected by the nature of any particular change.
+
+#### Recommended
+
+The following are good candidates for inclusion in this major version release
+
+- [Easier Identification of Experimental Modules and APIs](https://github.com/aws/aws-cdk-rfcs/issues/116)
+
+There have been long running discussions about separating experimental APIs into a separate module to make them a more explicit "opt in" feature. Users would either have to install a different module (i.e. `aws-cdk-lib-exerimental`) or import from a different entrypoint (`import * as s3 from 'aws-cdk-lib/experimental`). Since this could be performed at build time and be an optional setting thats turned on or off, it could be introduced in the 2.0 release without affecting the v1 modules, this seems like an ideal time to implement. The technical details of how to achieve this are yet to be determined.
+
```
eladb, line 235: what about stable APIs that extend/use experimental APIs?

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+1. Update Dependencies - Changing from referencing the old `@aws-cdk/` packages to the new `aws-cdk-lib` package in the corresponding target language's package manifest (package.json, .sln, pom.xml, setup.py).
+2. Update Imports - Since the current module structure is being changed, import statements in user's code will need to change to accommodate the new structure. This will be automated wherever needed with the [import rewriter](#import-rewriter).
+
+Users may have to make additional code changes when upgrading if they are relying on deprecated APIs or feature flags that are being removed. This cost can be minimized through good documentation and information in the [migration guide](#migration-guide).
+
+# Rationale and Alternatives
+
+## Rationale
+
+The main motivation for moving forward with a 2.0 release now is the desire to start moving towards the single module structure. The current multi module structure is known to cause a lot of confusion for users. For more details on why aws-cdk is moving to a single package, see the [MonoCDK RFC](https://github.com/aws/aws-cdk-rfcs/issues/6).
```
eladb, line 291: It's not only user confusion or a "belief" that this is better. It's basically impossible to use CDK modules as peer dependencies since every transitive dependency change will be breaking.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+Regardless, the point of using this change to push towards a 2.0 release, is that the single package structure is "the new cdk". Meaning, the long term plan is to only release the `aws-cdk-lib` package and no longer publish the `@aws-cdk/` namespaced ones. The [MonoCDK RFC](https://github.com/aws/aws-cdk-rfcs/issues/6) has more info on the problems currently experienced by users caused by the current packaging strategy and these problems will always exist unless we stop supporting the `@aws-cdk/` packages.
+
+Since we want to eventually supplant the current module structure with the new one, semver says we have to do it with the release of a major version.
+
+# Adoption Strategy
+
+All users will perform the following to adopt cdk v2
+
+1. Update Dependencies - Changing from referencing the old `@aws-cdk/` packages to the new `aws-cdk-lib` package in the corresponding target language's package manifest (package.json, .sln, pom.xml, setup.py).
+2. Update Imports - Since the current module structure is being changed, import statements in user's code will need to change to accommodate the change. This will be automated with the [import rewriter](#import-rewriter).
```
eladb, line 322: Hopefully not in all languages

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Unresolved questions
+
+The bulk of unresolved questions are listed in the [detailed design](#detailed-design) section. To recap:
+
+1. What additional features beyond adding the necessary tooling for releasing a major version going forward, should be included if any?
+2. As part of jsii submodule support, can namespace renaming mitigate the need for users of non-typescript target languages to restructure their import statements?
+3. Can `experimental` APIs be separated at build time to make their usage more explicitly "opt-in" for users?
+4. With regards to automatic exclusion of deprecated APIs, more discovery will be needed to converge on a specific technical strategy. It is currently thought this can be achieved by excluding classes/methods marked `deprecated` during JSII code generation. However, since the typescript/javascript packages are emitted by the typescript compiler, it will have to be done differently for the typscript language target. The current idea is to hide the deprecated APIs by removing their types from the `.d.ts` files emitted during build.
```
eladb, line 333: `/@deprecated/@internal/g` before compilation might do the trick

text/0079-cdk-2.0.md
MrArnoldPalmer: Yeah I'll add a section for library authors here.

text/0079-cdk-2.0.md
MrArnoldPalmer: it wont match `2.1.1-beta.2`. It only matches pre-releases with the same major, minor, and patch with a newer prelease tag. So it has to be `2.0.0-beta.x`.

We will have to test if we can version this way, matching the numbers of the pre-release tag to the minor.patch numbers of the current v1 release. I agree that would be ideal but looking through npm semver docs I don't see examples of additional numbers beyond the `-beta.x`.

text/0079-cdk-2.0.md
MrArnoldPalmer: This is just meant to show how the version of the node module translates to the targets.

Maven does allow defining ranges including prerelease numbers (which I thought it didn't). Will add.

text/0079-cdk-2.0.md
MrArnoldPalmer: I'm leaning towards agreeing with both of you. I think we should agree on a period of continuing support for v1 modules and announce that up front. Additionally the cost of hiding deprecated APIs during build time appears to become more complex the more we look into it.

We could implement whatever we can with the v2 modules building from the v1 modules. Then we can branch the two, without creating a git branch, by copying all of the v1 modules src into the `aws-cdk-lib` modules src directory and change the way we are building to just build from those local copies. Then we can implement breaking changes in the `aws-cdk-lib` module and continue to release v1 as is. This lets us go back and add any critical updates to v1 modules, but obviously ends new features and minor bugfixes etc for v1 during the overlap period.

text/0079-cdk-2.0.md
MrArnoldPalmer: I think a prototype is whats needed.

text/0079-cdk-2.0.md
MrArnoldPalmer: Thats what the design was meant to present. We would still want a # next to each deprecation to know if it was deprecated during the current, or previous cycle right?

text/0079-cdk-2.0.md
MrArnoldPalmer: No the `--no-deprecations` option is what you use when you want to remove deprecated APIs from the build you are performing. When you build without it, and JSII sees deprecated APIs from the `current - 1` version cycle, it warns you saying that you are in fact including APIs that shouldn't be.

text/0079-cdk-2.0.md
MrArnoldPalmer: Yeah we have to decide if that counts as stable or experimental for these purposes. Experimental feels right but does that remove too much useful stuff for the user? We should run some experiments to get metrics for how frequent stable apis calls experimental ones.

text/0079-cdk-2.0.md
MrArnoldPalmer: Will rethink the wording here. Mostly just trying to say "go read the mono rfc".

MrArnoldPalmer:
@eladb I editted the tracking issue to have the list of issues that are candidates for inclusion, and the current recommendation for whether they should be or not.

https://github.com/aws/aws-cdk-rfcs/issues/79

text/0079-cdk-2.0.md
eladb: I am concerned that duplicating the sources inside our repo and restructuring of our source repo will introduce risk and quite a lot of work into the critical path of v2.0. It will also require an awfully manual process to back port anything from 2.0 to 1.x and vice verse. 

I would also consider a git branch as a mechanism to maintain these two version lines temporarily. Perhaps we can even do that at an early stage in order to break the version number: branch off to a 2.x branch _today_ and have it continuously merge from master and build monocdk like today. Then, update the version number there to 2.x and create a separate pipeline that will release it.

Then, we can start applying changes and continue to merge from master until we officially release 2.x and perform the switch.

text/0079-cdk-2.0.md
eladb: By definition those APIs are stable (and jsii-diff will shout if you try to change them). I think we should add a validation to the jsii compiler that prohibits this (and switch all of those APIs to stable so we can enable this check from now on)

text/0079-cdk-2.0.md
rix0rrr: Spec says it's okay: https://semver.org/#spec-item-9

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+requirements.txt: 
+```
+aws-cdk.lib==2.0b1
+```
+
+All `@aws-cdk/` namespaced libraries will continue to be published with current `1.x.x` versions.
+
+### Stable
+
+After a period of testing and gathering feedback from users, `aws-cdk-lib` will be promoted to a stable release with the version `2.0.0`. All other modules, excluding those under the `@aws-cdk/` namespace, will also get promoted to v2.0, though will remain backwards compatible with `@aws-cdk/` modules.
```
rix0rrr, line 127: > All other modules, [...] will also get promoted to v2.0, though will remain backwards compatible with `@aws-cdk/` modules.

I would like to see a table with a list of modules (collapsing all the service packages into one), and what treatment they'll be getting. I'm afraid a couple of these things won't be like the others.

text/0079-cdk-2.0.md
rix0rrr: * Using: it's *super* useful to be able to add an experimental field to a stable struct. If we say that stable functions are not allowed to take structs with experimental members in them, we close this useful avenue. I don't think we should take the route of prohibiting that.

There is something to be said for the struct *itself* not being experimental, but given that the struct itself can be stable and literally every field in it could still be experimental, I'm not sure how useful it would be to have that additional restriction.

* Extending: I was all ready to say that an API class/interface not be able to extend an experimental class/interface, but after thinking about the previous issue I'm not so sure about that anymore. Inheriting from an experimental base just means that you'll inherit some experimental members. It should behave somewhat the same as if you just hand-declared a bunch of experimental members, and that shouldn't be disallowed, so why should this?

I guess there's something to be said about the stability of the class/interface name.

But on the other hand, if an interface of a class `ISomething` is experimental, then a consumer shouldn't be declaring variables of type `ISomething` to store that class instance on. Declare variables to be `const x: SomeConcreteSublass = gimmeObject();` instead.

All in all, I'm not so sure I have a problem with this anymore.

To my mind, feels like in the "stable" rendition of the library, jsii should behave as if the experimental types weren't `export`ed (same as it does now for private base classes and private base interfaces).

rix0rrr: What I'm more concerned about is the naming of types.

Types should have the same FQNs whether they are experimental or stable. If we don't have that, switching a class from experimental to stable breaks all consumers, and that would be shit.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+- [Public S3 Artifacts](https://github.com/aws/aws-cdk-rfcs/issues/39)
+
+A significant downside of packaging all of the aws service construct libraries together is that the size of dependencies shipped to the users in most cases will be increased significantly. Being able to remove assets, like custom resource lambda code, that are currently bundled with the modules, could mitigate some of the impact on package size. This could be accomplished by packaging said assets, uploading them to S3, and then creating publically accessible lambda function layers referencing them across all AWS regions. Whether v2 of the cdk could accomplish this without impacting v1 has not been determined.
```
rix0rrr, line 238: It would be quite scary to me to have this maintenance burden of a set of public artifacts.

All of a sudden we'd be responsible for replicating these all across the world, have monitoring on them, have backups, have a release pipeline, ...

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+- [Rename ID Parameter](https://github.com/aws/aws-cdk/issues/3203)
+
+`id` is a reserved keyword in python. Renaming this parameter in constructs prevents possible errors. This likely could be implemented as a feature flag with the flag removed as part of the normal major version release process.
```
rix0rrr, line 242: It's not a reserved keyword though, it's a built-in function. It might be a valid concern, but I find the linked ticket light on details and full of FUD.

> I think this could lead to potentials errors down the line for users that use the Python variant of CDK, since the id parameter could clash with the id() function.

Potential confusion how? What is the code he sees people typing that is wrong?

What's more likely is that his editor is yellow-underlining this `id` variable and saying "you shouldn't do this". 

That's fair enough, and might be a good enough reason to change it for that alone, but let's be accurate about why we're doing it: "prevent IDE linter warning on all constructs" is a lot sharper and easier to reason about than "something might go wrong somewhere".

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Not Recommended
```
rix0rrr, line 244: Agreed on all counts. This is going to be hard enough already, let's not make it worse than necessary.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Rationale and Alternatives
+
+## Rationale
+
+The main motivation for moving forward with a 2.0 release now is the desire to start moving towards the single module structure. The current multi module structure is known to cause a lot of confusion for users. For more details on why aws-cdk is moving to a single package, see the [MonoCDK RFC](https://github.com/aws/aws-cdk-rfcs/issues/6).
+
+This belief that releasing a single package will make user's lives easier is countered by a strong desire to limit the exposure of users to breaking changes. Though the aws-cdk is relatively young, one of the core tenants of the project is stability. Major version upgrades are an important part of building and maintaining libraries. However, limiting the cost of breaking changes included in major releases is important. It isn't in the interest of the project to drop a new major version, immediately stop supporting the previous version, and leave the onus on the user to "catch up".
```
rix0rrr, line 293:
```suggestion
This belief that releasing a single package will make user's lives easier is countered by a strong desire to limit the exposure of users to breaking changes. Though the aws-cdk is relatively young, one of the core tenets of the project is stability. Major version upgrades are an important part of building and maintaining libraries. However, limiting the cost of breaking changes included in major releases is important. It isn't in the interest of the project to drop a new major version, immediately stop supporting the previous version, and leave the onus on the user to "catch up".
```

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,337 @@
...
+
+If a deprecated API isn't removed from a version that it was inteded to be, and users start relying on the presence of that API, then by fixing the bug and removing the API in future versions we potentionally introduce breaking changes that weren't intended by us or expected by the users.
+
+Because of this reason, automated verification of our different builds and the presence/absence of APIs and feature flags is, as always, of high importance.
+
+The cost of migration for users varies depending on the state of the application being migrated. All users upgrading to v2 will have to perform the following steps:
+
+1. Update Dependencies - Changing from referencing the old `@aws-cdk/` packages to the new `aws-cdk-lib` package in the corresponding target language's package manifest (package.json, .sln, pom.xml, setup.py).
+2. Update Imports - Since the current module structure is being changed, import statements in user's code will need to change to accommodate the new structure. This will be automated wherever needed with the [import rewriter](#import-rewriter).
```
rix0rrr, line 283: Just a niggle, but an automatic rewriter is not guaranteed to work in the presence of removing deprecated APIs.

If you were using deprecated stuff, it's gone now. No way to rewrite around that.

text/0079-cdk-2.0.md
nija-at: Can we get some numbers?

MrArnoldPalmer:
The team has discussed further decoupling of the `monolithic packaging` and v2.0 release. Because of this a lot of the work discussed in this proposal won't be necessary. Gonna close this for now pending a rewrite to outline the new process which should be much simpler.

text/0079-cdk-2.0.md
ddneilson: Would something like this be feasible to keep it simple? Rather than trying to navigate a single branch driving two different release versions, we pick a date for the transition to a v2.0 release candidate, say the 'pre-release' date suggested above, and then on that date:

1. Create a v1.x branch of the repo. This will be the branch that the v1.x release(s) will be built from, and will only have security patches and important bugfixes backported into it. The v1.x releases of the CLI and Construct library are made off of the v1.x branch, and only the v1.x branch. This lets v1.x sticks around for a while for those that want to stay on it, but doesn't get any new features.

2. Mainline development shifts to a release-candidate process for v2.0. All new code targets the mainline branch. The CDK development team releases a new release candidate every N days/weeks, and when it looks stable enough for prime-time it does the official v2.0 release.

3. The reviewers for PRs on mainline would have to be cognizant of identifying changes that would need to be backported to the v1.x line, and there would have to be a process for that.

New v1.x releases are made as-needed, but their cadence should be drastically reduced as the vast majority of changes will be in mainline.  We don't break any existing apps that are out in the world because the v1.x line has a clean break from the v2.0 line.

RomainMuller:
I'd actually like for V2 to also be coordinated with the removal of all type union usage everywhere.

eladb:
> I'd actually like for V2 to also be coordinated with the removal of all type union usage everywhere.

@RomainMuller  Please add to https://github.com/aws/aws-cdk-rfcs/issues/79

MrArnoldPalmer:
Reopening this as we get ready to revise. The biggest change is that we are no longer trying to publish both 1.0 and 2.0 modules from the main branch and instead will branch off for 2.0 for some short time, then merge back when it is ready to publish.

Some questions that arise with that.
1. How do we continue to support v1 "micro" packages with bug/security fixes after we integrate the 2.0 branch into the main branch and what is our commitment to users for v1 support?
2. Is it best to remove deprecated APIs by hand or is still worth investing in some tooling to automatically remove them and enforce their absence/presence based on version numbers? (probably not).
3. Does branching allow us to include changes that we otherwise weren't able to, IE things that can't be implemented via feature flags or deprecations.

eladb:
@RomainMuller something that we should include as part of the LTS model for vPREV (v1.x) is a mechanism to track usage of each version during the transitional period. Perhaps the LTS model will have stages based on % of customers using the old version. I think this data will be critical to track users' migration to v2.x and announcing EOL for v1.x.

rix0rrr:
Here are 2 concerns I have with the current proposal, and ways to address them.

* Working on forks is going to be hell.
* Incompatible changes is going to prevent people from upgrading, fracturing the ecosystem.

# Working on forks is going to be hell

If we are going to be working on 2 completely separate codebases, *especially* if we're going to have a period where there is a "free-for-all, make all the breaking changes you want", our code is going to diverge and applying fixes and features to both of them is going to be a nightmare. I'm not looking forward to this world.

I'd prefer a more mechanical approach, where we vend both v1 and v2 from the same branch, and the only changes between them are automated in a mechanical way, and revolve around packaging. 

* v1 and v2 are identical, but v2 is packages as monocdk
* v1 and v2 are identical, except that we strip deprecated code from v2 automatically.

We'll only ever master 2 releases at the same time in the same branch. When we finally switch to v3, the now actually ACTUALLY deprecated code from v1 can finally and safely be removed from the code base.

This will have the advantage that we'll be able to support v1 for longer, and it also has the advantage that there's only one branch to work on, so no merge conflicts.

There is a downside in that there are some kind of breaking changes we can not make cutting over from v1 -> v2. For example, we cannot *repurpose* classes or properties (for example the `AutoScalingGroup` and `Vpc` we have now *should* have been or done something different). That is fine, we can still `@deprecate` the old one at v1, introduce a new one for v2, then when it comes to v3 `@deprecate` the v2 and introduce the *proper* `AutoScalingGroup` in v3.

Maybe we can even do some more automated tricks, like this:

```ts
/**
 * An AutoScalingGroup
 *
 * @rename v1 RealAutoScalingGroup
 */
class AutoScalingGroup { 
```

Or something.

# Incompatible changes is going to prevent people from upgrading

The worst thing we can do is make it hard for people to upgrade.

If there's even the littlest thing that will break for them in their current deployments that doesn't have an escape hatch/way out, then they will not upgrade, and stick happily to v1, and our ecosystem will fracture. I'm not talking about code changes--code is stateless and easy to change. The deployments that are already out there are another matter. Those exist, have been created with some configuration, and have to be maintained with that configuration.

That means that every flag and bit we flip by default for v2 should be unflippable, and we can't make any changes that impact the layout of deployments that aren't undoable by a flag.

pgollucci:
@rix0rrr gets my vote!

RomainMuller:
> I'd prefer a more mechanical approach, where we vend both v1 and v2 from the same branch, and the only changes between them are automated in a mechanical way, and revolve around packaging.

I feel this approach is going to severely limit our ability to perform changes that we'd consider desirable for a major version, which you've thought about, but your suggested "fix" only applies to a mildly uninteresting kind of breakage (type renaming)... Do you reckon a similar approach is going to work for:
- Removing type unions to be statically-typed-friendlier (I don't think deprecating the whole L1 layer and introducing a new one, with different type names, is going to be what we want)
- Sanitizing CLI argument names and shapes, changing the CLI's output formats, ...

> This will have the advantage that we'll be able to support v1 for longer, and it also has the advantage that there's only one branch to work on, so no merge conflicts.

I see the customer benefit for "supporting v1 for longer", however I also see how this gives customers zero incentive to upgrade to v2. Some balance has to be found.

---

> That means that every flag and bit we flip by default for v2 should be unflippable, and we can't make any changes that impact the layout of deployments that aren't undoable by a flag.

Are there specific flags that exist right now where people *cannot* "migrate over" without causing replacements? Is this documented somewhere? (It should be!)

rix0rrr:
> * Removing type unions to be statically-typed-friendlier (I don't think deprecating the whole L1 layer and introducing a new one, with different type names, is going to be what we want)

This would need to be a flippable flag. `jsii --no-type-unions`  for v2, and build without for v1.

> * Sanitizing CLI argument names and shapes, changing the CLI's output formats, ...

We'd have to introduce some versioning mechanism on the CLI as well.

> Are there specific flags that exist right now where people cannot "migrate over" without causing replacements? Is this documented somewhere? (It should be!)

Good question. I think all of the feature/future flags fall into this category, otherwise we would have just changed the code in-place and not bothered with the flag. This also applies to all changes we would make in the "free for all" period on the new branch (in case we disregard my concern #1 🙃)

RomainMuller:
> > * Removing type unions to be statically-typed-friendlier (I don't think deprecating the whole L1 layer and introducing a new one, with different type names, is going to be what we want)
> 
> This would need to be a flippable flag. `jsii --no-type-unions` for v2, and build without for v1.

How would this work? Erase all APIs that accept type unions? Somehow magically determine another type signature for the method/property? Based on what?

> This also applies to all changes we would make in the "free for all" period on the new branch (in case we disregard my concern #1 🙃)

Just because I break/clean an API up does not mean it is impossible to upgrade my codebase to the new API while not incurring a resource replacement. It literally only means users must change their code for it to **compile** with the new release.

README.md
```diff
@@ -1,218 +1,191 @@
 # AWS CDK RFCs
 
-This repo is a place to propose and track upcoming changes to [AWS CDK], [jsii], and
```
eladb, line 3: what happened here? revert please

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+
+## Changes considered for v2 inclusion
+
+- The following changes will be included, or are candidates for inclusion:
+
+  | Decision | RFC / Issue    | Description                                                                         |
+  | -------- | -------------- | ----------------------------------------------------------------------------------- |
+  | ✅ In    |                | Remove `@deprecated` APIs from the previous major version                           |
+  | ✅ In    | [RFC-55]       | Reset all feature flags from the previous major version                             |
+  | ✅ In    | [RFC-6]        | Combine all AWS Construct Libraries into a single package                           |
```
eladb, line 49:
```suggestion
  | ✅ In    | [RFC-6]        | Monolithic packaging                           |
```

rix0rrr:
> How would this work? Erase all APIs that accept type unions? Somehow magically determine another type signature for the method/property? Based on what?

In this specific case... I guess it either means `Object`, or maybe there's an annotation which says what type should be emitted. Spitballing here, but `JsiiUnion<CfnBloop, IResolvable, undefined>` which renders to `CfnBloop`.

All problems are solvable if we have chosen a strategy. My point was to propose a steady-and-safe migration/release strategy: one branch with feature flags, as opposed to different branches.

What feature flags are necessary and how we would support that can be decided on a case-by-case basis.

I'm not sure whether your objections stem from the general ("this strategy can never work because here's a change that you cannot release by using feature flags") or from the specific ("there is a specific change X I want to make for 2.0 which cannot be released this way").

> Just because I break/clean an API up does not mean it is impossible to upgrade my codebase to the new API while not incurring a resource replacement. It literally only means users must change their code for it to **compile** with the new release.

I think we are in agreement about what needs to be possible.

I think the consequences are:

* Looking critically at the flags we have.
* There will be those that incur behavior changes which means we can't get rid of them, we can only flip the defaults.
* Toss out a bit warning to all people planning to make changes in the free-for-all period, that they can't make behavioral changes without adding a flag to revert to the legacy behavior.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+repository-wide `version` configuration will be updated to `2.0.0-alpha` (the pre-release identifier could be
+different). A continuous integration pipeline will be configured on this new branch, so that pull requests can be
+automatically validated.
+
+As long as the codebases have not diverged too much, it should be possible to continue forward-porting new developments
+on top of the `next` branch by simply cherry-picking new commits from the `master` branch. Alternatively, it might be
+possible to simply periodically merge `master` into `next`.
+
+As the codebases grow further appart, manual work may be required in order to forward-port feature work merged in
+`master`. As a consequence we will strive to keep divergences to a minimum for as long as possible.
```
eladb, line 121: This is going to be quite a hassle, so we need to set up some ownership around it. It won't happen on volunteer basis.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+v2 codebase that is usable.
+
+At the same time, the work to update the [AWS CDK User Guide] should be scoped and planned out. The updates to the User
+Guide must also include a migration guide to help developers move from CDK v1 over to v2. Special attention will be
+spent in ensuring the v2 release notes contain a clear and accurate summary of breaking changes and how users should
+update their applications.
+
+[aws cdk user guide]: https://docs.aws.amazon.com/cdk/latest/guide/home.html
+
+## Prerelease Publishing
```
eladb, line 178: I think we should begin pre-release publishing as soon as we fork (or close to it) so it will be possible to test it in the wild. We don't have to announce it right away, but it's definitely something we should be able to point people to so they can give us early feedback. Trying to learn some lessons from our work on mono-cdk where we postponed this too long. For example, we will migrate delivlib and other internal projects to it.

There is no reason to wait for the documentation fix.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+- A finalized proposal for the CDK [v1 maintenance plan](#ongoing-maintenance-plan), and a call for developers to
+  provide feedback on that
+- A minimal migration guide for bravehearts who are inclined to try and migrate their existing (**non-production**)
+  applications
+
+## Construct Library Maintenance Tasks
+
+Now that the v2 framework is in place, _AWS Construct Library_ owners (as well as third party contributors) will be
+given some time to perform API-breaking maintenance tasks. The API-breaking changes should be limited to items that were
+scoped out during the [Prerelease Publishing](#prerelease-publishing) phase and listed in the prerelease announcepent,
```
eladb, line 198:
```suggestion
scoped out during the [Prerelease Publishing](#prerelease-publishing) phase and listed in the prerelease announcement,
```

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+prerelease announcement should at least mention the following elements:
+
+- A provisional list of breaking changes expected to be made to the _AWS Construct Library_ (`@aws-cdk/stdlib`) APIs
+  during the [Construct Library Maintenance Tasks](#construct-library-maintenance-tasks) phase
+- A finalized proposal for the CDK [v1 maintenance plan](#ongoing-maintenance-plan), and a call for developers to
+  provide feedback on that
+- A minimal migration guide for bravehearts who are inclined to try and migrate their existing (**non-production**)
+  applications
+
+## Construct Library Maintenance Tasks
```
eladb, line 194: I think I am leaning towards agreeing more and more with @rix0rrr on this. I think every breaking change needs to be vetted as part of this RFC and if it is approved it will be done. I am taking back my proposal to allow owners to just make breaking changes after the fork.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,384 @@
...
+Now that the v2 framework is in place, _AWS Construct Library_ owners (as well as third party contributors) will be
+given some time to perform API-breaking maintenance tasks. The API-breaking changes should be limited to items that were
+scoped out during the [Prerelease Publishing](#prerelease-publishing) phase and listed in the prerelease announcepent,
+however additional changes can be introduced if there is sufficient justification. All breaking changes (including those
+planned in the prerelease announcement) will be documented in the release notes for each subsequent prerelease version,
+per the standard process.
+
+## Developer Preview
+
+At the end of the [Constrcut Library Maintenance Tasks](#construct-library-maintenance-tasks) window, a moratorium on
```
eladb, line 205:
```suggestion
At the end of the [Construct Library Maintenance Tasks](#construct-library-maintenance-tasks) window, a moratorium on
```

README.md
RomainMuller: A reflow happened

text/0079-cdk-2.0.md
RomainMuller: > I think every breaking change needs to be vetted as part of this RFC and if it is approved it will be done.

Given the aspirational timeline, I reckon this means no Construct Library breaking changes (beyond dropping `@deprecated` APIs) will be included in v2. I don't see how library owners will be able to afford the due diligence of an RFC before we need to have a delivery plan signed off.

README.md
eladb: Please revert. The table is auto-generated

text/0079-cdk-2.0.md
eladb: We can give people a couple of weeks to propose minor breaking changes in their modules for review. Worst case we won't have any.

eladb: I am hopeful that most desired changes are already captured through deprecations so I am also fine with just picking those up.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+We will keep an eye on adoption of CDK v2 thanks to the [version reporting] dataset. Should adoption pick up too slowly,
+actions will be taken to understand the causes for slow adoption and identify opportunities to provide additional
+tooling to remove roadblocks and facilitate migration.
+
+[version reporting]: https://docs.aws.amazon.com/cdk/latest/guide/tools.html
+
+## Deprecation of v1
+
+Based on the v1 maintenance plan, feature development is to stop on CDK v1, as only bugs reported against CDK v1 are to
+be fixed. Back-porting of features is not necessary, although the community is still welcome to back-port features.
```
eladb, line 287: Is this the plan? To fix all bugs reported against 1.x both on 1.x and on 2.x? That's going to be pretty expensive. This is something we need to be very intentional about.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+  Candidates_ are ready for _General Availability_ unless significant bugs emerge, and will receive full AWS support.
+- **Full Support**: During this phase, software is fully supported. AWS will provide regular releases that include new
+  features, support for new services, API updates for existing services, as well as bug and security fixes.
+- **Maintenance Announcement**: AWS will make a public announcement at least 6 months before software enters the
+  **Maintenance** phase. During this period, software will continue to be fully supported. Typically, this announcement
+  happens at the same time as the next major version is introduced.
+- **Maintenance**: During this phase, AWS limits releases to address critical bug fixes and security issues only.
+  Software will no longer receive API updates for new or existing services, or be updated to support new regions. The
+  **Maintenance** phase has a default duration of 12 months, unless otherwise specified as part of the **Maintenance
+  Announcement**.
```
eladb, line 318: okay, this does *not* imply that any bug reported on 1.x will also be fixed on 1.x

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+
+# Rationale and Alternatives
+
+## Rationale
+
+The main motivation for moving forward with a v2 release now is the desire to start moving towards the single module
+structure. The current multi module structure is known to cause a lot of confusion for users. For more details on why
+aws-cdk is moving to a single package, see the [MonoCDK RFC](https://github.com/aws/aws-cdk-rfcs/issues/6).
+
+This belief that releasing a single package will make user's lives easier is countered by a strong desire to limit the
```
eladb, line 337: The main motivation for mono-cdk is to allow 3rd party libraries to declare CDK as a peer dependency without causing breaking changes any time additional transitive dependencies are added upstream. Basically monolithic packaging is the only viable solution.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+
+1. Update Dependencies - Changing from referencing the old `@aws-cdk/` packages to the new `@aws-cdk/stdlib` package in
+   the corresponding target language's package manifest (`package.json`, `.csproj`, `pom.xml`, `setup.py`, etc...).
+1. Update Imports - Since the current module structure is being changed in certain languages, import statements in
+   user's code will need to change accordingly:
+   - In TypeScript, this can be automated by a tool such as `@monocdk-experiment/import-rewriter`
+   - In Python, manual replacement of import statements migth be necessary
+   - In .NET languages (C#, F#, ...) and Java, the namespace structure from CDK v1 will be retained, removing the need
+     for any code change
+
```
eladb, line 394: link to the construct compat layer removal RFC which has more details on breaking changes

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+In order to reduce the codebase divergence between `master` and `next` (in order to minimize the amount of manual work
+involved in forward-porting features on `next`), existing feature flags must be made version-aware, such that they are
+enabled by default (and impossible to disable if a clean and reasonably simple migration is possible) when the current
+version is greater then `2.0.0-0`.
+
+The actual removal of code paths that support the _old behavior_ of feature flags that can no longer be disabled will be
+deferred to _after_ the new major version has reached _General Availability_.
+
+### Monolithic Packaging
```
eladb, line 171: Should be moved after the fork and before we start publishing

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+
+Additionally, all modules that are re-packaged as part of `@aws-cdk/stdlib` will be marked as `"private": true` in the
+`next` branch, making these available _only_ via `@aws-cdk/stdlib` in CDK v2.
+
+## Documentation Updates
+
+Before formally announcing the availability of CDK v2 for early adopters, a documentation source must be availble to
+facilitate use of the new repository. The minimum bar should be availability of an API reference documentation for the
+v2 codebase that is usable.
+
```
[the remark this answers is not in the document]
RomainMuller: Just the lines below:
- Finalize breaking changes to be expected
- Have signed off on the v1 maintenance plan
- Have a (preliminary) migration guide ready

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,405 @@
...
+1. Perform a _normal_ release
+1. Ceremoniously annouce CDK v2 is _Generally Available_, and repeat the v1 maintenance plan
+
+## Clean Up
+
+Now that the new major version has been declared _Generally Available_, the previous major (`v1`) has entered
+_maintenance_ mode. Forward porting is no longer exercised: instead, bug fixes on `master` will be back-ported to `v1`
+when necessary. This means `master` can diverge more significantly from `v1` than in earler stages of the process.
+
+The definition of `v1` feature flags and the code paths supporthing the _old behavior_ of those can be deleted.
```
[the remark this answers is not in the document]
RomainMuller: No the feature flags can be *neutralized* in `v2` (i.e: you CANNOT disable them), but the code paths must be kept in order to avoid tricky merge conflicts.

text/0079-cdk-2.0.md
RomainMuller: Either the bug exists in v1 AND v2, and then the fix is likely the same on both sides. Ergo I am assuming it will not be "significantly" more expensive to fix both.

Or the bug exists only in v1 and there is nothing to fix in v2.

text/0079-cdk-2.0.md
RomainMuller: Only if it's critical / security impacting, yes.

text/0079-cdk-2.0.md
RomainMuller: This paragraph isn't about "why monoCDK" (this is explained in the MonoCDK RFC as pointed to the line before), but more "why is this v2" / "why is v2 the way it is"

text/0079-cdk-2.0.md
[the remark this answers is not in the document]
eladb: What I meant to ask is what does API reference documentation support for 2.0 looks like? It would be useful to have a rough idea of the scope of work required. Things that come to mind: add support for multiple versions in the web UI, add support for submodules.

The reason I think this would be useful is to de-risk the project - if we know that this is going to be a major piece of work, we need to start tackling it sooner rather than later since it's in the critical path of the project.

text/0079-cdk-2.0.md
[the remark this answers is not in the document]
eladb: Okay got it. Makes sense. It wasn't super clear to me what "neutralized" means.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,624 @@
...
+  - `@aws-cdk/aws-wafv2`
+  - `@aws-cdk/aws-workspaces`
+- `@aws-cdk/app-delivery`
+- `@aws-cdk/assets`
+- `@aws-cdk/cloud-assembly-schema`
+- `@aws-cdk/cloudformation-include`
+- `@aws-cdk/custom-resource`
+- `@aws-cdk/cx-api`
+- `@aws-cdk/pipelines`
+- `@aws-cdk/region-info`
```
eladb, line 624: All of these (minor `app-delivery` which is deprecated) should be included.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,631 @@
...
+New GitHub labels will be added to differentiate bugs reported against `v1` from those reported against `v2`. Bugs
+reported against `v1` will be closed with a standard message during triaging unless they are identified as critical or
+security-impacting (meaning they qualify for `P0` handling).
+
+A similar provision applies to pull-requests, as they will only be accepted in the `v1` branch if they are non-breaking
+and fall into one of the following categories:
+
+- features dedicated to facilitating migration from `v1` to `v2`
+- features back-ported from `v2`
+- bug fixes (where relevant, the author is expected to submit the `v2` fix at the same time)
```
njlynch, line 357:
```suggestion
- `P0` bug fixes (where relevant, the author is expected to submit the `v2` fix at the same time)
```

Is this correct (based on the above paragraph)? We will only accept `P0` bugs on `v1`, so we'll only accept PRs for `P0` bug fixes?

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,631 @@
...
+Once sufficient documentation is in place, and the framework-level changes have been implemented and integration tested,
+we will be able to start promoting the upcoming new release and point early adopters to experimental builds of CDK v2.
+This will allow gathering early feedback on the updated experience.
+
+A prerelease announcement will be published, formally announcing our plan to release AWS CDK v2 in the near future. This
+prerelease announcement should at least mention the following elements:
+
+- A provisional list of breaking changes expected to be made to the _AWS Construct Library_ (`aws-cdk-lib`) APIs during
+  the [Construct Library Maintenance Tasks](#construct-library-maintenance-tasks) phase
+- A finalized proposal for the CDK [v1 maintenance plan](#ongoing-maintenance-plan), and a call for developers to
```
njlynch, line 267:
```suggestion
- A finalized proposal for the CDK [v1 maintenance plan](#aws-cdk-maintenance-policy), and a call for developers to
```

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,631 @@
...
+  - `@aws-cdk/aws-batch`
+  - `@aws-cdk/aws-budgets`
+  - `@aws-cdk/aws-cassandra`
+  - `@aws-cdk/aws-ce`
+  - `@aws-cdk/aws-certificatemanager`
+  - `@aws-cdk/aws-chatbot`
+  - `@aws-cdk/aws-cloud9`
+  - `@aws-cdk/aws-cloudformation`
+  - `@aws-cdk/aws-cloudfront`
+  - `@aws-cdk/aws-cloudtrail`
```
njlynch, line 511:
```suggestion
  - `@aws-cdk/aws-cloudfront`
  - `@aws-cdk/aws-cloudfront-origins`
  - `@aws-cdk/aws-cloudtrail`
```

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,631 @@
...
+<!-- replace the blockquoted sections with your content -->
+
+# Summary
+
+This RFC details the strategy for building and releasing v2 of the AWS CDK, in addition to changes to tooling required
+to support this strategy. For details on the specific features themselves, refer to the corresponding RFCs.
+
+# Motivation
+
+Since the CDK was announcned as "generally available", the team has tried to limit breaking changes for users. However,
```
njlynch, line 17:
```suggestion
Since the CDK was announced as "generally available", the team has tried to limit breaking changes for users. However,
```

text/0079-cdk-2.0.md
RomainMuller: No - I think we should accept bug fixes from the community for any severity, but we (the core team) will not author them.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,632 @@
...
+
+* The following changes were scoped out because a solution is likely to be achievable without breaking existing
+  customers:
+
+  | Decision | RFC / Issue | Description                                                                |
+  | -------- | ----------- | -------------------------------------------------------------------------- |
+  | 🚫 Out   | [#39]       | Distribute lambda code for bundled custom resources via a public S3 bucket |
+  | 🚫 Out   | [RFC-77]    | Generate CloudFormation Resources from the CloudFormation Registry schemas |
+  | 🚫 Out   | [RFC-193]   | Fix API for TypeScript type unions when used in statically typed languages |
+  | 🚫 Out   | [#116]      | Easier identification of `@experimental` modules and APIs                  |
```
ccfife, line 60: What is the plan for "experimental" modules and APIs in CDK v2.0? Too many customers have come to the wrong conclusion about the stability of the CDK because they were using experimental modules and didn't know it; presumably because they didn't read the API reference labels. The note above this table states the solution is "likely" achievable without breaking customers, but how we solve this particular problem is not stated. I like the proposal in issue #116 to have 2 versions of the CDK; one stable and one stable+experimental.

text/0079-cdk-2.0.md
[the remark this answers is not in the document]
ccfife: I support the notion of the API reference for CDK v2.0 have a similar layout to the current version, i.e. a separate page for each service/module. 

We've heard so much feedback that the non-TypeScript portions of the API reference are sub-optimal, we really need to address the pain points in v2. Most of the feedback is for the API ref boils down to:

1. the overview page is not the default landing page when you navigate to a module in the Python API re
2. not all of the code snippets translated from TypeScript to the other languages are 100% correct (debugging incorrect sample code in official AWS docs is a poor customer experience), 
3. there are no code snippets or meaningful README content for "cfn-only" modules, and some customers don't know how to use the L1 constructs available in these modules - leading them to think the module/service is not yet available in the CDK

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,632 @@
...
+
+Now that the new major version has been declared _Generally Available_, the previous major (`v1`) has entered
+_maintenance_ mode. Forward porting is no longer exercised: instead, bug fixes on `master` will be back-ported to `v1`
+when necessary. This means `master` can diverge more significantly from `v1` than in earler stages of the process.
+
+The definition of `v1` feature flags and the code paths supporthing the _old behavior_ of those can be deleted (they
+were previously simply _neutralized_ in `v2` so that they could no longer be disabled there).
+
+## Monitoring Adoption
+
```
ccfife, line 336: I don't think there is much work to do on the CDK CLI side of our metrics collection, but we'll need to rethink how we use the AWS::CDK::Metadata resource to track usage since all CDK constructs will be delivered in a single package/module. Ideally, we'll get more granular data on the actual constructs (L1, L2, or L3) used in CDK code.

text/0079-cdk-2.0.md
```diff
@@ -0,0 +1,632 @@
...
+problems currently experienced by users caused by the current packaging strategy and these problems will always exist
+unless we stop supporting the `@aws-cdk/` packages.
+
+Since we want to eventually supplant the current module structure with the new one, semver says we have to do it with
+the release of a major version.
+
+# Adoption Strategy
+
+All users will perform the following to adopt CDK v2
+
```
ccfife, line 451: What's the feasibility of a migration tool to help customers update their dependencies and imports?

text/0079-cdk-2.0.md
rix0rrr: The non-breaking thing we can do here quickly is add `"(experimental)"` to the front of every experimental API's docstring.

Two different libs won't work well as soon as we involve 3rd party libraries.

text/0079-cdk-2.0.md
RomainMuller: All of this is *easy* if you consider **one** language (**TypeScript**/**Javascript**). Things become **much** hairier once you start to think about **Python**, **C#**, **Java**, ... I think if we go into tooling we should put ourselves in the position to offer an equally managed experience for **all** languages (and I don't think this is realistic).

PR #169, opened 2020-06-04
RFC: 0164 construct library segments

nija-at:
[Rendered version](/aws/aws-cdk-rfcs/blob/nija-at/segments/text/0164-construct-library-segments.md)

---

### DIFF
--- a/text/0164-construct-library-segments.md
+++ b/text/0164-construct-library-segments.md
@@ -1,0 +1,243 @@
+---
+feature name: Construct Library Segments
+start date: 26/05/2020
+rfc pr: [#169](https://github.com/aws/aws-cdk-rfcs/pull/169)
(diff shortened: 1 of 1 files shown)


nija-at:
> I might have missed something but I am failing to understand what are the implications from a user’s standpoint besides roadmap granularity and the table in documentation. Are there any implications when importing/defining resources?

There are no other implications besides better customer communication.

eladb:
> > I might have missed something but I am failing to understand what are the implications from a user’s standpoint besides roadmap granularity and the table in documentation. Are there any implications when importing/defining resources?
> 
> 
> 
> There are no other implications besides better customer communication.

So basically the proposal is to add a stability label for each module’s README with a list of segments, and also manage the roadmap at the segment level.

Maybe worth adding to the beginning of the RFC

PR #170, opened 2020-06-07
Resource Type Providers

eduardomourar:
[Rendered version](/eduardomourar/aws-cdk-rfcs/tree/feature/rfc-0158-resource-types/text/0158-resource-type-providers.md)

---

### DIFF
--- a/text/0158-resource-type-providers.md
+++ b/text/0158-resource-type-providers.md
@@ -1,0 +1,75 @@
+---
+feature name: resource-type-providers
+start date: 2020-06-07
+rfc pr: [#170](https://github.com/aws/aws-cdk-rfcs/issues/170)
+related issue: [#158](https://github.com/aws/aws-cdk-rfcs/issues/158)
+---
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


text/0158-resource-type-providers.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Design Summary
+
+
+
+# Detailed Design
+
+
+
+# Drawbacks
```
eladb, line 56: The main drawback as far as I see it is that if we release those custom resources to the registry it means their APIs become public APIs that we must maintain with backwards compatibility, versioning and assurances we are not set out to provide. 

The benefit of embedded custom resources is that they can be treated as an "implementation detail" of the construct, and the API stability can only be managed at the construct level. 

In a sense every public registry resource is a fully fledged service that must be maintained from an API perspective, and I doubt that we have the bandwidth to do that.

eduardomourar: Agreed and I thought of pointing out that this should initially be implemented for single versioned services in mind. So services somewhat "static" (like IAM OIDC provider, DNS certificate validation, etc) can be easily migrated, but not for "dynamic" services (like EKS cluster creation, etc).

eladb:
I am closing this for now. Please reopen when ready to continue

Issue #171, opened 2020-06-15
CloudFront Module Redesign


njlynch:
|PR|Champion|
|--|--------|
|[#174](/aws/aws-cdk-rfcs/pull/174) |  @iliapolo  |

## Description

See https://github.com/aws/aws-cdk/issues/572 for some background.
> There's a lot we can do in order to make it friendlier, less error-prone and less flat and declarative. At the moment, it's basically "leaking" the low level API provided by CloudFormation.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
- [x] Resolved

iliapolo:
@njlynch @rix0rrr @skinny85 

This is what I had in mind:

```typescript
// 90th percentile: only default behavior
new cloudfront.Distribution(this, 'Distribution', {
  defaultBehavior: {
    origin: cloudfront.Origin.fromBucket(myBucket, {}),
    allowedMethods: ALL,
    compress: true
  }
})

// 10th percentile: 1 default behavior, 3 custom behaviors

// at construction
new cloudfront.Distribution(this, 'Distribution', {
  defaultBehavior: {
    origin: cloudfront.Origin.fromBucket(myBucket),
    allowedMethods: ALL,
    compress: true
  },
  additionalBehaviors: {
    '/api/*': {
      origin: cloudfront.Origin.fromHttpServer({
        domainName: 'www.example.com',
        protocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
      }),
      allowedMethods: ALL,
      compress: true
    },
    '/api/_admin/*': {
      origin: cloudfront.Origin.fromHttpServer({
        domainName: 'www.example.com',
        protocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
      }),
      allowedMethods: [GET],
      compress: false
    },
    '/assets/*': {
      origin: cloudfront.Origin.fromBucket(myBucket),
      allowedMethods: ALL,
      compress: true
    }
  }
})

// post construction
const distribution = new cloudfront.Distribution(this, 'Distribution', {
  defaultBehavior: {
    origin: cloudfront.Origin.fromBucket(myBucket),
    allowedMethods: ALL,
    compress: true
  }
})

distribution.addBehavior('/api/*', {
  origin: cloudfront.Origin.fromHttpServer({
    domainName: 'www.example.com',
[...]

PR #174, opened 2020-06-16
RFC: 171 CloudFront Redesign

njlynch:
---
title: "RFC: #0171 CloudFront Redesign"
labels: management/rfc
---

Proposal to redesign the @aws-cdk/aws-cloudfront module.

The current module does not adhere to the best practice naming conventions or ease-of-use patterns that are present in the other CDK modules. A redesign of the API will allow for friendly, easier access to common patterns and usages.

This RFC does not attempt to lay out the entire API; rather, it focuses on a complete re-write of the module README with a focus on the most common use cases and how they work with the new design. More detailed designs and incremental API improvements will be tracked as part of GitHub Project Board once the RFC is approved.

[Rendered version](/njlynch/aws-cdk-rfcs/blob/njlynch/rfc171/text/0171-cloudfront-redesign.md)

---

### DIFF
--- a/text/0171-cloudfront-redesign.md
+++ b/text/0171-cloudfront-redesign.md
@@ -113,25 +113,21 @@ new cloudfront.Distribution(this, 'myDist', {
 
 Note that in the above example the aliases are inferred from the certificate and do not need to be explicitly provided.
 
-## Caching Behaviors
+## Behaviors
 
-Each distribution has a default cache behavior which applies to all requests to that distribution; additional cache behaviors may be specified for a
-given URL path pattern. Cache behaviors allowing routing with multiple origins, controlling which HTTP methods to support, whether to require users to
-use HTTPS, and what query strings or cookies to forward to your origin, among other behaviors.
+Each distribution has a default behavior which applies to all requests to that distribution; additional behaviors may be specified for a
+given URL path pattern. Behaviors allow routing with multiple origins, controlling which HTTP methods to support, whether to require users to
+use HTTPS, and what query strings or cookies to forward to your origin, among others.
 
-The properties of the default cache behavior can be adjusted as part of the distribution creation. The following example shows configuring the HTTP
+The properties of the default behavior can be adjusted as part of the distribution creation. The following example shows configuring the HTTP
 methods and viewer protocol policy of the cache.
 
 ```ts
 const myWebDistribution = new cloudfront.Distribution(this, 'myDist', {
-  origin: cloudfront.Origin.fromHTTPServer({
-    domainName: 'www.example.com',
-    protocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
-  }),
-  behavior: {
+  origin: cloudfront.Origin.fromLoadBalancerV2(myLoadBalancer, {
     allowedMethods: AllowedMethods.ALL,
     viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
-  }
+  }),
 });
 ```
 
@@ -196,13 +192,12 @@ The following shows a Lambda@Edge function added to the default behavior and tri
 ```ts
 const myFunc = new lambda.Function(...);
 new cloudfront.Distribution(this, 'myDist', {
-  origin: cloudfront.Origin.fromBucket(myBucket),
-  behavior: {
+  origin: cloudfront.Origin.fromBucket(myBucket, {
     edgeFunctions: [{
       functionVersion: myFunc.currentVersion,
       eventType: EventType.VIEWER_REQUEST,
-    }]
-  },
+    }],
+  }),
 });
 ```
 
@@ -424,15 +419,14 @@ new CloudFrontWebDistribution(this, 'dist', {
 
 ```ts
 const dist = new cloudfront.Distribution(this, 'MyDistribution', {
-  origin: cloudfront.Origin.fromLoadBalancerV2(lbFargateService.loadBalancer),
-  behavior: {
+  origin: cloudfront.Origin.fromLoadBalancerV2(lbFargateService.loadBalancer, {
     allowedMethods: AllowedMethods.ALL,
     forwardQueryString: true,
     edgeFunctions: [{
       function: myFunctionVersion,
       eventType: EventType.ORIGIN_RESPONSE,
-    }]
-  }
+    }],
+  }),
 });
 dist.addOrigin(cloudfront.Origin.fromBucket(sourceBucket), { pathPattern: 'static/*' });
 ```
--- a/text/0171-cloudfront-redesign.md
+++ b/text/0171-cloudfront-redesign.md
@@ -77,7 +77,7 @@ new cloudfront.Distribution(this, 'myDist', {
 
 // Creates a distribution for an HTTP server.
 new cloudfront.Distribution(this, 'myDist', {
-  origin: cloudfront.Origin.fromHTTPServer({
+  origin: cloudfront.Origin.fromHttpServer({
     domainName: 'www.example.com',
     protocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
   })
@@ -250,11 +250,10 @@ The following is an incomplete, but representative, listing of the API:
 
 ```ts
 class Distribution extends BaseDistribution {
-  static fromArn(scope: Construct, id: string, distributionArn: string): IDistribution;
-  static fromAttributes(scope: Construct, id: string, distributionArn: string): IDistribution;
+  static fromDistributionAttributes(scope: Construct, id: string, attributes: DistributionAttributes): IDistribution;
 
-  static forBucket(scope: Construct, id: string, bucket: IBucket): IDistribution;
-  static forWebsiteBucket(scope: Construct, id: string, bucket: IBucket): IDistribution;
+  static forBucket(scope: Construct, id: string, bucket: IBucket): Distribution;
+  static forWebsiteBucket(scope: Construct, id: string, bucket: IBucket): Distribution;
 
   constructor(scope: Construct, id: string, props: DistributionProps) {}
 
--- a/text/0171-cloudfront-redesign.md
+++ b/text/0171-cloudfront-redesign.md
@@ -9,170 +9,186 @@ related issue: https://github.com/aws/aws-cdk-rfcs/issues/171
 
 (Draft) Proposal to redesign the @aws-cdk/aws-cloudfront module.
 
-The current module could be enhanced to be friendlier and less error-prone.
+The current module does not adhere to the best practice naming conventions or ease-of-use patterns
+that are present in the other CDK modules. A redesign of the API will allow for friendly, easier
+access to common patterns and usages.
+
+This RFC does not attempt to lay out the entire API; rather, it focuses on a complete re-write of
+the module README with a focus on the most common use cases and how they work with the new design.
+More detailed designs and incremental API improvements will be tracked as part of GitHub Project Board
+once the RFC is approved.
+
+---
 
 # README
 
-Example usages:
+Amazon CloudFront is a web service that speeds up distribution of your static and dynamic web content, such as .html, .css, .js, and image files, to
+your users. CloudFront delivers your content through a worldwide network of data centers called edge locations. When a user requests content that
+you're serving with CloudFront, the user is routed to the edge location that provides the lowest latency, so that content is delivered with the best
+possible performance.
 
-## Use case #1 - Simple S3-bucket distribution
+## Creating a distribution
 
-Creates a distribution based on an S3 bucket, and sets up an origin access identity so the
-distribution can access the contents of the bucket (if non-public).
+CloudFront distributions deliver your content from one or more origins; an origin is the location where you store the original version of your
+content. Origins can be created from S3 buckets or a custom origin (HTTP server).
 
-**Before:**
+An S3 bucket can be added as an origin. If the bucket is configured as a website endpoint, the distribution can use S3 redirects and S3 custom error
+documents.
 
 ```ts
-new CloudFrontWebDistribution(this, 'dist', {
-  originConfigs: [{
-    s3OriginSource: {
-      s3BucketSource: myBucket,
-      originAccessIdentity: OriginAccessIdentity.fromOriginAccessIdentityName(this, 'oai', 'distOAI'),
-    },
-    behaviors: [{ isDefaultBehavior: true }],
-  }]
+// Creates a distribution for a S3 bucket.
+const myBucket = new s3.Bucket(...);
+new Distribution(this, 'myDist', {
+  origin: Origin.fromBucket(this, 'myOrigin', myBucket)
+});
+
+// Creates a distribution for a S3 bucket that has been configured for website hosting.
+const myWebsiteBucket = new s3.Bucket(...);
+new Distribution(this, 'myDist', {
+  origin: Origin.fromWebsiteBucket(this, 'myOrigin', myBucket)
 });
 ```
 
-**After:**
+Both of the S3 Origin options will automatically create an origin access identity and grant it access to the underlying bucket. This can be used in
+conjunction with a bucket that is not public to require that your users access your content using CloudFront URLs and not S3 URLs directly.
+
+Origins can also be created from other resources (e.g., load balancers, API gateways), or from any accessible HTTP server.
 
 ```ts
-new Distribution(this, 'dist', {
-  origins: [Origin.fromS3Bucket(myBucket)],
+// Creates a distribution for an application load balancer.
+const myLoadBalancer = new elbv2.ApplicationLoadBalancer(...);
+new Distribution(this, 'myDist', {
+  origin: Origin.fromLoadBalancerV2(this, 'myOrigin', myLoadBalancer)
+});
+
+// Creates a distribution for an HTTP server.
+new Distribution(this, 'myDist', {
+  origin: Origin.fromHTTPServer(this, 'myOrigin', {
+    domainName: 'www.example.com',
+    originProtocolPolicy: OriginProtocolPolicy.HTTPS_ONLY
+  })
 });
 ```
 
-## Use case #2 - S3-bucket distribution with a Lamda@Edge function
+## Domain Names and Certificates
 
-Creates a distribution based on an S3 bucket, and adds a Lambda function as a Lambda@Edge association
-to origin responses.
-
-**Before:**
+When you create a distribution, CloudFront returns a domain name for the distribution, for example: `d111111abcdef8.cloudfront.net`. If you want to
+use your own domain name, such as `www.example.com`, you can add an alternate domain name to your distribution.
 
 ```ts
-new CloudFrontWebDistribution(this, 'dist', {
-  originConfigs: [{
-    s3OriginSource: {
-      s3BucketSource: myBucket,
-      originAccessIdentity: OriginAccessIdentity.fromOriginAccessIdentityName(this, 'oai', 'distOAI'),
-    },
-    behaviors: [{
-      isDefaultBehavior: true,
-      lambdaFunctionAssociations: [{
-        lambdaFunction: myFunctionVersion,
-        eventType: LambdaEdgeEventType.ORIGIN_RESPONSE,
-      }]
-    }],
-  }]
-});
+new Distribution(this, 'myDist', {
+  origin: Origin.fromBucket(myBucket),
+  aliases: ['www.example.com']
+})
 ```
 
-**After:**
+CloudFront distributions use a default certificate (`*.cloudfront.net`) to support HTTPS by default. If you want to support HTTPS with your own domain
+name, you must associate a certificate with your distribution that contains your domain name. The certificate must be present in the AWS Certificate
+Manager (ACM) service in the US East (N. Virginia) region; the certificate may either be created by ACM, or created elsewhere and imported into ACM.
 
 ```ts
-const myBucketOrigin = Origin.fromS3Bucket(myBucket);
-new Distribution(this, 'dist', {
-  origins: [myBucketOrigin],
-  edgeFunctions: [{
-    lambdaFunctionVersion: myFunctionVersion,
-    eventType: LambdaEdgeEventType.ORIGIN_RESPONSE,
-    origin: myBucketOrigin,
-  }],
+const myCertificate = new certmgr.DnsValidatedCertificate(this, 'mySiteCert', {
+  domainName: 'www.example.com',
+  hostedZone,
 });
-```
-
-## Use Case #3 - S3 bucket distribution with certificate and custom TTL
-
-**Before:**
-
-```ts
-new CloudFrontWebDistribution(this, 'dist', {
-  originConfigs: [{
-    s3OriginSource: {
-      s3BucketSource: myBucket,
-      originAccessIdentity: OriginAccessIdentity.fromOriginAccessIdentityName(this, 'oai', 'distOAI'),
-    },
-    behaviors: [{
-      isDefaultBehavior: true,
-      defaultTtl: Duration.minutes(10),
-    }],
-  }],
-  viewerCertificate: ViewerCertificate.fromAcmCertificate(myCertificate),
-});
-```
-
-**After:**
-
-```ts
-new Distribution(this, 'dist', {
-  origins: [Origin.fromS3Bucket(myBucket)],
-  behaviors: [{defaultTtl: Duration.minutes(10)}],
+new Distribution(this, 'myDist', {
+  origin: Origin.fromBucket(myBucket),
   certificate: myCertificate,
 });
 ```
 
-## Use case #4 - Complicate multi-origin setup with multiple behaviors per origin and a Lambda@Edge function
+Note that in the above example the aliases are inferred from the certificate and do not need to be explicitly provided.
 
-Stress test. This is what the config looks like with more stuff. An S3 bucket origin with the default and
-one custom behavior, and a LoadBalancedFargateService origin that is HTTPS only.
+## Caching Behaviors
 
-**Before:**
+Each distribution has a default cache behavior which applies to all requests to that distribution; additional cache behaviors may be specified for a
+given URL path pattern. Cache behaviors allowing routing with multiple origins, controlling which HTTP methods to support, whether to require users to
+use HTTPS, and what query strings or cookies to forward to your origin, among other behaviors.
+
+The properties of the default cache behavior can be adjusted as part of the distribution creation. The following example shows configuring the HTTP
+methods and viewer protocol policy of the cache.
 
 ```ts
-new CloudFrontWebDistribution(this, 'dist', {
-  originConfigs: [{
-    s3OriginSource: {
-      s3BucketSource: websiteBucket,
-      originAccessIdentity: OriginAccessIdentity.fromOriginAccessIdentityName(this, 'oai', 'distOAI'),
-    },
-    behaviors: [
-      { isDefaultBehavior: true },
-      {
-        forwardedValues: { queryString: true },
-        pathPattern: 'images/*.jpg',
-      }
-    ],
-  },
-  {
-    customOriginSource: {
-      domainName: lbFargateService.loadBalancer.loadBalancerDnsName,
-      originProtocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
-    },
-    behaviors: [{ isDefaultBehavior: true }],
-  }],
+const myWebDistribution = new Distribution(this, 'myDist', {
+  origin: Origin.fromHTTPServer(this, 'myOrigin', {
+    domainName: 'www.example.com',
+    originProtocolPolicy: OriginProtocolPolicy.HTTPS_ONLY
+  }),
+  behavior: {
+    allowedMethods: AllowedMethods.ALL,
+    viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS
+  }
 });
 ```
 
-**After:**
+Additional cache behaviors can be specified at creation, or added to the origin(s) after the initial creation. These additional cache behaviors enable
+customization on a specific set of resources based on a URL path pattern. For example, we can add a behavior to `myWebDistribution` to override the
+default time-to-live (TTL) for all of the images.
 
 ```ts
-const myBucketOrigin = Origin.fromS3Bucket(myBucket);
-new Distribution(this, 'dist', {
-  origins: [
-    myBucketOrigin,
-    Origin.fromApplicationLoadBalancer(lbFargateService.loadBalancer),
-  ],
-  behaviors: [{
-    origin: myBucketOrigin,
-    forwardedValues: { queryString: true },
-    pathPattern: 'images/*.jpg',],
-  },
+myWebDistribution.origin.addBehavior('/images/*.jpg', {
+  viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+  defaultTtl: cdk.Duration.days(7),
+})
+```
+
+## Multiple Origins
+
+A distribution may have multiple origins in addition to the default origin; each additional origin must have (at least) one behavior to route requests
+to that origin. A common pattern might be to serve all static assets from an S3 bucket, but all dynamic content served from a web server. The
+following example shows how such a setup might be created:
+
+```ts
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new Distribution(this, 'myDist', {
+  origin: Origin.fromWebsiteBucket(this, 'myOrigin', myBucket),
+  additionalOrigins: [Origin.fromLoadBalancerV2(this, 'myOrigin', myLoadBalancer, {
+    pathPattern: '/api/*',
+    allowedMethods: AllowedMethods.ALL,
+    forwardQueryString: true,
+  })];
 });
 ```
+
+You can specify an origin group for your CloudFront origin if, for example, you want to configure origin failover for scenarios when you need high
+availability. Use origin failover to designate a primary origin for CloudFront plus a second origin that CloudFront automatically switches to when the
+primary origin returns specific HTTP status code failure responses. An origin group can be created and specified as the primary (or additional) origin
+for the distribution.
+
+```ts
+const myOriginGroup = Origin.groupFromOrigins(
+  primaryOrigin: Origin.fromLoadBalancerV2(this, 'myOrigin', myLoadBalancer),
+  fallbackOrigin: Origin.fromBucket(this, 'myFallbackOrigin', myBucket),
+  fallbackStatusCodes: [500, 503]
+);
+new Distribution(this, 'myDist', { origin: myOriginGroup });
+```
+
+The above will create both origins and a single origin group with the load balancer origin falling back to the S3 bucket in case of 500 or 503 errors.
+
+## Lambda@Edge
+
+Lambda@Edge is an extension of AWS Lambda, a compute service that lets you execute functions that customize the content that CloudFront delivers. You
+can author Node.js or Python functions in the US East (N. Virginia) region, and then execute them in AWS locations globally that are closer to the
+viewer, without provisioning or managing servers. Lambda@Edge functions are associated with a specific behavior and
+
+**TODO:** Design the Lambda@Edge API.
+
+---
 
 # Motivation
 
 > Why are we doing this? What use cases does it support? What is the expected
 > outcome?
 
-***TODO***
+**_TODO_**
 
 # Design Summary
 
 > Summarize the approach of the feature design in a couple of sentences. Call out
 > any known patterns or best practices the design is based around.
 
-***TODO***
+**_TODO_**
 
 # Detailed Design
 
@@ -188,45 +204,6 @@ new Distribution(this, 'dist', {
 > - [Graphviz](http://graphviz.it/#/gallery/structs.gv)
 > - [PlantText](https://www.planttext.com)
 
-Proposed interfaces:
-
-```ts
-export interface DistributionProps {
-  origins?: Origin[];
-  behaviors?: Behavior[];
-  edgeFunctions?: LambdaEdgeFunction[];
-  aliases?: string[];
-  comment?: string;
-  defaultRootObject?: string;
-  enableIpV6?: boolean;
-  httpVersion?: HttpVersion;
-  priceClass?: PriceClass;
-  viewerProtocolPolicy?: ViewerProtocolPolicy;
-  webACLId?: string;
-  errorConfigurations?: ErrorConfiguration[];
-  loggingBucket?: IBucket;
-  loggingIncludeCookies?: boolean;
-  loggingPrefix?: string;
-  certificate?: certificatemanager.ICertificate;
-  sslMinimumProtocolVersion?: SslProtocolVersion;
-  sslSupportMethod?: SslSupportMethod;
-  geoLocations?: string[]; // Is there a CDK element for country codes?
-  geoRestrictionType?: GeoRestrictionType;
-}
-
-export interface BaseOriginProps {
-  domainName: string,
-
-  connectionAttempts?: int,
-  connectionTimeoutSeconds?: int,
-  id?: string,
-  originCustomHeaders?: { [key: string]: string },
-  originPath?: string,
-}
-
-...
-```
-
 # Drawbacks
 
 > Why should we _not_ do this? Please consider:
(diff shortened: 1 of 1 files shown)


text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+
+```ts
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new Distribution(this, 'myDist', {
+  origin: Origin.fromWebsiteBucket(myBucket),
+  additionalOrigins: [Origin.fromLoadBalancerV2(myLoadBalancer, {
+    pathPattern: '/api/*',
+    allowedMethods: AllowedMethods.ALL,
+    forwardQueryString: true,
+  })];
```
skinny85, line 149: Comma here, not `;` 🤪

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+
+A distribution may have multiple origins in addition to the default origin; each additional origin must have (at least) one behavior to route requests
+to that origin. A common pattern might be to serve all static assets from an S3 bucket, but all dynamic content served from a web server. The
+following example shows how such a setup might be created:
+
+```ts
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new Distribution(this, 'myDist', {
+  origin: Origin.fromWebsiteBucket(myBucket),
+  additionalOrigins: [Origin.fromLoadBalancerV2(myLoadBalancer, {
```
skinny85, line 145: I would indent it like so:

```ts
  additionalOrigins: [
    cloudfront.Origin.fromLoadBalancerV2(myLoadBalancer, {
      // ...
    }),
  ],
});
```

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+});
+```
+
+You can specify an origin group for your CloudFront origin if, for example, you want to configure origin failover for scenarios when you need high
+availability. Use origin failover to designate a primary origin for CloudFront plus a second origin that CloudFront automatically switches to when the
+primary origin returns specific HTTP status code failure responses. An origin group can be created and specified as the primary (or additional) origin
+for the distribution.
+
+```ts
+const myOriginGroup = Origin.groupFromOrigins(
```
skinny85, line 159: How about `fromOriginGroup()`?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+
+Lambda@Edge functions can also be associated with additional behaviors, either at behavior creation or after the fact, either by attaching
+directly to the behavior, or to the distribution and referencing the behavior.
+
+```ts
+// Assigning at behavior creation.
+myOrigin.addBehavior('/images/*.jpg', {
+  viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+  defaultTtl: cdk.Duration.days(7),
+  functionAssociation: [{
```
skinny85, line 195: Super minor: I wonder if we should lean-in into the `Lambda@Ege` vocabulary here? How about `edgeLambda` instead of `functionAssociation`? Seems a little less "stiff" too me :)

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+myWebDistribution.origin.addBehavior('/images/*.jpg', {
+  viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+  defaultTtl: cdk.Duration.days(7),
+});
+```
+
+## Multiple Origins
+
+A distribution may have multiple origins in addition to the default origin; each additional origin must have (at least) one behavior to route requests
+to that origin. A common pattern might be to serve all static assets from an S3 bucket, but all dynamic content served from a web server. The
```
skinny85, line 138: > A distribution may have multiple origins in addition to the default origin; each additional origin must have (at least) one behavior to route requests to that origin.

I find this a little non-intuitive. Above, `behavior` was a separate thing, that you passed when creating `Distribution` next to origin:

```ts
new Distribution(this, 'Distr', {
  origin: ...
  behavior: {
    ...
  },
});
````

But below, you have `additionalOrigins` which don't specify any behaviors... So is this example correct, or is it still missing something?

File: text/0171-cloudfront-redesign.md [hunk 6]
```diff
@@ -0,0 +1,477 @@
...
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new Distribution(this, 'myDist', {
+  origin: Origin.fromWebsiteBucket(this, 'myOrigin', myBucket),
+  additionalOrigins: [Origin.fromLoadBalancerV2(this, 'myOrigin', myLoadBalancer, {
+    pathPattern: '/api/*',
+    allowedMethods: AllowedMethods.ALL,
+    forwardQueryString: true,
+  })];
+});
+myMultiOriginDistribution.origin.addBehavior('/api/errors/*', ...);
```
skinny85, line 287: This (the fact that you have behaviors both on Distribution and Origin) wrecks my poor little brain 😕. Any chance we can simplify it, so that you can only add Behavior to Origin?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+## Interaction with other L2s
+
+The existing @aws-cdk/aws-cloudfront module is used in three other modules of the CDK: (1) aws-route53-patterns, (2) aws-route53-targets, and
+(3) aws-s3-deployment.
+
+1. In aws-route53-patterns, the CloudFrontWebDistribution is used internally to the HttpsRedirect class; no CloudFront
+properties or classes are exposed to the consumer. This usage can be swapped out when the new L2s are ready without impact to customers.
+2. In aws-route53-targets, the CloudFrontTarget constructor takes a CloudFrontWebDistribution as the sole parameter. This usage could actually
+be replaced with an IDistribution, and then work for both Distribution and CloudFrontWebDistribution. I _believe_ this is a backwards-compatible
+change; please correct me if I'm wrong.
```
skinny85, line 303: I confirm I would consider this a backwards-compatible change as well :).

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+  }]
+});
+```
+
+**After:**
+
+```ts
+new Distribution(this, 'MyDistribution', {
+  origin: Origin.fromBucket(sourceBucket)
+});
```
skinny85, line 331: Love it - so much better!

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+
+Once created, the new L2s can be used by existing CDK developers for new use cases, or by converting their existing CloudFrontWebDistribution usages
+to the new Distribution resource.
+
+# Unresolved questions
+
+1. Are `fromBucket` and `fromWebsiteBucket` (potentially) redundant? There isn't enough information
+on the `IBucket` interface to determine if the bucket has been configured for static web hosting and
+we should treat as such. However, we could have an additional parameter to `fromBucket` trigger this
+behavior (e.g., `isConfiguredAsWebsite`?).
```
skinny85, line 454: I like `fromBucket` and `fromWebsiteBucket`, I don't have much problems with those being here.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+2. What are the advantage/disadvantages to making both `Origin` and `Behavior` extend `cdk.Construct` (or not)?
```
skinny85, line 455: I don't think there are many advantages. If we need `scope` to use there, we could always resort to our `bind()` pattern.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+3. Other examples of constructs which modify the underlying Cfn* resources post-creation? For example, calling `addBehavior` after
+construction alters the synthesized construct (not by adding new resources, but by changing the original one). Any best practices or
+patterns to be aware of here for the implementation?
```
skinny85, line 458: Lazy values! Example: https://github.com/aws/aws-cdk/blob/3b5640e9b1c41ec464d31c426f0dfadc4c296239/packages/%40aws-cdk/aws-codepipeline/lib/pipeline.ts#L256-L258

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+4. Should the current (CloudFrontWebDistribution) construct be marked as "stable" to indicate we won't be making future updates to it? Any other
```
skinny85, line 459: I'm 1,000% for this.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+One extension point for this redesign is building all-in-one "L3s" on top of the new L2s. One extremely common use case for CloudFront is
+to be used as a globally-distributed front on top of a S3 bucket configured for website hosting. One potential L3 for CloudFront would wrap
+this entire set of constructs into one easy-to-use construct. For example:
+
+```ts
+// Creates the hosted zone, S3 bucket, CloudFront distribution, ACM certificate, and wires everything together.
+new StaticWebsiteDistribution(this, 'SiteDistribution', {
+  domainName: 'www.example.com',
+  source: s3.Source.asset('static-website/'),
+});
```
skinny85, line 473: `@aws-ckd/aws-cloudfront-patterns`? 🙂

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,477 @@
...
+```
+
+Additional cache behaviors can be specified at creation, or added to the origin(s) after the initial creation. These additional cache behaviors enable
+customization for a specific set of resources based on a URL path pattern. For example, we can add a behavior to `myWebDistribution` to override the
+default time-to-live (TTL) for all of the images.
+
+```ts
+myWebDistribution.origin.addBehavior('/images/*.jpg', {
+  viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+  defaultTtl: cdk.Duration.days(7),
```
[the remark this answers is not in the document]
njlynch: Yes, there is also a minimumTtl and maximumTtl, both with their own meanings and used in different scenarios. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-cachebehavior.html#cfn-cloudfront-distribution-cachebehavior-defaultttl

text/0171-cloudfront-redesign.md
njlynch: The CloudFront documentation uses `LambdaFunctionAssocation`, which was my starting point, but I opted to drop the service name, the same way I treated buckets (`Origin.fromBucket`, not `Origin.fromS3Bucket`). I like `edgeLambdas` or `edgeFunctions`, but I was thinking of the design guidelines that state: "We prefer the terminology used by the official AWS service documentation over new terminology, even if you think it's not ideal". The CloudFront documentation talks about "Lambda@Edge" when speaking generically about the functionality, and "LambdaFunctionAssociation" when speaking about APIs or in the console experience. I'm open to feedback though; curious to see what @iliapolo thinks as well.

text/0171-cloudfront-redesign.md
njlynch: So this is … tricky. To create a distribution, there must be one origin and a default behavior; this default behavior must have a pathPattern of “*”. In order for any additional origins to receive traffic, there must be a new behavior with a pathPattern defined; so while not strictly required, it makes sense generally for each origin to have a behavior associated with it at creation. There are exceptions to this which I’ll get back to.

So the obvious thing would be to always have the `Origin.from*` methods create a behavior to associate with the origin, which would require a pathPattern. But for the default origin/behavior, those should be optional. So one option would be to alter the factory methods to look like:

```ts
static fromBucket(bucket: s3.IBucket, pathPattern?: string, cacheBehaviorProps?: BehaviorProps): Origin
```

For the “default” origin, the pathPattern and cacheBehaviorProps can be omitted, but for all others they would be required. This would have to be enforced by the Distribution when the origin is added (if an “additionalOrigin” with a “*” behavior is added, throw an error).

There are two caveats with this approach: (1) Behavior creation order can influence the origin routing (see detailed design and rationale for more details). Implicitly creating behaviors may make this approach more confusing. (2) Origin groups — with origin groups, only one behavior is needed which points to the primary origin, and the fallback origin is routed to only in case of failure of the primary. In that case, the auto-created behavior of the fallback origin would be ignored.

I’m going to call in @iliapolo for his feedback on this one as well.

skinny85: Before I chip in, let me ask 2 clarifying question.

I understand the default origin always has to have the `pathPattern` equal to `"*"` (so ideally, we don't want to burden the customer with providing it). But:

1. Can the default behavior have a custom cache behavior?
2. Can the non-default behaviors skip providing the cache behavior part?

njlynch: > 1. Can the default behavior have a custom cache behavior?

Yes, you can customize the default cache behavior the exact same as any other behavior.

> 2. Can the non-default origins skip providing the cache behavior part?

Yes, except for the 'pathPattern' part. So you can define a second origin and accept all of the default behaviors, and only specify what path redirects to the second origin. So this would be valid:

```ts
mySecondOrigin.addBehavior('/images/*.jpg');
```

This makes sense for the first behavior on any additional origin; however, adding a second behavior on an origin _without_ customization wouldn't make sense.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+distributions, origins, and behaviors, and has been in an "experimental" state for years. This proposal aims to bring a friendlier, more ergonic
+interface to the module, and advance the module to a GA-ready state.
+
+# Design Summary
+
+The approach will create a new top-level Construct (`Distribution`) to replace the existing `CloudFrontWebDistribution`, as well as new constructs
+to represent the other logical resources for a distribution (i.e., `Origin`, `Behavior`). The new L2s will be created in the same aws-cloudfront
+module and no changes will be made to the existing L2s to preserve the existing experience. Unlike the existing L2, the new L2s will feature a
+variety of convenience methods (e.g., `addBehavior`) to aid in the creation of the distribution, and provide several out-of-the-box defaults for
+building distributions off of other resources (e.g., buckets, load balanced services).
```
iliapolo, line 233: I would mention the fact that this new construct is somewhat optimized for the common use-case of creating a distribution with a single origin, reducing the cognitive load on the user.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+  eventType: EventType.VIEWER_REQUEST,
+});
+
+myDist.addFunctionAssociation({
+  functionVersion: myFunc.currentVersion,
+  eventType: EventType.ORIGIN_REQUEST,
+  behavior: myImagesBehavior,
+});
+```
+
```
iliapolo, line 218: Would be nice to see how a full example looks like, i.e a single origin distribution with a custom domain serving https and an edge lambda.

text/0171-cloudfront-redesign.md
iliapolo: Yeah, you discussion aligns with some of the comments I made along the way...

Its a thinker. 

I definitely think its worth coming up with a way to move behavior declaration to the origin level only. As far as I can tell, the problem is the `pathPattern`, which is fixed for the default origin, but configurable for the additional origins. Intuitively, the API I would expect is:

```typescript
const myWebDistribution = new cloudfront.Distribution(this, 'myDist', {
  origin: cloudfront.Origin.fromHTTPServer({
    domainName: 'www.example.com',
    originProtocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
    behavior: {
      allowedMethods: AllowedMethods.ALL,
      viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
    }
  }),
});
```

Which will use `*` as the `pathPattern`. Then, configure additional origins using:

```typescript
myWebDistribution.addOrigin('/api/*', Origin.fromBucket(myBucket, {
  allowedMethods: AllowedMethods.ALL,
  forwardQueryString: true,
}));
```

Basically, removing `pathPattern` from `BehaviorProps`.  Can this work? 

For the constructor, we can have:

```typescript
const myWebDistribution = new cloudfront.Distribution(this, 'myDist', {
  origin: cloudfront.Origin.fromHTTPServer({
    domainName: 'www.example.com',
    originProtocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
    behavior: {
      allowedMethods: AllowedMethods.ALL,
      viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
    }
  }),
  additionalOrigins: {
    '/api/*': Origin.fromBucket(myBucket, {allowedMethods: AllowedMethods.ALL, forwardQueryString: true});
  }
});
```

One thing I don't understand is, does it make sense to have multiple behavior on the same origin? As far as I can tell, a behavior is configured to a path, which cloudfront routes to some origin. Right?

text/0171-cloudfront-redesign.md
njlynch: > Notice it says AWS Service Documentation, not necessarily CloudFormation Documentation.

I wasn't referencing the CloudFormation docs. The CloudFront docs themselves use "LambdaFunctionAssociation" in their API docs and for the label in the console. That being said, given they also use Lambda@Edge in their user docs, I'm fine with 'edgeLambdas' (or 'edgeFunctions').

text/0171-cloudfront-redesign.md
njlynch: > One thing I don't understand is, does it make sense to have multiple behavior on the same origin? As far as I can tell, a behavior is configured to a path, which cloudfront routes to some origin. Right?

Yes, having multiple behaviors on the same origin does make sense. The behavior does route to an origin, but also has all of the controls around caching, headers, forwarded values, etc. You might have one behavior on an origin for longer-cached, mostly static content and another for more dynamic content that needs query parameters and cookies.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+An S3 bucket can be added as an origin. If the bucket is configured as a website endpoint, the distribution can use S3 redirects and S3 custom error
+documents.
+
+```ts
+import * as cloudfront from '@aws-cdk/aws-cloudfront';
+
+// Creates a distribution for a S3 bucket.
+const myBucket = new s3.Bucket(...);
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromBucket(myBucket),
```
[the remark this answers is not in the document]
njlynch: Taking that suggestion of syntactic sugar a bit further, would changing `Origin.fromBucket` to `Origin.forBucket` reduce some of the dissonance here? Also, I like top-level sugar (`Distribution.forBucket`); would you suggest that for just the most common (e.g., forBucket / forWebsiteBucket) or would we end up with one `Distribution.for*` for every `Origin.for*`?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+  domainName: 'www.example.com',
+  hostedZone,
+});
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromBucket(myBucket),
+  certificate: myCertificate,
+});
+```
+
+Note that in the above example the aliases are inferred from the certificate and do not need to be explicitly provided.
```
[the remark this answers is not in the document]
njlynch: They could; this might make sense if the certificate has a ton of SANs on it, but the distribution only needs to respond on a single name.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+Manager (ACM) service in the US East (N. Virginia) region; the certificate may either be created by ACM, or created elsewhere and imported into ACM.
+
+```ts
+const myCertificate = new certmgr.DnsValidatedCertificate(this, 'mySiteCert', {
+  domainName: 'www.example.com',
+  hostedZone,
+});
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromBucket(myBucket),
+  certificate: myCertificate,
```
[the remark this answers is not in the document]
njlynch: > As far as I can tell, the certificate and aliases property basically only have to do with http vs https.

_Mostly_ correct. You technically can add aliases to a distribution without having a cert to cover it, but then HTTPS won't work. The official documentation is a bit inconsistent on whether this is supported or not, but the general recommendation is that you must have a cert to cover any alternative names, in which case I could just drop the alias property altogether and only support adding names via certs. That being said, there are situations where a user might have a cert with 10 names on it, but only want the distribution to respond/use 5 of those names. In that case, some additional property somewhere would be needed to limit that down.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: > would changing Origin.fromBucket to Origin.forBucket reduce some of the dissonance here?

I considered that, but `Origin.forBucket` doesn't really feel right. `Distribution.forBucket` makes sense because the distribution is serving the bucket in a way. I prefer the `from` terminology on `Origin`, the are actually semantically related :)

Lets see where we end with on the `from` usage and we can change later on. 

> would you suggest that for just the most common (e.g., forBucket / forWebsiteBucket) or would we end up with one Distribution.for* for every Origin.for*?

I think just for the common ones, those that we need minimum configuration for. I'm ok with it also being just for bucket. Definitely not a must to do it for all.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+viewer, without provisioning or managing servers. Lambda@Edge functions are associated with a specific behavior and event type. Lambda@Edge can be
+used rewrite URLs, alter responses based on headers or cookies, or authorize requests based on headers or authorization tokens.
+
+By default, Lambda@Edge functions are attached to the default behavior:
+
+```ts
+const myFunc = new lambda.Function(...);
+const myDist = new cloudfront.Distribution(...);
+myDist.addLambdaFunctionAssociation({
+  functionVersion: myFunc.currentVersion,
```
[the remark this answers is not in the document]
njlynch: You must specify a version (and not $LATEST) when providing a function for Lambda@Edge (we need an `IVersion`, not an `IFunction`). We could change this to accept an (I)Function, and then use (or re-create) the Function.currentVersion method, but that would limit the customer from ever specifying a specific version (if needed). Thoughts on this trade-off?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+# Motivation
+
+The existing aws-cloudfront module doesn't adhere to standard naming convention, lacks convenience methods for more easily interacting with
+distributions, origins, and behaviors, and has been in an "experimental" state for years. This proposal aims to bring a friendlier, more ergonic
+interface to the module, and advance the module to a GA-ready state.
+
+# Design Summary
+
+The approach will create a new top-level Construct (`Distribution`) to replace the existing `CloudFrontWebDistribution`, as well as new constructs
+to represent the other logical resources for a distribution (i.e., `Origin`, `Behavior`). The new L2s will be created in the same aws-cloudfront
```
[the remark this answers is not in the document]
njlynch: I think it is confusing when we also use the v2 suffix for when a service actually releases a v2 (e.g., API Gateway, WAF, ELB). If CloudFront then released a v2, we'd need to create a v3 that mapped to CloudFront v2. It is a bit confusing to have both in the same package, but I think it's the best of the alternatives to provide new functionality "in-line" for customers without breaking existing consumers.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: I think what you described fits nicely to the API I proposed. Eventually I think using HTTP and HTTPS terminology is friendlier then using `aliases` and `certs` directly. We could just have `forHttps` accept an optional aliases array.

text/0171-cloudfront-redesign.md
iliapolo: > I wasn't referencing the CloudFormation docs. 

Whoops, sorry about that. I was looking at *CloudFront* and reading *CloudFormation* :)

Yeah, lets go with `edgeLambda`.

text/0171-cloudfront-redesign.md
iliapolo: Can you post an elaborate example of the current API with multiple origins and multiple behaviors per origin?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+const myWebsiteBucket = new s3.Bucket(...);
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromWebsiteBucket(myBucket),
+});
+```
+
+Both of the S3 Origin options will automatically create an origin access identity and grant it access to the underlying bucket. This can be used in
+conjunction with a bucket that is not public to require that your users access your content using CloudFront URLs and not S3 URLs directly.
+
+Origins can also be created from other resources (e.g., load balancers, API gateways), or from any accessible HTTP server.
```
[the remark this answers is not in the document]
njlynch: I'm ok with the bucket/http separation. Past that, there's one small sentence per header, which is overkill.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+
+A distribution may have multiple origins in addition to the default origin; each additional origin must have (at least) one behavior to route requests
+to that origin. A common pattern might be to serve all static assets from an S3 bucket, but all dynamic content served from a web server. The
+following example shows how such a setup might be created:
+
+```ts
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromWebsiteBucket(myBucket),
+  additionalOrigins: [
```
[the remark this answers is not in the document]
njlynch: > What is the order of evaluation of these expressions?

It's actually interesting you call that out here, because in this example it doesn't matter. There is a default origin with a default behavior created, and an additional origin with a behavior on '/api/*' created. You could create these in any order you wanted, and you'd end with the same result.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: 👍

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: Wait, does the default origin path map to `/` or `/*`?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+```ts
+// Creates a distribution for an application load balancer.
+const myLoadBalancer = new elbv2.ApplicationLoadBalancer(...);
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromLoadBalancerV2(myLoadBalancer),
+});
+
+// Creates a distribution for an HTTP server.
+new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromHTTPServer({
```
rix0rrr, line 72: Nit: no uppercase abbreviations. So it will be `fromHttpServer`

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
rix0rrr: I think I'd prefer the simple and obvious:

```ts
  domainNames?: string[];
  certificate?: Certificate;
```

Over more a more complex, non-obvious `Domain` case class.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+  origin: cloudfront.Origin.fromBucket(myBucket),
+  certificate: myCertificate,
+});
+```
+
+Note that in the above example the aliases are inferred from the certificate and do not need to be explicitly provided.
+
+## Caching Behaviors
+
+Each distribution has a default cache behavior which applies to all requests to that distribution; additional cache behaviors may be specified for a
```
rix0rrr, line 110: Is this about caching or not? If not, maybe just calling it "behaviors" makes more sense (maybe indicate that upstream uses "cache behaviors" but that this no longer covers the cases).

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+to that origin. A common pattern might be to serve all static assets from an S3 bucket, but all dynamic content served from a web server. The
+following example shows how such a setup might be created:
+
+```ts
+const myWebsiteBucket = new s3.Bucket(...);
+const myMultiOriginDistribution = new cloudfront.Distribution(this, 'myDist', {
+  origin: cloudfront.Origin.fromWebsiteBucket(myBucket),
+  additionalOrigins: [
+    cloudfront.Origin.fromLoadBalancerV2(myLoadBalancer, {
+      pathPattern: '/api/*',
```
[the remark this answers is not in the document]
rix0rrr: I think rather than double down on the distinction between `Origin` and `AdditionalOrigin` (which is just going to explode cases), can we do the reverse? Treat all origins syntactically the same:

```
origins: Origin[]
```

But do a runtime check to make sure the array contains exactly one `Origin` without a `pathPattern`?

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+});
+```
+
+You can specify an origin group for your CloudFront origin if, for example, you want to configure origin failover for scenarios when you need high
+availability. Use origin failover to designate a primary origin for CloudFront plus a second origin that CloudFront automatically switches to when the
+primary origin returns specific HTTP status code failure responses. An origin group can be created and specified as the primary (or additional) origin
+for the distribution.
+
+```ts
+const myOriginGroup = cloudfront.Origin.fromOriginGroup({
```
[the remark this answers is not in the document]
rix0rrr: I can think of a couple of other syntaxes:

```
new Distribution(/* ... */, {
  origin: Origin.fromLoadBalancerV2(lb).withFallback(Origin.fromBucket(bucket));
});

new Distribution(/* ... */, {
  origin: Origin.fromLoadBalancerV2(lb, {
    fallback: Origin.fromBucket(bucket)
  }),
});
```

Just throwing that out there. I'm not really opposed to the original proposal, except maybe that I question the value of needing to know Origin Groups as a concept.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+    cloudfront.Origin.fromLoadBalancerV2(myLoadBalancer, {
+      pathPattern: '/api/*',
+      allowedMethods: AllowedMethods.ALL,
+      forwardQueryString: true,
+    }),
+  ],
+});
+```
+
+You can specify an origin group for your CloudFront origin if, for example, you want to configure origin failover for scenarios when you need high
```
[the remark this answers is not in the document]
rix0rrr: Seems that this mechanism has a singular purpose (but correct me if I'm wrong).

Isn't

### Specifying a failover origin

A more useful header if you're scanning the document outline?

But it kinda depends if Origin Groups are worthwhile know about for any other reason. Compromise:

### Specifying a failover origin (origin groups)

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,484 @@
...
+  eventType: EventType.VIEWER_REQUEST,
+});
+```
+
+---
+
+# Motivation
+
+The existing aws-cloudfront module doesn't adhere to standard naming convention, lacks convenience methods for more easily interacting with
+distributions, origins, and behaviors, and has been in an "experimental" state for years. This proposal aims to bring a friendlier, more ergonic
```
rix0rrr, line 228:
```suggestion
distributions, origins, and behaviors, and has been in an "experimental" state for years. This proposal aims to bring a friendlier, more ergonomic
```

text/0171-cloudfront-redesign.md
rix0rrr: I'd also like to see a complex example.

To reiterate my previous comment, I think I'd like to see the distinction between primary and additional origins disappear.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
njlynch: The default behavior has a path pattern of '*' (i.e., a catch-all). All paths not explicitly matching another behavior go to the default behavior. In the above case, no matter what order you define these objects in, all calls to /api/* will go to the additional origin, and all other calls will go to the default.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
njlynch: @rix0rrr  - I actually started out with the flat list of origins, but @eladb pointed out that if 90% of the time the user only configured a single origin, we should make that friendlier and not require an array.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: >  except maybe that I question the value of needing to know Origin Groups as a concept.

@rix0rrr Thats an interesting point. And im a big fan of fluent API's. We just need to be careful not to cause confusion in case users are already mainly familiar with this concept.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
eladb: Yes, we usually try to optimize for the common use case. I like the current design where the primary original is modeled as `origin` and is required.  and also has some other requirements that can be enforced at compile time.

Another way to look at it is that in practice CloudFront itself has a concept of the primary origin (whether semantically defines or not)

text/0171-cloudfront-redesign.md
njlynch: This is already present. See "Use case #3" in the Comparison section.

text/0171-cloudfront-redesign.md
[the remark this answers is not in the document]
iliapolo: It might also be worth renaming ‘origin’ to ‘primaryOrigin’

text/0171-cloudfront-redesign.md
njlynch: A complex example already exists in the "Comparison" section (Use case 3). 

After thinking about this a bit, I think the simplest way to resolve this is to remove the `behavior` property from the Distribution to model the default behavior. This aligns with @skinny85 's original comment of making it so behaviors can only be attached to Origins. I've updated the RFC and examples to match this.

*Implementation notes:* What this means is that all of the `Origin.from*` methods accept the BehaviorProps (as they already did) as an optional argument. If no BehaviorProps are provided, then the Origin will originally not be assigned a Behavior. At Distribution creation, a default behavior will be created for the default origin, and an error will be thrown for any other origin without a behavior (as no sane default exists). This is a reasonably minor change to the design -- as evidenced by how easy it was to make the change to the RFC -- but I think resolves the open issues here. Lazy-creating/binding the behaviors solves my concern about origin groups having redundant behaviors, and ordering isn't a strict concern because the default behavior is always last anyway.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,489 @@
...
+
+The following is an incomplete, but representative, listing of the API:
+
+```ts
+class Distribution extends BaseDistribution {
+  static fromArn(scope: Construct, id: string, distributionArn: string): IDistribution;
+  static fromAttributes(scope: Construct, id: string, distributionArn: string): IDistribution;
+
+  static forBucket(scope: Construct, id: string, bucket: IBucket): IDistribution;
+  static forWebsiteBucket(scope: Construct, id: string, bucket: IBucket): IDistribution;
```
skinny85, line 257: Isn't this a problem? If `forBucket` and `forWebiteBucket` returns an `IDistribution`, that interface doesn't have the `addOrigin` method, so you can't add additional origins to a Distribution created that way.

text/0171-cloudfront-redesign.md
skinny85: > I'd also like to see a complex example.
> 
> To reiterate my previous comment, I think I'd like to see the distinction between primary and additional origins disappear.

But there is a very important distinction between primary and secondary origins, right?

A secondary origin must have a behavior associated with it with some `pathPattern`. While a primary origin has a default behavior (= with `pathPatter` = `*`) associated with it automatically.

text/0171-cloudfront-redesign.md
njlynch: Yup, bad copy/paste. Those should definitely both return a Distribution, not IDistribution.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,495 @@
...
+complex interface with diminishing benefits. However, feedback is welcome on a more elegant way to give users control of behavior ordering.
+
+# Adoption Strategy
+
+Once created, the new L2s can be used by existing CDK developers for new use cases, or by converting their existing CloudFrontWebDistribution usages
+to the new cloudfront.Distribution resource.
+
+# Unresolved questions
+
+1. What level of breaking changes are acceptable for the existing `IDistribution` and `CloudFrontWebDistribution` resources? Notably, the
```
iliapolo, line 472: Have we dropped the approach of publishing this as a new module? If so - why? I feel like putting both new and old in the same module is going to confuse users.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,495 @@
...
+`IDistribution` interface should extend `IResource`, and `CloudFrontWebDistribution` changed from extending `Construct` to `Resource`. Is this
+worth it, given the breaking changes to existing consumers? The alternative is to leave both `IDistribution` and `CloudFrontWebDistribution` as-is,
+and have `Distribution` directly extend `Resource`.
+2. Related to the above, should the current (CloudFrontWebDistribution) construct be marked as "stable" to indicate we won't be making future
```
iliapolo, line 476: I feel like we should mark it as stable with a nice banner that says we are actively working on a revamped version.

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Origin Groups
+
+You can specify an origin group for your CloudFront origin if, for example, you want to configure origin failover for scenarios when you need high
+availability. Use origin failover to designate a primary origin for CloudFront plus a second origin that CloudFront automatically switches to when the
+primary origin returns specific HTTP status code failure responses. An origin group can be created and specified as the primary (or additional) origin
+for the distribution.
+
+```ts
+const myOriginGroup = cloudfront.Origin.fromOriginGroup({
```
rix0rrr, line 173: Also, I think `Origin.fromOriginGroup()` is stretching the definition of `from`. It's not like an "origin group" already exists and you're turning it into an origin here.

It feels much more like you're making a new Origin Group, in which case I feel `from` is wrong. 

OTOH, it is kinda discoverable and we generally use `from` for factory functions. Ttssschhh naming is hard!

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,495 @@
...
+  origin: cloudfront.Origin.fromLoadBalancerV2(lbFargateService.loadBalancer, {
+    allowedMethods: AllowedMethods.ALL,
+    forwardQueryString: true,
+    edgeFunctions: [{
+      function: myFunctionVersion,
+      eventType: EventType.ORIGIN_RESPONSE,
+    }],
+  }),
+});
+dist.addOrigin(cloudfront.Origin.fromBucket(sourceBucket), { pathPattern: 'static/*' });
```
iliapolo, line 436: Is the `pathPattern` required at this point?

njlynch: Yes. That example also has a mis-placed parenthesis. It should be:
```suggestion
dist.addOrigin(cloudfront.Origin.fromBucket(sourceBucket, { pathPattern: 'static/*' });
```

text/0171-cloudfront-redesign.md
```diff
@@ -0,0 +1,495 @@
...
+
+```ts
+const dist = new cloudfront.Distribution(this, 'MyDistribution', {
+  origin: cloudfront.Origin.fromLoadBalancerV2(lbFargateService.loadBalancer, {
+    allowedMethods: AllowedMethods.ALL,
+    forwardQueryString: true,
+    edgeFunctions: [{
+      function: myFunctionVersion,
+      eventType: EventType.ORIGIN_RESPONSE,
+    }],
```
iliapolo, line 433: Is this just an inline of the `behavior` from the previous way?

wcauchois:
I like this design! I'm currently using the cloudfront module and I feel like the existing ergonomics leave tons of room for improvement :)

Apologies if this is a naive question, but my understanding is that this is only the RFC - would there be a place where I could track implementation/release of these constructs? Thanks!

njlynch:
@wcauchois - Thanks for the interest and feedback!

We are proceeding forward with the implementation, and using this milestone to track the initial "developer preview" launch: https://github.com/aws/aws-cdk/milestone/5.

Issue #193, opened 2020-07-06
Fixing of type unions


RomainMuller:
|PR|Champion|
|--|--------|
|#194 | @RomainMuller |

## Description

Usage of TypeScript type unions has proven to be a source of problems when representing APIs in statically typed languages, such as Java and C#. This proposal is about fixing type unions to address these developer experience issues.

This is a candidate change for [v2](https://github.com/aws/aws-cdk-rfcs/issues/79)

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

mrgrain:
There is a significant proposal, but no current owner. Closing as stale.

PR #194, opened 2020-07-07
RFC: 193 Fixing type unions

RomainMuller:
* Tracking issue: #193
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/feature/remove-type-unions/text/0193-fixing-type-unions.md )

---


text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,484 @@
...
+features offered by the AWS CDK.
+
+An option is to generate specific wrapper types that effectively encapsulate the _type union_, however such wrapper
+types must then be used for _all_ properties in order to avoid incurring an API breaking change if some _CloudFormation
+Resource_ schema adds support for alternate value types on a given property (this is possible since CloudFormation
+describes resource schemas using [JSON Schema]). This would result in a significant deterioration of the ergonomics of
+the _CloudFormation Resource_ classes, but provides an API that is safe to use from any programming language.
+
+[escape hatches]: https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html
+[json schema]: https://json-schema.org
```
skinny85, line 177: I've read this several times, but I still don't understand what the proposed solution is here.

Can you show:

a) What the new generated code for Cfn* classes will look like in TypeScript.
b) What will the experience for the Cfn* classes look like for other languages (probably just Java and Python are representative enough to cover 99% of the cases).

skinny85: Feel free to use something like `CfnBucket` as a reference.

RomainMuller: Fair enough, will illustrate stuff a lot more.

The fact a bunch of places lists options and doesn't quite make recommendations was intentional at this point (I needed a base to discuss with several folks before I'd narrow it down). But I wholeheartedly agree it does not help understanding the proposal, so I'll revise.

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+      rules: ArrayOfRulesPropertyValue.immediate([
+        RulePropertyValue.immediate({ expirationInDays: 7 }),
+      ]),
+    }),
+  });
+  ```
+
+[escape hatches]: https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html
+[json schema]: https://json-schema.org
+
```
skinny85, line 269: I have another alternative.

What if we forced union types in JSII-compatible TypeScript to go through a type alias, with a JSII-specific annotation? Something like:


```ts
@JsiiTypeUnion
type LifecycleConfigurationPropertyValue = LifecycleConfigurationProperty | IResolvable;
```

And the type of `lifecycleConfiguration` in `CfnBucketProps` will be `readonly lifecycleConfiguration: LifecycleConfigurationPropertyValue`. Using type unions without an alias would be a JSII error.

Then, for Java, we would generate the "union-like" pattern class like you showed above, but the TypeScript / JavaScript experience would be untouched.

Is this feasible?

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+_Type unions_ is a feature of the TypeScript language that allows typing values as "one of several candidate types".
+This is achieved by specifying a `|`-delimited list of candidate types (e.g: `Foo | Bar`). While this feature is common
+in dynamic languages (**TypeScript**, **Python**, ...), it is almost always absent from statically typed languages
+(**Java**, **C#**, ...). In those languages, the impossibility to represent _type unions_ natively results in degraded
+developer experience.
+
+This RFC proposes to remove all usage of _type unions_ from the AWS CDK, and eventually the removal of the feature from
+the `jsii` type model.
+
+# README
```
eladb, line 19: Which README is this? Is this jsii readme? CDK?

Seems like this should probably be a "Release notes" type working backwards an not a README

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+easily. In **TypeScript** and other dynamic languages (such as **Python**), the `Lazy.asAny` method can be used to wrap
+an `IResolvable` and evade the type checker. This approach is however no workable in **C#**, **Go** and **Java** (and
+likely other statically typed languages we may wish to support in the future). For example the `s3.CfnBucket` accepts
+(and exposes) properties such as:
+```ts
+lifecycleConfiguration: CfnBucket.LifecycleConfigurationProperty | cdk.IResolvable | undefined;
+```
+
+#### Recommended Approach
+
```
eladb, line 190: I would lead with the more common TypeScript use case (L2s) and show how this can be resolved there.

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+#### API Extension
+
+The `@jsii/kernel` API could be extended so that instances can be tagged using unions, instead of necessarily narrowing
+down to a single type. This way, the kernel would not have to make any decision beyond the statically available context
+of a serialization.
+
+This requires more extensive work to be performed in each individual language runtime library.
+
+### Language-specific union modeling
```
eladb, line 357: Is this needed here really? Feels like a future RFC for *adding support* for type unions, no?

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+
+This feature is a candidate for rollout as part of [AWS CDK v2]. Since the removal of type unions will incur significant
+breaking in the API of the generated CloudFormation Resource classes, it naturally requires a major version bump.
+
+A migration guide will be provided on release of [AWS CDK v2] that explains how to migrate from the old API patterns
+over to the new ones, although the majority of users may not have to change anything in their code for it to continue
+working.
+
+[aws cdk v2]: https://github.com/aws/aws-cdk-rfcs/issues/79
+
```
eladb, line 464: Is there anything we can do over the 1.x branch to "prepare" for this change?

text/0193-removal-of-type-unions.md
eladb: I think that's definitely worth exploring as an alternative

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,484 @@
...
+The recommended substitution pattern for APIs where _type unions_ would be the natural candidate in dynamic languages is
+the introduction of a wrapper type with a private constructor and static factories for each of the candidate types:
+
+```ts
+// recommended replacement for `Foo | Bar` (the name is illustrative only):
+export class FooOrBar {
+  /**
+   * Static factory method for the `Foo` variant of the union.
+   */
+  public static ofFoo(value: Foo): FooOrBar {
```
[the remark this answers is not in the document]
RomainMuller: I've seen `from` used for "import" factories. I think this is a different use-case entirely, so might warrant a different name. That said, names and such are 100% illustrative.

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,484 @@
...
+the `jsii` type model.
+
+# README
+
+Use of TypeScript _type unions_ is prohibited by `jsii` due to the lack of support of this language feature in several
+of the target languages supported (or planned to be supported) by `jsii`: **C#**, **Go**, **Java**, and probably more
+languages to come. Proper support of _type unions_ in those languages requires generating code that produces a degraded
+developer experience in those languages.
+
+The recommended substitution pattern for APIs where _type unions_ would be the natural candidate in dynamic languages is
```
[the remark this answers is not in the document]
RomainMuller: I kind of feel that a proposal to remove a feature should include enough evidence of the feasibility of an implementation that does not use said feature?

text/0193-removal-of-type-unions.md
RomainMuller: Yeah that does make sense.

text/0193-removal-of-type-unions.md
RomainMuller: Alternatives to removing type unions include **not** removing type unions, and only fixing them up so the current problems no longer exist.

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+collaborate on the finalized configuration of a resource.
+
+### Where existing APIs already support the use-case in a safe manner
+
+All _type unions_ where late-bound values can be represented in the literal type of the value (`string`, `number` and
+`any`) can be simplified to that literal type, since the required APIs already exist and the migration path for those
+use-cases is already very well defined: using `Lazy` factory methods to manually wrap `IResolvable` instances:
+- Current instances of:
+  ```ts
+  public propertyName: string | IResolvable;
```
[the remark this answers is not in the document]
RomainMuller: You are correct, this was already done previously.

text/0193-removal-of-type-unions.md
RomainMuller: A `@decorator` **cannot** be applied to entities that have no materialization in Javascript (the `@decorator` is applied at runtime, not compile time). But that isn't limiting this (instead of annotating it with a `@decorator` it should simply be `export`ed to achieve the desired intent -- providing a name for the union). This suggestion is already documented under *Alternatives Considered*:

https://github.com/aws/aws-cdk-rfcs/blob/54295bf96daf868967a235edd96333bc084d9ef8/text/0193-removal-of-type-unions.md#L431-L443

---

This definitely allows generating union-like types in target languages where this is necessary.

However, that only solves the code generation problem, but does not solve the kernel ambiguity (which is the cause of the currently observed bugs caused by the presence of type unions): how can we reliably offer inspection methods that can conclusively tell overlapping interface (or struct) instances apart, so as to allow users to author *safe* code.

This would need to be complemented by enhancements in the *jsii kernel* to support such types, understanding that it is not possible to distinguish homonym methods with different argument signatures (since those are not available in the javascript runtime) without introducing additional code (possibly via a TypeScript transform - but this is going to be quite an endeavor).

RomainMuller:
@skinny85 thanks for your feedback! I'm providing some answers below...
---

> In my opinion, forcing casting things to `any` and using parameter overrides is too big of a drop in the customer experience of the CDK. One of the foundational promises that we make in the CDK is that everything that is achievable in CloudFormation is achievable also in the CDK (albeit, maybe only on the L1 level). Resorting to type hacks to implement CloudFormation idioms seems like is pretty much breaking that promise (yes, it's technically possible, but the experience is so bad, and so poorly discoverable, that I don't feel comfortable in making that change).

I agree this provides arguments in favor of the argument that claims "TypeScript is the only first class citizen language in CDK", and so far was willing to take this as an interim step. I only reluctantly got to this recommendation from lack of evidence that a more complex solution would be necessary, and I'm absolutely willing to reconsider the recommendation as concerns are raised over the general safety/usability of the current proposal.

---

> I would bet if I did a similar search for TypeScript, I would get ten times as many results. So this capability is actively being used by our customers.

Well *TypeScript* does not suffer from requiring `as any` usage, so the fact this feature is widely leveraged on the user side in *TypeScript* isn't much of an argument against the proposal. However I worry about the consequences of using this approach in the *AWS Construct Library*, as this could made it impossible to inspect property values (for example, in an `Aspect`, though a native method override) from *Java* and *C#* (since the static type would be declared to `CfnBucket.LifecyclePolicyProperty`, but the actual value returned would be of type `IResolvable` -- leading to a `ClassCastException`).

---

> It's also very important to our future plan with `cloudformation-include`. if we want to write methods like `Bucket.fromCfnBucket(cfnBucket: CfnBucket): IBucket`, we need type safety that the properties of the `CfnBucket` are actually what the types say they are, not some hidden `IResolvable`s. If that's the case, writing those methods will be, for all practical intents and purposes, impossible.

How is it different from the current situation where those methods are typed as unions of `IResolvable` and some complex type? If such a `fromCfnBucket` can be written against the current types, the exact same code can be leveraged with values shoe-horned using an `as any`. I agree this makes it more difficult to write such code, and it might be a lit more error prone... But **impossible** sounds like an overstatement.

---

> I understand that this fixes some problems in JSII, but, in my opinion, this is putting JSII before the customer experience of CDK users, which is not the correct tradeoff.

This is not about fixing problems *in jsii*. The reality today is this is fixing problems **in CDK** due to bugs *in jsii* that are inherent to type unions, and which cannot be fixed in the current state of affairs.

---

> I'm open to discussing the "Alternative Approach". I've also posted my own, second, alternative, but I'm not sure how feasible it is. Would love any feedback on that.

As I commented, this alternative is already discussed in the document, and likely would require a complete overhaul of the `jsii` compiler, which might not fit in our desired timeline (although if that is the right thing to do, then we probably should bend the timeline and not the other way around).

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+  bucket.lifecycleConfiguration = value as any;
+  ```
+
+- Developers in **Java** (and other statically typed languages) would instead use:
+  ```java
+  CfnBucket bucket;
+  IResolvable resolvable;
+  // ...
+  bucket.addPropertyOverride('LifecycleConfiguration', resolvable);
+  ```
```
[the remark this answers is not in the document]
nija-at: Given that this issue is only with L1s and not the L2s are easy to fix, how many different types (for lack of a better word) of type unions exist in our current L1s? i.e., are they all union-ed with `IResolvable` vs unions are across significantly different types (`number | string`, etc.) vs unions of scalar and vector (`string | string[]`)?

Given CF has a certain level of types inherently, there can't be **that many** variations, right?

I'm wondering if we can also change our L1 generation so that this type erasure is not lost. Taking your example as an illustration -

```ts
lifecycleConfiguration: CfnBucket.LifecycleConfigurationProperty;
lifecycleConfigurationLazy: cdk.IResolvable;
```

skinny85:
> > I would bet if I did a similar search for TypeScript, I would get ten times as many results. So this capability is actively being used by our customers.
> 
> Well _TypeScript_ does not suffer from requiring `as any` usage, so the fact this feature is widely leveraged on the user side in _TypeScript_ isn't much of an argument against the proposal. 

Sorry, I don't understand this comment. In the same message, you also write:

> If such a `fromCfnBucket` can be written against the current types, the exact same code can be leveraged with values shoe-horned using an `as any`.

Your current proposal contains the following TypeScript example:

> #### Recommended Approach
> Where **TypeScript** developers could use:
>  ```ts
>  let cfnBucket: CfnBucket;
>  let value: IResolvable;
 > // ...
 > bucket.lifecycleConfiguration = value as any;
 > ```

So TypeScript definitely seems to suffer from `as any` usage, or I'm really not understanding something fundamental about the proposal here.

text/0193-removal-of-type-unions.md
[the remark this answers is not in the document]
eladb: I see. Perhaps make this section a bit more concise (e.g reference docs or a GitHub about implementing union-like classes) because it buries the important stuff.

text/0193-removal-of-type-unions.md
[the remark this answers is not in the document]
eladb: Nope, we use `fromX` in almost all union-like classes. The import factories are a different pattern

text/0193-removal-of-type-unions.md
eladb: Okay

text/0193-removal-of-type-unions.md
```diff
@@ -0,0 +1,575 @@
...
+  compilation failure, since they share the same _erasure_ (e.g: `void call(List)`) and a _Java Virtual Machine_ would
+  be unable to determine which implementation to dynamically dispatch to.
+
+Additionally, certain static languages (such as **Go**) do not support overloading at all, rendering this approach
+entirely unsuitable.
+
+# Design Summary
+
+Removal of _type unions_ from the AWS CDK requires designing new APIs for those places where they are in use. This
+includes two kinds of APIs:
```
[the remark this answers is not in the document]
RomainMuller: `| undefined` is actually not a type union in *jsii*. It's only a `null`-able type.

RomainMuller:
> > > I would bet if I did a similar search for TypeScript, I would get ten times as many results. So this capability is actively being used by our customers.
> > 
> > 
> > Well _TypeScript_ does not suffer from requiring `as any` usage, so the fact this feature is widely leveraged on the user side in _TypeScript_ isn't much of an argument against the proposal.
> 
> Sorry, I don't understand this comment. In the same message, you also write:
> 
> > If such a `fromCfnBucket` can be written against the current types, the exact same code can be leveraged with values shoe-horned using an `as any`.
> 
> Your current proposal contains the following TypeScript example:
> 
> > #### Recommended Approach
> > Where **TypeScript** developers could use:
> > ```ts
> > let cfnBucket: CfnBucket;
> > let value: IResolvable;
> > // ...
> > bucket.lifecycleConfiguration = value as any;
> > ```
> 
> So TypeScript definitely seems to suffer from `as any` usage, or I'm really not understanding something fundamental about the proposal here.

What I mean is that the usage of `as any` is not "harmful" in **TypeScript**: you can use this language feature to cheat with the type system, in ways that are purely impossible in statically typed languages (there is no way I'll pass `(Object)foo` to a **Java** method that accepts a `Bar`-typed parameter).

However, I discovered scenarios that make certain usage patterns blow up in statically typed languages if something in **TypeScript** (or **Javascript**) used `as any` to force a value in... Since it makes that value impossible to read back from **Java** without getting a `ClassCastException`. I'm working on an updated version of the document.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+  overridden from a class in another language)
+
+On the other hand, a solution may involve generating code that is substantially different from one language to the
+other. This means statically typed languages can leverage additional generated code in order to satisfy the requirements
+mentioned above without introducing unwanted boilerplate in dynamic languages that do not need that.
+
+# Design Summary
+
+As this feature is a candidate for inclusion in [AWS CDK v2], the design should maximize the amount of changes that can
+be rolled out withon CDK v1 without breaking existing customers. The solution is hence described in two main milestones,
```
skinny85, line 299: Typo ('withon').

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+  export type InvalidUnion = cdk.Construct | constructs.Construct;
+  ```
+- no candidate type can have a name starting with `ListOf` or `MapOf` followed by an upper-case letter:
+  ```ts
+  // Invalid: ListOfFoo would collide with generated union-like accessors for `Foo[]`
+  export type InvalidUnion = ListOfFoo | Bar;
+  ```
+
+In order to preserve backwards-compatbility, and to retain ability to use this new feature against the current CDK v1
+codebase, the `jsii` compiler will not **require** that _type unions_ are systematically aliased. Similarly,
```
skinny85, line 346: 1. The bolding is a little weird here... maybe **not require** makes more sense?
2. I think the word "systematically" can be safely removed from this sentence without changing its meaning.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+
+- In **Java**:
+
+  ```java
+  // UnionType.java
+  public final class UnionType {
+    public static UnionType fromFoo(Foo value) { /* ... */ }
+    public static UnionType fromMapOfBar(Map<String, Bar> value) { /* ... */ }
+    public static UnionType fromListOfBaz(Baz... value) { /* ... */ }
+    public static UnionType fromListOfBaz(List<Baz> value) { /* ... */ }
```
skinny85, line 397: It's probably worth to mention that `UnionType` will have a private constructor, so that it can be only created through the static factory methods.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+
+- In **Go**:
+
+  > :construction: TBD
+
+- In **Java**:
+
+  ```java
+  // UnionType.java
+  public final class UnionType {
```
skinny85, line 393: Are you sure you want this to be final? You can also use a pattern similar to what Kotlin does for its data classes:

```java
public class UnionType {
    public static UnionType fromFoo(Foo value) { /* ... */ }
    public static UnionType fromMapOfBar(Map<String, Bar> value) { /* ... */ }
    public static UnionType fromListOfBaz(Baz... value) { /* ... */ }
    public static UnionType fromListOfBaz(List<Baz> value) { /* ... */ }

    private UnionType(...) { ... }

    // works, because inner classes have access to the private constructor of UnionType
    private static final class FooUnionType extends UnionType { ... }
    private static final class MapOfBarUnionType extends UnionType { ... }
    private static final class ListOfBazUnionType extends UnionType { ... }
}
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+This aliasing offers multiple benefits for library authors and users alike:
+
+- _type unions_ are no longer repeated in multiple locations, so that new candidates can be added in a single location
+- better code can be generated to support _type union_ APIs in languages without native support for type unions (most
+  statically typed languages, such as **C#**, **Go**, **Java**, ...)
+- safer code can be authored, since _type unions_ no longer have to be frowned upon: all unsafe usage of
+  **TypeScript**'s `as any` to force a `Lazy` value can be replaced with a safer API using a _type union_ with
+  `IResolvable`
+
+An important think to keep in mind is that changing the type of a parameter to a _type union_ is a breaking change for
```
ericzbeard, line 43: thing

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+```
+
+This makes it more difficult for **Java** developers to write _correct_ and _safe_ code, since static type analysis
+cannot help identifying issues. Mistakes will typically result in a `ClassCastException` being thrown. This is also a
+degradation in the documentation experience, since the type signature of methods is no longer accurate.
+
+### Type-safe overloads cannot always be generated
+
+Since most statically typed languages support _overloads_ (although **Go** does not). This means we are often able to
+leverage overloads to provide type-safe entry points, for example in the **Java** fluent builders:
```
ericzbeard, line 165: Since most statically typed languages support _overloads_ (although **Go** does not), we are often able to
leverage overloads to provide type-safe entry points, for example in the **Java** fluent builders:

(Combine the sentences. Also is "fluent" correct?)

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+}
+```
+
+A similar pattern is used for methods, where an overload must be generated for each possible combination of _type union_
+candidates. This can be problematic if a method accepts multiple _union-typed_ parameters, as this can lead to an
+explosion of overloads (although this is not a common use-case and is mostly a stylistic concern).
+
+Certain statically typed languages are however unable to express certain combinations of _type union_ candidates because
+of the way _generics_ (while _generics_ are not supported in _jsii_ libraries, they are typically used in code
+generation for emitting _collection_ types such as `List<T>` and `Map<String, T>`) are _reified_.
```
ericzbeard, line 196: Might be good to define reify.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+    }
+    // ...
+  }
+  // ...
+}
+```
+
+### Bugs in the Kernel
+
+Currently, the _jsii Kernel_ protocol imposes each object instance is annotated with exactly one type, and this type is
```
ericzbeard, line 258: imposes?

text/0193-fixing-type-unions.md
eladb, line 43:
```suggestion
An important thing to keep in mind is that changing the type of a parameter to a _type union_ is a breaking change for
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+export type ShinyUnion = Foo | Bar | Baz;
+
+export interface FancyProps {
+  readonly union: ShinyUnion;
+}
+```
+
+This aliasing offers multiple benefits for library authors and users alike:
+
+- _type unions_ are no longer repeated in multiple locations, so that new candidates can be added in a single location
```
eladb, line 36:
```suggestion
- _Type unions_ are no longer repeated in multiple locations, so that new candidates can be added in a single location
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+- better code can be generated to support _type union_ APIs in languages without native support for type unions (most
```
eladb, line 37:
```suggestion
- Better code can be generated to support _type union_ APIs in languages without native support for type unions (most
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+  statically typed languages, such as **C#**, **Go**, **Java**, ...)
+- safer code can be authored, since _type unions_ no longer have to be frowned upon: all unsafe usage of
```
eladb, line 39:
```suggestion
- Safer code can be authored, since _type unions_ no longer have to be frowned upon: all unsafe usage of
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+This is achieved by specifying a `|`-delimited list of candidate types (e.g: `Foo | Bar`). While this feature is common
+in dynamic languages (**TypeScript**, **Python**, ...), it is almost always absent from statically typed languages
+(**Java**, **C#**, ...). In those languages, the impossibility to represent _type unions_ natively results in degraded
+developer experience due to the reduced fidelity of the generated typings.
+
+This RFC proposes to introduce new features in the _jsii_ compiler to support generation of additional code that will
+allow statically typed languages to benefit from the flexibility of type unions without having to endure the experience
+degradations they are currently exposed to.
+
+# Release Notes
```
eladb, line 20: This is the jsii RELEASE NOTES, please also add the CDK's "BREAKING CHANGE" section for 2.x which describes for e.g. .NET and Java users what's changed for them (before/after)

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+
+An important think to keep in mind is that changing the type of a parameter to a _type union_ is a breaking change for
+those languages that do not natively support _type unions_. The `jsii-diff` tool can be used to automatically check that
+you incurred no such breakage on an API that was declared `@stable`. If you need to change a `@stable` API to support
+_type unions_ when it previously accepted only a single type, the easiest solution is to deprecate the current `@stable`
+API and introduce a new one that accepts or returns the _type union_.
+
+# Motivation
+
+## Existing use-cases
```
eladb, line 51: I feel this section should go into the appendix and referenced from here so you can keep the motivation section more concise (basically just name the issues and reference an appendix which dives deeper). It buries the important part of the doc, which is the solution.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+
+In those cases, the current _jsii kernel_ will actually annotate _all_ instances with the first candidate type it
+investigates (regardless of whether the instance conforms to it's specification or not). While schema validation could
+be performed to only use a particular option if the instance conforms to it's schema, it would not allow solving the
+ambiguity around items without any required property (rejecting the presence of additional property would make the
+runtime validation stricter than what the **TypeScript** language does). Additionally, it is impossible to differentiate
+method implementations from one another at run-time, since the **Javascript** runtime does not retain type information
+for method arguments and return types (making homonymous methods impossible to tell apart).
+
+## Desired Features
```
eladb, line 280:
```suggestion
## Requirements
```

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+```ts
+export type UnionType = CandidateA | CandidateB;
+```
+
+In order to ensure ability to generate stable code for _union-like_ types in statically typed languages, the possible
+_candidate types_ of an aliased _type union_ must however be restricted to prevent name collisions:
+
+- all candidates **must** have distinct _unqalified names_:
+  ```ts
+  // Invalid: both types have the unqualified name `Construct`
```
eladb, line 336: Wait! so how is this achieved?

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+  export type InvalidUnion = cdk.Construct | constructs.Construct;
+  ```
+- no candidate type can have a name starting with `ListOf` or `MapOf` followed by an upper-case letter:
```
eladb, line 339: This means that we can't allow these prefixes for *any* jsii type because they could potentially be needed in a union in the future. I am okay with this but should call out that this is a new compiler limitation.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+  // Invalid: ListOfFoo would collide with generated union-like accessors for `Foo[]`
+  export type InvalidUnion = ListOfFoo | Bar;
+  ```
+
+In order to preserve backwards-compatbility, and to retain ability to use this new feature against the current CDK v1
+codebase, the `jsii` compiler will not **require** that _type unions_ are systematically aliased. Similarly,
+`jsii-pacmak` will be modified in such a way that it continues producing the same code whether a _type union_ is aliased
+or not.
+
+Once this done, the `cfn2ts` tool, which is used in the CDK to generate _CloudFormation Resources_ (also known as _L1
```
eladb, line 350: Why not start with this and then introduce the compiler restriction?

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+constructs_) will be modified so that it emits aliases for any _type union_ it generates.
+
+Gearing up for the [AWS CDK v2] release, `jsii` will be updated to accept a new option (`--strict=type-unions`) that
+will allow users to opt into **requiring** that any visible _type union_ is aliased. This feature will be gradually
+turned on in the AWS CDK v1 development branch, as all existing _type unions_ are aliased. Once the whole codebase has
```
eladb, line 355: Asides from L1s, the CDK codebase itself almost doesn't have type unions. Can we just get rid of them early? I don't see a huge need for this gradual phase out.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+    public static UnionType fromListOfBaz(Baz... value) { /* ... */ }
+    public static UnionType fromListOfBaz(List<Baz> value) { /* ... */ }
+
+    public @Nullable Foo asFoo() { /* ... */ }
+    public @Nullable Map<String, Bar> asMapOfBar() { /* ... */ }
+    public @Nullable List<Baz> asListOfBaz() { /* ... */ }
+  }
+  ```
+
+The _jsii Kernel_ will need adjustments in order to better support those types. The generated libraries will provide a
```
eladb, line 405: Can we do this without a flag?

Generally, I am wondering - since today Java/.NET code just uses `Object` for union types, we can technically move to generate the union-like classes and accept them as inputs for those `Object` properties and then, the only thing we do in v2.x is replace `Object` with the union name. 

The benefit is that we basically already generate all the classes and allow them to be used freely and only introduce the breaking API change in 2.x.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+proven to be challenging from an API design pespective: it tends to make the experience significantly worse for dynamic
+languages (as boilerplate-heavy patterns must be leveraged as a replacement for _type unions_), and would reduce the
+flexibility in language-specific code generation for _union-like_ types (as those would be bound to the **TypeScript**
+API). Full support of _type unions_ as a type kind in the _jsii type model_ achieves better outcomes:
+
+- it ensures consistency around _union-like_ types
+- it does not degrade the developer experience in languages with _native type_ union support
+- it requires significantly less effort from library authors
+
+This feature could however not be introduced as part of CDK v1, as the change of static (and dynamic) type it requires
```
eladb, line 430: As mentioned above, the only thing that can't be introduce on 1.x is changing the types of .NET and Java from Object to the union-like. But we can technically already generate and accept the union like on v1.x, no?

Another idea is to only generate union-like classes when the typescript code uses a union type alias. Then, change our L1 code generators to generate the union type aliases only for new properties introduced after some CFN spec version (not sure how to do that). 

The benefit is that we can basically start migrating towards this (and baking it!) without breaking anyone. Then, when we are ready, in the next CDK major version we can flip the switch and generate union type aliases for everything.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,558 @@
...
+in statically typed languages would be breaking to existing customers. The release of [AWS CDK v2] is a prime
+opportunity to introduce this change and to address a well-known customer pain point.
+
+# Adoption Strategy
+
+This feature is a candidate for rollout as part of [AWS CDK v2]. Since the removal of type unions will incur significant
```
eladb, line 436: I am wondering if we should wait for 3.x to "flip the switch" because there is quite a lot of work over the 1.x line (both CDK and jsii) that can be done before we are ready to introduce the breaking change to the CDK.

I would recommend to kick-off the non-breaking work on jsii and CDK and then flip the switch in the next CDK major release after this code has been baked and used on the 1.x lines (of both projects) for a few months.

text/0193-fixing-type-unions.md
RomainMuller: I'd actually argue the important part of the doc is the problem statement. The solution is worthless if it does not address the problem! (Which was my issue before). That said - I'm all for moving the use-case characterization down, since it only provides context for the problem statement.

text/0193-fixing-type-unions.md
RomainMuller: By a check in the `jsii` compiler.

text/0193-fixing-type-unions.md
RomainMuller: I had the idea of allowing non-union type aliases to allow providing a local alternate name for some type, which would allow disambiguating when there are unqualified-name clashes, and to work around the limitation here...

I'd probably be inclined to issue a compilation time warning that a type called `ListOf<foo>` or `MapOf<foo>` will not (at least in the then-current state of things) be useable within a type union. Those kind of type names are actually relatively poor (class) API design, particularly in **TypeScript** (typically, a more expressive name should be used if it is worth giving a list or map a name), so the warning is probably not a bad idea.

text/0193-fixing-type-unions.md
RomainMuller: Philosophically it does not matter much. I will **need** to gate this with a flag as this is otherwise a breaking change for `jsii` customers!

Whether we fix this flag early or late... doesn't change the flag must be here.

text/0193-fixing-type-unions.md
RomainMuller: Except this consists in a breaking change for Java customers. Right now, they coud get the `Object` and in many cases down-cast that to the correct dynamic type / check it with an `instanceof` guard.

If we change the dynamic type of return values of `Object`-returning methods, then all of these will suddenly break.

text/0193-fixing-type-unions.md
RomainMuller: > Another idea is to only generate union-like classes when the typescript code uses a union type alias. Then, change our L1 code generators to generate the union type aliases only for new properties introduced after some CFN spec version (not sure how to do that).

So technically, we can have `jsii-diff` treat converting from an "inline union" to an aliased one as a breaking change. We can then introduce an `awslint` rule that requires aliased unions (and add exceptions for all existing inline unions) so we stop proliferation.

It has the benefit of allowing out-of-band (meaning directly on v1) breaking changes on those APIs that are returning struct/interface unions, and are **known** to currently be unusable in **C#** or **Java** right now (since their broken nature would mean we cannot be breaking any usage - or at least not more than it already was).

text/0193-fixing-type-unions.md
RomainMuller: Oh looks like this is a leftover from the previous iteration. The intention is quite a bit different now :)

text/0193-fixing-type-unions.md
RomainMuller: Yeah [introduction to the Fluent Builder Pattern](https://dzone.com/articles/fluent-builder-pattern).

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+an assigned _jsii FQN_) can be annotated with their _real_ dynamic type without problems. However in the case where the returned value is an object
+literal (either a struct literal, or an anonymous interface implementation), the **Javascript** runtime provides no way to determine which struct or
+interface of the _type union_ candidates is implemented by the value.
+
+Taking the example of `AWS::SAM::Function`'s `events` property, the union of possible event sources types only includes structs, and one of the
+candidates, `AlexaSkill` has no required properties, meaning _any_ object actually satisfies it's specification.
+
+In those cases, the current _jsii kernel_ will actually annotate _all_ instances with the first candidate type it investigates (regardless of whether
+the instance conforms to it's specification or not). While schema validation could be performed to only use a particular option if the instance
+conforms to it's schema, it would not allow solving the ambiguity around items without any required property (rejecting the presence of additional
```
rix0rrr, line 217: It would somewhat. We could (probably, need to think on this) the types so that they are ordered from most to least restrictive.

In that case, the first one that matches is [probably] the intended type, with the most permissive "anything goes" type coming last and not accidentally swallowing the real type.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+    private UnionType(/* ... */) { /* ... */ }
+
+    private static final class FooUnionType extends UnionType { /*...*/ }
+    private static final class MapOfBarUnionType extends UnionType { /*...*/ }
+    private static final class ListOfBazUnionType extends UnionType { /*...*/ }
+  }
+  ```
+
+The _jsii Kernel_ will need adjustments in order to better support those types. The generated libraries will provide a new argument to the _jsii
+kernel_'s `load` API if it's code was generated with `--union-like-types` enabled. This way, the kernel will be bale to reference the declared _type
```
rix0rrr, line 355: I'd still like to see what the TypeScript code that produces a `UnionType` looks like, and a description of how the kernel uses that.

As far as I can tell, you're still stuck with the:

```ts
export function banana(): UnionType {
  return {
    fu: 'manchu'
  };
}
```

Problem of needing to do field analysis on object literals.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+union_'s information when annotating instances that are passed from **Javascript** to the host application, removing all ambiguity regarding which
+type is serialized.
+
+# Drawbacks
+
+This approach may allow developers in statically typed languages to cause _undefined behavior_ to occur if they are not provided with a way to
```
rix0rrr, line 361: Wait what? I don't understand this entire section.

> if they are not provided with a way to validate

In statically typed languages that use our generated union class, we can GIVE them a way to validate the type, no? We can add an `isFoo(): boolean` method.

Further, you initially defined `asFoo()` to return `null` if the enclosed type wasn't a `Foo`, so that could serve as a type check (albeit one I disagree with).

> developers may use an invalid `as<TypeName>` accessor and retrive a proxy that may behave unexpectedly

You defined it to return `null`, which is not a proxy at all. If we throw an exception instead, we don't return a proxy at all, which is also not a problem.

> This behavior is however identical to what happens in **TypeScript** when developers use `as any`

It is not? In TypeScript you'll get "'8' does not have a method 'toLowerCase()'", whereas this would (as currently specced) return `null`. 

-----

I think you might either have meant to say "dynamically typed languages" or "statically typed languages with union type support so that we don't use our generated union class" ? Otherwise I don't really understand what you're trying to say here.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+replacement for _type unions_), and would reduce the flexibility in language-specific code generation for _union-like_ types (as those would be bound
+to the **TypeScript** API). Full support of _type unions_ as a type kind in the _jsii type model_ achieves better outcomes:
+
+- it ensures consistency around _union-like_ types
+- it does not degrade the developer experience in languages with _native type_ union support
+- it requires significantly less effort from library authors
+
+# Adoption Strategy
+
+Leveraging aliased _type unions_ in all existing CDK APIs would be a breaking change for all statically typed bindings. As such, the initial rollout
```
rix0rrr, line 379: Simpler proposed rollout strategy:

* Rewrite all current code to use **named unions** (most of them are generated so shouldn't be too hard).
* Add a pacmak flag to generate v1 bindings treating the unions as "inline types", and v2 bindings generating the union classes.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+    private static final class FooUnionType extends UnionType { /*...*/ }
+    private static final class MapOfBarUnionType extends UnionType { /*...*/ }
+    private static final class ListOfBazUnionType extends UnionType { /*...*/ }
+  }
+  ```
+
+The _jsii Kernel_ will need adjustments in order to better support those types. The generated libraries will provide a new argument to the _jsii
+kernel_'s `load` API if it's code was generated with `--union-like-types` enabled. This way, the kernel will be bale to reference the declared _type
+union_'s information when annotating instances that are passed from **Javascript** to the host application, removing all ambiguity regarding which
+type is serialized.
```
rix0rrr, line 357: I'd like to see some more detail about the kernel changes here.

Describe the wire format of a union type on the wire, and lists of union types etc.

Is there anything different between input parameters, return types, callback parameters, etc we should think about?

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+
+# Release Notes
+
+## For the _jsii_ project
+
+Use of **TypeScript** _type unions_ on any visible APIs (`public` or `protected` members of `export`ed types) within _jsii libraries_ must be named
+through the use of an exported type alias, and all those visible APIs must refer to the union using the alias:
+
+```ts
+export type ShinyUnion = Foo | Bar | Baz;
```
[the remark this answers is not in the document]
RomainMuller: The type system actually has those visible for ya... under `Type.aliasSymbol`. It's pretty dope :)

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+This _union-like_ type is also returned from methods instead of `Object`:
+
+```java
+final CfnBucket bucket = createBucket();
+
+// Attempt getting the value as an `IResolvable` instance (returns null if the instance is *definitely not* an IResolvable).
+@Nullable
+final IResolvable value = bucket.getLoggingConfiguration().asIResolvable();
```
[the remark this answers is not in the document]
RomainMuller: Yeah. My POC so far actually throws a ClassCastException by default :)

text/0193-fixing-type-unions.md
RomainMuller: The issue is that in case two types only have optional properties, a value that has no properties matches both. If an API has a contract such as "if foo is bar, then baz is <some type>", but the kernel elects the other empty-compatible candidate, then user will have a (bad) surprise.

It's not always possible to generate a dynamic proxy that represents "both" or "either" of those types (and the Kernel API definitely does not support such a wire type currently).

text/0193-fixing-type-unions.md
RomainMuller: No - we'd wire the value as a `UnionType` typed. So we're not actually promising any more than what the declared type tells us.

text/0193-fixing-type-unions.md
RomainMuller: My argument is we may not always be able to type validate completely (e.g: I cannot tell two homonymous methods apart at runtime). In such cases, runtime could allow a forceful downcast that is technically incorrect, and you're in the same situation as if you had used `as any` in TypeScript.

Of course, if the API includes a type assertion method, this can be used to confirm whether your cast is valid or not (either before or after having done it, I suppose).

---

> It is not? In TypeScript you'll get "'8' does not have a method 'toLowerCase()'", whereas this would (as currently specced) return null.

I mean identical to what happens if you use `as any` in TypeScript in the sense that you could get an error if you're attempting to do things to this value that it does not support/allow. The error could be different (but it'd be an error nonetheless).

Also - if we craft an "arbitrary type" proxy for this, we'd forward the call straight to the JS process, who'd then fail it the same way as it would in TypeScript.

text/0193-fixing-type-unions.md
RomainMuller: This was the proposal before moving to that. The inconvenient of the approach you are proposing is that it does not allow fixing the **currently broken** things by generating better type signatures in Java/C#. Those methods could be broken right now since they cannot currently be used from those languages.

RomainMuller:
> Potentially we can [start with] disallowing object literals from being sent out by JS where the type is declared as a union.

@rix0rrr - this would be a breaking change, which cannot easily be guarded against at compile-time (so users **in non-TypeScript** would be the only ones enjoying the nasty surprise).

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,581 @@
...
+feature name: fixing-type-unions
+start date: 2020-07-06
+rfc pr: 194
+related issue: 193
+---
+
+# Summary
+
+_Type unions_ is a feature of the TypeScript language that allows typing values as "one of several candidate types". This is achieved by specifying a
+`|`-delimited list of candidate types (e.g: `Foo | Bar`). While this feature is common in dynamic languages (**TypeScript**, **Python**, ...), it is
```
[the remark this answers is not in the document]
RomainMuller: Sorry, but:

- **Rust** unions are unions in the `C` union sense (and hence they require a name). The semantics are very similar, but while in TypeScript you can pass "Foo" or "Bar" to a parameter typed "type FooBar = Foo | Bar", you **must** pass a FooBar value in **Rust**.
- **Swift** enumerations are not type unions, and they must be named. The semantics are similar, and I can agree they are union-like in the same sense that the classes we'd generate per this RFC in Java/C# are union-like.
- **Kotlin** appears to apply ["type decaying"](https://kotlinlang.org/spec/type-system.html#type-decaying) to union types. I found I am not smart enough to understand the description of this process, so I have no idea what it entails 🤷🏻‍♂️.

I honestly don't know enough about the `OCaml` and `Scala` scenarios.

Anyhow - I guess I see how the statement can be perceived as excessively broad, largely because the "type unions" feature is under-qualified (in TypeScript the unions can be anonymous, they imply a value typed as any candidate can be passed through or returned from a union-typed API), and the language groups aren't qualified well enough (as you had noted previously).

text/0193-fixing-type-unions.md
rix0rrr: We can disallow unions where it is impossible to discriminate between the types.

Anyway, I'd like to see more detail on this specific problem in the RFC as a whole. It's a little undertreated right now.

text/0193-fixing-type-unions.md
rix0rrr: Okay. But then what does a Java receiver do? How will it decide between `FuType | BarType` in this instance?

text/0193-fixing-type-unions.md
rix0rrr: Can you give examples of things that are today outright *broken* though? The only change that's acceptable is things that simply *cannot be used* today (*can be used but with ugly syntax* is not good enough since presumably we'd be changing the semantics).

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,604 @@
...
+be provided as a way to maintain compatibility with the old name at very little cost.
+
+# Implementation Plan
+
+- [ ] Implement support for aliased _type unions_ in `jsii`
+- [ ] Implement _union-like_ type emission in `jsii-pacmak`
+  - [ ] in **C#**
+  - [ ] in **Java**
+  - [ ] in **Go**
+- [ ] Make `jsii-diff` aware of aliased _type-unions_ (changing from inline to aliased, or back, is a breaking change)
```
rix0rrr, line 434: Also need to add support to `jsii-reflect` for this.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,604 @@
...
+leveraged to disambiguate class names in certain contexts. For example, when renaming a class as part of a refactoring, a `@deprecated` type alias can
+be provided as a way to maintain compatibility with the old name at very little cost.
+
+# Implementation Plan
+
+- [ ] Implement support for aliased _type unions_ in `jsii`
+- [ ] Implement _union-like_ type emission in `jsii-pacmak`
+  - [ ] in **C#**
+  - [ ] in **Java**
+  - [ ] in **Go**
```
rix0rrr, line 433: Still need to do something for Python. As specced, an object literal will contain the FQN for the union type, which Python will need to handle. Although if we change the spec I *think* we won't need to do more for Python than handle the union type in pacmak and ignore it :).

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,604 @@
...
+    private UnionType(/* ... */) { /* ... */ }
+
+    private static final class FooUnionType extends UnionType { /*...*/ }
+    private static final class MapOfBarUnionType extends UnionType { /*...*/ }
+    private static final class ListOfBazUnionType extends UnionType { /*...*/ }
+  }
+  ```
+
+The _jsii Kernel_ will be aware of the _aliased type unions_, and values passed from **JavaScript** to the _host application_ will be annotated with
+the _union type_'s fully qualified name if the value is not an instance of a known class (this is the case for values of _struct_ types). On the other
```
rix0rrr, line 356: > the _union type_'s fully qualified name if the value is [an object literal]

I don't think this will work, and break the wire protocol unnecessarily.

Looks like currently, the union types are a proposal that mostly affects statically typed host languages without native type union support. In the type:

```ts
type SomeUnion = string | Foo | IBar;
```

* A `string` would be serialized as `{ "jsii$native": "string", "value": "some_string" }` (or whatever the serialization is), same as today.
* A `Foo` would be serialized as `{ "jsii$objid": "Foo@123" }`, same as today

For both those cases, the only thing a language kernel would need to do is deserialize as usual, and then wrap the value in a `SomeUnion` instance.

If we were to switch object literals from:

```
{ "jsii$objid": "Object@123", "interfaces": ["IBar"] }

// to

{ "jsii$objid": "SomeUnion@123" }
```

This doesn't actually help the language kernel deserialize the right value. Worse, it now has broken the Python kernel, which used to be able to deserialize the value fine but now needs to grow support for union types (or know to ignore them, or something).

Union types should have zero effect on the wire protocol, only on the kernels on the respective sides.

text/0193-fixing-type-unions.md
```diff
@@ -0,0 +1,604 @@
...
+hand, if the value is the instance of a _jsii_ class, the instance will continue to be serialized using this class' fully qualified name, and the
+languages' runtime libraries will wrap that using the _union-like type_ as necessary.
+
+It does not make sense for a _host application_ to use a _type union_ fully qualified name with the _jsii Kernel_'s `create` call: one of the
```
rix0rrr, line 360: If this is true for communication the one way, then it feels like it should be equally true the other way.

text/0193-fixing-type-unions.md
rix0rrr: I'd still like to see whether you consider this problem in scope or out of scope for the current RFC, and if in scope what the proposed handling of:

```
type PropsUnion = FooProps | BarProps;
```

Is going to be.

text/0193-fixing-type-unions.md
RomainMuller: As far as I'm concerned this is an integral part of adding support in `@jsii/spec`. I don't mind clapping it out specifically.

text/0193-fixing-type-unions.md
RomainMuller: > Still need to do something for Python. As specced, an object literal will contain the FQN for the union type, which Python will need to handle

This is naturally required as new tests are added to validate those APIs can be used and behave in the expected way.

> Although if we change the spec I think we won't need to do more for Python than handle the union type in pacmak and ignore it :)

This was my intention (aka de-sugar the type aliases in Python).

text/0193-fixing-type-unions.md
RomainMuller: The point is to continue serializing definitely typed stuff in the same way they currently are, and to only use the union's FQN in the object ID when that is not possible (i.e: when the value returned is a struct or anonymous interface implementation).

text/0193-fixing-type-unions.md
RomainMuller: They can be passed both ways.

The only call you cannot use with is `create`, which will never return a type union either.

text/0193-fixing-type-unions.md
RomainMuller: This is documented... You'll get the union in the same way than you would in **TypeScript**, and you can either use a type inspection API (`isFooProps(v: PropsUnion): v is FooProps`) and you can "safely" force it, or you don't have this and you have to opportunistically force it... Either way you get similar ergonomics & safety to what you have in **TypeScript** in the same scenario.

rix0rrr:
The naive implementation of this in C# would look like (passing a union type as a struct parameter, which is very common because all L1s have that):

```
new CfnUserPool(..., new CfnUserPoolProps {
  UserNameConfiguration = UserNameConfigurationUnion.fromUserNameConfiguration(new UsernameConfiguration { 
    Field = ...
  })
})
```

Which is not great.

I'd like to see added to this RFC: examples in every language we are planning to generate union types for (Java, C#, Go), for the following 3 cases:

* Passing a union typed value as a positional argument
* Passing a union typed value into a Props type
* Reading a union typed value as a return type (<-- we've mostly been focusing on this so far I believe)

eladb:
@RomainMuller what's the status here?

Issue #192, opened 2020-07-05
Removal of the "constructs" compatibility layer (v2.0)


eladb:
|PR|Champion|
|--|--------|
|#195 | @RomainMuller         |

## Description

As part of [v2.0](https://github.com/aws/aws-cdk-rfcs/issues/156) we would like to deprecate the compatibility layer we have in @aws-cdk/core for the [constructs](https://github.com/aws/constructs) library.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

eladb:
This is completed in the v2.0 branch of the CDK. Yeah!

PR #195, opened 2020-07-09
RFC: 0192 Removal of the "constructs" compatibility layer

eladb:
* Tracking issue: #192
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/benisrae/remove-constructs/text/0192-remove-constructs-compat.md)

---

### DIFF
--- a/text/0192-remove-constructs-compat.md
+++ b/text/0192-remove-constructs-compat.md
@@ -30,6 +30,7 @@ This RFC describes the motivation, implications and plan for this project.
 - [Summary](#summary)
 - [Table of Contents](#table-of-contents)
 - [Release Notes](#release-notes)
+  - [00-DEPENDENCY: Declare a dependency on "constructs"](#00-dependency-declare-a-dependency-on-constructs)
   - [01-BASE-TYPES: Removal of base types](#01-base-types-removal-of-base-types)
   - [02-ASPECTS: Changes in Aspects API](#02-aspects-changes-in-aspects-api)
   - [03-DEPENDABLE: Changes to IDependable implementation](#03-dependable-changes-to-idependable-implementation)
@@ -45,6 +46,7 @@ This RFC describes the motivation, implications and plan for this project.
   - [3. Composability with other domains](#3-composability-with-other-domains)
   - [4. Non-intuitive dependency requirement](#4-non-intuitive-dependency-requirement)
 - [Design](#design)
+  - [00-DEPENDENCY](#00-dependency)
   - [01-BASE-TYPES](#01-base-types)
   - [02-ASPECTS](#02-aspects)
   - [03-DEPENDABLE](#03-dependable)
@@ -86,7 +88,7 @@ import { Construct } from '@aws-cdk/core';
 With this:
 
 ```ts
-import { Cosntruct } from 'constructs';
+import { Construct } from 'constructs';
 ```
 
 The following table summarizes the API changes between 1.x and 2.x:
@@ -100,19 +102,60 @@ The following table summarizes the API changes between 1.x and 2.x:
 `@aws-cdk/core.ConstructOrder` | `constructs.ConstructOrder`
 `@aws-cdk/core.ConstructNode` | `constructs.Node`
 `Construct.isConstruct(x)` | `x instanceof Construct`
-`construct.node.applyAspect(aspect)` | `Aspects.of(construct).apply(aspect)`
+`construct.node.applyAspect(aspect)` | `Aspects.of(construct).add(aspect)`
 `@aws-cdk/core.IDependable` | `constructs.IDependable`
 `@aws-cdk/core.DependencyTrait` | `constructs.Dependable`
 `@aws-cdk.core.DependencyTrait.get(x)` | `constructs.Dependable.of(x)`
 `construct.node.dependencies` | Is now non-transitive
 `construct.addMetadata()` | Stack trace not attached by default
-`ConstructNode.prepare(node)`, `onPrepare()`, `prepare()` | Not supported
-`ConstructNode.synthesize(node)`, `onSynthesize()`, `synthesize()` | Not supported
+`ConstructNode.prepareTree()`, `node.prepare()`, `onPrepare()`, `prepare()` | Not supported, use aspects instead
+`ConstructNode.synthesizeTree()`, `node.synthesize()`, `onSynthesize()`, `synthesize()` | Not supported
 `construct.onValidate()`, `construct.validate()` hooks | Implement `constructs.IValidation` and call `node.addValidation()`
 `ConstructNode.validate(node)` | `construct.node.validate()`
 
 The following sections describe all the related breaking changes and details
 migration strategies for each change.
+
+## 00-DEPENDENCY: Declare a dependency on "constructs"
+
+As part of migrating your code to AWS CDK 2.0, you will need to declare a
+dependency on the `constructs` library (in addition to the `aws-cdk-lib` library
+which now includes the entire AWS CDK).
+
+For libraries, this should be a peer dependency, similarly to your dependency on
+the AWS CDK. You will likely also want to declare those as `devDependencies` in
+order to be able to run tests in your build environment.
+
+To increase interoperability of your library, the recommendation is to use the
+lowest possible __caret__ version:
+
+```json
+{
+  "peerDependencies": {
+    "aws-cdk-lib": "^2.0.0",
+    "constructs": "^10.0.0"
+  },
+  "devDependencies": {
+    "aws-cdk-lib": "^2.0.0",
+    "constructs": "^10.0.0"
+  }
+}
+```
+
+For apps, you should declare these as direct dependencies, and you would
+normally want to use the highest version available:
+
+```json
+{
+  "dependencies": {
+    "aws-cdk-lib": "^2.44.0",
+    "constructs": "^10.787.0"
+  }
+}
+```
+
+NOTE: Due to it's foundational nature, the `constructs` library is committed to
+never introduce breaking changes. Therefore, it's version will be `10.x`.
 
 ## 01-BASE-TYPES: Removal of base types
 
@@ -128,7 +171,7 @@ See [examples](https://github.com/aws/aws-cdk/pull/9056/commits/e4dff913d486592b
 ## 02-ASPECTS: Changes in Aspects API
 
 Aspects are not part of the "constructs" library, and therefore instead of
-`construct.node.applyAspect(aspect)` use `Aspects.of(construct).apply(aspect)`.
+`construct.node.applyAspect(aspect)` use `Aspects.of(construct).add(aspect)`.
 
 The `Tag.add(scope, name, value)` API has been removed. To apply AWS tags to a
 scope, use:
@@ -172,17 +215,22 @@ const myConstruct = new MyConstruct(stack, 'MyConstruct');
 This is still a supported idiom, but in 2.x these root stacks will have an
 implicit `App` parent. This means that `stack.node.scope` will be an `App`
 instance, while previously it was `undefined`. The "root" stack will have a
-construct ID of `Stack` unless otherwise specified.
+construct ID of `Default` unless otherwise specified.
 
-> Technically, this change should have resulted in the `node.path` and
-> `node.uniqueId` of all constructs defined within such trees to change, but we
-> are utilizing `node.relicate()` to maintain the current behavior.
+Please note that this also means that the value of `node.path` for all
+constructs in the tree would now have a `Default/` prefix (if it was `Foo/Bar`
+it will now be `Default/Foo/Bar`).
+
+> In contrast, the value of `node.uniqueId` will _not_ change because `Default`
+> is a special ID that is ignored when calculating unique IDs (this feature
+> already exists in 1.x).
 
 ## 05-METADATA-TRACES: Stack traces no longer attached to metadata by default
 
 For performance reasons, the `construct.node.addMetadata()` method will *not*
-attach stack traces to metadata entries. You can explicitly request to attach
-stack traces to a metadata entry using the `stackTrace` option:
+attach stack traces to metadata entries. Stack traces will still be associated
+with all `CfnResource` constructs and can also be added to custom metadata using
+the `stackTrace` option:
 
 ```ts
 construct.node.addMetadata(key, value, { stackTrace: true })
@@ -205,7 +253,7 @@ Although we recommend that you rethink the use of "prepare", you can use this
 idiom to implement "prepare" using aspects:
 
 ```ts
-Aspects.of(this).apply({ visit: () => this.prepare() });
+Aspects.of(this).add({ visit: () => this.prepare() });
 ```
 
 The `ConstructNode.prepare(node)` method no longer exists. To realize references
@@ -316,19 +364,29 @@ constructs with constructs from other domains.
 It is currently impossible to define AWS CDK constructs within a non-AWS-CDK
 construct scope.
 
-For example, consider the
-[Terrastack](https://github.com/TerraStackIO/terrastack) project or a similar
-one, which uses CDK constructs to define stacks through Terraform. Say we want to
-use an AWS CDK L2 inside a Terraform stack construct:
+For example, consider the [CDK for
+Terraform](https://github.com/hashicorp/terraform-cdk) or a similar project,
+which uses constructs to define stacks through Terraform.
+
+We are currently working with HashiCorp to enable the following use case in the
+Terraform CDK:
 
 ```ts
-const stack = new terrastack.Stack(...);
+import * as cdktf from 'cdktf';        // <=== Terraform CDK
+import * as s3 from '@aws-cdk/aws-s3'; // <=== AWS CDK
 
-// COMPILATION ERROR: `this` is not a `cdk.Construct`.
+const stack = new cdktf.TerraformStack(...);
+
+// COMPILATION ERROR: `this` is of type `constructs.Construct` and not a `@aws-cdk/core.Construct`.
 new s3.Bucket(this, 'my-bucket');
 ```
 
-Being able to create construct compositions from multiple domains is a powerful
+In order to enable this usage, we will need `s3.Bucket` to accept any object
+that implements `constructs.Construct` as its `scope`. At the moment, this will
+fail compilation with the error above, because the `scope` in `s3.Bucket` is
+`core.Construct`.
+
+Being able to create compositions from multiple CDK domains is a powerful
 direction for the CDK ecosystem, and this change is required in order to enable
 these use case.
 
@@ -341,12 +399,17 @@ to take a _peer dependency_ on both the CDK library (`aws-cdk-lib`) and
 > Currently, the AWS CDK also takes `constructs` as a normal dependency (similar
 > to all dependencies), but this is about to change with mono-cdk.
 
-The reason `constructs` will also be required (whether we leave the
-compatibility layer or not) is due to the fact that all CDK constructs
-eventually extend the base `constructs.Construct` class. This means that this
-type is part of their public API and therefore a peer dependency is required
-(otherwise, there could be incompatible copies of `Construct` in the node
-runtime).
+The reason `constructs` will have to be defined as a peer-dependency of the AWS
+CDK, whether we leave the compatibility layer or not, is due to the fact that
+all AWS CDK constructs eventually extend the base `constructs.Construct` class.
+This means that this type is part of their public API, and therefore must be
+defined as a peer dependency (otherwise, there could be incompatible copies of
+`Construct` in the dependency closure).
+
+The removal of the compatibility layer means that now anyone who uses the AWS
+CDK will need to explicitly use the `constructs.Construct` type (even for
+trivial apps), and therefore it would "make sense" for them to take a dependency
+on the `constructs` library.
 
 See the RFC for [monolithic packaging] for more details.
 
@@ -362,6 +425,25 @@ our strategy for front-loading the change into the 1.x branch to reduce forking
 costs and/or alert users of the upcoming deprecation.
 
 This design is based on this [proof of concept](https://github.com/aws/aws-cdk/pull/9056).
+
+## 00-DEPENDENCY
+
+In order to enable composability with other CDK domains (Terraform, Kubernetes),
+all constructs must use the same version of the `Construct` base class.
+
+As long as all libraries in a closure take a __peer dependency__ on a compatible
+version of `constructs`, the npm package manger will include a single copy of
+the library, and therefore all constructs will derive from the same `Construct`
+(and more importantly, accept the same `Construct` for `scope`).
+
+Practically this means that we can never introduce a major version of
+`constructs` because any major version will require a new major version of all
+CDKs, and that is impossible to require or coordinate given the decentralized
+nature of the ecosystem.
+
+We propose to take a commitment to __never introduce breaking changes in
+"constructs"__. This implies that we will never introduce another major version.
+To symbolize that to users, we will use the major version `10.x`.
 
 ## 01-BASE-TYPES
 
@@ -424,7 +506,7 @@ apply aspects to scopes for AWS CDK apps. To do that, we propose to use the
 scopes:
 
 ```ts
-Aspects.of(scope).apply(aspect);
+Aspects.of(scope).add(aspect);
 ```
 
 The major downside of this change is discoverability, but
@@ -443,7 +525,7 @@ pattern: `Tags.of(x).add(name, value)`.
 ## 03-DEPENDABLE
 
 The `constructs` library supports dependencies through `node.addDependency` like
-in 1.x, but the API to implement `IDependable` has been slightly changed.
+in 1.x, but the API to implement `IDependable` has been changed.
 
 The `constructs` library also introduces `DependencyGroup` which is a mix
 between `CompositeDependable` and `ConcreteDependable`.
@@ -519,24 +601,27 @@ synthesized output if called twice, we will also need to introduce a
 `stage.synth({ force: true })` option. This will be the default behavior when
 using `expect(stack)` or `SynthUtils.synth()`.
 
-### Preserving paths and unique IDs using `node.relocate()`
+### Preserving unique IDs using an ID of `Default`
 
-A side effect of adding a `App` parent to "root" stacks is that construct paths
-and unique IDs in unit tests will now change. In the above example,
-`foo.node.path` will change from `MyFoo` to `Stack/MyFoo`. Additionally, tests
-for resources that utilized `node.uniqueId` to generate names will also change
-given `uniqueId` is based on the path.
+A side effect of adding a `App` parent to "root" stacks is that we now have an
+additional parent scope for all constructs in the tree. The location of the
+construct in the tree is taken into account when calculating `node.path` and
+`node.uniqueId`.
 
+Since `uniqueId` is used in several places throughout the AWS Construct Library
+to allocate names for resources, and we have multiple unit tests that expect
+these values, we will use the ID `Default` for the root stack.
 
-To address this, we propose to introduce a new capability in `constructs` to
-"**relocate**" a scope to a different path. This capability will also be useful
-generally when refactoring code (for example, if I want to add a level of
-encapsulation to my app without breaking production deployments).
+The `uniqueId` algorithm in the constructs library (see [reference]()) ignores
+any node with the ID `Default` for the purpose of calculating the unique ID,
+which allows us to perform this change without breaking unique IDs.
 
-Using this capability, we will relocate the "root" stack to the `""` path so that
-all paths and unique IDs will be preserved.
+We will accept the fact that `node.path` is going to change for this specific
+use case (only relevant in tests).
 
 ### Alternative considered
+
+#### Alternative 1: Breaking unique IDs
 
 We explored the option of fixing all these test expectations throughout the CDK
 code base and back port this change over the 1.x behind a feature flag in order
@@ -550,10 +635,21 @@ The downsides of this approach are:
 1. We currently don't have a way to implicitly run all our unit tests behind a
    feature flag, and it is not a trivial capability to add.
 
+#### Alternative 2: `node.relocate()`
+
+We also explored the option of introducing an additional capability to
+constructs called `node.relocate(newPath)` which allows modifying the path of a
+scope such that all child scopes will automatically be "relocated" to a new
+path. This would have allowed avoiding the breakage in `node.path` but would
+have also introduced several other idiosyncrasies and potential violations of
+invariants such as the fact that a path is unique within the tree.
+
 ### What can we do on 1.x?
 
-We will add `node.relocate()` to `constructs` 3.x and implement this change over
-1.x.
+We will introduce this change over the 1.x branch as-is, acknowledging that we
+are technically breaking the behavior of `node.path` in unit tests which use
+`Stack` as the root. Since we are not breaking `uniqueId`, we expect this to be
+tolerable over the 1.x branch.
 
 ## 05-METADATA-TRACES
 
@@ -615,7 +711,7 @@ dependencies](https://github.com/awslabs/cdk8s/blob/ef95b9ffce8a39200e028c2fe8ac
 required that the name of the output manifest will be determined based on the
 topologic order at the app level. Here again, the generic approach failed.
 
-In leu of those failures, we decided that there is no additional value in
+In lieu of those failures, we decided that there is no additional value in
 actually offering a synthesis mechanism at the `constructs` level. Each
 CDK-domain implements synthesis at the "right" level. This does not mean that
 specific domains can't offer a decentralized approach (i.e. call a method called
@@ -644,19 +740,20 @@ this as soon as the CDK app is created). Therefore, the solution for
 
 ### Implications on end-users
 
-Participation in synthesis is an "advanced" feature of the CDK framework and se
+Participation in synthesis is an "advanced" feature of the CDK framework and we
 assume most end-users don't use this directly.
 
-If they need "last minute processing", they would likely use `prepare()` (which
-is also being [removed](#06-no-prepare-the-prepare-hook-is-no-longer-supported))
-but not `synthesize()`.
+If they need "last minute processing", they can add an aspect to the node which
+will be applied before synthesis (the alternative to "prepare").
 
 The use case of emitting arbitrary files into the cloud assembly directory is
 weak. The cloud assembly is a well-defined format, and is currently "closed".
-There are no extension points that tools can identify. To that end, just writing
-files to the cloud assembly output directory does not make tons of sense. Having
-said that, it is still possible to do in the same way we plan for
-`AssetStaging`.
+There are no extension points that tools can identify.
+
+To that end, just writing files to the cloud assembly output directory does not
+make tons of sense. Yet, if there is still a use case for writing files during
+initialization, it is possible to find out the output directory through
+`Stage.of(scope).outdir`. This is how asset staging will be implemented.
 
 ### What can we do on 1.x?
 
@@ -886,3 +983,5 @@ for constructs 4.x.
   from the POC branch into it.
 - [ ] Write a migration guide with guidance on `synthesize` and `prepare` in
   <https://github.com/aws/aws-cdk/issues/8909>
+- [ ] Updates to Developer Guide (add 2.x section)
+- [ ] Updates to READMEs across the library (add 2.x section)
--- a/text/0192-remove-constructs-compat.md
+++ b/text/0192-remove-constructs-compat.md
@@ -318,7 +318,12 @@ Possible solutions:
    violate an eslint rule, but we can disable it for the time being.
 
 We can start with #1 for a while and see how bad it is. If it is, we can try to
-introduce some automation or backporting.
+introduce some automation or back-porting.
+
+Something that we can definitely do on 1.x is replace all the variants such as
+`core.Construct` and `cdk.Construct` to just `Construct`. This will slightly
+help.
+
 
 ## Removal of "synthesize"
 
@@ -493,8 +498,9 @@ The prepare hook was used in the CDK in a few cases:
 
 The first two use cases have already been addressed by centralizing the "prepare" logic at the stage level (into [prepare-app.ts](TODO)).
 
-The 3rd item can be addressed using `Lazy` tokens, and will be addressed on 1.x
-prior to the 2.x fork.
+The 3rd item can be addressed using `Lazy` tokens (see
+[example](https://github.com/aws/aws-cdk/pull/8962/files#diff-51d435d71a31c2607f923fc4d96cac56R140)),
+and will be addressed on 1.x prior to the 2.x fork.
 
 ## Validation changes
 
@@ -658,3 +664,4 @@ merge conflict potential.
 - [ ] Remove the use of "prepare" and "synthesize" in 1.x
 - [ ] Implicit `App` for `Stack`s without a scope behind a feature flag and
       enable in our unit tests in 1.x
+- [ ] Normalize reference to base types (`cdk.Construct` => `Construct`).
--- a/text/0192-remove-constructs-compat.md
+++ b/text/0192-remove-constructs-compat.md
@@ -25,18 +25,48 @@ This RFC describes the motivation, implications and plan for this project.
 [AWS CDK v2.0]: https://github.com/aws/aws-cdk-rfcs/issues/156
 [construct-compat.ts]: https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/core/lib/construct-compat.ts
 
+# Table of Contents
+
+- [Summary](#summary)
+- [Table of Contents](#table-of-contents)
+- [Release Notes](#release-notes)
+  - [Removal of base types](#removal-of-base-types)
+  - [Changes in Aspects API](#changes-in-aspects-api)
+  - [Changes to IDependable implementation](#changes-to-idependable-implementation)
+  - [Stacks as root constructs](#stacks-as-root-constructs)
+  - [Stack traces no longer attached to metadata by default](#stack-traces-no-longer-attached-to-metadata-by-default)
+  - [Lifecycle hooks removal](#lifecycle-hooks-removal)
+  - [Summary](#summary-1)
+- [Motivation](#motivation)
+- [Design](#design)
+  - [Removal of the base types (`cdk.Construct`, `cdk.IConstruct`, ...)](#removal-of-the-base-types-cdkconstruct-cdkiconstruct-)
+  - [Removal of "synthesize"](#removal-of-synthesize)
+  - [Removal of "prepare"](#removal-of-prepare)
+  - [Validation changes](#validation-changes)
+  - [Stack trace settings](#stack-trace-settings)
+  - [Info/warning/error message metadata key changes](#infowarningerror-message-metadata-key-changes)
+- [Drawbacks](#drawbacks)
+  - [User migration effort](#user-migration-effort)
+  - [CDK codebase migration efforts](#cdk-codebase-migration-efforts)
+  - [Staging of the 2.x fork](#staging-of-the-2x-fork)
+- [Rationale and Alternatives](#rationale-and-alternatives)
+  - [Alternatives considered](#alternatives-considered)
+- [Adoption Strategy](#adoption-strategy)
+- [Unresolved questions](#unresolved-questions)
+- [Future Possibilities](#future-possibilities)
+- [Implementation Plan](#implementation-plan)
+
+
 # Release Notes
 
 > This section "works backwards" from the v2.0 release notes in order to
 > describe the user impact of this change.
 
+**BREAKING CHANGE**
+
 As part of CDK v2.0, all types related to the *constructs programming model*
 have been removed from the AWS CDK and should be used directly from the
 [constructs](https://github.com/aws/constructs) library.
-
-```
-BREAKING CHANGE
-```
 
 For the majority of CDK libraries and apps, you will simply need to add
 `constructs@^4` to your `package.json` and replace:
@@ -52,7 +82,7 @@ import { Cosntruct } from 'constructs';
 ```
 
 The following sections describe all the breaking changes related to this
-project.
+project:
 
 ## Removal of base types
 
--- a/text/0192-remove-constructs-compat.md
+++ b/text/0192-remove-constructs-compat.md
@@ -23,7 +23,7 @@ and new projects such as CDK for Kubernetes.
 This RFC describes the motivation, implications and plan for this project.
 
 [AWS CDK v2.0]: https://github.com/aws/aws-cdk-rfcs/issues/156
-[construct-compat.ts]: https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/core/lib/construct-compat.ts
+[construct-compat.ts]: https://github.com/aws/aws-cdk/blob/fcca86e82235387b88371a0682cd0fc88bc1b67e/packages/%40aws-cdk/core/lib/construct-compat.ts
 
 # Table of Contents
 
@@ -92,7 +92,6 @@ The following `@aws-cdk/core` types have stand-in replacements "constructs":
 - The `@aws-cdk/core.IConstruct` type has been replaced with `constructs.IConstruct`
 - The `@aws-cdk/core.ConstructOrder` class has been replaced with `constructs.ConstructOrder`
 - The `@aws-cdk/core.ConstructNode` class has been replaced with `constructs.Node`
-- `Construct.isConstruct(x)` should be replaced with `x instanceof Construct`.
 
 ## Changes in Aspects API
 
@@ -182,10 +181,16 @@ values or Aspects if appropriate.
 The **synthesis** hook (`construct.onSynthesize()` and `construct.synthesize()`)
 is no longer supported. Synthesis is now implemented only at the app level.
 
-The `ConstructNode.synthesize(node)` method no longer exists. To prepare an app
-or a stage, simply call `app.synth()` or `stage.synth()`.
+If your use case for overriding `synthesize()` was to write into the cloud
+assembly directory, you can now find the current cloud assembly output directory
+using `Stage.of(this).outdir`.
 
-To write files into the cloud assembly directory, use `Stage.of(this).outdir`.
+The `ConstructNode.synthesize(node)` method no longer exists. However, since now
+`Stage.of(construct)` is always defined and returns the enclosing stage/app, you
+can can synthesize a construct node through `Stage.of(construct).synth()`.
+
+For additional questions/guidance on how to implement your use case without this
+hook, please post a comment on this GitHub issue: [TODO].
 
 ### Validation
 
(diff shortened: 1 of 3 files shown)


text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+did not hold. In `cdk8s`, the support for [TODO: chart dependencies] required
+that the name of the output manifest will be determined based on the topologic
+order at the app level. Here again, the generic approach failed.
+
+In leu of those failures, we decided that there is no additional value in
+actually offering a synthesis mechanism at the `constructs` level. Each
+CDK-domain implements synthesis at the "right" level. This does not mean that
+specific domains can't offer a decentralized approach (i.e. call a method called
+"synthesize" on all constructs in the tree), it just means that this is not
+provided at the base `Construct` class.
```
RomainMuller, line 350: Conceptually okay with this, but I would love `constructs` to provide documentation on how to "properly" to that. In the current state of things, it's a little hand-wavy.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+method to `_synthesizeTemplate()` and `_synthesizeTree()` and will call them
+from the centralized synthesis function.
+
+The `AssetStaging` construct does not really need to wait until synthesis in
+order to stage the asset. In fact, all the information required already exists
+during initialization. The only missing information is the cloud assembly output
+directory, and this information is actually known during initialization (we know
+this as soon as the CDK app is created). Therefore, the solution for
+`AssetStaging` is to move the staging logic to the constructor and use
+`Stage.of(this).outdir` to find the output directory of the current stage.
```
RomainMuller, line 370: I wish we capitalized that `D` (`outDir`)

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+uses this as the stack name when the stack does not have an ID (see [TODO]).
+
+This change will allow us to remove any special casing we have for stacks in the
+testing framework and throughout the synthesis code path (we have quite a lot of
+that), because we will be able to assume that `Stage.of(construct.node.root)` is
+never `undefined` and has a `synth()` method which returns a cloud assembly.
+
+Since unit tests sometimes use "incremental tests" for synthesized templates,
+and `stage.synth()` would reuse the synthesized output if called twice, we will
+also need to introduce a `stage.synth({ forceResynth: true })` option. This will
```
RomainMuller, line 433: I think `{ force: true }` conveys the meaning already. I don't really love the aesthetics of `Resynth`.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+be the default behavior when using `expect(stack)` or `SynthUtils.synth()`.
```
RomainMuller, line 434: Not clear why. If I `expect(stack)` multiple times on the same stack, I most likely don't expect it to be synthesized over and over again.

For `SynthUtils.synt()` I'd say it likely makes sense (since in this case, the user clearly expressed they intention to synthesize).

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+----
+
+
+
+
+constructs 4.x:
+
+- [x] Migrate to projen
+- [x] Branch to 4.x and release as @next
+- [x] Reintroduce `c.node` instead of `Node.of(c)`
```
RomainMuller, line 468: I kinda liked `Node.of(c)` because it keeps the API of `Construct`/`IConstruct` super clean. However I don't really love how `IConstruct` is just a marker interface and it has no bearing in how `Node.of(IConstruct)` achieves getting the `Node` from there (which makes it unsafe).

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+Since stack traces are not attached to metadata entries by default in constructs
+4.x, we will need to pass `stackTrace: true` for `CfnResource`s. This will
+preserve the deploy-time stack traces which are very important for users.
+
+Other metadata entries will not get stack traces by default, and that's a
+reasonable change.
+
+When stack traces are disabled (either through the CDK context or through
+`CDK_STACK_TRACE_DISABLE`), we will need to set the appropriate context key in
```
RomainMuller, line 523: I wish we had made this also available as a CLI flag, but then I'd have wanted this to look like `--[no-]stack-taces`, which I believe would have made the "default" ENV for it `CDK_STACK_TRACES=0|1`

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+Since this change involves modifications to the CDK's source code, it may cause
+merge conflicts as during the period in which we need to forward-port or
+back-port code between the v1.x branch and the v2.x branches.
+
+The key would be to continuously merge between the branches.
+
+# Rationale and Alternatives
+
+As a general rule, software layers which do not provider value to users should
```
RomainMuller, line 571:
```suggestion
As a general rule, software layers which do not provide value to users should
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+The [repository migration](#repository-migration-efforts) efforts and
+[co-existence of 2.x/1.x](#co-existence-of-2x1x) are both one-off costs this
+proposal suggests ways to reduce the chance for merge conflicts across these
+branches.
+
+## Alternatives considered
+
+At a high-level, we may consider to postpone this change to v3.x or to never
+take it, leaving this compatibility layer in place for eternity.
```
RomainMuller, line 595: ... and eternity gets pretty long towards the end :)

iliapolo:
@eladb Notice: https://github.com/aws/aws-cdk/issues/8645

Should we really have the `addInfo`, `addWarning`, ... methods in `constructs`? These are tied to the `aws-cdk` implementation, and don't have any meaning outside its context (for example for `cdk8s`). 

Can we move this to `@aws-cdk/core`? Perhaps its worth extending `constructs.Construct` in `@aws-cdk/core` and providing this API, and have users still extend `core.Construct` rather than `constructs.Construct`?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+API shim so that CDK code will continue to function without change.
+
+As part [AWS CDK v2.0], we plan to remove this redundant layer and make some
+improvements to a few APIs in "constructs" based on our learnings from CDK v1.x
+and new projects such as CDK for Kubernetes.
+
+This RFC describes the motivation, implications and plan for this project.
+
+[AWS CDK v2.0]: https://github.com/aws/aws-cdk-rfcs/issues/156
+[construct-compat.ts]: https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/core/lib/construct-compat.ts
```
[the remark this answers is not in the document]
eladb: Fixed.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+## Removal of base types
+
+The following `@aws-cdk/core` types have stand-in replacements "constructs":
+
+- The `@aws-cdk/core.Construct` class has been replaced with `constructs.Construct`
+- The `@aws-cdk/core.IConstruct` type has been replaced with `constructs.IConstruct`
+- The `@aws-cdk/core.ConstructOrder` class has been replaced with `constructs.ConstructOrder`
+- The `@aws-cdk/core.ConstructNode` class has been replaced with `constructs.Node`
+- `Construct.isConstruct(x)` should be replaced with `x instanceof Construct`.
```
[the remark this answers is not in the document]
eladb: I am not sure I agree given "constructs" must be used as a peer dependency, but I don't really mind restoring this.

Done.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+## Changes in Aspects API
+
+Aspects are not part of the "constructs" library, and therefore instead of
+`construct.node.applyAspect(aspect)` use `Aspects.of(construct).apply(aspect)`.
+
+To apply AWS tags to a scope, prefer `Tag.add(scope, name, value)`.
+
+> TODO: should we change to `Tags.of(scope).add(name, value)`?
```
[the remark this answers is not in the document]
eladb: Yes I agree, `Tag.add` is the current API, as mentioned here we can consider changing to `Tags.of(x).add()`, but it's not a breaking change nor required for this migration at this point.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+The method `construct.node.addDependency(otherConstruct)` did not change, and it
+is likely the primary usage, and therefore most users won't be impacted.
+
+If you need to implement `IDependable`, read on:
+
+- The `@aws-cdk/core.IDependable` type has been replaced with `constructs.IDependable`
+- The `@aws-cdk/core.DependencyTrait` class has been replaced with `constructs.Dependable`
+- `@aws-cdk.core.DependencyTrait.get(x)` is now `constructs.Dependable.of(x)`
+- `construct.node.dependencies` is now **non-transitive** and returns only the dependencies added to the current node.
```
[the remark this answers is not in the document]
eladb: Yes, the `dependencyGraph` thing should have all the goodies.

See https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s/src/dependency.ts

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+> TODO: You can use `construct.node.dependencyGraph` to access a rich object
+> model for reflecting on the node's dependency graph.
```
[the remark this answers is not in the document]
eladb: Will be adopted from https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s/src/dependency.ts

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+### Synthesis
+
+The **synthesis** hook (`construct.onSynthesize()` and `construct.synthesize()`)
+is no longer supported. Synthesis is now implemented only at the app level.
+
+The `ConstructNode.synthesize(node)` method no longer exists. To prepare an app
+or a stage, simply call `app.synth()` or `stage.synth()`.
+
+To write files into the cloud assembly directory, use `Stage.of(this).outdir`.
```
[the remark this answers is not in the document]
eladb: ok, revised phrasing.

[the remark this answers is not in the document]
eladb: Also added a link to a GitHub issue for additional guidance.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+### Validation
+
+Validation is still supported, but instead of a protected base class method it
+is defined through an interface called `constructs.IValidation`.
```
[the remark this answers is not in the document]
eladb: Yes, but that's a stupid name. I think `IValidation` is ok.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+`ConstructNode.validate(node)` | `construct.node.validate()`
+
+# Motivation
+
+There are various motivations for this change:
+
+1. Removal of a redundant layer
+2. User confusion caused by multiple `Construct` types
+3. Inability to compose AWS CDK constructs into other domains
+4. Non-intuitive dependency requirement on `constructs`
```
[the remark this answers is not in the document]
eladb: Not really, #4 is about the fact that regardless of whether you are using `constructs.Construct` or a subclass thereof, you will always need a dependency on "constructs". If you use a subclass (i.e. you never reference `Construct` explicitly), requiring a dependency feels weird.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+Being able to create construct compositions from multiple domains is a powerful
+direction for the CDK ecosystem, and this change is required in order to enable
+these use case.
+
+### 4. Non-intuitive dependency requirement
+
+As we transition to
+[monolithic packaging]
+as part of v2.x, CDK users will have to take a _peer dependency_ on both the CDK
+library (`aws-cdk-lib`) and `constructs`.
```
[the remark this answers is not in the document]
eladb: We can reopen this discussion as part of the v2.0 RFC (#79). Not related to this RFC though I believe.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+
+The reason `constructs` will also be required (whether leave the compatibility
```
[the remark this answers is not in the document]
eladb: Fixed.

text/0192-remove-constructs-compat.md
iliapolo: As opposed to the "default App" proposed earlier, which also has some benefits for our users, it doesn't feel right to enable this solely due to testing considerations. Are we ready for users to start using this? possibly abusively? 

What would be the effort of changing the tests to not need this?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+branch until we switch over to 2.x. Since this change includes modifications to
+all `import` sections in almost all of our files, merge conflicts are imminent.
+
+Possible solutions:
+1. Take the hit and resolve the conflicts manually every time. These are
+   mechanical fixes and easy to do.
+2. Add some automation using eslint to fix these merge conflicts.
+3. Perform the change on the 1.x codebase such that the soon-to-be-moved types
+   are imported separately from other types of `@aws-cdk/core`. This will
+   violate an eslint rule, but we can disable it for the time being.
```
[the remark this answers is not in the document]
eladb: sure, will add the name at some point

text/0192-remove-constructs-compat.md
eladb: I agree that this should be discussed in the docs for `constructs`.

text/0192-remove-constructs-compat.md
eladb: We already use this capitalization in `CloudAssemblyBuilder` and I don't see this warrants a breaking change at this point.

text/0192-remove-constructs-compat.md
eladb: okay, I will change

text/0192-remove-constructs-compat.md
eladb: This is the current behavior (re-synth) and I rather try to avoid breaking additional tests. It's mainly a performance optimization.

text/0192-remove-constructs-compat.md
eladb: Exactly, and that's why I am reintroducing `IConstruct.node` so it's not a marker interface anymore. Also, not having `this.node` as a way to access the node is going to break _a lot_ of CDK code.

text/0192-remove-constructs-compat.md
eladb: See https://github.com/aws/aws-cdk/issues/8792

Assigned to @shivlaks

text/0192-remove-constructs-compat.md
eladb: Fixed

text/0192-remove-constructs-compat.md
eladb: :-)

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+and 2.x we propose to introduce this change on the 1.x branch prior to forking
+off 2.x.
+
+However, this could be perceived as a breaking change by end-users since
+`node.path` and `node.uniqueId`, and their derivatives, will change for trees
+rooted by a `Stack` and unit tests will need to be updated.
+
+Therefore we propose to introduce this change as a feature flag over the 1.x codebase and enable it in our unit tests (I don't believe we have a way to enable feature flags for all unit tests, but we can devise one).
+
+This will allow us to update our tests in 1.x and avoid the merge conflicts
```
iliapolo, line 462: 👍

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+- [x] Reintroduce dependencies
+- [x] Change `node.dependencies` to return the list of node dependency (non recursive) and add `node.depgraph` which returns a `Graph` object from cdk8s.
+- [x] Stack trace control
+
+CDK changes:
+
+- [x] cfn2ts
+- [ ] Consider aliasing in TypeScript to reduce potential merge conflicts.
+
+- [ ] Require `app` when defining a `Stack`.
```
iliapolo, line 485: What do you mean? Didn't we say an app will be implicitly created?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+`CDK_STACK_TRACE_DISABLE`), we will need to set the appropriate context key in
+`App` so that this will propagate to "constructs".
+
+## Info/warning/error message metadata key changes
+
+The construct metadata keys for `addInfo()`, `addWarning()` and `addError()` are
+`aws:cdk:info`, etc. These are not the keys used by default in "constructs"
+since the library is not part of the AWS CDK.
+
+To address this, the "constructs" library allows customizing these keys through a settings module. We will need to add that configuration to the `App` level so this behavior will be preserved.
```
iliapolo, line 538: Makes sense. And addresses https://github.com/aws/aws-cdk-rfcs/pull/195#issuecomment-656611597

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+> describe the user impact of this change.
+
+**BREAKING CHANGE**
+
+As part of CDK v2.0, all types related to the *constructs programming model*
+have been removed from the AWS CDK and should be used directly from the
+[constructs](https://github.com/aws/constructs) library.
+
+For the majority of CDK libraries and apps, you will simply need to add
+`constructs@^4` to your `package.json` and replace:
```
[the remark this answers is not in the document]
eladb: Rephrased to:

For most CDK libraries and apps, you will probably just need to add
`constructs@^4` to your `package.json` and replace:

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+`construct.node.applyAspect(aspect)` use `Aspects.of(construct).apply(aspect)`.
+
+To apply AWS tags to a scope, prefer `Tag.add(scope, name, value)`.
+
+> TODO: should we change to `Tags.of(scope).add(name, value)`?
+
+## Changes to IDependable implementation
+
+The method `construct.node.addDependency(otherConstruct)` did not change, and it
+is likely the primary usage, and therefore most users won't be impacted.
```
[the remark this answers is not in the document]
eladb: Rephrased:

The method `construct.node.addDependency(otherConstruct)` __did not change__,
which is likely the only API you are using related to construct dependencies.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+This has implications on the value of `construct.node.path` (it will be prefixed
+with `Stack/`) and the value of `construct.node.uniqueId` (and any derivatives
+of these two).
+
+This means that as you migrate to v2.x, some unit tests may need to be updated
+to reflect this change.
+
+## Stack traces no longer attached to metadata by default
+
+For performance reasons, the `construct.addMetadata()` method will *not* attach
```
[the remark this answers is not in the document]
eladb: Fixed.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+To write files into the cloud assembly directory, use `Stage.of(this).outdir`.
+
+### Validation
+
+Validation is still supported, but instead of a protected base class method it
+is defined through an interface called `constructs.IValidation`.
+
+Practically this means that if you wish to implement validation for a custom
+construct, implement the `IValidation` interface and change your `validate`
+method from `protected` to `public`.
```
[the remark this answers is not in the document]
eladb: Yes, good catch. added.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+introduced in order to avoid breaking v1.x users. As we release v2.0 we are able
+to clean up this layer, and with it, improve maintainability and code hygiene.
+
+### 2. Multiple `Construct` types
+
+The current situation is error-prone since we have two `Construct` classes in
+the type closure. For example, when a developer types `"Construct"` and uses
+VSCode to _automatically add import statements_, the IDE will actually add an
+import for `constructs.Construct`, which is _not_ the type they need. Using the
+wrong type will fail during compilation (similar error as above).
```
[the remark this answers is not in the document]
eladb: Cleaned this up.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+For example, consider the
+[Terrastack](https://github.com/TerraStackIO/terrastack) project or a similar
+one, which uses CDK constructs to define stacks through Terraform. Say we want to
+use an AWS CDK L2 inside a Terraform stack construct:
+
+```ts
+const stack = new terrastack.Stack(...);
+
+// COMPILATION ERROR: `this` is not a `cdk.Construct`.
+new s3.Bucket(this, 'my-bucket');
```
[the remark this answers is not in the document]
eladb: > Have we considered properly defining the constructs.IConstruct interface

What do you mean by properly defining it?

> using that as the type of the scope argument

I have considered it but I think it will increase the cognitive load in defining custom constructs (which is very important to keep low) and is also potentially error prone (i.e. if I *didn't* define the interface, it would still work but it will limit composability).

I rather keep the current model in which implementing a custom construct is just `extend Construct` + `scope: Construct`.

> avoid having to take a peerDependency on constructs?

The peer dependency is required because `construct.Construct` is used as a base class and therefore it's public API is by definition part of the public API of any derived classes.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+```
+
+Being able to create construct compositions from multiple domains is a powerful
+direction for the CDK ecosystem, and this change is required in order to enable
+these use case.
+
+### 4. Non-intuitive dependency requirement
+
+As we transition to
```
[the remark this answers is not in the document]
eladb: Today `constructs` is also defined as a normal dependency. As part of our change in mono-cdk, we will stop taking a normal dependency on it.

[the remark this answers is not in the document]
eladb: added a clarification

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+assume most end-users don't use this directly.
+
+If they need "last minute processing", they would likely use `prepare()` (which
+is also being [TODO: removed]) but not `synthesize()`.
+
+The use case of emitting arbitrary files into the cloud assembly directory is
+weak. The cloud assembly is a well-defined format, and is currently "closed".
+There are no extension points that tools can identify. To that end, just writing
+files to the cloud assembly output directory does not make tons of sense. Having
+said that, it is still possible to do in the same way we plan for
```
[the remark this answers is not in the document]
eladb: What do you mean? Example?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,667 @@
...
+Since only the root construct may have an empty ID, we will also need to assign
+an ID. We propose to use `"Stack"` since we already have fallback logic that
+uses this as the stack name when the stack does not have an ID (see [TODO]).
+
+This change will allow us to remove any special casing we have for stacks in the
+testing framework and throughout the synthesis code path (we have quite a lot of
+that), because we will be able to assume that `Stage.of(construct.node.root)` is
+never `undefined` and has a `synth()` method which returns a cloud assembly.
+
+Since unit tests sometimes use "incremental tests" for synthesized templates,
```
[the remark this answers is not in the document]
eladb: added

text/0192-remove-constructs-compat.md
eladb: yes, old list

text/0192-remove-constructs-compat.md
eladb: I actually decided that I agree with you that `addWarning/error/...` should not be part of `constructs` and moved them to `Logging.of(x).addWarning(msg)`.

text/0192-remove-constructs-compat.md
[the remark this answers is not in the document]
iliapolo: 👍

text/0192-remove-constructs-compat.md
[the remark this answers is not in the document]
iliapolo: Yes. How would I write something to the assembly directory in case i need the full constructed tree? for example the dependency graph, or maybe just a serialization of a sub-tree. (Just wild ideas, I don't really have an explicit use case..)

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,882 @@
...
+});
+
+// now add a "foo" and verify that the synthesized template contains two items
+c1.addFoo('baz');
+expect(stack).toHaveResource('AWS::Resource', {
+  Foos: [ 'bar', 'baz' ]
+});
+```
+
+Since `stage.synth()` (which is called by `expect(stack)`) would reuse the
```
NetaNir, line 521: Why is that true? I assume it does no happen today since we call `stack.synth()`?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,882 @@
...
+off 2.x.
+
+However, this is technically a breaking (behavioral) change for end-users since
+`node.path` and `node.uniqueId`, and their derivatives, will change for trees
+rooted by a `Stack`, and unit tests will need to be updated.
+
+Therefore we propose to introduce this change as a feature flag over the 1.x
+codebase and migrate all of our unit tests (I don't believe we have a way to
+enable feature flags for all unit tests, but we can devise one).
+
```
NetaNir, line 547: can we mitigate this inside the `match`/`haveResourceLike`  implementations? ignoring the `Stack` prefix?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,882 @@
...
+this as soon as the CDK app is created). Therefore, the solution for
+`AssetStaging` is to move the staging logic to the constructor and use
+`Stage.of(this).outdir` to find the output directory of the current stage.
+
+### Implications on end-users
+
+Participation in synthesis is an "advanced" feature of the CDK framework and se
+assume most end-users don't use this directly.
+
+If they need "last minute processing", they would likely use `prepare()` (which
```
NetaNir, line 643: What about a third party construct that requires "last minute changes", the construct vendors does not have access to the `App` (as the construct in created in a consumer application), how will they preform these changes?

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,882 @@
...
+
+> Currently, the AWS CDK also takes `constructs` as a normal dependency (similar
+> to all dependencies), but this is about to change with mono-cdk.
+
+The reason `constructs` will also be required (whether we leave the
+compatibility layer or not) is due to the fact that all CDK constructs
+eventually extend the base `constructs.Construct` class. This means that this
+type is part of their public API and therefore a peer dependency is required
+(otherwise, there could be incompatible copies of `Construct` in the node
+runtime).
```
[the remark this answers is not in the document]
eladb: yes, I'll clarify

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,882 @@
...
+"trait" pattern, which is becoming a common idiom for offering APIs over CDK
+scopes:
+
+```ts
+Aspects.of(scope).apply(aspect);
+```
+
+The major downside of this change is discoverability, but
+`construct.node.applyAspect` is not necessarily more discoverable. We will make
+sure documentation is clear.
```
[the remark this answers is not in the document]
eladb: I like `.add()`. Good call.

text/0192-remove-constructs-compat.md
eladb: Yes, the current implementation has a special-case for trees with stacks as roots.

text/0192-remove-constructs-compat.md
eladb: I found a way to work around this breaking change so it's not breaking any more ("relocate").

text/0192-remove-constructs-compat.md
eladb: They can use aspects.

text/0192-remove-constructs-compat.md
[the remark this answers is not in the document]
eladb: Resolving unless. @RomainMuller unresolve if you are not satisfied

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+### What can we do on 1.x?
+
+- We will migrate the 1.x branch to `Aspects.of(x)` and add a deprecation
+  warning to `this.node.applyAspect()`.
+- Introduce `Tags.of(x).add()` and add a deprecation warning to `Tag.add()`.
+
+## 03-DEPENDABLE
+
+The `constructs` library supports dependencies through `node.addDependency` like
+in 1.x, but the API to implement `IDependable` has been slightly changed.
```
nija-at, line 446:
```suggestion
in 1.x, but the API to implement `IDependable` has slightly changed.
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+([synthesis.ts](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/core/lib/private/synthesis.ts)).
+The main reason was that we needed to "break" the recursion in various
+domain-specific points (e.g. stages, nested stacks) which meant that the generic
+logic of "traverse the entire tree and call `synthesize`" did not hold. In
+`cdk8s`, the support for [chart
+dependencies](https://github.com/awslabs/cdk8s/blob/ef95b9ffce8a39200e028c2fe8acc55a9915161c/packages/cdk8s/src/app.ts#L39)
+required that the name of the output manifest will be determined based on the
+topologic order at the app level. Here again, the generic approach failed.
+
+In leu of those failures, we decided that there is no additional value in
```
nija-at, line 618:
```suggestion
In lieu of those failures, we decided that there is no additional value in
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+order to stage the asset. In fact, all the information required already exists
+during initialization. The only missing information is the cloud assembly output
+directory, and this information is actually known during initialization (we know
+this as soon as the CDK app is created). Therefore, the solution for
+`AssetStaging` is to move the staging logic to the constructor and use
+`Stage.of(this).outdir` to find the output directory of the current stage.
+
+### Implications on end-users
+
+Participation in synthesis is an "advanced" feature of the CDK framework and se
```
nija-at, line 647:
```suggestion
Participation in synthesis is an "advanced" feature of the CDK framework and we
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+If they need "last minute processing", they would likely use `prepare()` (which
+is also being [removed](#06-no-prepare-the-prepare-hook-is-no-longer-supported))
+but not `synthesize()`.
+
+The use case of emitting arbitrary files into the cloud assembly directory is
+weak. The cloud assembly is a well-defined format, and is currently "closed".
+There are no extension points that tools can identify. To that end, just writing
+files to the cloud assembly output directory does not make tons of sense. Having
+said that, it is still possible to do in the same way we plan for
+`AssetStaging`.
```
nija-at, line 659: Would suggest dropping the last sentence. Otherwise it's a cliff hanger.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+  - [Alternatives considered](#alternatives-considered)
+- [Adoption Strategy](#adoption-strategy)
+- [Unresolved questions](#unresolved-questions)
+- [Future Possibilities](#future-possibilities)
+- [Implementation Plan](#implementation-plan)
+  - [Preparation of 1.x](#preparation-of-1x)
+  - [constructs 4.x](#constructs-4x)
+  - [2.x Work](#2x-work)
+
+# Release Notes
```
nija-at, line 71: What about the developer guide, and our documentation.

We should include a section in this RFC (or maybe in the one for v2) on how that is going to look when both v1 and v2 are present, just to make sure we don't forget about it.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+
+A side effect of adding a `App` parent to "root" stacks is that construct paths
+and unique IDs in unit tests will now change. In the above example,
+`foo.node.path` will change from `MyFoo` to `Stack/MyFoo`. Additionally, tests
+for resources that utilized `node.uniqueId` to generate names will also change
+given `uniqueId` is based on the path.
+
+
+To address this, we propose to introduce a new capability in `constructs` to
+"**relocate**" a scope to a different path. This capability will also be useful
```
nija-at, line 532: Sounds dangerous as a feature available to customers.
Maybe more dangerous than stack mutation during `prepare()`?

text/0192-remove-constructs-compat.md
eladb: added tasks

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+For most CDK libraries and apps, you will likely just need change this:
+
+```ts
+import { Construct } from '@aws-cdk/core';
+```
+
+With this:
+
+```ts
+import { Cosntruct } from 'constructs';
```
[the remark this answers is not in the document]
eladb: Fixed

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+`@aws-cdk/core.ConstructNode` | `constructs.Node`
+`Construct.isConstruct(x)` | `x instanceof Construct`
+`construct.node.applyAspect(aspect)` | `Aspects.of(construct).apply(aspect)`
+`@aws-cdk/core.IDependable` | `constructs.IDependable`
+`@aws-cdk/core.DependencyTrait` | `constructs.Dependable`
+`@aws-cdk.core.DependencyTrait.get(x)` | `constructs.Dependable.of(x)`
+`construct.node.dependencies` | Is now non-transitive
+`construct.addMetadata()` | Stack trace not attached by default
+`ConstructNode.prepare(node)`, `onPrepare()`, `prepare()` | Not supported
+`ConstructNode.synthesize(node)`, `onSynthesize()`, `synthesize()` | Not supported
```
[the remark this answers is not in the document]
eladb: Ok added.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+`construct.onValidate()`, `construct.validate()` hooks | Implement `constructs.IValidation` and call `node.addValidation()`
+`ConstructNode.validate(node)` | `construct.node.validate()`
+
+The following sections describe all the related breaking changes and details
+migration strategies for each change.
+
+## 01-BASE-TYPES: Removal of base types
+
+The following `@aws-cdk/core` types have stand-in replacements in `constructs`:
```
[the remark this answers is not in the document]
eladb: The ones listed in this section all have stand-in replacements

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+```
+
+This is still a supported idiom, but in 2.x these root stacks will have an
+implicit `App` parent. This means that `stack.node.scope` will be an `App`
+instance, while previously it was `undefined`. The "root" stack will have a
+construct ID of `Stack` unless otherwise specified.
+
+> Technically, this change should have resulted in the `node.path` and
+> `node.uniqueId` of all constructs defined within such trees to change, but we
+> are utilizing `node.relicate()` to maintain the current behavior.
```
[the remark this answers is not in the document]
eladb: fixed

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+
+## 05-METADATA-TRACES: Stack traces no longer attached to metadata by default
```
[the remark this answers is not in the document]
eladb: This is the new phrasing:

For performance reasons, the `construct.node.addMetadata()` method will *not*
attach stack traces to metadata entries. Stack traces will still be associated
with all `CfnResource` constructs and can also be added to custom metadata using
the `stackTrace` option:

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+```
+
+After:
+
+```ts
+class MyConstruct extends Construct {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    this.node.addValidation({ validate: () => [ 'validation-error' ] });
```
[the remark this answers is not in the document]
eladb: That's not a predicate. A predicate usually returns `true/false`. This guy returns a list of error messages.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+The main motivation for this change is to enable composition of AWS CDK
+constructs with constructs from other domains.
+
+It is currently impossible to define AWS CDK constructs within a non-AWS-CDK
+construct scope.
+
+For example, consider the
+[Terrastack](https://github.com/TerraStackIO/terrastack) project or a similar
+one, which uses CDK constructs to define stacks through Terraform. Say we want to
+use an AWS CDK L2 inside a Terraform stack construct:
```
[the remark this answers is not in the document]
eladb: Yes, we will need to implement a mapping layer that will map CFN resources to TF resources. We are working with HashiCorp to define what this looks like. Nevertheless, the construct composition is required.

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,888 @@
...
+
+> Currently, the AWS CDK also takes `constructs` as a normal dependency (similar
+> to all dependencies), but this is about to change with mono-cdk.
+
+The reason `constructs` will also be required (whether we leave the
+compatibility layer or not) is due to the fact that all CDK constructs
+eventually extend the base `constructs.Construct` class. This means that this
+type is part of their public API and therefore a peer dependency is required
+(otherwise, there could be incompatible copies of `Construct` in the node
+runtime).
```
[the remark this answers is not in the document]
eladb: added clarification:

The removal of the compatibility layer means that now anyone who uses the AWS
CDK will need to explicitly use the `constructs.Construct` type (even for
trivial apps), and therefore it would "make sense" for them to take a dependency
on the `constructs` library.

text/0192-remove-constructs-compat.md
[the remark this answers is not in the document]
eladb: done

text/0192-remove-constructs-compat.md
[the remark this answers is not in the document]
eladb: Changed.

text/0192-remove-constructs-compat.md
eladb: Fixed

text/0192-remove-constructs-compat.md
eladb: Reverted and decided to use `Default` as the name of the root stack, which is already ignored by `uniqueId` yey!

text/0192-remove-constructs-compat.md
eladb: Thanks. Fixed.

text/0192-remove-constructs-compat.md
eladb: Fixed

text/0192-remove-constructs-compat.md
eladb: Added.

text/0192-remove-constructs-compat.md
eladb: Modified to include the details.

text/0192-remove-constructs-compat.md
```diff
@@ -485,6 +517,76 @@ Alternatives considered:
 - **Automatic merge resolution and import organization**: requires research and
   development, not necessarily worth it.
 
+## 10-CONSTRUCT-NODE
+
+As we realize that we won't be able to release additional major versions of the
+"constructs" in order to ensure interoperability between domains is always
+possible, we need to closely examine the API of this library.
```
eladb, line 524:
```suggestion
As we realize we won't be able to release additional major versions of the
"constructs" in order to ensure that interoperability between domains is always
possible, we need to closely examine the API of this library.
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+- **Do nothing in 1.x**: will incur an ongoing maintenance cost of the 1.x -> 2.x
+  merge conflicts.
+- **Automatic merge resolution and import organization**: requires research and
+  development, not necessarily worth it.
+
+## 10-CONSTRUCT-NODE
+
+As we realize that we won't be able to release additional major versions of the
+"constructs" in order to ensure interoperability between domains is always
+possible, we need to closely examine the API of this library.
```
eladb, line 524:
```suggestion
Since we won't be able to release additional major versions of the
"constructs" library (in order to ensure interoperability between domains is always
possible), we need to closely examine the API of this library.
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+In particular, the API of the `Construct` class, which is the base of all
+constructs in all CDKs, should be as small as possible in order not to "pollute"
+domain-specific APIs introduced in various domains.
+
+In many cases (cdk8s, cdktf), constructs are *generated* on-demand from
+domain-specific API specifications. In such cases, we need to ensure that the
+API in `Construct` does not conflict with generated property names or methods.
+
+The current API of `Construct` in the base class (2.x, 3.x) only includes a
+bunch of protected `onXxx` methods (`onPrepare`, `onValidate` and
```
eladb, line 535:
```suggestion
few protected `onXxx` methods (`onPrepare`, `onValidate` and
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+`onSynthesize`). Those methods will be removed in 10.x
+([prepare](#06-no-prepare) and [synthesize](#07-no-synthesize) are no longer
+supported and [validate](#08-validation) will be supported through
+`addValidation()`).
+
+In AWS CDK 1.x, to construct API, is available under `myConstruct.node`. This
```
eladb, line 541:
```suggestion
In AWS CDK 1.x the construct API is available under `myConstruct.node`. This
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+API has been intentionally removed from "constructs" 3.x in order to allow the
```
eladb, line 542:
```suggestion
API has been intentionally removed when we extracted "constructs" from the AWS CDK in order to allow the
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+compatibility layer in AWS CDK 1.x to use property name and expose the shim
+type. The base library currently offers `Node.of(scope)` as an alternative - but
+this API is cumbersome to use and not discoverable. In evidence, in CDK for
+Terraform, they chose to offer
+[`constructNode`](https://github.com/hashicorp/terraform-cdk/blob/5becfbc699180adfe920dec794200bbf56dda0a7/packages/cdktf/lib/terraform-element.ts#L21)
+in `TerraformElement` as a sugar for `Node.of()`.
```
eladb, line 548:
```suggestion
API has been intentionally removed when we extracted "constructs" from the AWS CDK
in order to allow the compatibility layer in AWS CDK 1.x to use the same property name
and expose the shim type (jsii does not allow changing the type of a property in a subclass). 

The base library currently offers `Node.of(scope)` as an alternative - but this API is cumbersome
to use and not discoverable. In evidence, in CDK for Terraform, they chose to offer [`constructNode`]
in `TerraformElement` as a sugar for `Node.of()`.

[`constructNode`]: https://github.com/hashicorp/terraform-cdk/blob/5becfbc699180adfe920dec794200bbf56dda0a7/packages/cdktf/lib/terraform-element.ts#L21
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+domain-specific APIs, we propose to introduce this API under the name
+`construct` (of type `Node`).
+
+This has a few benefits:
+
+1. It signals that "this is the construct API" and more uniquely identifies with
+   it.
+3. The chance for conflicts with domain-specific names is minimized.
+3. We can introduce this API without breaking existing code while deprecating
+   `node` in the AWS CDK.
```
eladb, line 569:
```suggestion
1. It's semantically signals that "this is the construct API".
2. The chance for conflicts with domain-specific names is low ("construct" is not prevalent).
3. We can introduce this API while deprecating `node` in AWS CDK 1.x.
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+
+The main downside is that is is **a breaking change** in AWS CDK 2.x. There is
+likely quite a lot of code out there (a [few
+hundred](https://github.com/search?q=cdk+node.addDependency++extension%3Ats&type=Code&ref=advsearch&l=&l=)
+results for an approximated GitHub code search).
```
eladb, line 574:
```suggestion
The main downside is that it is **a breaking change** in AWS CDK 2.x. There is
likely quite a lot of code out there (a [few
hundred](https://github.com/search?q=cdk+node.addDependency++extension%3Ats&type=Code&ref=advsearch&l=&l=)
results for an approximated GitHub code search).
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
+---
+feature name: remove-constructs-compat
+start date: 2020-07-05
+rfc pr: (leave this empty)
```
eladb, line 4:
```suggestion
rfc pr: [#195](https://github.com/aws/aws-cdk-rfcs/pull/195)
```

text/0192-remove-constructs-compat.md
```diff
@@ -0,0 +1,1095 @@
...
+structural typing (it's structurally identical to `any`).
+
+As we evaluate this use case for constructs 10.x, we would like to restore the
+ability to access the construct API from a property of `Construct`, and use that
+property as the single marker that represents a construct type (`IConstruct`).
+
+To reduce the risk of naming conflicts (e.g. see [Terraform CDK
+issue](https://github.com/hashicorp/terraform-cdk/pull/230)) between `node` and
+domain-specific APIs, we propose to introduce this API under the name
+`construct` (of type `Node`).
```
RomainMuller, line 563: So... let me summarize this: all classes that extend `Construct` will have a `construct` property that returns an instance of `Node`? I would naturally have expected `construct` to return an instance of `Construct` (basically, `this`), and the naming you're proposing strikes me as surprising and counter-intuitive.

eladb: The property name should be considered in the context of its usage.

Do you find this counterintuitive:

```ts
bucket.construct.addDependency(role);
```

As oppose to:

```ts
bucket.node.addDependency(role);
```

I actually think the former is much more intuitively and reads much better.

We actually have a naming guideline that says that names should _not_ include the type name because it's redundant information in strongly typed environments.

Do you have an alternative suggestion?

Perhaps we can rename `Node` to something else?

RomainMuller: As discussed... I don't feel *extremely* strong against this. It does feel counter-intuitive to me but I have no idea whether I'm representative of users or not in this respect.

For the record, I think `this.construct.addDependency(role)` reads **super good**, but for some reason the magic weakens when the receiver is not `this`.

Additionally, I cannot come up with a *better* suggestion for a name... so there's that, too.

rix0rrr: I know it's a little late, but I don't agree with this rename.

I would expect a field called `construct` to point to an object that IS a Construct. I like what they did in TerraCDK, `constructNode` seems like a good compromise.

eladb: The main reason we have this property is in order to group all the API of `Construct` under a single namespace so it won't "pollute" the APIs of subclasses, which are oftentimes generated classes in CDKs.

The conceptual mental model is that this is "the construct API". The fact that the type is `Node` is actually not in line with this mental model.

If I am looking at this from a usage point of view, I think `construct` makes more sense:

```ts
this.construct.id
this.construct.path
this.construct.addDependency(other)
this.construct.addMetadata
```

This is actually more natural then:

```ts
this.node.id
this.node.path
...
```

Additionally, we actually have a guideline *against* using the type name in symbol names because type information is an intrinsic part of the API (we prefer `handler: lambda.Function` over `handlerFunction: lambda.Function`).

I would rather rename the `Node` class (maybe `ConstructApi`) instead of renaming this property.

rix0rrr: For the record, after some discussion I think I would have preferred for this member to be called `constructTree`, as this is the member that consumers use to access the features of the "construct tree" (which is a CDK concept that people are, or should be, familiar with which is distinct from the concept of a "construct").

We get things like:

```ts
// All perfectly natural
bucket.constructTree.findChild()
bucket.constructTree.removeChild()
bucket.constructTree.defaultChild
bucket.constructTree.findAll()
bucket.constructTree.children
bucket.constructTree.addWarning()

// Better than what we have now, this name emphasizes the tree traversal involved in these operations
bucket.constructTree.setContext()
bucket.constructTree.applyAspect()
bucket.constructTree.addDependency() 
 
// These now have not such great names, because they apply to the "current node" and don't really have
// strong tree-ish connotations. Don't think any of them are *horrible* but they are certainly not fantastic.
bucket.constructTree.id
bucket.constructTree.path
bucket.constructTree.uniqueId
```

eladb: I don't think CDK users are required to understand the tree concept. In fact, we have traditionally tried to talk about "scopes" (I.e "the construct is created within the scope of another construct"). From a CS perspective this is indeed a tree, but I don't see how this adds much value to users.

As mentioned above, the primary reason for this property is to group all the construct APIs under an api namespace so they don't pollute or conflict the public api of subclasses, which is the primary programming pattern we use in CDKs.

rix0rrr: There's a difference between not needing to know that there is a tree structure involved (until you do), and pretending that the tree structure doesn't exist by studiously avoiding calling it a tree.

You might be under the impression that my argument is one about purity of CS concepts. It is not. 

It is about acknowledging the fact that this structure exists, that it underpins the way the CDK works in a way which you can oftentimes ignore, but sometimes is quite relevant and requires people to understand it (see Stages, Stacks, Aspects, LogicalIDs, default children and escape hatches, see `tree.json` and the VSCode plugin, ...) and that there is a term available which accurately describes the semantics, expectations, and the sort of API that people can expect from this structure that they are already intimately familiar with from other programming tasks.

I think being vague and coy and pretending there is no tree when we all full well know that there is, or using the word "construct" to refer to two distinct concepts, is not going to help users form a mental model.

I'm also not saying they won't be able to. Given enough exposure and enough examples, over time people are going to eventually be trained to write the right code. In that sense, ultimately it doesn't even matter what it's called, be it `node`, `construct`, or `abc`, people will learn to cope. 

------

I think I've laid out my reasoning for why I think it's advantageous to call the beast by its name. It's completely fine if you disagree.

eladb: Cool, thanks for the effort in explaining.

The plan is to introduce `construct` and deprecate `node` over the `1.x` branch in short order. This will give us some "bake time" of a few months until we completely remove `node` in 2.x. This will allow us to get feedback from users in case the mental model is confusing or insufficient.

skorfmann: Just stumbled upon this - thanks for making the effort of changing the name 👍

Issue #107, opened 2020-01-31
Publish a Construct Library Module Lifecycle document


ccfife:
|PR|Champion|
|--|--------|
|#107 | @ccfife |

## Description
The purpose of this document is to outline the process used to prioritize and mature modules in the AWS-CDK Construct Library. It outlines how we graduate modules from CloudFormation-only resources (L1) to modules that contain high-level resource abstractions (L2) and stabilize them for GA release.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

Issue #28, opened 2019-12-09
Construct library graduation process


eladb:
|PR|Champion|
|--|--------|
|# |        |

## Description

Devise a methodology to graduate a module

- Bake time (age)
- Bugs
- Feature-set
- API sign-off
- Documentation
- Sample
- Bug Bash
- Usability

Devise a metric of "amount" of experimentation we have, and track it so we allocate capacity to maintain continuously.

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

ccfife:
Closing because this is a duplicate of: https://github.com/aws/aws-cdk-rfcs/issues/107

PR #215, opened 2020-07-29
chore: remove duplicate entry for construct library graduation

shivlaks:
duplicated by RFC #107 is in this table which is classified as `done`

---


shivlaks:
> This table is generated. All you need to do is remove the `status/proposed` label from #28 and then a PR will be created with an updated table.

ah neat, i didn't know thanks for the heads up!

shivlaks:
discarding this as i've dropped the labem `status/proposed` from #28

PR #241, opened 2020-08-25
rfc 79: cdk v2.0 - change strategy on how deprecated APIs are handled

nija-at:
Two changes to the RFC:
- The branch name `next` has been updated to `v2` throughout the RFC.
The name `v2` is a lot clearer on what the branch indicates than slightly
more vague `next`.
- The strategy on how deprecated APIs are handled has been revamped to
be less manual and more automated. A one-time clean up towards the end
of the project is included.

---

### DIFF
--- a/text/0079-cdk-2.0.md
+++ b/text/0079-cdk-2.0.md
@@ -131,7 +131,7 @@ v1 will be front-loaded. This includes all work that can be prepared without bre
 
 ## Forking v2 out of v1
 
-The forking point will be achieved by cutting a new `next` branch from the current tip of the `master` branch, where the
+The forking point will be achieved by cutting a new `v2` branch from the current tip of the `master` branch, where the
 following changes will be made right away:
 
 1. the repository-wide `version` configuration will be updated to `2.0.0-alpha` (the pre-release identifier could be
@@ -168,8 +168,8 @@ validated. Releases will be automatically published, however using the `next` NP
 [github packages]: https://github.com/features/packages
 
 As long as the codebases have not diverged too much, it should be possible to continue forward-porting new developments
-on top of the `next` branch by simply cherry-picking new commits from the `master` branch. Alternatively, it might be
-possible to simply periodically merge `master` into `next`. This task can easily be automated.
+on top of the `v2` branch by simply cherry-picking new commits from the `master` branch. Alternatively, it might be
+possible to simply periodically merge `master` into `v2`. This task can easily be automated.
 
 A new standard operating procedure (SOP / MCM) will be made to support manually releasing a CDK `v2` release until this
 process is fully automated. Additionally, canaries and integration tests need to be prepared to ensure the ongoing
@@ -184,9 +184,9 @@ _Thursdays_), manually resolving merge conflicts as needed.
 
 The AWS Construct Library will be released as a single artifact, instead of a collection of more than 100 libraries. The
 `monocdk-experiment` package has the tools necessary to re-package the existing _hypermodular_ libraries into a single
-packaging. Re-using that same process will allow keeping the difference between `master` and `next` as constrained as
+packaging. Re-using that same process will allow keeping the difference between `master` and `v2` as constrained as
 possible (reducing the need for manual intervention in forward-porting features). No additional packaging work is needed
-once the `next` branch has been forked from `master`, as the renaming of `monocdk-experiment` to `aws-cdk-lib` and
+once the `v2` branch has been forked from `master`, as the renaming of `monocdk-experiment` to `aws-cdk-lib` and
 marking all bundled packages `private` is taken care of at that time.
 
 However, since users will all use a single `aws-cdk-lib` library, the granularity of information collected via the
@@ -210,15 +210,23 @@ The definition of `@deprecated` in the CDK specifies that annotated APIs will be
 release. Consequently, we will systematically remove all `@deprecated` APIs from the codebase upon cutting a new major
 version.
 
-This task is expected to be easy to carry out by hand. Searching the codebase for `@deprecated` and deleting the
-annotated code, as well as any private function that is no longer used. In the event non-`@deprecated` code is found to
-depend on `@deprecated` code, the implementation in `master` will be fixed to no longer defer into the `@deprecated`
-feature, and that fix will be forward-ported into the `next` branch in the exact same way that other changes are.
+Tooling will enforce that no deprecated APIs methods, properties, interfaces or classes will be used within the CDK
+repository. This also applies to tests, with an option for individual overrides, in the case of testing the deprecated
+APIs themselves.
 
-In order to prevent accidental introduction (or re-introduction) of `@deprecated` code in the `next` branch while
-forward-porting new developments from `master`, a built-time check will be added that scans the codebase for
-`@deprecated` (for example using `grep`), failing the build if occurrences are found. This is not done through
-inspection of the _jsii assemblies_ as `@deprecated` code can exist outside of exported API signatures.
+The `jsii` binary will accept a `--no-deprecated` flag that will skip deprecated APIs from the `.jsii` binary. In
+conjunction, deprecated APIs will be stripped off the typescript files as part of producing the monolithic package,
+`aws-cdk-lib`, similar to [import rewrites].
+
+When CDK v2.0 become generally available, the source code will be updated to remove all use of `@deprecated`.
+This will occur as part of flipping `master` to point to the `v2` branch, with a code freeze of at most 2 days
+instituted.
+Additionally, the `--no-deprecated` flag and the typescript transform will be dropped, now enabling API deprecations on
+CDKv2.
+
+**Why not remove these by hand?** See [alternatives section](#removal-of-deprecated-apis---why-not-remove-these-by-hand).
+
+[import rewrites]: https://github.com/aws/aws-cdk/tree/master/packages/%40monocdk-experiment/rewrite-imports
 
 ### Resetting v1 feature flags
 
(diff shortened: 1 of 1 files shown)


nija-at:
Moving this to draft as I work through some of the details.

text/0079-cdk-2.0.md
```diff
@@ -131,7 +131,7 @@ v1 will be front-loaded. This includes all work that can be prepared without bre
 
 ## Forking v2 out of v1
 
-The forking point will be achieved by cutting a new `next` branch from the current tip of the `master` branch, where the
+The forking point will be achieved by cutting a new `v2` branch from the current tip of the `master` branch, where the
```
eladb, line 134: What's the rationale for changing from `v2` to `next`? Please indicate in the PR description

text/0079-cdk-2.0.md
```diff
@@ -210,15 +210,23 @@ The definition of `@deprecated` in the CDK specifies that annotated APIs will be
 release. Consequently, we will systematically remove all `@deprecated` APIs from the codebase upon cutting a new major
 version.
 
-This task is expected to be easy to carry out by hand. Searching the codebase for `@deprecated` and deleting the
-annotated code, as well as any private function that is no longer used. In the event non-`@deprecated` code is found to
-depend on `@deprecated` code, the implementation in `master` will be fixed to no longer defer into the `@deprecated`
-feature, and that fix will be forward-ported into the `next` branch in the exact same way that other changes are.
+Tooling will enforce that no deprecated APIs methods, properties, interfaces or classes will be used within the CDK
```
eladb, line 213: Can you describe these tools? Are you referring to the `Annotations` API that I recently added?

nija-at: Clarified in the new revision.

text/0079-cdk-2.0.md
```diff
@@ -210,15 +210,47 @@ The definition of `@deprecated` in the CDK specifies that annotated APIs will be
...
+Before creating the `v2` branch,
+
+- The `jsii` assembler will be updated to accept a `--no-deprecated` flag that will strip out all deprecated methods,
+properties, classes and interfaces. This will be achieved via a typescript transform that is run before assembly
+generation. A proof of concept for this transform is available [here](https://github.com/nija-at/cdk-deprecated-api-poc).
+
+- `cdk-build` will be updated to recognize the environment variable `CDK_BLOCK_DEPRECATIONS` and set the
+`--no-deprecated` flag on the `jsii` assembler.
+
+- All current usage of deprecated APIs within the CDK repository will be removed. In addition, a new check will be
```
skinny85, line 232: What about things that are deprecated because we think it's a bad idea to use them, but that we don't want to remove? (One example that springs to my mind are [unversioned Engines in RDS](https://github.com/aws/aws-cdk/blob/34f40b9fb4b467c9c7af290400efa2c8ac8c8e10/packages/%40aws-cdk/aws-rds/lib/cluster-engine.ts#L449-L471), but I'm sure there's more of them)

We need some mechanism for preserving deprecated things that we don't want to remove in v2.0.

text/0079-cdk-2.0.md
```diff
@@ -210,15 +210,47 @@ The definition of `@deprecated` in the CDK specifies that annotated APIs will be
...
+added to pull requests on the `master` branch, that will run a full CDK build **and test** with the
+`CDK_BLOCK_DEPRECATIONS` environment variable set. The typescript compilation and execution of the CDK's comprehensive
+unit test suite should highlight any usage of deprecated APIs.
+
+- In order to support unit tests that verify the behaviour of deprecated APIs, `cdk-test` will be updated to recognize
+a new pragma `!/// !cdk-test:allow-deprecations` and the `CDK_BLOCK_DEPRECATIONS` environment variable. Tests holding
+this pragma will be skipped when `CDK_BLOCK_DEPRECATIONS` is set.
```
skinny85, line 239: Sorry, why would they be skipped? I think they should still be executed, just not causing the build to fail because of using deprecated APIs?

nija-at: These **will be** executed as part of regular PR validation build and skipped as part of the '2.0 compat' validation build.

They will be skipped in the latter because the APIs won't exist in the typescript output and JSII assembly when `--no-deprecations` flag is set.

text/0079-cdk-2.0.md
nija-at: What's this new category of deprecated APIs? 

If it's a bad idea for customers to use them, they should be removed, and customers should use the better APIs that should now be available.

skinny85: > What's this new category of deprecated APIs?
> 
> If it's a bad idea for customers to use them, they should be removed, and customers should use the better APIs that should now be available.

That's not really an option, because L2s should model the entire service area of the service, even things we consider not to be good practices.

nija-at: Still not clear what you mean.

In the example you've provided, they would use the versioned engines API, no?

skinny85: Yes, but that would mean it's impossible to create an Instance/Cluster without a version, violating the above tenet.

nija-at: There should be a way for customers to do this using a non-deprecated API. Either this API should be removed as deprecated or another option needs to be provided.

Depreciation in the cdk, means that a better API or customer experience is available, or the deprecated option is no longer maintained and will not be supported in the future. 

If neither, the API is not deprecated. Maybe you're looking for a new category, although I'm still not clear what that is, from this discussion thus far.

rix0rrr (dismissed):
I think I found a situation where this strategy is going to break:

https://github.com/aws/aws-cdk/blob/443948164e09aaa81c094c013b32aa1f67b69570/packages/%40aws-cdk/core/lib/stack-synthesizers/legacy.ts#L75

This is where we purposely call a deprecated API for backwards compatibility, and that will need to stay in.

I am okay with the general gist of this strategy, but we're going to need an exclude list of some sort.

nija-at:
> This is where we purposely call a deprecated API for backwards compatibility, and that will need to stay in.

Why not change that line to `this.stack.synthesizer.addFileAsset(asset);`?

text/0079-cdk-2.0.md
skinny85: I don't think you're correct here, and I'm not sure where did you find this definition of "deprecated in cdk" that you're citing here.

Another example: old runtimes in Lambda (I don't see a good reason why would we ever delete those, as customers who have Lambdas created with them can still edit them). We will have the same situation in RDS for old versions of Oracle.

skinny85: Another example: the very widely used "old" `CloudFrontWebDistribution` construct from the `@aws-cdk/aws-cloudfront` module (probably don't want to remove that one).

nija-at: I see what you mean. So, there are indeed two kinds of deprecated - one that comes from the cdk and the other that is bubbled up from what AWS services consider deprecated.

We have 3 options to handle such cases - (1) build an exclusion list, (2) remove the use of the `@deprecated` tags and replace it with a text such as 'DEPRECATED' in the method's documentation, or (3) still remove them, forcing customers to use an escape hatch or a different more cumbersome API.

eladb: I vote for not using `@deprecated` for such cases but instead document that this is not a recommended practice and/or add some synthesis warning.

The `@deprecated` label means that this API is going to be removed in the next major version.

rix0rrr:
The point is: people used to be able to override a method on `Stack`. That's `@deprecated`, but still supported.

text/0079-cdk-2.0.md
rix0rrr: We probably shouldn't overload `@deprecated` then, but it is probably useful to still have a mechanism to annotate them as such.

nija-at: For the time being, limiting the scope for v2 work, I'm going to replace the `@deprecated` tag with a 'DEPRECATED:' text on the tsdoc. We can consider a different annotation and/or synthesis warnings separately.

skinny85: I don't like that, because a `'DEPRECATED'` text in the JSDocs will not sure anything in the IDE when using those classes/methods.

How about something in `@deprecated` that the automation respects, and doesn't remove? Like `@deprecated(preserve)`? (exact name can be changed of course). It would also serve @rix0rrr's need from https://github.com/aws/aws-cdk-rfcs/pull/241#issuecomment-686301830.

eladb: If the api will not be removed in the next major version is not deprecated. Maybe it's "NOT RECOMMENDED TO USE" but the term deprecated should not mean two different things. Keep it simple please

nija-at: +1 to @eladb. It's not clear how much the IDE helps the user today anyway.

skinny85: Usually, IDEs mark the usage of something deprecated pretty clearly (here's an example in Intellij).
Screen Shot 2020-09-04 at 8 51 12 AM

skinny85: > If the api will not be removed in the next major version is not deprecated. Maybe it's "NOT RECOMMENDED TO USE" but the term deprecated should not mean two different things. Keep it simple please

This is not the definition of "deprecated" that I know. For example, there are [APIs still available in Java](https://docs.oracle.com/javase/6/docs/api/org/xml/sax/Parser.html) that have been deprecated since the 90s.

eladb: > Usually, IDEs mark the usage of something deprecated pretty clearly (here's an example in Intellij).
> 
> Screen Shot 2020-09-04 at 8 51 12 AM
> 
> 

That's not the case for VSCode unfortunately, which is probably the most used by our users.

> For example, there are APIs still available in Java that have been deprecated since the 90s.

The fact that java decided to not remove these between major versions doesn't mean we should leave deprecated "debt" behind.

I think this is just about the semantics and I really prefer that keep the term "deprecated" reserved for APIs that will (most likely) be removed in the next major release (unless there are very unique circumstances that may require keeping it for another round). The api you are describing is not being deprecated but perhaps not recommended to use. If possible we should deprecate and remove it, if not, we should state in the docs and even add a warning if it's used.

Issue #244, opened 2020-08-26
Migration path for EKS Developer Preview


iliapolo:
|PR|Champion|
|--|--------|
|#245 |   @eladb      |

## Description

When we introduce breaking changes to `experimental` libraries that require resource replacements,
or in general are considered highly intrusive, we should provide a clear migration path.

## Progress

- [x] Tracking Issue Created
- [X] RFC PR Created 
- [X] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

PR #245, opened 2020-08-26
RFC: Migration Path For EKS Developer Preview

iliapolo:
* Tracking issue: #244
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/epolon/eks-dev-preview-migration/text/0244-eks-developer-preview-migration-path.md)

---


text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,91 @@
...
+When users upgrade their version of `aws-eks`, they will now receive the new intrusive version. This version will detect that the user had previously been using the old `aws-eks` (which is now `aws-eks-experimental`) module, and display the following error:
+
+```console
+The aws-eks module has undergone a name change, to use the previously named aws-eks module, please change your imports and dependencies to use aws-eks-experimental.
+If you'd like to use the new aws-eks module, add the following key to your cdk.json file: "@aws-cdk/aws-eks:dev-preview: true".
+```
+
+The user will then choose to either change imports to point to the `aws-eks-experimental` module, and continue working without interruption. Or, add the necessary feature flag to `cdk.json` and continue with the new module which will replace the cluster.
+
+# Future Possibilities
```
eladb, line 85: One thing that's worth talking about is how do we make sure that the stuff we actually release as developer preview is working? It feels like there might be risk in accumulating many changes on the `eks-dev-preview` branch and then releasing all of them together as "developer preview" and then realizing something is horribly broken and having to release a breaking change to fix.

eladb (changes requested):

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,102 @@
...
+That is, rename the current `eks.Cluster` to `eks.ExperimentalCluster` and implement everything we need in `eks.Cluster`.
+We have actually already done this once with `eks.LegacyCluster`.
+
+The benefit here is that users won't need to change dependencies. However, this change would also require a name change and
+duplication of any class that interacts with `eks.Cluster` since those classes may not work with both `eks.Cluster` and `eks.ExperimentalCluster`.
+
+This approach feels more fragile and doesn't provide much value since the user would still incur a breaking change.
+It would just be in the form of a class name change, instead of a module name change.
+
+## Rename new module
```
ericzbeard, line 77: Alternately, we could put the new developer preview module in a separate repository and vend it as an independent package. When we go to GA with CDK 2, the old module could be replaced entirely. The CLI warning is probably still a good idea. This way, anyone who does not want to replace their clusters stays on 1.x until they are ready.

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Rename `eks.Cluster`
+
+Instead of spinning off to a new module, we can spin off to a new construct.
+
+That is, rename the current `eks.Cluster` to `eks.ExperimentalCluster` and implement everything we need in `eks.Cluster`.
+We have actually already done this once with `eks.LegacyCluster`.
+
+The benefit here is that users won't need to change dependencies. However, this change would also require a name change and
+duplication of any class that interacts with `eks.Cluster` since those classes may not work with both `eks.Cluster` and `eks.ExperimentalCluster`.
```
rix0rrr, line 72: Should have been an `ICluster`. Is it still true in that case?

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,102 @@
...
+
+When users upgrade their version of `aws-eks`, they will now receive the new intrusive version. This version will detect that the user
+had previously been using the old `aws-eks` (which is now `aws-eks-experimental`) module, and display the following error:
+
+```console
+The aws-eks module has undergone a name change, to use the previously named aws-eks module, please change your imports and dependencies to use aws-eks-experimental.
+If you'd like to use the new aws-eks module, add the following key to your cdk.json file: "@aws-cdk/aws-eks:dev-preview: true".
+```
+
+The user will then choose to either change imports to point to the `aws-eks-experimental` module, and continue working without interruption.
```
rix0rrr, line 95: To be clear, you will ALWAYS get this error if you don't add that key, is that right? There is no actual check on anything?

(I mean, how could there be...)

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,102 @@
...
+related issue: https://github.com/aws/aws-cdk-rfcs/issues/244
+---
+
+# Summary
+
+This RFC proposes a migration path for existing users following the upcoming transition of the `aws-eks` module from *Experimental* to *Developer Preview*.
+
+# Motivation
+
+As part of transitioning the `aws-eks` module to *Developer Preview*, we need to introduce a few breaking changes that require
```
rix0rrr, line 14: I'm pretty interested in the specifics of this case. Breaking  changes and replacement don't *have* to coincide. Can you go a little into the changes that prompted this RFC? Have we already done the work to see if we can avoid this process completely?

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Rationale and Alternatives
+
+### Rename `eks.Cluster`
+
+Instead of spinning off to a new module, we can spin off to a new construct.
+
+That is, rename the current `eks.Cluster` to `eks.ExperimentalCluster` and implement everything we need in `eks.Cluster`. We have actually already done this once with `eks.LegacyCluster`.
+
+The benefit here is that users won't need to change dependencies. However, this change would also require a name change and duplication of any class that interacts with `eks.Cluster` since those classes may not work with both `eks.Cluster` and `eks.ExperimentalCluster`.
```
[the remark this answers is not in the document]
iliapolo: 👍

text/0244-eks-developer-preview-migration-path.md
iliapolo: Following our talk today, we decided to go with releasing an `@aws-cdk/aws-eks-next` that will incrementally introduce those changes. This will give us the option to use the module exactly as users would and validate it. We can also expose this module to a select few, or to those who choose to use it, knowing fully that it may introduce these types of changes frequently.

text/0244-eks-developer-preview-migration-path.md
iliapolo: What is the benefit of putting it in a different repo as opposed to just using a different name in the current repo? Also, are you suggesting the EKS library will be vended separately from the CDK until we release V2? Under which name? It feels wrong from a user's perspective.

text/0244-eks-developer-preview-migration-path.md
njlynch: Does that decision for `aws-eks-next` effectively invalidate the rest of this PR? Is this equivalent to the `Rename new module` alternative, except that we're not (yet) declaring the new module dev preview?

text/0244-eks-developer-preview-migration-path.md
```diff
@@ -0,0 +1,102 @@
...
+
+    - This will be released with a BREAKING CHANGE notice and cause users to change their dependency.
+
+2. Create a staging branch `eks-dev-preview` which contains `aws-eks` as a copy of `aws-eks-experimental`.
+
+    - This will be a long living branch we will branch out of to and merge to.
+
+3. Implement necessary changes onto the `eks-dev-preview` branch.
+
+    - We branch out of the `aws-dev-preview` branch to separate feature branches so that we can review the changes.
```
njlynch, line 56:
```suggestion
    - We branch out of the `eks-dev-preview` branch to separate feature branches so that we can review the changes.
```

text/0244-eks-developer-preview-migration-path.md
iliapolo: Unfortunately some constructs currently cannot accept an `ICluster`. If they were able to, you might be right.

text/0244-eks-developer-preview-migration-path.md
iliapolo: Yeap

text/0244-eks-developer-preview-migration-path.md
iliapolo: 👍

text/0244-eks-developer-preview-migration-path.md
iliapolo: > Breaking changes and replacement don't have to coincide

You're right, and actually the changes we talk about are not breaking the API. But since they require cluster replacement, we treat them with even more care than API breakage.

> Can you go a little into the changes that prompted this RFC?

Essentially we weren't using the right mechanisms to configure the cluster `connections` and its interaction with our custom resource that runs `kubectl` commands against it. To make it work properly, we need to change the VPC configuration of the cluster, which requires replacement.

I'm not sure if you wanted more deep technical details, if so let me know, happy to talk about it. 

You can have a look at this [issue](https://github.com/aws/aws-cdk/issues/9754), which kind of prompted this whole thing. 

> Have we already done the work to see if we can avoid this process completely?

We have. Strictly speaking, we can probably get away without requiring cluster replacement. However, as we dove deeper into how EKS works, we realized that the our current configuration is messy, fragile, and in some (albeit rare) cases, is just plain wrong. In addition, we already had a feature that required cluster replacement a while ago, and eventually managed to avoid it at the expense of keeping a complicated stack modeling and introducing a **1 cluster per stack** limitation. 

At the end of the day, we decided we should do the right thing for GA, and make the cluster as robust as possible so that we minimize the change of having to do cluster replacements post GA.

iliapolo:
@rix0rrr @ericzbeard 

After a discussion with @eladb - we've decided to go in a slightly different direction. The reasoning behind dropping the previous approach is described [here](https://github.com/aws/aws-cdk-rfcs/blob/epolon/eks-dev-preview-migration/text/0244-eks-developer-preview-migration-path.md#rename-old-module).

text/0244-eks-developer-preview-migration-path.md
iliapolo: Pretty much

iliapolo:
Closing - we wound up not going in this migration path at all.

PR #250, opened 2020-09-08
RFC 249: V2 Experiments

ericzbeard:
* Tracking issue: 249
* [Rendered version](https://github.com/ericzbeard/aws-cdk-rfcs/blob/v2-experimental/text/0249-v2-experiments.md)

---


text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  supported for the lifetime of the major version.
+
+# Alternatives
+
+Some alternatives to this strategy are:
+
+- Make no changes and continue to experiment in the main release. Improve documentation so that it is more obvious to
+  users that they are importing experimental code. We might also make it a policy to actually include the name
+  'experimental' in all experimental APIs.
+- Use clever build tricks to hide experimental modules. This might be possible in Typescript, where the type files are
```
rix0rrr, line 105: Here's another alternative: use runtime checks.

Calling an experimental API throws an exception which can be silenced by flipping a flag `ALLOW_EXPERIMENTAL=1`.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  modules, development will be more complicated.
+- In some cases, experimentation will not be possible. New features introduced to the main repository will have to be
+  vetted much more carefully before they are merged, since any public API released as part of v2 will have to be
+  supported for the lifetime of the major version.
+
+# Alternatives
+
+Some alternatives to this strategy are:
+
+- Make no changes and continue to experiment in the main release. Improve documentation so that it is more obvious to
```
rix0rrr, line 102: ...and this is not good enough because...?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  documentation to point users to the new stable version in the main repository. The lifecycle of experimental modules
+  in the preview repository will be exactly the same as it is currently in 1.x, with the same metadata markers in the
+  code.
+- The contributor guide for the CDK will be modified so that users know where to introduce new experimental code.
+- If it becomes absolutely necessary to add a new feature to v2 that would be a breaking change for users, the feature
+  will be hidden behind a feature flag. The code path will not be accessible without a specific opt-in by the user.
+
+# Drawbacks
+
+- In several places in the current 1.x CDK branch, stable modules refer to or contain experimental code. This will no
```
rix0rrr, line 88: This is listed as a drawback, but what I'd like you to do is reason through the consequences of this.

I know your focus with this doc is mostly on service libraries. Take Cognito, which is purposely only half-implemented right now. Describe to me how we would implement the other half.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  longer be possible after the release of v2.
+- When we graduate a module, users who have participated in the Developer Preview will have to update their imports to
+  make use of the stable version in the main repository.
+- For developers who are working on experimental code that has complex dependencies with the core and with other
+  modules, development will be more complicated.
+- In some cases, experimentation will not be possible. New features introduced to the main repository will have to be
```
rix0rrr, line 94: You're throwing this drawback out pretty casually as a statement of fact.

Convince me why it is okay to lose development speed.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+> major releases can have breaking changes. Each module's stability designation is available on its Overview page in the
+> AWS CDK API Reference. For more information, see Versioning in the CDK Developer Guide.
+
+The existing contributor guide does not mention experimental code, so the following will be added after the intro
+paragraph:
+
+> If you are contributing to a 2.x release, keep in mind that you may not introduce breaking changes to any public
+> facing API. Experimental modules are developed in a separate repository. [link]
+
+The new aws-cdk-previews repository will have the following Readme:
```
rix0rrr, line 53: How/where is the API reference of the packages in this new repo going to appear? In line with the rest? On a completely different site? How is it going to be distinguished from the other API ref?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
+---
+feature name: V2 Experiments
+start date: 2020-09-08
+? rfc pr
+? related issue
+---
+
+# Summary
+
+When CDK v2 is released to General Availability (GA), the main release branch will no longer contain experimental code.
```
skinny85, line 10: I actually don't like this summary. How about we state what problem are we solving first? Something like "With CDK v2.0, we want to strictly adhere to semantic versioning. That means all experimental code will be in libraries with versions starting with 0.x, and this means the CDK 2.x libraries can only contain stable code".

I don't like that we're starting with a solution, instead of a problem. There are many ways to achieve the above goal, and details like "separate repository" are kind of orthogonal to that concern.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+> facing API. Experimental modules are developed in a separate repository. [link]
+
+The new aws-cdk-previews repository will have the following Readme:
+
+> This repository contains experimental code related to the AWS Cloud Development Kit (AWS CDK). CDK is an open-source
+> software development framework to define cloud infrastructure in code and provision it through AWS CloudFormation.
+> Please refer to the main repository here [link]. [TODO...]
+
+The contributor guide can be copied from the main repository, since the toolchain for building the preview repository
+will be basically the same.
```
skinny85, line 60: I think this whole `# README` paragraph is way too low-level. I don't think we need to get into such details in this RFC.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+
+- Upon release of v2.0 to GA, all experimental modules will be automatically marked as stable (in fact, the metadata in
+  the code regarding stability can be removed, since everything will be stable). This means that the opportunity for the
+  team and for the community to influence the shape of existing experimental APIs is between now and then. Once we go
+  GA, we have to live with our design decisions.
+- Any APIs that cannot be graduated with confidence in time for the v2 release will be deprecated in 1.x, removed from
+  the v2 branch, and moved to the preview repository. This option should be minimized, since some customers may have
+  deployed production stacks that depend on these APIs.
+- Create a new repository called aws-cdk-previews that will be the new home for experimental development and developer
+  preview releases. The packages in this repository will be vended in the same way that 1.x packages are vended today,
```
skinny85, line 72: > The packages in this repository will be vended in the same way that 1.x packages are vended today,

I don't understand why would we vend these as separate packages? This brings back the same dependency problem that we started v2.0 to solve...

Why not one `aws-cdk-lib-experimental` library?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+# Design Summary
+
+- Upon release of v2.0 to GA, all experimental modules will be automatically marked as stable (in fact, the metadata in
+  the code regarding stability can be removed, since everything will be stable). This means that the opportunity for the
+  team and for the community to influence the shape of existing experimental APIs is between now and then. Once we go
+  GA, we have to live with our design decisions.
+- Any APIs that cannot be graduated with confidence in time for the v2 release will be deprecated in 1.x, removed from
+  the v2 branch, and moved to the preview repository. This option should be minimized, since some customers may have
+  deployed production stacks that depend on these APIs.
+- Create a new repository called aws-cdk-previews that will be the new home for experimental development and developer
```
skinny85, line 71: Why do we need a new repository? Why not vend them as separate packages from the same repository?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  preview releases. The packages in this repository will be vended in the same way that 1.x packages are vended today,
+  in a variety of languages, with metadata that draws a line between new experiments and more mature developer preview
+  packages. This repo will take a dependency on the core CDK package just like any other independent library. This
```
skinny85, line 74: > , with metadata that draws a line between new experiments and more mature developer preview
  packages

Can you expand what this means exactly?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  avoids the complications associated with experimenting on a fork or a branch of the main repository. The repository
```
skinny85, line 75: > avoids the complications associated with experimenting on a fork or a branch of the main repository

Why not experiment on `master` of the main repository, and just vend separate packages?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+  will be clearly marked as experimental.
+- When a package in the preview repository has reached maturity and is ready for GA, it will be copied into the main
+  repo and made a part of CDK 2. The module in the preview repository will then be marked as deprecated with clear
+  documentation to point users to the new stable version in the main repository. The lifecycle of experimental modules
+  in the preview repository will be exactly the same as it is currently in 1.x, with the same metadata markers in the
+  code.
+- The contributor guide for the CDK will be modified so that users know where to introduce new experimental code.
+- If it becomes absolutely necessary to add a new feature to v2 that would be a breaking change for users, the feature
```
skinny85, line 83: Sorry, how is this point relevant to this RFC?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+
+- In several places in the current 1.x CDK branch, stable modules refer to or contain experimental code. This will no
+  longer be possible after the release of v2.
+- When we graduate a module, users who have participated in the Developer Preview will have to update their imports to
+  make use of the stable version in the main repository.
+- For developers who are working on experimental code that has complex dependencies with the core and with other
+  modules, development will be more complicated.
+- In some cases, experimentation will not be possible. New features introduced to the main repository will have to be
+  vetted much more carefully before they are merged, since any public API released as part of v2 will have to be
+  supported for the lifetime of the major version.
```
skinny85, line 96: Another downside: third-party libraries effectively cannot use the experimental modules (because of the dependency problem that mono-CDK solves for stable modules).

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+the top of their source files, rebuilding, and redeploying. This upgrade should not result in any resources being
+destroyed and recreated, except in very specific cases where we have made it clear to users, such as with the EKS L2
+construct library. The same should be true for users upgrading from the preview repository to the stable repository as
+modules are graduated from Developer Preview to GA.
+
+# Unresolved questions
+
+- One debated decision was related to the preview repository. Should it be a single repository or should each new module
+  get its own repository? Separate repositories are attractive because it gives us a way to demonstrate a canonical way
+  to develop an independent construct library. But using a single repository is much simpler logistically.
```
skinny85, line 120: In my opinion, multiple repositories will make it effectively impossible for the experimental modules to depend on each other, because of dependency problems. This is a huge problem in my opinion.

skinny85 (commented):
Some initial comments.

In general, I feel like I'm missing some details. What are the names of the experimental modules? What are their versions? Are they released together, or does each and every one have a separate version? How does this affect our usage statistics gathering?

In my opinion, releasing them as separate modules is a mistake. This will bring back the same dependency problems that mono-CDK was designed to solve, except just for the experimental modules. What this will effectively mean is that 1) experimental modules cannot depend on other experimental modules, and 2) third-party modules cannot depend on experimental modules. This will make experimental modules much less useful than today, which will affect their usage (I suspect it will be considerably lower than today), which will make graduating them more difficult (less customer feedback and engagement). I consider that a serious risk.

I also don't see why do we need a separate repository for this. Why can't we release from the one we have today, just separately for the experimental modules? This hasn't been addressed in the doc at all.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,127 @@
...
+> This repository contains experimental code related to the AWS Cloud Development Kit (AWS CDK). CDK is an open-source
+> software development framework to define cloud infrastructure in code and provision it through AWS CloudFormation.
+> Please refer to the main repository here [link]. [TODO...]
+
+The contributor guide can be copied from the main repository, since the toolchain for building the preview repository
+will be basically the same.
+
+# Design Summary
+
+- Upon release of v2.0 to GA, all experimental modules will be automatically marked as stable (in fact, the metadata in
```
skinny85, line 64: > Upon release of v2.0 to GA, all experimental modules will be automatically marked as stable.

Why? Why not move experimental packages out of the mono-CDK? `2.0` will be a breaking change to its consumers anyway.

text/0249-v2-experiments.md
ericzbeard: Adding that to the next commit.

text/0249-v2-experiments.md
ericzbeard: I don't think we can rely on documentation. Many (most?) users will find some sample code or something from a repository like Solutions Constructs, and if it works, they won't have any reason to go back and read the docs. I don't think we can rely on intellisense either, since not everyone uses an IDE with that functionality.

text/0249-v2-experiments.md
ericzbeard: Why not create a new package in the preview repository named something like cognito_additions with the new classes? Then when we GA the additional features, it would be a simple import change. If we need to make changes on the stable repo to incorporate the new classes, we wait until they are merged to do it.

text/0249-v2-experiments.md
ericzbeard: I think the loss in speed on the stable repo will be compensated by much faster development on the preview repository. I also anticipate more contributors there, since it will be much easier to get started.

text/0249-v2-experiments.md
ericzbeard: I put that as the summary because the RFC is recommending that solution. I feel that mixing experiments into the stable repo is a slippery slope, and enforcing the separation is the best way to make sure the 2.x releases are all stable.

text/0249-v2-experiments.md
ericzbeard: This was more about reducing effort with the transition. I'm open to a single package, but I don't know how much extra work that would entail.

text/0249-v2-experiments.md
ericzbeard: I think it's a slippery slope. How would we guarantee there is no mixing of stable and experimental in a 2.x release? Also, why not separate repos? Why is that a problem?

text/0249-v2-experiments.md
ericzbeard: Which point? Feature flags? Seems like we need a plan for complex new features that can't be developed outside the stable repo.

text/0249-v2-experiments.md
ericzbeard: Why not? There are several 3rd party repositories out there like Solutions Constructs that reference experimental packages.

text/0249-v2-experiments.md
ericzbeard: I think we need to minimize the breaks. We hope that for the vast majority of customers, upgrading to 2.0 will be as simple as changing a few imports. We don't know how many customers are using experimental modules in production stacks. We do plan to move some of them out, but it should be reserved for code that we feel is nowhere near production quality. This should be rare, since everything has gone through the PR process and we fix any high priority bugs that get reported.

text/0249-v2-experiments.md
skinny85: > Which point? Feature flags?

Yes, the feature flags point.

> Seems like we need a plan for complex new features that can't be developed outside the stable repo.

Perhaps, but I don't see how that's relevant to this RFC.

text/0249-v2-experiments.md
skinny85: From what I know, all third-party packages are pinned to specific versions of the CDK libraries, which prevents their consumers from upgrading. This is exactly the problem that mono-CDK was meant to solve. I feel like we're re-introducing it here, just for the experimental modules.

text/0249-v2-experiments.md
skinny85: > How would we guarantee there is no mixing of stable and experimental in a 2.x release?

I feel like that's build-time validation that shouldn't be very hard to implement.

> Also, why not separate repos? Why is that a problem?

I can think of like 8 different reasons a separate repo can be a problem:

1. Customers will be confused which repo they need to contribute to.
2. There will be differences in working on the two repos (different build commands, code organization, linter configuration, etc.).
3. The second repo will be starved by the core team, because the main repo will take priority.
4. Changes cannot be coordinated between stable and experimental modules anymore.
5. We need to duplicate the entire release pipeline.
6. etc. (I'm sure there are more reasons if I thought about it more)

NetaNir:
> Some initial comments.
> 
> In general, I feel like I'm missing some details. What are the names of the experimental modules? What are their versions? Are they released together, or does each and every one have a separate version? How does this affect our usage statistics gathering?
> 
> In my opinion, releasing them as separate modules is a mistake. This will bring back the same dependency problems that mono-CDK was designed to solve, except just for the experimental modules. What this will effectively mean is that 1) experimental modules cannot depend on other experimental modules, and 2) third-party modules cannot depend on experimental modules. This will make experimental modules much less useful than today, which will affect their usage (I suspect it will be considerably lower than today), which will make graduating them more difficult (less customer feedback and engagement). I consider that a serious risk.
> 
> I also don't see why do we need a separate repository for this. Why can't we release from the one we have today, just separately for the experimental modules? This hasn't been addressed in the doc at all.

@skinny85  experimental modules will be just like any other third party construct, any limitation they will have will also effect the eco system, which is what worries me in your statement. Maybe we need to think about a solution to that (huge) problem.

text/0249-v2-experiments.md
ericzbeard: Looks like solutions constructs upgrades whenever we release a new version. https://github.com/awslabs/aws-solutions-constructs/blob/master/source/patterns/%40aws-solutions-constructs/aws-apigateway-sqs/package.json

skinny85:
> > Some initial comments.
> > In general, I feel like I'm missing some details. What are the names of the experimental modules? What are their versions? Are they released together, or does each and every one have a separate version? How does this affect our usage statistics gathering?
> > In my opinion, releasing them as separate modules is a mistake. This will bring back the same dependency problems that mono-CDK was designed to solve, except just for the experimental modules. What this will effectively mean is that 1) experimental modules cannot depend on other experimental modules, and 2) third-party modules cannot depend on experimental modules. This will make experimental modules much less useful than today, which will affect their usage (I suspect it will be considerably lower than today), which will make graduating them more difficult (less customer feedback and engagement). I consider that a serious risk.
> > I also don't see why do we need a separate repository for this. Why can't we release from the one we have today, just separately for the experimental modules? This hasn't been addressed in the doc at all.
> 
> @skinny85 experimental modules will be just like any other third party construct, any limitation they will have will also effect the eco system, which is what worries me in your statement. Maybe we need to think about a solution to that (huge) problem.
> (yeah I know this it's a bold, rabbit hole statement)

I don't think that statement is true though. The Construct Library is special in the ecosystem - if it was the same as other third-party modules, we wouldn't be doing mono-CDK in the first place.

text/0249-v2-experiments.md
skinny85: I'm not sure what your argument here is @ericzbeard . Are you saying that there is no need for mono-CDK, and that the current third-party ecosystem works fine?

skinny85:
I pulled some data from our dependencies. Here's all instances of stable modules depending on unstable ones:

```
⚠️  Stable package '@aws-cdk/assert' depends on unstable package '@aws-cdk/cloudformation-diff'
⚠️  Stable package '@aws-cdk/assert' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-apigateway' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-apigateway' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-applicationautoscaling' depends on unstable package '@aws-cdk/aws-autoscaling-common'
⚠️  Stable package '@aws-cdk/aws-autoscaling' depends on unstable package '@aws-cdk/aws-autoscaling-common'
⚠️  Stable package '@aws-cdk/aws-autoscaling' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-codebuild' depends on unstable package '@aws-cdk/aws-ecr-assets'
⚠️  Stable package '@aws-cdk/aws-codebuild' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-codebuild' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-codepipeline' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-ec2' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-ec2' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-ec2' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-ecs' depends on unstable package '@aws-cdk/aws-ecr-assets'
⚠️  Stable package '@aws-cdk/aws-ecs' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-ecs' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-elasticloadbalancingv2' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-elasticloadbalancingv2-actions' depends on unstable package '@aws-cdk/aws-cognito'
⚠️  Stable package '@aws-cdk/aws-events-targets' depends on unstable package '@aws-cdk/aws-batch'
⚠️  Stable package '@aws-cdk/aws-iam' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-lambda' depends on unstable package '@aws-cdk/aws-efs'
⚠️  Stable package '@aws-cdk/aws-lambda' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-lambda' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package '@aws-cdk/aws-logs' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/aws-route53-patterns' depends on unstable package '@aws-cdk/aws-cloudfront'
⚠️  Stable package '@aws-cdk/aws-route53-patterns' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-route53-targets' depends on unstable package '@aws-cdk/aws-cloudfront'
⚠️  Stable package '@aws-cdk/aws-route53-targets' depends on unstable package '@aws-cdk/aws-cognito'
⚠️  Stable package '@aws-cdk/aws-route53-targets' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package '@aws-cdk/aws-stepfunctions-tasks' depends on unstable package '@aws-cdk/aws-batch'
⚠️  Stable package '@aws-cdk/aws-stepfunctions-tasks' depends on unstable package '@aws-cdk/aws-ecr-assets'
⚠️  Stable package '@aws-cdk/aws-stepfunctions-tasks' depends on unstable package '@aws-cdk/aws-glue'
⚠️  Stable package '@aws-cdk/aws-stepfunctions-tasks' depends on unstable package '@aws-cdk/aws-s3-assets'
⚠️  Stable package '@aws-cdk/core' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package 'aws-cdk' depends on unstable package '@aws-cdk/cloudformation-diff'
⚠️  Stable package 'aws-cdk' depends on unstable package '@aws-cdk/cx-api'
⚠️  Stable package 'aws-cdk' depends on unstable package '@aws-cdk/region-info'
⚠️  Stable package 'aws-cdk' depends on unstable package 'cdk-assets'
```

(this is after I've marked `@aws-cdk/assert` as `"stable"`, as we have hundreds more of those)

None of those dependencies would be possible after this change.

skinny85:
Another serious dependency issue is that we have a lot of `private` modules used in the repository for common things like build, test lint, etc., and if you want to split experimental packages into a separate repository, all those `private` modules will now have to be published to NPM.

skinny85:
And here is the list of unstable modules depending on each other - relevant to the discussion whether there should be a single experimental package released, or multiple ones:

```
ℹ️️  Unstable package '@aws-cdk/aws-appsync' depends on unstable package '@aws-cdk/aws-cognito'
ℹ️️  Unstable package '@aws-cdk/aws-appsync' depends on unstable package '@aws-cdk/aws-s3-assets'
ℹ️️  Unstable package '@aws-cdk/aws-backup' depends on unstable package '@aws-cdk/aws-efs'
ℹ️️  Unstable package '@aws-cdk/aws-backup' depends on unstable package '@aws-cdk/aws-rds'
ℹ️️  Unstable package '@aws-cdk/aws-cloudfront-origins' depends on unstable package '@aws-cdk/aws-cloudfront'
ℹ️️  Unstable package '@aws-cdk/aws-docdb' depends on unstable package '@aws-cdk/aws-efs'
ℹ️️  Unstable package '@aws-cdk/aws-ecr-assets' depends on unstable package '@aws-cdk/cx-api'
ℹ️️  Unstable package '@aws-cdk/aws-efs' depends on unstable package '@aws-cdk/cx-api'
ℹ️️  Unstable package '@aws-cdk/aws-s3-assets' depends on unstable package '@aws-cdk/cx-api'
ℹ️️  Unstable package '@aws-cdk/aws-s3-deployment' depends on unstable package '@aws-cdk/aws-cloudfront'
ℹ️️  Unstable package '@aws-cdk/aws-s3-deployment' depends on unstable package '@aws-cdk/aws-s3-assets'
ℹ️️  Unstable package '@aws-cdk/aws-ses-actions' depends on unstable package '@aws-cdk/aws-ses'
ℹ️️  Unstable package '@aws-cdk/aws-synthetics' depends on unstable package '@aws-cdk/aws-s3-assets'
ℹ️️  Unstable package '@aws-cdk/pipelines' depends on unstable package '@aws-cdk/aws-s3-assets'
ℹ️️  Unstable package '@aws-cdk/pipelines' depends on unstable package '@aws-cdk/cx-api'
️ℹ️  Unstable package '@aws-cdk/pipelines' depends on unstable package '@aws-cdk/aws-ecr-assets'
ℹ️️  Unstable package 'cdk-assets' depends on unstable package '@aws-cdk/cx-api'
```

NetaNir:
> > > Some initial comments.
> > > In general, I feel like I'm missing some details. What are the names of the experimental modules? What are their versions? Are they released together, or does each and every one have a separate version? How does this affect our usage statistics gathering?
> > > In my opinion, releasing them as separate modules is a mistake. This will bring back the same dependency problems that mono-CDK was designed to solve, except just for the experimental modules. What this will effectively mean is that 1) experimental modules cannot depend on other experimental modules, and 2) third-party modules cannot depend on experimental modules. This will make experimental modules much less useful than today, which will affect their usage (I suspect it will be considerably lower than today), which will make graduating them more difficult (less customer feedback and engagement). I consider that a serious risk.
> > > I also don't see why do we need a separate repository for this. Why can't we release from the one we have today, just separately for the experimental modules? This hasn't been addressed in the doc at all.
> > 
> > 
> > @skinny85 experimental modules will be just like any other third party construct, any limitation they will have will also effect the eco system, which is what worries me in your statement. Maybe we need to think about a solution to that (huge) problem.
> > (yeah I know this it's a bold, rabbit hole statement)
> 
> I don't think that statement is true though. The Construct Library is special in the ecosystem - if it was the same as other third-party modules, we wouldn't be doing mono-CDK in the first place.

Why is that true? What's special about the L2s? We are doing monocdk because of the inevitable dependencies between L2s which stems from the dependencies between AWS services (i.e you need S3 bucket for triggering a lambda etc.). Every L2 will depend on aws-cdk/aws-s3, aws-cdk/aws-iam, aws-cdk/aws-cloudwatch... for the most basic construct you will need to depend on at  least 30 aws-cdk/* modules, this why for these modules monocdk make sense. On the other hand you will have modules like aws-cdk/aws-synthetics that are more like third party constructs in that aspect - much less construct will depend on them. Are we saying that third party constructs can't depend on each others as well? What is the technical difference between  aws-cdk/aws-synthetics and any other third party construct

skinny85:
So is your argument that `@aws-cdk/aws-synthetics` should _not_ be in mono-CDK @NetaNir ?

NetaNir:
> So is your argument that `@aws-cdk/aws-synthetics` should _not_ be in mono-CDK @NetaNir ?

I don't know if it shouldn't be in monocdk forever, overall it is nice that everything is in the same package. I was giving it as an example to an L2 which is similar to a third party construct in terms of dependencies.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,129 @@
...
+  vetted much more carefully before they are merged, since any public API released as part of v2 will have to be
+  supported for the lifetime of the major version.
+
+# Alternatives
+
+Some alternatives to this strategy are:
+
+- Make no changes and continue to experiment in the main release. Improve documentation so that it is more obvious to
+  users that they are importing experimental code. We might also make it a policy to actually include the name
+  'experimental' in all experimental APIs.
```
NetaNir, line 104: > We might also make it a policy to actually include the name
  'experimental' in all experimental APIs.

This should be in its own bulletin as it's a viable options and it is a change from the way we write experimental APIs today. It will solve the  "no one reads the docs" problem and set the right expectations.

README.md
```diff
@@ -82,6 +82,7 @@ future state of the libraries and to discover projects for contribution.
 [204](https://github.com/aws/aws-cdk-rfcs/issues/204)|[Go language bindings](https://github.com/aws/aws-cdk-rfcs/issues/204)||💡 proposed
 [217](https://github.com/aws/aws-cdk-rfcs/issues/217)|[Alternative Infrastructure Providers](https://github.com/aws/aws-cdk-rfcs/issues/217)||💡 proposed
 [219](https://github.com/aws/aws-cdk-rfcs/issues/219)|[ECS Patterns Service Builder](https://github.com/aws/aws-cdk-rfcs/issues/219)||💡 proposed
+[249](https://github.com/aws/aws-cdk-rfcs/issues/249)|[V2 Experiments](https://github.com/aws/aws-cdk-rfcs/issues/249)||💡 proposed
```
eladb, line 85: This is updated automatically by a GitHub action. No need to edit.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,129 @@
+---
+feature name: V2 Experiments
```
eladb, line 2: The name is a bit odd. I wonder if something like "Experimental APIs post 1.x"?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,129 @@
...
+start date: 2020-09-08
+? rfc pr
```
eladb, line 4: plz update

text/0249-v2-experiments.md
eladb: I agree with @ericzbeard. The summary should succinctly describe the "result" of the RFC assuming some people will only read this to determine if they are actually interested to dive deeper.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,129 @@
...
+? related issue
+---
+
+# Summary
+
+When CDK v2 is released to General Availability (GA), the main release branch will no longer contain experimental code.
+All experimental code, including modules in Developer Preview, will be developed in a separate repository.
+
```
eladb, line 12: I think it's worth separating experimental modules and experimental APIs. I believe those are quite different and it would simplify to think about each separately. I think we can't afford to loose the ability to introduce experimental *APIs* to stable modules because this will dramatically hinder our agility. To that end, we should come up with a solution that allows users to **opt-in** to experimental APIs (e.g. a CLI flag that enables experimental APIs, otherwise, runtime error).

I do think that we can remove experimental *modules* from the library though which I believe is actually the main user pain.

text/0249-v2-experiments.md
eladb: I think this is my preferred approach for enabling experimentation within stable modules. It requires some form of enforcement (awslint/jsii) which will make sure that people always include some runtime check when an API is annotated as @experimental (or jsii injecting some code somehow).

text/0249-v2-experiments.md
eladb: My preferred approach would be to put each experimental module in it's own repository (it's very easy to set up a fully-featured jsii project nowadays (`npx projen new jsii`). It will be a good forcing function for us to publish any remaining build tools from the CDK repo as public libraries so that anyone can enjoy them.

eladb (changes requested):

skinny85:
Is it even feasible to go with the multi-repository approach here at all? Let's say a customer wants to submit a PR adding a new L2 to a module that was previously CFN-only. They can't do that, because there is no experimental module for this AWS service, and thus no GitHub repository for it, so there is nowhere to submit the Pull Request...

eladb:
> Is it even feasible to go with the multi-repository approach here at all? Let's say a customer wants to submit a PR adding a new L2 to a module that was previously CFN-only. They can't do that, because there is no experimental module for this AWS service, and thus no GitHub repository for it, so there is nowhere to submit the Pull Request...

Yeah you are right. But how does this connect with the Uluru import solution? Eventually we are working towards a point where L1s are going to be generated on-demand by users same as in `cdk8s import` (`cdk import AWS::S3::Bucket`).

So we won't need these empty L1 modules anyway... we will just have a projen project type for AWS CDK modules that will generate a module repo and people can submit that as their PR.

skinny85:
Re: the discussion on separate vs one module, and what should go into mono-CDK. I want to summarize the existing points.

We can do one of two things.

1. Decide that all our L2s are "special" in some way, and they should all go to one of two NPM packages: either the mono-CDK (in case of stable modules), or a single, separate module for all experimental modules (to avoid the same dependency issue that mono-CDK was designed to solve). The name and version of this umbrella packages for all experimental modules is TBD. This pretty much kills the "separate Git repository per module" idea (in fact, I think we should release both from one Git repository, and just divide them at build time). We can go through a process for modules that are still not stable right before we release 2.0, to assign them to one of the 2 buckets. When a module graduates from the "unstable" bucket to "stable", it gets added to mono-CDK.

2. Decide that only some L2s are "special", and only those should be in mono-CDK. I assume the set of all "special" L2s is pretty much the same as the set of current stable modules (we can go through all modules right before releasing 2.0, and I guess, if there is a module that we want to include but which is still "unstable", we can graduate it to stable immediately). In this scheme, a module graduating from "unstable" to "stable" just means a new version; no new modules will be added to mono-CDK (unless a module can go from being "not special" to "special" somehow...?), and mono-CDK will not contain the L1s for non-"special" modules. In this scheme, we can go with either one, or separate Git repositories for the unstable modules. I don't know what criteria we would use to call a module "special" - @NetaNir , seems like you had some ideas here, care to comment?

I'm personally leaning towards 1., but I don't care that much. What I really don't want is the current proposal in this RFC, which essentially buckets the modules as "special" on the criteria that "if it has an L2 contributed to it before the 2.0 release, then it's automatically "special" (even if it's a brand new L2 that was pushed a week before), and every module that doesn't have an L2 before that date is not special".

ericzbeard:
> Let's say a customer wants to submit a PR adding a new L2 to a module that was previously CFN-only.

I'm a little confused by the discussion around this point, so let's use a concrete example. If a member of the community wants to submit an L2 for, let's say Neptune, after we go to v2, what do they do? If we go with one repo per experimental module, they would need to ask us to first create the repo. If we go with a single repo for all experiments, then they could submit a PR with the new code. It's logistically more work for us to create a repo per module, but maybe it's a good thing to force a discussion about the new construct library before accepting a new PR out of nowhere. Do we want contributors to do the initial work in their own personal repositories first? What would the process be to convert an L2 that someone wrote into an "official" AWS experimental L2? 

Adam makes good points for not creating any new repositories at all, and using the build and package process to make sure we never introduce any breaking changes to a 2.x minor release, but I worry that it will be difficult to enforce.

text/0249-v2-experiments.md
ericzbeard: > I think it's worth separating experimental modules and experimental APIs.

Can you give an example of an experimental API that would be made available by opt-in?

text/0249-v2-experiments.md
ericzbeard: No, just providing a data point. We can't depend on an external group to consistently upgrading every time we do a release. If they stopped, or pinned to a certain version, then their users would be stuck on an old version of CDK.

skinny85:
> If we go with one repo per experimental module, they would need to ask us to first create the repo. If we go with a single repo for all experiments, then they could submit a PR with the new code. It's logistically more work for us to create a repo per module

I don't think it's a matter of which is more work.... I don't think the "Git repository per package" is viable at all - where would the customer "ask us to first create the repo"...? I don't think we have a communication channel with customers  that can be reliably used for that...

> but maybe it's a good thing to force a discussion about the new construct library before accepting a new PR out of nowhere. Do we want contributors to do the initial work in their own personal repositories first? What would the process be to convert an L2 that someone wrote into an "official" AWS experimental L2?

I feel like that will actually be _more _ work for us, not less. If they submit PRs with us starting from the beginning, we can make sure they're small, and they get feedback from us early in the process. If we have to move an entire, already existing L2 into our repository, that will be a huge PR (that might involve dramatic changes to the library, that we didn't have the opportunity to specify earlier). In fact, we already have few of those (SageMaker?) that we refused to do, exactly for that reason.

> using the build and package process to make sure we never introduce any breaking changes to a 2.x minor release, but I worry that it will be difficult to enforce.

We do this already for stable modules in 1.x, so this is a non-issue.

text/0249-v2-experiments.md
skinny85: Exactly. Which leads to my comment, which is: "If we release every experimental module as a separate package, we will cause exactly the same pain points the third-party libraries are going through today, just limited to the experimental modules", which is another downside of this approach (which is the section I was commenting on).

text/0249-v2-experiments.md
ericzbeard: Isn't that by definition an L3, though, when we write a construct that has dependencies on several other libraries? We have talked about limiting aws-cdk-lib to L2s, which should be for the most part focused on a single service, and moving L3s out to their own repositories.

skinny85: I don't follow... [`@aws-cdk/aws-ecs` has 23 dependencies](https://github.com/aws/aws-cdk/blob/857acbbb7f26feecca938dc881add57fe5cae7e4/packages/%40aws-cdk/aws-ecs/package.json#L78-L100), does that mean it's not an L2, but an L3, and we should move it out of mono-CDK...?

skinny85: Also, I don't think the numbers matter here. The moment you have even a single dependency between two separate experimental modules, you introduce this problem (inability to upgrade the CDK version) for NPM consumers.

text/0249-v2-experiments.md
RomainMuller: > (or jsii injecting some code somehow)

I'm not entirely against the idea of adding a `jsii` compiler flag to trigger the injection...  I'm just wondering if this is a good fit for it (should we consider this is a "language primitive"?

Frankly, I'm thinking it might be more appropriate then to swap out the `@experimental` TS-Doc tag with an `@experimental` decorator (however decorators are experimental themselves.....)

text/0249-v2-experiments.md
RomainMuller: At the same time I agree with @skinny85  that it's super helpful to have a clear & crisp problem statement somewhere.

text/0249-v2-experiments.md
RomainMuller: > it's very easy to set up a fully-featured jsii project nowadays (`npx projen new jsii`)

Yeah it's awesome except for all the drawbacks that come with it, including inability to keep your dependencies up-to-date automatically.

text/0249-v2-experiments.md
RomainMuller: @skinny85 I actually think

> 4. Changes cannot be coordinated between stable and experimental modules anymore.

 is a good thing, not a problem. Since the "stable" part is supposed to be... well... stable... then you shouldn't need coordination! If you need it, something smells bad.

text/0249-v2-experiments.md
RomainMuller: I agree this is a big problem. If nobody can use the experimental modules, they are effectively useless. It's important to not lose sight of *why* we have experimental APIs in the first place (that is - to enable early adopters to provide us with valuable feedback before it's too late).

RomainMuller:
> Another serious dependency issue is that we have a lot of private modules used in the repository for common things like build, test lint, etc., and if you want to split experimental packages into a separate repository, all those private modules will now have to be published to NPM (or copy-pasted, I guess, but that doesn't sound like a great idea).

Or `git submodule`'d

text/0249-v2-experiments.md
skinny85: I just meant changing things in stable modules that are then used in experimental ones (of course, preserving backwards compatibility of the stable module). Today, you can do it in one commit; if we tear out the experimental modules to a separate repository, you have to first do it in the "stable" repository, wait at least a week for the release, and only then do the change in the "experimental" repository; which is another downside of the "separate repository" approach proposed here.

skinny85:
> Or `git submodule`'d

God, please no. Git submodules are terrible. I would avoid them like the plague.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,128 @@
...
+The new aws-cdk-previews repository will have the following Readme:
+
+> This repository contains experimental code related to the AWS Cloud Development Kit (AWS CDK). CDK is an open-source
+> software development framework to define cloud infrastructure in code and provision it through AWS CloudFormation.
+> Please refer to the main repository here [link]. [TODO...]
+
+The contributor guide can be copied from the main repository, since the toolchain for building the preview repository
+will be basically the same.
+
+# Design Summary
```
nija-at, line 61: Having read until this point and including the 'design summary' section, I still don't get how this is a better customer experience.

It seems that customers, now, instead of depending on just `aws-cdk-lib` will also take a dependency on `aws-cdk-previews` (or its individual packages). 

If `aws-cdk-previews` is going to follow (approximately) the same versioning scheme and release cycle, they're still going to see the same breaking changes.

Of course, they can choose not take a dependency on the `-previews` packages, but is that a practical option or how customers are going to behave? More likely than not, they will use the `-previews` packages, or resort to using the L1s. Now, we're back to the same problem again, right?

Did I miss something?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,128 @@
...
+  preview releases. The packages in this repository will be vended in the same way that 1.x packages are vended today,
+  in a variety of languages, with metadata that draws a line between new experiments and more mature developer preview
+  packages. This repo will take a dependency on the core CDK package just like any other independent library. This
+  avoids the complications associated with experimenting on a fork or a branch of the main repository. The repository
+  will be clearly marked as experimental.
+- When a package in the preview repository has reached maturity and is ready for GA, it will be copied into the main
+  repo and made a part of CDK 2. The module in the preview repository will then be marked as deprecated with clear
+  documentation to point users to the new stable version in the main repository. The lifecycle of experimental modules
+  in the preview repository will be exactly the same as it is currently in 1.x, with the same metadata markers in the
+  code.
```
nija-at, line 80: This statement, while seems simple to say, I would guess is going to be way more complicated than perceived.
From the top of my head, I can think of a few things that can get complicated. 

For instance, are the package names in the different package managers going to be the same or different? Is a packages within the previews repo allowed to depend on another package in the same repo? 

Just a few examples, I'm sure there will be more if we go deeper.

Expand this in a separate section called 'Graduation' and get into the details.

nija-at (changes requested):

text/0249-v2-experiments.md
rix0rrr: Yes I think decorators will the superior solution

text/0249-v2-experiments.md
rix0rrr: Please put that motivation in the document.

text/0249-v2-experiments.md
rix0rrr: If we do that, are any of the currently stable classes going to need to interact with types from the new library? Were we planning to do `UserPool.fromIdentityPool(...)` (or the other way around)?

If so, how would that work?

text/0249-v2-experiments.md
rix0rrr: > simply a part of the repository itself.

I don't know what that means.

Do we do cross-linking between the references?

text/0249-v2-experiments.md
rix0rrr: Also a single repo is:

* Easier to onboard with OSDS
* Easier to configure the GitHub project the way we want it to
* Easier to point people to (go THERE for experimental modules, single location)

eladb: projen now allows dependency versions to be maintained in package.json (so dependabot and `ncu` just work)

text/0249-v2-experiments.md
eladb: Asset bundling

text/0249-v2-experiments.md
eladb: I am also comfortable with including experimental modules in monocdk under `xxx-experimental` names (@ccfife's original proposal) and also requiring that users opt-in to use these APIs (like any experimental api).

text/0249-v2-experiments.md
ericzbeard: `cognito.UserPoolClientIdentityProvider.custom("MyProviderName")]` can be used for new IDP providers developed externally. Once merged in to the stable package, we can go back and add convenience properties similar to existing ones like `public static readonly FACEBOOK = new UserPoolClientIdentityProvider('Facebook')`.

That being said, I do think we will find places where the stable package has to be modified first, in order to make it flexible enough to interact with things being developed externally, which is a drawback.

text/0249-v2-experiments.md
ericzbeard: Maybe the name should be more obvious, like `aws-cdk-experiments`, since one of our motivations is to solve for users who do not read the docs carefully.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+> breaking API changes in any release. After a module is designated Stable, it adheres to semantic versioning, and only
+> major releases can have breaking changes. Each module's stability designation is available on its Overview page in the
+> AWS CDK API Reference. For more information, see Versioning in the CDK Developer Guide.
+
+The existing contributor guide does not mention experimental code, so the following will be added after the intro
+paragraph:
+
+> If you are contributing to a 2.x release, keep in mind that you may not introduce breaking changes to any public
+> facing API. Experimental modules are developed in a separate package, and experimental APIs in stable modules are
+> hidden behind a feature flag. [link]
```
eladb, line 50: We can't really "hide" the experimental APIs, only document them and issue runtime errors if users did not opt-in.

Also, we need to decide if we want a single feature flag for all experimental APIs, or a flag per API. The former is likely the pragmatic choice. The latter will incur quite a lot of overhead I believe.

text/0249-v2-experiments.md
eladb: `aws-cdk-labs`?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+construct library. The same should be true for users upgrading from the preview package to the stable package as modules
+are graduated from Developer Preview to GA.
+
+# Implementation Plan
+
+This section contains examples covering common scenarios:
+
+## 1. A brand new L2, where there was only an L1 before.
+
+- Create a new package in packages/@aws-cdk-previews exactly the same we we would have for v1
```
eladb, line 103: are we planning previews to be a single module or multiple modules? This implies that @aws-cdk-previews is an npm scope with multiple modules.

We should also make sure that we never take a dependency on experimental modules (either from the stable module or between themselves). Otherwise, we are back in peer-dependency hell again.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+
+Some alternatives to this strategy are:
+
+- Make no changes and continue to experiment in the main release.
+  - Improve documentation so that it is more obvious to users that they are importing experimental code.
+  - Make it a policy to actually include the name 'experimental' in all experimental APIs.
+- Use clever build tricks to hide experimental modules. This might be possible in Typescript, where the type files are
+  separate from implementation, but might prove difficult or impossible in other languages such as Python.
+- Call the experimental package @aws-cdk-experiments instead of @aws-cdk-previews.
+
```
eladb, line 88: Where is the alternative to separate experimental code into *submodules* within `aws-cdk-lib`.

Something like this:

```ts
import * as eks from 'aws-cdk-lib/aws-eks-experimental';
// or we could do
import * as eks from 'aws-cdk-lib/experimental/aws-eks';
```

The benefit of this approach is that we will still vend everything as a single module so we can avoid the potential risk of peer-deps hell

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+TODO
+
+## 4. How to work around stable modules that want to refer to experiments
+
+TODO - stepfunctions-tasks?
+
+## 5. Changing the entire implementation for a core module.
+
+TODO - asset bundling?
```
eladb, line 137: ?

eladb (changes requested):

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+understand the documentation simply install packages, copy sample code, make a few tweaks and put the code into
+production. When they later upgrade to a new version, they are surprised to find that their code no longer works. The
+perception of instability has driven some users away, or caused them to limit their usage of CDK to the fundamental L1
+constructs, which do not provide them with the benefits of higher level abstractions.
+
+This RFC proposes that we stop releasing breaking changes in the main packages that we vend that are not hidden by a
+feature flag. A user that installs aws-cdk-lib using npm or pip or any other package manager should be confident that
+there will be no intentional breaking changes in the 2.x line of releases for its lifetime.
+
+# README
```
nija-at, line 36: As mentioned in a prior comment, this section is too small and doesn't work properly here.

We should expand this into what an expanded user guide explaining to a new user of the CDK and an existing user of the CDK on the new way to use these packages.

nija-at (dismissed):

text/0249-v2-experiments.md
ericzbeard: I think I will remove this section, since writing new user docs are part of implementation.

text/0249-v2-experiments.md
ericzbeard: I need someone to help me with these sections, since I don't have enough experience with implementation.

text/0249-v2-experiments.md
ericzbeard: Not sure but didn't we have a strong reaction from others against submodules?

text/0249-v2-experiments.md
ericzbeard: A single experimental package with multiple modules.

text/0249-v2-experiments.md
ericzbeard: It might not always be practical to enable them all at once. What if two experimental features conflict with each other?

text/0249-v2-experiments.md
nija-at: This is key to describing the change in customer experience. Not sure how you will effectively describe the customer experience without this section.

IIUC, @elad's new working backwards RFC format was written to emphasize on the customer experience.

text/0249-v2-experiments.md
eladb: What's the dev ex for enabling these experimental APIs? How do we separate framework usage of experimental APIs from consumer usage of these APIs?

text/0249-v2-experiments.md
eladb: #What kind of reaction? (I mean submodules within `AWS-CDK-lib`, not git submodules)

text/0249-v2-experiments.md
eladb: Okay we need to dive deeper into these sections to see the full picture

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,137 @@
...
+
+An example would be Cognito User Pools, when we want to add additional Identity Pool Providers.
+
+- Create the implementation file and export it.
+  - packages/@aws-cdk-previews/cognito/lib/user-pool-idps/OpenIdConnect.ts
+- A user that wants to try it would import both the stable and experimental package
+
+```typescript
+import * as cognito from '@aws-cdk-lib/aws_cognito';
+import * as cognitoPreview from '@aws-cdk-previews/aws_cognito';
```
skinny85, line 119: I believe there will be no `@` here, so this should be:

```suggestion
import * as cognito from 'aws-cdk-lib/aws_cognito';
import * as cognitoPreview from 'aws-cdk-previews/aws_cognito';
```

skinny85 (commented):
Some comments from me. I see some naming inconsistencies that I think we need to sort out, and some details look left over from the previous versions.

General note: I would suggest using Markdown code backtics to enclose names used in code (so, `aws-cdk-lib`, not aws-cdk-lib). Makes it much easier to read that way, as the names stand out much more.

I think Elad's suggestion ("What if experimental packages are just in a sub-namespace of the main package?") is a great one, and worth exploring. It would allow us to keep the capability of stable modules to depend on experimental ones, which the current proposal bans.

text/0249-v2-experiments.md
ericzbeard: I can't think of a good example, but for the sake of simplicity we should go with a single flag, and if any weird situations come up where that doesn't work, we can introduce new flags on an ad-hoc basis if absolutely necessary.

text/0249-v2-experiments.md
ericzbeard: Decorators would definitely be a great solution, but what are the chances of things changing before it's a stable feature? Seems like too much of a risk for now.

text/0249-v2-experiments.md
ericzbeard: Adding this alternative to the RFC. If we can be sure we won't end up shooting ourselves in the foot, we should definitely explore the idea.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+Because of a combination of how `aws-cdk-lib` will be depended on by third-party
+libraries (through peer dependencies), and the goals of ths RFC, it will no
+longer be possible to vend unstable code as part of `aws-cdk-lib`'s main modules
+(so, modules that are marked "stable" will no longer have the capability to have
+unstable APIs inside of them).
+
+We will need to add validations to our build scripts that make sure this rule is
+never broken.
+
+### Why can't we vend unstable code in mono-CDK stable modules?
```
eladb, line 90: Move to appendix

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+their advantages and disadvantages.
+
+**Note**: I purposefully don't mention the issue of how should this decision
+affect the number of Git repositories the CDK project uses. I consider that an
+orthogonal concern to this one, and more of an implementation detail. The number
+of Git repositories is also a two-way door, unlike the package structure
+decision, which I believe cannot be changed without bumping the major version of
+CDK.
+
+### Option 1: sub-package of `aws-cdk-lib`
```
eladb, line 136: This option conflicts with "No unstable code in the main mono-CDK modules". I don't see a reason to propose it.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+  ℹ️️  Unstable package '@aws-cdk/aws-docdb' depends on unstable package '@aws-cdk/aws-efs'
+  ℹ️️  Unstable package '@aws-cdk/aws-s3-deployment' depends on unstable package '@aws-cdk/aws-cloudfront'
+  ℹ️️  Unstable package '@aws-cdk/aws-ses-actions' depends on unstable package '@aws-cdk/aws-ses'
+  ```
+
+Disadvantages:
+
+- Might be considered less explicit, as a customer never says they want to
+  depend on a package containing unstable APIs, or with `0.x` for the version.
+- If a third-party package depends on an unstable API in a non-obvious way (for
```
eladb, line 191: I think this makes this a no starter

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+
+- It's not possible for stable modules to depend on unstable ones. This has
+  serious side implications:
+  - All unstable modules that have stable dependents today will have to be
+    graduated before `v2.0` is released.
+  - Before a module is graduated, all of its dependencies need to be graduated.
+  - It will not be possible to add new dependencies on unstable modules to
+    stable modules in the future (for example, that's a common need for
+    StepFunction Tasks).
+- Graduating a module to stable will be a breaking change for customers. We can
```
eladb, line 241:
```suggestion
- Graduating a module to stable will be a breaking change for customers who consume the unstable module from the experimental library. We can
```

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+  solve.
+- Graduating a module to stable will be a breaking change for customers. We can
+  mitigate this downside by keeping the old unstable package around, but that
+  leads to duplicated classes.
+
+### Option 4: separate V3 that's all unstable
+
+In this option, we will fork the CDK codebase and maintain 2 long-lived
+branches: one for version `2.x`, which will be all stable, and one for version
+`3.x`, which will be all unstable.
```
eladb, line 286: I think using "3.x" is a bit misleading because this is not necessarily staging the next major version but rather just the experiments

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+## 3. Extra unstable precautions
+
+This chapter discusses additional precautions we can choose to implement to
+re-inforce goal #1 above. These are orthogonal to the decision on how to divide
+the stable and experimental modules (meaning, we could implement any of these
+with each of the options above).
+
+### Require a feature flag for unstable code
+
+In this variant, we would add a runtime check into all unstable APIs that
```
eladb, line 333: Didn't we say there are no unstable APIs in aws-cdk-lib? what am I missing?

eladb (dismissed):
- Missing final recommendation
- How do we enable the goal "The CDK team can still perform API experiments" for APIs in stable modules?

text/0249-v2-experiments.md
ericzbeard: What about `aws-cdk-lib@3.0.0-alpha.x`? I think that's the semver convention for pre-releases.

eladb: It's also a bit "wrong" because this is not the alpha for 3.x, it's a staging module for experimental APIs for 2.x, no?

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+    graduated before `v2.0` is released.
+  - Before a module is graduated, all of its dependencies need to be graduated.
+  - It will not be possible to add new dependencies on unstable modules to
+    stable modules in the future (for example, that's a common need for
+    StepFunction Tasks).
+- Graduating a module to stable will be a breaking change for customers. We can
+  mitigate this downside by keeping the old unstable module around, but that
+  leads to duplicated classes.
+
+### Option 3: separate multiple unstable packages
```
nija-at, line 245: I like this option. The constraints (disadvantages) feel reasonable.

> We can mitigate this downside by keeping the old unstable package around, but that leads to duplicated classes.

Re: duplicate classes - we can completely delete the code off the code base and mark the released packages as deprecated, so no code duplication. Existing customers can continue to use them but won't get newer features.

text/0249-v2-experiments.md
```diff
@@ -0,0 +1,380 @@
...
+  doing that brings us back to the dependency hell that mono-CDK was designed to
+  solve.
+- Graduating a module to stable will be a breaking change for customers. We can
+  mitigate this downside by keeping the old unstable package around, but that
+  leads to duplicated classes.
+
+### Option 4: separate V3 that's all unstable
+
+In this option, we will fork the CDK codebase and maintain 2 long-lived
+branches: one for version `2.x`, which will be all stable, and one for version
```
nija-at, line 285: Not a fan of this option. Maintaining 2 long-lived branches is not easy and is going to cause us a lot of maintainance pain. Separate repos or folders are much easier to manage.

nija-at: Plus, the disadvantages section points the other problems quite well.

skinny85 (approved):

Issue #253, opened 2020-09-23
CDK Metadata v2


rix0rrr:
|PR|Champion|
|--|--------|
|#254 |  @nija-at      |

## Description

We need to overhaul CDK metadata for v2.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [x] Resolved

PR #254, opened 2020-09-24
RFC 253: CDK Metadata v2

rix0rrr:
* Tracking issue: #253
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/huijbers/0253-cdk-metadata-v2/text/0253-cdk-metadata-v2.md)

---


text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+will not be collected.
+
+# Design
+
+There are two subproblems to address:
+
+* Collecting construct identifiers: library name, version, and qualified class name.
+* Encoding that information into the metadata resource.
+
+## Collecting construct identifiers
```
nija-at, line 53: Would be nice to start with an example somewhere and walk the example alongside the algorithm, or something else to that effect.

There's a few things here that I'm not entirely familiar with or why certain steps in the algorithm will work, such as "most qualified file name" and why we need to crawl up the inheritance hierarchy.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+
+* We can get the simple class name by reading `obj.constructor.name`.
+* We can find the file a class is defined in by by examining `require.cache[...].exports` and
+  which one the class (`object.constructor`) is in. If the class is non-exported
+  (such as an `class Import implements IBucket { ... }`) it will not show up though, and we
+  won't be able to figure out what file it's from.
+  * A constructor may be in multiple module's `exports` (because of
+    re-exporting). If it is, take the one with the most qualified filename.
+* We crawl up the inheritance hierarchy to find the first class that is exported in this way.
+  For unrecognized classes, we'll end up seeing `Construct`.
```
nija-at, line 67: Is unrecognized classes the same as the previously mentioned non-exported constructs?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+* We can find the package a class is from by crawling up the directory tree from the filename
+  we identified and looking for `package.json`, getting the `name` and `version` from that file.
+
+### Submodule names
+
+The only thing we can't get from this is a qualified "submodule" name which is going to
+become relevant in the monolibrary v2, where both the EKS and ECS submodules have a `Cluster` class,
+for example, which are otherwise indistinguishable apart from the filename. We also cannot
```
nija-at, line 75:
```suggestion
for example, which are otherwise indistinguishable apart from the file path. We also cannot
```

nija-at (changes requested):
Have not gone through entirely. Comments so far...

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+
+There are two subproblems to address:
+
+* Collecting construct identifiers: library name, version, and qualified class name.
+* Encoding that information into the metadata resource.
+
+## Collecting construct identifiers
+
+We need a sufficiently unique string representation of a construct('s class). Since
+we don't have manual annotation of construct class types at the moment, we will
```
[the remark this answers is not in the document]
rix0rrr: I don't mean anything. I literally only mean that a human must type some code somewhere to indicate the construct type. Haven't gotten into the mechanism yet and since I don't think we want any solution that requires human annotation, I don't want to spend any more cycles on thinking up a plausible mechanism to do this only to then discard it.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+have to make do with the reflection capabilities provided by NodeJS.
+
+* We can get the simple class name by reading `obj.constructor.name`.
+* We can find the file a class is defined in by by examining `require.cache[...].exports` and
+  which one the class (`object.constructor`) is in. If the class is non-exported
+  (such as an `class Import implements IBucket { ... }`) it will not show up though, and we
+  won't be able to figure out what file it's from.
```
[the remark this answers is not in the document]
rix0rrr: The solution is human annotation.

The second question is a good one.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+If an identifier is estimated to be a maximum of 100 bytes (which seems
+reasonable be plenty of room) the resulting data set will be ~40kB.
+
+We can encode the elements as a list straight in the template, but it will
+take up a lot of space in the template: vertical space while looking at it,
+as well as bytes that count toward the template size (maximum of 460kB).
+
+Experiments with a random sampling of 400 resources from the list of classes:
+
+| Description | Raw size | Base64ed size |
```
nija-at, line 134: Table is not rendered correctly.

Nit: could make this ascii tables, i.e., cells padded with whitespace, so it's easier to read in markdown format as well.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+        9C6ysN6r5aRlt+6/LgqDxGUSeP3lBDJHyl0G/VJzdGQ/ph4rNyFutAyZiGw+NS0Xy2k+Lj722pBGiiTkmqG52jgVy9gwCR+Qx3iYS0LwlEzUtpvwd9Evqjzy
+        1RHh+KehLsry4s8poRwVw9f55yvaR3swoXDcZg4qUeIwesj1Q99FuVa0iHPR88CxOQQtn/M0OLhvycrwNQs47KJ6ZPhonZFw7DRJxZVSGShXJtHLouTzVAyB
+        Tcb6JLsH831taVm6w3Zjny5t4MOtJbwo0yELp/JTOGLXrnqv62gRLQxUEDCQLl3YwPi2rvnHAfdP1ui4cNteKFyASguTUUyko4eCdYbP7nkkSNmQ7MVh807U
+        IFWKnOgczoUf+wNUgLcFOl48yLziJSBDGw0hpN5jaZwELR6AkhizgLORMh1MAux759Y8jqyK/oMZwImJHz0DSCp8KksnI6h14WYGwO/gt68rzcLLjH8pHuI3
+        wKVE/HOCgIQrRofRC0suQYuRg8T3OdNGnYRCKHIdgiLIMChV+y7NS+bxfnx8CSbXytQkXLx3iX9OSl+OmlTeeQY5awKMNM/dScR7tpbE8tmEXvBFgcqnTUUs
+        nW5r/rjfQtFwk5w+1cifgmylq45Nk8al8K5Y4aur
+```
+
+We should also prefix with an identifier or version number for the encoding
+strategy so that we are free to change it in the future.
```
nija-at, line 198: Yes! If we apply compression & encoding, we should treat it as a protocol.

Worth its own section.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+Experiments with a random sampling of 400 resources from the list of classes:
+
+| Description | Raw size | Base64ed size |
+|--|--|
+| Plaintext | 23k | - |
+| gzipped | 3.6k | 4.8k |
+| Zopfli | 3.4k | 4.6k |
+| Bloom Filter (1% error) | 720b | 960b |
+| Tree-encoded (plaintext) | 7.4k | - |
+| Tree-encoded gzipped | 3.0k | 4.0k |
```
nija-at, line 141: 4-5K seems like reasonable amount of metadata to be added. What does this look like in terms of percentage of the template's size?

Are you happy with these numbers?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
+---
+feature name: CDK Metadata v2
+start date: 2020-09-23
+rfc pr: (leave this empty)
+related issue: #253
+---
+
+# Abstract
```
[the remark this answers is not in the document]
rix0rrr: Hmm I thought this term was commonly understood to mean that, but sure I'll use Summary instead

https://en.wikipedia.org/wiki/Abstract_(summary)

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+Stack's synthesis. For every construct used in a Stack from an AWS-authored
+library we will collect the library name, version and qualified class name
+(using a generated lookup table to obtain submodule names) and submit these
+to the metadata resource in a gzipped blob.
+
+# Motivation
+
+The goal of CDK Metadata is twofold:
+
+* Track what library versions people are using, so that if there are severy (security)
```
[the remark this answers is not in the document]
rix0rrr: I'm not going to say "when" here.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+  problems with a library, we can contact them in a targeted fashion and suggest
+  they upgrade.
+* See what construct libraries are being used by people to inform future development
+  activities (example insight: are there areas where people are using L1s
```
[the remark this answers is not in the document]
rix0rrr: Again `identify` implies to me that this is definitely true, whereas my phrasing was intended to keep the option open that there might not be any.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+The current CDK Metadata is based on libraries loaded into the NodeJS process. When
+we move to CDK v2, the entire CDK library will be vended as one library, destroying
+our ability to measure in the same way.
+
+At the same time, our current method fails in the face of multiple distinct
+stacks in the same application. For example, in the case of a CDK Pipelines
+application, the `@aws-cdk/pipelines` library would be counted `N+1` times:
+in addition to the pipeline stack itself, once for every environment deployed
+via that pipeline because the library happened to be in memory while those
+stacks were being synthesized.
```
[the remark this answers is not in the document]
rix0rrr: Not sure that I agree. Let's say a CDK Pipelines app contains a `PipelineStack` and an `ApiGateway` stack (the latter to be deployed using the former).

Would you prefer this be counted as:

* 1 time a stack using `@aws-cdk/pipelines` (the `PipelineStack`)
* 1 time a stack using `@aws-cdk/api-gateway` (the `ApiGateway` stack)

Or

* 2 times a stack using `@aws-cdk/pipelines` (both stacks, because they both happen to be synthesized by an application that uses `@aws-cdk/pipelines`).
* 2 times a stack using `@aws-cdk/api-gateway` (both stacks, because they both happen to be synthesized by an application that uses `@aws-cdk/pipelines`).

You seem to be arguing you prefer those stacks be counted as the latter, but I'm not sure that would be all that intuitive to most people.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+
+> Implementation note: looking up the source file from a class is a potentially
+> expensive operation as it takes a linear scan through all sources for every class.
+> We should build a reverse index in order to speed this up.
+
+## Metadata resource encoding
+
+A rough estimate of the maximum number of constructs used in a stack will be
+about 400 (maximum of 200 L1s with a corresponding L2 wrapper). We will
+therefore need to encode roughly 400 strings of the form
```
ericzbeard, line 116: That's the worst case, right? We aren't encoding all 400 in each stack?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+
+The CDK Metadata resource will start tracking individual constructs used in a
+Stack's synthesis. For every construct used in a Stack from an AWS-authored
+library we will collect the library name, version and qualified class name
+(using a generated lookup table to obtain submodule names) and submit these
+to the metadata resource in a gzipped blob.
+
+# Motivation
+
+The goal of CDK Metadata is twofold:
```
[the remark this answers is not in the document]
rix0rrr: I was intending to describe "business metrics" in a general sense (as distinct from security motivations for version tracking), and then a legitimate example use case for business metrics.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+  to cut down on it, what clever strategies can we come up with?
+
+# Implementation Plan
+
+This can be implemented in v1 already, no need to wait for v2.
+
+* Update metadata service backend to accept new `Constructs` field. This is where
+  decoding should happen back to a flat list again.
+* Update canary to exercise new field.
+* Update reporting tool to report on new information.
```
ericzbeard, line 223: Is the new metrics extraction format going to be backwards compatible? There are other teams consuming the data now, which will complicate releasing the changes if not.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+* Add submodule-map generator to monocdk and v2 metapackages.
+* Emit both fields in the metadata resource
+* Verify metrics are coming in
+* Remove the old `Modules` field from the metadata resource.
+
+# Appendix 1: Reference implementation for tree-encoding used above
```
ericzbeard, line 229: Don't really understand what this code is doing, or why it's necessary in the RFC.

text/0253-cdk-metadata-v2.md
rix0rrr: That's the worst case, correct.

text/0253-cdk-metadata-v2.md
rix0rrr: Might come in useful for the final implementation.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,267 @@
...
+Cluster | .../node_modules/aws-cdk-lib/lib/aws-ecs/lib/cluster.js | aws-cdk-lib.aws_ecs.Cluster@2.1.0
+Cluster | .../node_modules/aws-cdk-lib/lib/aws-eks/lib/cluster.js | aws-cdk-lib.aws_eks.Cluster@2.1.0
+LambdaToSqs | .../node_modules/@aws-solutions-constructs/aws-lambda-sqs/lib/index.js | @aws-solutions-constructs/aws-lambda-sqs.LambdaToSqs@1.63.0
+
+It would be nice if we used submodule naming that's consistent with the jsii
+naming. There is nothing in the jsii manifest that could reliably be used for
+this, but we can use an additional lookup table that can be generated by the
+monocdk/v2 build tool, as that is in charge of the submodule names anyway.
+
+There will be a `submodules.json` next to `package.json` with the following structure:
```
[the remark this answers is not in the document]
rix0rrr: Ooooohh I like that! Good call!

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,508 @@
...
+.../node_modules/aws-cdk-lib/lib/aws-ecs/lib/cluster.js                | "aws_ecs"
+.../node_modules/aws-cdk-lib/lib/aws-eks/lib/cluster.js                | "aws_eks"
+.../node_modules/@aws-solutions-constructs/aws-lambda-sqs/lib/index.js | ""
+
+We could obtain submodule names by establishing conventions on monocdk's
+directory structure, using rules as "ignore `/lib/` and replace `-` with `_`
+to obtain the module name", for example. This reliance on convention seems
+brittle and unflexible. Instead, let's use a configuration file.
+
+One possible candidate is the jsii manifest. It contains the right module names
```
nija-at, line 203: I'm looking at the jsii manifest for `monocdk-experiment` and here's a snippet for eks Cluster and ecs Cluster -

```
      "name": "Cluster",
      "namespace": "aws_ecs",
      "properties": [
        {
          "docs": {
            "stability": "experimental",
            "summary": "The Amazon Resource Name (ARN) that identifies the cluster."
          },
          "immutable": true,
          "locationInModule": {
            "filename": "lib/aws-ecs/lib/cluster.ts",
            "line": 85
          },
          "name": "clusterArn",
          "overrides": "monocdk-experiment.aws_ecs.ICluster",
          "type": {
            "primitive": "string"
          }
        },
```

```
      "name": "Cluster",
      "namespace": "aws_eks",
      "properties": [
        {
          "docs": {
            "remarks": "This role also has `systems:master` permissions.",
            "stability": "experimental",
            "summary": "An IAM role with administrative permissions to create or update the cluster."
          },
          "immutable": true,
          "locationInModule": {
            "filename": "lib/aws-eks/lib/cluster.ts",
            "line": 685
          },
          "name": "adminRole",
          "type": {
            "fqn": "monocdk-experiment.aws_iam.Role"
          }
        },
```

It seems to me that from here we can derive the construct name (`name` property), module it is in (`namespace` property) and source file (`locationInModule` property).

Maybe we can update the JSII manifest to contain a `locationInModule` property for where the construct itself is, instead of individual properties.

What am I missing?

nija-at (commented):
Comments so far...

rix0rrr:
> You use the phrase "For example". I assume this is a unique problem today for @aws-cdk/pipelines.

You assume wrong. I can construct the same example for any pairs of stacks:

- Stack A: uses `aws-dynamodb`.
- Stack B: uses `aws-lambda`.

Will be counted as: 2 stacks of `aws-dynamodb`, 2 stacks of `aws-lambda`.

 it's just that CDK Pipelines does it by definition (and so will make the problem more pronounced).

text/0253-cdk-metadata-v2.md
nija-at: > what if for some reason some sister team doesn't want to, but does want tracking?

I don't see a use case yet. This will be a nice-to-have, but we probably shouldn't index on this.

> To improve loading speed by a factor of 10 I'd want us to webpack the individual modules in the future (NodeJS takes a rather long time to load each and every individual source file), which means the jsii manifest will contain lib/aws-cdk/lib/cluster.ts but the runtime path we detect will look like aws-cdk/lib/aws-eks/index.js. Now we have to map that in addition.

Isn't your current algorithm also subject to the same problems if we change how we pack our output artifacts (webpack or otherwise). Things like - determining module name from file path, using the deepest export - come to mind.

Between this comment and https://github.com/aws/aws-cdk-rfcs/pull/254#discussion_r500150191, an option to generate the necessary metadata at compile time or even pack time (so that the path-to-module mapping is consistent) feels worth exploring. Don't you think? At least called out as an alternative.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+We could obtain submodule names by establishing conventions on monocdk's
+directory structure, using rules as "ignore `/lib/` and replace `-` with `_`
+to obtain the module name", for example. This reliance on convention seems
+brittle and inflexible. Instead, let's use a configuration file.
+
+#### jsii manifest
+
+What about the jsii manifest? It contains the right module names already, and
+it seems to have a `locationInModule` entry that seems to do what we need:
```
nija-at, line 250: This should go into a section called 'Alternatives Considered' so as to not muddy the proposed design.

text/0253-cdk-metadata-v2.md
nija-at: Just getting into the `cdk-metadata.json` section. Maybe this is what you're already doing that file.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+Size of template, number of resources are obvious candidates.
+
+# Unresolved questions
+
+- How acceptable is the giant base64-encoded blob in the template? If we want
+  to cut down on it, what clever strategies can we come up with?
+
+# Implementation Plan
+
+This can be implemented in v1 already, no need to wait for v2.
```
nija-at, line 493: Your RFC has v2 in it. It's confusing to say v1 and v2 again.

```suggestion
These changes are backwards compatible. There is no need to wait for CDKv2.
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+Decoding will happen on the receiving end before the data hits long-term storage,
+in order to make consuming it as trivial as possible. There is no need to be frugal
+with bits anymore once the data hits an AWS datacenter.
+
+# Extensions
+
+We will be able more information into additional fields of the metadata resource.
+Size of template, number of resources are obvious candidates.
+
```
nija-at, line 485: There needs to be a section around end-to-end verification. 

How will we know when the data collection logic has an error (not counting or mis-counting), but none of our unit tests fail?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
+---
+feature name: CDK Metadata v2
```
nija-at, line 2: Nit: would be nice to name this RFC after the problem it's solving or the solution. Something like 'construct-usage-metadata'.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+|----------------------------|-------------|---------------|
+| Plaintext                  | 23k         | -             |
+| gzipped                    | 3.6k        | 4.8k          |
+| Zopfli                     | 3.4k        | 4.6k          |
+| bzip2                      | 3.4k        | 4.6k          |
+| Prefix-grouped (plaintext) | 7.4k        | -             |
+| Prefix-grouped gzipped     | 3.0k        | 4.0k          |
+| Bloom Filter (1% error)    | 720b        | 960b          |
+
+Prefix-grouped, gzipped data seems to give the best results that are still
```
nija-at, line 419: Assuming the same 400 constructs, this comes to 8% of the template if it were going inline (max 50K) as part of `createStack`, and <1% if the template was sent via S3 (max 450K).

I assume a template with 400 constructs cannot fit into the 50K limit anyway. I did a quick experiment and ~85 construct *instances* were needed to fill up to 50K. The actual construct types will be much lesser. 

The compression reached by prefix-grouped gzipped feels fine to me.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+convenient and generic to work with (see Appendix 2 for a description of
+prefix grouping).
+
+See Appendix 3 for an example blob to be added to a template. I'm not super
+happy with this, but this is the most straightfoward option available.
+
```
eladb, line 425: Perhaps we can upload the analytics to S3 as an asset and allow our system to read it through cross-account access?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+# Extensions
+
+We will be able more information into additional fields of the metadata resource.
+Size of template, number of resources are obvious candidates.
+
+# Unresolved questions
+
+- How acceptable is the giant base64-encoded blob in the template? If we want
+  to cut down on it, what clever strategies can we come up with?
```
eladb, line 489: assets?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+* Library and version
+* Qualified class name; the qualified class name itself consists of two components:
+  * Submodule
+  * Class name
+
+Construct identifiers will be formatted like this:
+
+```text
+library@version:[submodule.]classname
```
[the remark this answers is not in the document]
rix0rrr: There is. I moved version to the front because it compresses better though.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+  "locationInModule": {
+    "filename": "lib/aws-eks/lib/cluster.ts",
+    "line": 685
+  },
+}
+```
+
+However, I don't think this is appropriate to use for the following reasons:
+
+Reading the jsii manifest ties us to using jsii in a way that I don't think
```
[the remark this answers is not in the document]
rix0rrr: I agree, but I'm not convinced it simplifies appreciably much. Since that's the case, I'd rather also not be tied to it to retain flexibility.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+Currently, CDK core contains a hardcoded list of module names for which
+metadata should be reported. We can make this an open opt-in protocol by
+triggering this reporting off purely off of the presence of the `cdk-metadata.json`
+file, which removes the need for the hard-coded names found here:
+
+https://github.com/aws/aws-cdk/blob/1f7311f56457556a6f229e745cd24e3f1e5fe1d3/packages/%40aws-cdk/core/lib/private/runtime-info.ts#L5-L8
+
+Other library vendors like the AWS Solutions Architects Team simply needs
+to add an empty `cdk-metadata.json` to their libraries to get metadata
```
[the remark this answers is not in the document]
rix0rrr: I don't really see the problem with an opt-in system.

* The data goes to us, so unless you've got a tie to us you're not reading it (and hence your reasons for opting in are zero).
* There's not really any security to speak of today, as the system is wide open and if anyone wants to spoof they're free to do so at any point, and there's not really anything we can do about it. Library-based opt-in or allow list in the core library doesn't change that.

At the same time, an open opt-in system will make it easier for other future construct vendors to onboard onto this system (especially internal construct vendors).

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+resource.
+
+The Metadata resource looks like this:
+
+```yaml
+Resources:
+  CDKMetadata:
+    Type: AWS::CDK::Metadata
+    Properties:
+      Modules: <v1 module information, deprecated by this RFC>
```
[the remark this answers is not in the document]
rix0rrr: That makes sense. Feels like the receiver is able to do that.

text/0253-cdk-metadata-v2.md
rix0rrr: I'm not keen on the operational/security overhead of a system where we will be able to read a user's buckets.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+### Collecting construct identifiers
+
+There are two strategies for getting a construct's identifier:
+
+* Manually annotate each construct in some way, adding a unique identifier for it
+  (either by passing additional strings, adding decorators, or some other mechanism).
+* Try to automatically infer the identifier components from the reflection
+  capabilities available to us in NodeJS.
+
```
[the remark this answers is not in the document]
rix0rrr: I guess we could, but there's not really any precedent in jsii currently for modifying the JavaScript classes that get emitted.

text/0253-cdk-metadata-v2.md
rix0rrr: Also we would lose the ability to deploy (small stacks) without bootstrapping.

rix0rrr:
Points to discuss:

[ ] require jsii or not
[ ] hardcoded package list or open system
[ ] compression

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+**Construct**: an abstract, reusable specification of Cloud Infrastructure.
+In practice, implemented by a `class`
+**Construct type**: same as "Construct".
+**Construct instance**: a concrete single definition of a construct. In
+practice, a class instance.
+**Construct identifier**: a string describing a specific version of a
+*construct.
+**RTTI**: Run-Time Type Information; a mechanism by which a compiler emits
+information about the types it has processed which can be read by programs
+during execution, used to power reflection.
```
nija-at, line 82: Fix rendering

```suggestion
**Construct**: an abstract, reusable specification of Cloud Infrastructure.
In practice, implemented by a `class`

**Construct type**: same as "Construct".

**Construct instance**: a concrete single definition of a construct. In
practice, a class instance.

**Construct identifier**: a string describing the specific version of a
construct.

**RTTI**: Run-Time Type Information; a mechanism by which a compiler emits
information about the types it has processed which can be read by programs
during execution, used to power reflection.
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+The analytics payload itself will be preceded by an encoding scheme identifier,
+followed by a `:` and the transmitted data:
+
+```yaml
+Resources:
+  CDKMetadata:
+    Type: AWS::CDK::Metadata
+    Properties:
+      Analytics: <payload-encoding>:<payload>
```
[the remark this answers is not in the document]
nija-at: This comment is not addressed.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+value is a `string[]`).
+
+> If a value does not start with > /^[a-z0-9]{,10}:/ it is considered to
+> be not encoded (the encoding is considered to be "plain").
+
+## Metadata resource decoding
+
+Decoding will happen on the receiving end before the data hits long-term storage,
+in order to make consuming it as trivial as possible. There is no need to be frugal
+with bits anymore once the data hits an AWS datacenter.
```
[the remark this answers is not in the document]
nija-at: This comment is not addressed.

text/0253-cdk-metadata-v2.md
nija-at: Please address this comment. 

Are we confident all parts of the system can be tested? Could we have a situation where metrics are not counted and we're flying blind?

How about e2e tests/canaries from specific accounts?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+* Update metadata service backend to accept new `Analytics` field. This is where
+  decoding should happen back to a flat list again. Metadata service should extract
+  packages from the construct list to continue emitting the old information as well.
+* Update canary to exercise new field.
+* Update metadata reporting job to report on both new and old information.
+* Update metadata resource construction implementation in CDK to switch to RTTI
+  implementation, and start emitting `Analytics` field.
+* Verify metrics are coming in
+
+# Appendix 1: jsii Run-Time Type Information
```
nija-at, line 306:
```suggestion
# Appendix A: jsii Run-Time Type Information
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+Effectively, every class will be emitted as this in the `.js` file (the
+`.d.ts` file remains unchanged).
+
+```js
+// This will be added to the top of every source file
+const vfqnSym = Symbol.for('jsii.vfqn');
+
+export class SomeClass {
+  // Every class gets an additional field added to it
+  private static [vfqnSym] = 'module.SomeClass@1.2.3';
```
nija-at, line 320: How do we get the module in the case of `monocdk` vs a normal jsii package? 
Please add them here.

I suppose the version comes from `package.json`?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+constructs, authored and vended by AWS.
+
+We decided against an open system, instead preferring to have a specific
+allow-list of packages (or package prefixes) we are going to report on. For
+example, the current list looks like this and can be found
+[here](https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/core/lib/private/runtime-info.ts):
+
+```ts
+const WHITELIST_SCOPES = ['@aws-cdk', '@aws-solutions-konstruk', '@aws-solutions-constructs', '@amzn'];
+const WHITELIST_PACKAGES = ['aws-rfdk'];
```
RomainMuller, line 137:
```suggestion
const ALLOWLIST_SCOPES = ['@aws-cdk', '@aws-solutions-konstruk', '@aws-solutions-constructs', '@amzn'];
const ALLOWLIST_PACKAGES = ['aws-rfdk'];
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+```
+
+Having a PR opened against the CDK to update this list will be a clear
+interaction moment between an AWS team trying to onboard with this mechanism
+and us, and will prevent 3rd party authors from accidentally (or maliciously)
+onboarding themselves as well and sending us information that we don't want to
+be responsible for.
```
RomainMuller, line 144:
```suggestion
onboarding themselves as well and sending us information that we don't have a
legitimate use-case for.
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+Trying various methods of compression on it yields the following (keeping in mind
+that the result of binary compression will need to be base64-encoded to transmit via
+a CloudFormation template):
+
+| Description                | Raw size    | Base64ed size |
+|----------------------------|-------------|---------------|
+| Plaintext                  | 23k         | -             |
+| gzipped                    | 3.6k        | 4.8k          |
+| Zopfli                     | 3.4k        | 4.6k          |
+| bzip2                      | 3.4k        | 4.6k          |
```
RomainMuller, line 197: Curious if `zstd` is significantly different to those... Although this is probably just misplaced curiosity 🙃

Additionally - how's those configured? Default? There's a world of difference between `gzip -9` and `gzip -1` for example.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+  CDKMetadata:
+    Type: AWS::CDK::Metadata
+    Properties:
+      Analytics: <payload-encoding>:<payload>
+```
+
+The only valid payload encodings at the moment are:
+
+* `deflate64`: base64-encoded, gzipped data.
+* `plain`: plaintext data.
```
RomainMuller, line 235:
```suggestion
* `plain`: plaintext data (encoded as UTF-8).
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+The only valid payload value encodings at the moment are:
+
+* `afx`: affix-encoded string list (see Appendix B).
+* `plain`: plaintext data.
+
+Encoding may change the type (the encoded value is a `string` but the decoded
+value is a `string[]`).
+
+> If a value does not start with > /^[a-z0-9]{,10}:/ it is considered to
+> be not encoded (the encoding is considered to be "plain").
```
RomainMuller, line 256: I'd rather consider it is invalid in that case.

If the value does not match expected format, it may well contain data that you did not expect, and may not be safe to drop in logs / storage / etc...

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+Though, that would require us to gain read access to a customer's S3 bucket
+from our account, which has security implications; it would also require a
+customer to have a bootstrapped account for *every* deployment, even small
+ones (though we could lift *that* limitation by hosting bootstrap buckets on behalf
+of users).
+
+# Unresolved questions
+
+- How acceptable is the giant base64-encoded blob in the template? If we want
+  to cut down on it, what clever strategies can we come up with?
```
RomainMuller, line 288: I think a key thing is we want to make sure *customers* are able to decode this data locally with great ease, so they can independently check that we are not collecting anything they would not want to disclose.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+* Inform dependent teams that they need to upgrade to the jsii version that supports
+  RTTI.
+* Update metadata service backend to accept new `Analytics` field. This is where
+  decoding should happen back to a flat list again. Metadata service should extract
+  packages from the construct list to continue emitting the old information as well.
+* Update canary to exercise new field.
+* Update metadata reporting job to report on both new and old information.
+* Update metadata resource construction implementation in CDK to switch to RTTI
+  implementation, and start emitting `Analytics` field.
+* Verify metrics are coming in
```
RomainMuller, line 304:
```suggestion
* [ ] Implement RTTI support in in jsii.
* [ ] Inform dependent teams that they need to upgrade to the jsii version that supports
  RTTI.
* [ ] Update metadata service backend to accept new `Analytics` field. This is where
  decoding should happen back to a flat list again. Metadata service should extract
  packages from the construct list to continue emitting the old information as well.
* [ ] Update canary to exercise new field.
* [ ] Update metadata reporting job to report on both new and old information.
* [ ] Update metadata resource construction implementation in CDK to switch to RTTI
  implementation, and start emitting `Analytics` field.
* [ ] Verify metrics are coming in
```

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+  public static forObject(object: any): TypeInformationForObject | undefined;
+
+  /** Type FQN */
+  public readonly fqn: string;
+
+  /** Module version */
+  public readonly version: string;
+
+  /** Base class type information */
+  public readonly baseClass: TypeInformation | undefined;
```
RomainMuller, line 344: Given this format, I actually think we'll be significantly better of modeling the `[Symbol.for('jsii.vfqn')]` property, so we save having to parse it out (i.e: actually set it to some object such as `{ fqn, version }`).

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+
+We will start by having the jsii compiler add FQN information to every
+JavaScript class it emits.
+
+Effectively, every class will be emitted as this in the `.js` file (the
+`.d.ts` file remains unchanged).
+
+```js
+// This will be added to the top of every source file
+const vfqnSym = Symbol.for('jsii.vfqn');
```
RomainMuller, line 316: I gens the naming here is 100% illustrative... But I kinda wonder why you'd not simply call that symbol `Symbol.for('jsii.rtti')`?

text/0253-cdk-metadata-v2.md
RomainMuller, line 320: How does this work for "pure" interfaces? This includes cases where we return a private type that implements a public interface?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,488 @@
...
+  /** Type information for the first base class that has type information */
+  readonly typeInformation: TypeInformation;
+
+  /**
+   * How many classes were skipped before finding an ancestor with type information
+   *
+   * This number will be 0 if the object itself was an instance of a class with
+   * type information.
+   */
+  readonly classesSkipped: number;
```
RomainMuller, line 357: What's the use-case for this information? Smells of future-proofing?

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,639 @@
...
+
+The current CDK Metadata is based on libraries loaded into the NodeJS process. When
+we move to CDK v2, the entire CDK library will be vended as one library, destroying
+our ability to measure in the same way.
+
+At the same time, our current method fails in the face of multiple distinct
+stacks in the same application. If two different stacks exist in the same
+application, each using a single construct library, then we will count both
+construct libraries as "having been used" for both stacks (because they were
+loaded into memory at the time the stacks were synthesized).
```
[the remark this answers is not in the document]
rix0rrr: Your suggested rewrite would change the meaning of the sentence to something I don't intend to say.

text/0253-cdk-metadata-v2.md
rix0rrr: It's all max compression

text/0253-cdk-metadata-v2.md
rix0rrr: jsii has all this information already

rix0rrr: It does not. This is already explained elsewhere, and is a regrettable problem.

We *might* conceivably add a feature to `jsii-rtti` to manually annotate an anonymous object instance. But that seems like a v2 concern.

text/0253-cdk-metadata-v2.md
rix0rrr: Sure

text/0253-cdk-metadata-v2.md
rix0rrr: Given an inheritance chain that potentially looks like this:

```
Root -> ClassA -> ClassB -> (unknown class) -> (unknown class) -> INSTANCE
                     ^
                     |
              this is the first one we can return
```

Just saying it's a `ClassB` is not accurate.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,523 @@
...
+    Properties:
+      Modules: <v1 module information, deprecated by this RFC>
+      Analytics: <v2 construct information, introduced by this RFC>
+```
+
+### Construct list size
+
+The list of construct identifiers we have discovered may be sizeable. A rough
+estimate of the maximum number of constructs used in a stack will be about
+400 (maximum of 200 L1s with a corresponding L2 wrapper). We will therefore
```
eladb, line 165: Does it make sense to record both L2 and L1 in case the L2 is used? It actually makes sense to omit the L1 in this case I think so we can easily identify direct L1 usage (which is of particular interest for our construct library planning).

The simplest solution would be to omit L1s that are named `Resource` and that their patent construct inherits from the Resource base class.

text/0253-cdk-metadata-v2.md
```diff
@@ -0,0 +1,523 @@
...
+The only valid value encodings at the moment are:
+
+* `afx`: string list, affix-encoded as a string (see Appendix B).
+* `literal`: literal JSON value.
+
+The `$encoding` key may be missing, in which case `literal` is implied.
+
+Encoding may change the type (for `afx` the encoded value is a `string` but the decoded
+value is a `string[]`).
+
```
eladb, line 262: One concern this might raise is that the metadata sent to aws will become opaque. As mitigation maybe we can emit an additional file to the cloud assembly (`analytics.json`) with the unencoded value. This way people will have visibility to it if they examine `cdk.out`

text/0253-cdk-metadata-v2.md
eladb: Did you consider using assets?

text/0253-cdk-metadata-v2.md
rix0rrr: We can do that, and I'm not opposed to it. I'm not convinced that it'll be necessary though, given that the behavior of the afx-encoded string will be such that if you paste it into `bash` by doing  `echo <afx-encoded-string-here>`, `bash` brace expansion is going to reconstruct the full list.

It's not like it's a secret, we're publishing the spec out in the open. 

But sure, if you feel it should be MORE readable than that, we can also emit the file.

text/0253-cdk-metadata-v2.md
rix0rrr: I've been on the fence about this, and trying to come up with a more general rule that would also cover this (but failing to).

But I guess it makes sense to just special-case it, you're right.

text/0253-cdk-metadata-v2.md
eladb: I think it will add a level of trust

PR #261, opened 2020-10-14
Updates to "RFC 249: v2.0 Experiments"

skinny85:
- Move out discarded options to an appendix
- Move out the explanation of why can't mono-CDK main modules break backwards compatibility into an appendix
- Clarified what does `@experimental` mean in mono-CDK
- Moved out the data on modules with unstable dependencies into an appendix

[Rendered version](https://github.com/skinny85/aws-cdk-rfcs/blob/experiments/review-comments/text/0249-v2-experiments.md)

---

### DIFF
--- a/text/0249-v2-experiments.md
+++ b/text/0249-v2-experiments.md
@@ -6,11 +6,10 @@ rfc pr: https://github.com/aws/aws-cdk-rfcs/pull/250
 
 # Summary
 
-When CDK version `2.0` is released to General Availability (GA), the main
-Construct Library package that we vend will no longer contain unstable code.
-Unstable modules, including modules in Developer Preview, will be vended
-separately. Users will not experience breaking changes in minor releases of the
-CDK unless they explicitly opt-in to using unstable APIs.
+When CDK version `2.0` is released to General Availability (GA),
+the single monolithic Construct Library package we vend will no longer allow breaking changes in its main modules.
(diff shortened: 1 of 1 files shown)


rix0rrr (approved):

text/0249-v2-experiments.md
```diff
@@ -364,17 +244,167 @@ Disadvantages:
...
+which was exactly the problem mono-CDK was designed to prevent in the first place.
+
+# Appendix B - modules depending on unstable modules
+
+This section contains the snapshot of the interesting dependencies between Construct Library modules as of writing this document.
+
+## Stable modules depending on unstable modules
+
+```
+⚠️  Stable module '@aws-cdk/aws-applicationautoscaling' depends on unstable module '@aws-cdk/aws-autoscaling-common'
```
njlynch, line 296: I know this has already been merged, but for the next revision...

I think it would be really helpful to identify 1-3 categories of these dependencies, and then have a very quick statement about how Option N would handle this.

Potential categories:
* Implementation detail -- the module relies on an experimental construct as an implementation detail, completely hidden from the user.
* Targets/"Join modules" -- The dependency exists to bridge between two services (e.g., stepfunctions-tasks and batch). As new services are added, new targets will need to be created.
* Exposed as props/`add*` methods - E.g., Lambda/EFS

How would option 2/5 tackle each of the above?

PR #262, opened 2020-10-15
Add option 5 to experiments doc

rix0rrr:
* Tracking issue: #
* [Rendered version](/aws/aws-cdk-rfcs/tree/huijbers/v2-experiments-option-5/text/0249-v2-experiments.md)

---


text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+from the **public interface** of the stable release. The implementation
+(`.js` files) will be the same in both releases, so stable APIs in
+`aws-cdk-lib` can safely use experimental APIs internally (though not in
+their public API surfaces).
+
+**How does stripping APIs from the public work?**
+
+| TypeScript            | JavaScript | Java, Python, C#, Go            |
+|-----------------------|------------|---------------------------------|
+| Stripped from `.d.ts` | N/A        | No bindings generated by `jsii` |
```
skinny85, line 218: Minor: I feel like this table should be organized that each language is a row, not a column... kind of difficult to read right now.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
+In TypeScript, the compiler will prevent usage of APIs that are not advertised
+in the `.d.ts` files. Users will be able to cast this away using `as any` though.
```
skinny85, line 221: > Users will be able to cast this away using `as any` though.

That's true also of `private` properties and methods in TypeScript, right?

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+They will also be able to confuse the compiler enough that they will get to pass
+struct properties that are technically `@experimental`.
+
+For JavaScript, an IDE will (hopefully) not autocomplete APIs missing from
```
skinny85, line 225: Why the "hopefully" here? Are you saying there's a chance the IDE _can_ autocomplete APIs missing from the `.d.ts` files? If so, how can we solve that?

If we're certain it can't happen, I would remove the "hopefully", it muddles the message here quite a bit.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+the `.d.ts` files, but otherwise nothing prevents users from calling the
+APIs, or passing struct properties that are technically experimental. On the
```
skinny85, line 227: > passing struct properties that are technically experimental

Again, I feel like it's very similar to the above sentence, and I still don't understand what this means.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+**How does stripping APIs from the public work?**
+
+| TypeScript            | JavaScript | Java, Python, C#, Go            |
+|-----------------------|------------|---------------------------------|
+| Stripped from `.d.ts` | N/A        | No bindings generated by `jsii` |
+
+In TypeScript, the compiler will prevent usage of APIs that are not advertised
+in the `.d.ts` files. Users will be able to cast this away using `as any` though.
+They will also be able to confuse the compiler enough that they will get to pass
+struct properties that are technically `@experimental`.
```
skinny85, line 223: Sorry, I don't understand this sentence... can you show a quick code example inline here that illustrates what you mean?

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
+For JavaScript, an IDE will (hopefully) not autocomplete APIs missing from
+the `.d.ts` files, but otherwise nothing prevents users from calling the
+APIs, or passing struct properties that are technically experimental. On the
+other hand, JavaScript devs are by definition required to read the docs,
+so they can't miss the *experimental* banners displayed there.
+
+For jsii client languages, the bindings for experimental APIs will simply not
+exist so there's no way to work around that (though we'll have to check
+whether for example Python makes it possible to pass struct values that
```
skinny85, line 233: > though we'll have to check whether for example Python makes it possible to pass struct values that aren't in the declared API surface

You lost me here again... can you expand on this point?

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
+**Vending**
+
+There are a number of different ways we can choose to distinguish the two builds.
+See Appendix D for an evaluation of all the possibilities.
+
+- Distinguish by package name
+  - Can't be used in **pip**: there's no way to have an app that uses the experimental
+    library work with a library that uses the stable library (no way to prevent the
+    transitive dependency package with a different name from being installed).
```
skinny85, line 244: I don't understand exactly why that is... I thought that if a JSII-enabled JS/TS library had a peer dependency on `aws-cdk-lib`, then its `pip` version would **not** have a dependency on the Python version of `aws-cdk-lib`, because the customer of the app is supposed to bring in a dependency on the Python version of `aws-cdk-lib`, in whatever version they want? Is my understanding incorrect, and a peer-only dependency in JavaScript gets translated in Python to a regular dependency?

Or is the problem here a Python-only construct library?

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+- Distinguish by prerelease tag
+  - Can't be used in **Maven**: an app can't depend on stable versions only
+    (a range of `[1.60.0,2.0.0)` would also match `1.61.0-experimental`).
+- Distinguish by major version
+  - All package managers have a way to bring in a different version of a package,
+    although many of them will complain that declared dependencies aren't satisified.
+
+Distinguishing by major version is the only possibility that works across all
+package managers.
```
skinny85, line 253: If the conclusion here is "we have to do it by major version", then I would state that here, and move the explanation of why that is into the appendix itself.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+There are a number of different ways we can choose to distinguish the two builds.
+See Appendix D for an evaluation of all the possibilities.
+
+- Distinguish by package name
+  - Can't be used in **pip**: there's no way to have an app that uses the experimental
+    library work with a library that uses the stable library (no way to prevent the
+    transitive dependency package with a different name from being installed).
+- Distinguish by prerelease tag
+  - Can't be used in **Maven**: an app can't depend on stable versions only
+    (a range of `[1.60.0,2.0.0)` would also match `1.61.0-experimental`).
```
skinny85, line 247: 1. By "app", do you mean "library" here? I think it's important to be precise in this RFC, this is complicated as-is 😃.
2. Not sure if that's very important. We can always make the Java packages depend on a fixed version. In Maven, the closest dependency to the root always wins, so depending on a fixed version shouldn't be an issue.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+Distinguishing by major version is the only possibility that works across all
+package managers.
+
+Something like:
+
+| Scheme   | Stable   | Experimental            |
+|----------|----------|-------------------------|
+| Even/odd | `2.60.1` | `3.60.1`                |
+| +100     | `2.60.1` | `102.60.1`              |
+| <1       | `2.60.1` | `0.2.60-experimental.1` |
```
skinny85, line 261: Do we really need both `0.x` _and_ `experimental` in the version? I think just `0.x` should make that pretty clear?

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
+**Summary**
+
+Advantages:
+
+1. Stable code can depend on unstable code.
```
skinny85, line 267: Can we use the same wording as in the other options?

```suggestion
1. It's possible for stable modules to depend on unstable ones
  (see Appendix B for data on how necessary that is for the CDK currently).
```

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+2. Stabilizing code will not be a breaking change for customers.
+3. Stabilizing code will be a simple operation for CDK developers.
+4. No additional management overhead of multiple packages/repos/etc.
```
skinny85, line 270: We're not interested in the repos argument here really. So I would just leave it as "No additional management overhead from splitting the code into multiple packages".

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
```
skinny85, line 271: I believe you missed an advantage, which is "It's possible for unstable modules to depend on other unstable modules".

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+Disadvantages:
+
+1. Protection not offered for JavaScript users, can be bypassed in TypeScript.
```
skinny85, line 274: How important is this? Aren't TypeScript users one `as any` away from breaking all compile-time guarantees like `private` today?

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+
+We have the following options to pick from:
+
+- Different package names (`aws-cdk-lib` vs `aws-cdk-lib-experimental`).
+- Pre-release version tag
+- Different major versions
+
+We need to evaluate all of these, for each package manager, on the following criteria:
+
+- Can applications depend on an upwards unbounded range of *stable* builds? (important)
```
skinny85, line 498: "applications" - do you mean "libraries" here?

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+  "peerDependencies": {
+      "aws-cdk-lib": "1.60.0"
+      // Note: ideally you'd like to express "aws-cdk-lib-experiments": "1.60.0" here, but
+      // if you do that "npm ls" will complain about an unmet peer dependency
+  },
+  "devDependencies": {
+      "aws-cdk-lib": "npm:aws-cdk-lib-experiments@1.60.0"
+  }
+}
+```
```
skinny85, line 575: Sorry, I don't understand the problem here. If a library peer depends on `aws-cdk-lib-experiments`, then the application _should_ declare a dependency on `aws-cdk-lib-experiments`, and NPM complaining about missing peer dependency if the application only depends on `aws-cdk-lib` is a _good_ thing I'm pretty sure?

What am I missing?

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+
+**Maven**
+
+No need to change the sources when using a stable library with an
+experimental app, as the namespaces and class names will be the same.
+
+Using a stable library in an experimental app can be done in two ways:
+
+* The library uses the `provided` dependency flavor (similar to `peerDependencies` in NPM).
+  Requires the app to bring the right version of CDK.
```
skinny85, line 590: `provided` seems like slam-dunk here.

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+
+No need to change the sources when using a stable library with an
+experimental app, as the namespaces and class names will be the same.
+
+`pip` transitively fetches from the library's `setup.py`, and there's no way
+to override that.
+
+Our only option around that is to have stable libraries not declare any CDK
+dependency in `setup.py`, but then we also lose the ability for libraries to
+specify their dependency version, which is not really acceptable.
```
skinny85, line 607: Is it really such a big deal? I feel like declaring the CDK dependency version is just meant to provide a different error message when you violate it? So is it really such a big deal that it would fail at runtime?

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+Stable ranges: yes, `^1.60.0` does not include `1.61.1-experimental`.
+
+Experimental app uses stable lib: can be done, the lib only uses a peerDependency. `npm ls` will complain about
+a missing peerDependency though.
+
+**Maven**
+
+Stable ranges: no, `[1.60.0,)` *will* include `1.61.1-experimental`.
+
+Experimental app uses stable lib:
```
skinny85, line 656: Simple - works fine if we use `provided` in the library.

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+**NPM**
+
+Stable ranges: yes, `^1.60.0` does not include `1.61.1-experimental`.
+
+Experimental app uses stable lib: can be done, the lib only uses a peerDependency. `npm ls` will complain about
+a missing peerDependency though.
+
+**Maven**
+
+Stable ranges: no, `[1.60.0,)` *will* include `1.61.1-experimental`.
```
skinny85, line 654: Again, I would question whether using a range for Maven is really interesting here.

(Also, is this for the library? Or app? It's really important to be precise here).

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+
+- `aws-cdk-lib@2.60.0`
+- `aws-cdk-lib@3.60.0` (odd ranges are experimental)
+
+Or:
+
+- `aws-cdk-lib@2.60.0`
+- `aws-cdk-lib@102.60.0` (100+ versions are experimental)
+
+| .     | Stable apps | Experimental app uses Stable lib | Without complaints | Lib advertises Experimental |
```
skinny85, line 687: I gotta say, the "Without complaints" category is hard, because you have to remember it's only for the  "experimental app with stable lib" scenario. Maybe call it "...without complaints", to make it more obvious?

skinny85 (commented):
Looks very promising! Quite complicated to get an answer for every package manager, but looks like there's a way forward, which is very exciting.

I'd love making the language a little more precise in a few places, and I think the constraints you have for NPM are too strong in package managers that do not allow multiple versions of the same package to exist in the same dependency closure.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+2. Stabilizing code will not be a breaking change for customers.
+3. Stabilizing code will be a simple operation for CDK developers.
+4. No additional management overhead of multiple packages/repos/etc.
+
+Disadvantages:
+
+1. Protection not offered for JavaScript users, can be bypassed in TypeScript.
+2. Stable/experimental versioning scheme is not based on any well-known industry
+   standard, we're going to have to clearly explain it to people.
+
```
skinny85, line 277: I would add "Requires JSII changes" to the list of downsides.

text/0249-v2-experiments.md
```diff
@@ -408,3 +484,240 @@ Disadvantages:
...
+Like pre-release tags, except we don't get the mixing in of pre-release versions with
+regular versions.
+
+**pip**
+
+Like pre-release tags.
+
+**NuGet**
+
+...pending responses from .NET SDK team...
```
RomainMuller, line 721: AFAIK, .NET does not support supplying multiple version constraints... All my attempts so far resulted in constraints being entirely ignored, resulting in .NET resolving to the lowest existing version (not the desired behavior).

text/0249-v2-experiments.md
eladb: Non issue

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+For JavaScript, an IDE will (hopefully) not autocomplete APIs missing from
+the `.d.ts` files, but otherwise nothing prevents users from calling the
+APIs, or passing struct properties that are technically experimental. On the
+other hand, JavaScript devs are by definition required to read the docs,
+so they can't miss the *experimental* banners displayed there.
+
+For jsii client languages, the bindings for experimental APIs will simply not
+exist so there's no way to work around that (though we'll have to check
+whether for example Python makes it possible to pass struct values that
+aren't in the declared API surface).
```
eladb, line 234: Of course it will allow, but that's akin to javascript.

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+**Vending**
+
+There are a number of different ways we can choose to distinguish the two builds.
+See Appendix D for an evaluation of all the possibilities.
+
+- Distinguish by package name
+  - Can't be used in **pip**: there's no way to have an app that uses the experimental
+    library work with a library that uses the stable library (no way to prevent the
+    transitive dependency package with a different name from being installed).
+- Distinguish by prerelease tag
```
eladb, line 245: What about "snapshot" versions in maven? Aren't those similar to pre-release in npm?

text/0249-v2-experiments.md
eladb: I think we can omit

text/0249-v2-experiments.md
```diff
@@ -199,6 +199,82 @@ Disadvantages:
...
+
+1. Stable code can depend on unstable code.
+2. Stabilizing code will not be a breaking change for customers.
+3. Stabilizing code will be a simple operation for CDK developers.
+4. No additional management overhead of multiple packages/repos/etc.
+
+Disadvantages:
+
+1. Protection not offered for JavaScript users, can be bypassed in TypeScript.
+2. Stable/experimental versioning scheme is not based on any well-known industry
```
eladb, line 275: still feel like there should be an idiomatic way in any package manager to publish a pre-release and it feels like that's the correct mental model for the experimental flavor

text/0249-v2-experiments.md
```diff
@@ -243,37 +243,47 @@ See Appendix D for an evaluation of all the possibilities.
     library work with a library that uses the stable library (no way to prevent the
     transitive dependency package with a different name from being installed).
 - Distinguish by prerelease tag
-  - Can't be used in **Maven**: an app can't depend on stable versions only
-    (a range of `[1.60.0,2.0.0)` would also match `1.61.0-experimental`).
+  - Most logical
```
skinny85, line 246: What's this line...? Did you forget to finish a sentence perhaps? 🙂

text/0249-v2-experiments.md
rix0rrr: An IDE can use other mechanisms than reading the `.d.ts` files. It can parse the JavaScript code, it can do dynamic evaluations of the code and inspect the objects at runtime.

I don't know which ones use what mechanism.

text/0249-v2-experiments.md
rix0rrr: To my understanding, we have `peerDependencies` only because NPM allows multiple copies, which we don't want. A `peerDependency` still allows to model a dependency with an explicit version number though.

In other package managers, we don't have the problem of multiple copies so we don't need the solution of `peerDependencies`. In fact it doesn't exist, and the only way to achieve the same effect in `pip` would be NOT declare a dependency at all. Then you still require the consumer to bring the dependency (good but unnecessary), but the package cannot declare what version it needs (bad).

-------------

All this is a long way to say: `peerDependencies` in npm will be translated into *regular* dependencies in other package managers.

text/0249-v2-experiments.md
rix0rrr: I think a short explainer with a pointer to the appendix for more details reads better. It's not like it takes up giant swathes of space.

text/0249-v2-experiments.md
rix0rrr: No, I specifically mean app.

text/0249-v2-experiments.md
rix0rrr: I guess, but those `0.x` versions already exist and we have to not conflict with them.

In any case, this is water under the bridge.

text/0249-v2-experiments.md
rix0rrr: It's fine that we don't think it's an option with a heavy weight, but I'd still want to keep it in for completeness.

It's definitely a distinction with the partitioned packages variant. 

I will move it further down though.

text/0249-v2-experiments.md
rix0rrr: I do not. We want application builders to migrate along with us to new CDK versions quickly. That is, I would want app builders to depend on ranges, not on point versions.

text/0249-v2-experiments.md
rix0rrr: I see I left something out that was (to me) so obvious that I didn't think it would need writing down. Failure of judgement :).

text/0249-v2-experiments.md
rix0rrr: It's just the 'most logical' :)

text/0249-v2-experiments.md
skinny85: Does this imply that the solution you gave in the Maven part (`provided` scope) is not viable then?

text/0249-v2-experiments.md
skinny85: Interesting. I feel like I have the *exact* opposite approach (you should only ever depend on pinned versions, and any upgrades should be done through a commit that you can revert if anything goes wrong),...

I feel like most tools agree it's better to pin versions, with how popular lock files have become. So maybe this is not an important point for apps? (I agree it's important for libs though)

text/0249-v2-experiments.md
skinny85: I guess this is the same discussion as https://github.com/aws/aws-cdk-rfcs/pull/262#discussion_r509589600

skinny85 (approved):
@rix0rrr since we said we need to have a Proof of Concepts for these, I'm fine with merging Option 5 now, and doing the final edits later.

Issue #264, opened 2020-11-05
Generating L1s from CloudFormation Registry Schema Format


rix0rrr:
|PR|Champion|
|--|--------|
|# |        |

## Description

Instead of consuming the CloudFormation Resource Specification format to generate the set of L1s, we should we consuming the CloudFormation Registry Schema Format instead.

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

eladb:
Duplicate with #77

Issue #77, opened 2019-04-18
CloudFormation Registry Support


eladb:
## Description

This proposal is to stop generating the CloudFormation Resources (also known as *`Cfn*` classes* / *L1 Constructs*) from the legacy [CloudFormation Resource Specification] document, and instead to leverage the improved schema that is provided as part of the [CloudFormation Registry].

[CloudFormation Resource Specification]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html
[CloudFormation Registry]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html

We recommend to use the `import` terminology and `imports/` directory to align with [cdk8s](https://cdk8s.io).

```shell
$ cdk import My::Resource::Example [--language python|dotnet|java|typescript]
imports/my/resource/example.ts
```

The CloudFormation resource spec is evolving and plans to support arbitrary JSON schema for resource properties. This will require that we will be able to generate code that models things like JSON unions (what else?).

* JSON schema to JSII structs: https://github.com/eladb/json2jsii
* Model is here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @eladb 
| Author(s)           | @rix0rrr, @eladb
| API Bar Raiser      | @eladb 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

eladb:
@workeitel wrote:

### Use Case

[...]

PR #266, opened 2020-11-05
RFC 77, RFC 264: CloudFormation Registry Imports

rix0rrr:
These are two distinct RFCs for two distinct features, but we might as
well work on the docs at the same time. Hence, one PR.

* Tracking issue: #77 - importing external resources
* [Rendered version](/huijbers/0264-0077-cloudformation-registry/text/0077-import-external-resources.md)

* Tracking issue: #264 - codegen off the new registry schema
* [Rendered version](/huijbers/0264-0077-cloudformation-registry/text/0264-registry-schema-codegen.md)

---

### DIFF
--- a/text/0264-registry-schema-codegen.md
+++ b/text/0264-registry-schema-codegen.md
@@ -47,7 +47,8 @@ for:
 * What properties are used to hold secrets/passwords/etc (`writeOnlyProperties`).
 * Deprecated properties (`deprecatedProperties`).
 
-Here are extensions that are now possible that will pose problems for us:
+There are things that customers are now able to specify that will pose
+problems for us:
 
 * Anonymous object types (example: see `AWS::Kendra::DataSource`, `DataSourceToIndexFieldMapping`).
   We will need to invent names for these types, which is going to lead to issues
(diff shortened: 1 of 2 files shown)


text/0264-registry-schema-codegen.md
```diff
@@ -0,0 +1,303 @@
...
+  our build. Since a number of resources are automatically being converted
+  from old format to new format, they are missing additional information. We
+  can enhance the spec with information obtained from the [cfn-lint] project.
+* A tool to two versions of the spec. This is necessary to generate the
+  changelog every time we import a new version of the resources. It should be
+  able to identify breaking changes, so that we can extend this tool to
+  resource developers and they can prevent themselves from accidentally
+  releasing breaking changes (reducing churn on downstream systems).
+* A tool to generate code the type and resource definitions found
+  in the schema. This can mostly be based on `json2jsii`, although we will
```
eladb, line 144: Let's definitely try to bake as much of the capabilities into Json2jsii so we get the most leverage for other CDKs (and jsii-based projects).

eladb (changes requested):
Great RFCs! Let's definitely have a discussion as part of these two RFCs around whether we want to continue to vend CFN resources as part of the AWS construct library or just use them as an implementation detail.

eladb (changes requested):
We should make sure to take the recently launched CloudFormation modules into the `cdk import` feature. I'd like to be able to do something like `cdk import MyCompany::S3::Bucket::MODULE`

nija-at:
@rix0rrr - can you check if this is superceded by https://github.com/aws/aws-cdk-rfcs/pull/356.

rix0rrr:
> @rix0rrr - can you check if this is superceded by #356.

From the title, #356 would build on the work proposed here, but the work here would also be useful to people importing constructs in their own accounts.

eladb:
FYI, this is now being implemented and published under [cdklabs/cdk-import](https://github.com/cdklabs/cdk-import).

@rix0rrr I am wondering if it's worth leaving this RFC open or we can merge it for now and continue iterating on the code?

PR #279, opened 2020-12-27
Add option 6 to experiments doc

NetaNir:
* Tracking issue: #
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/neta/option-6/text/0249-v2-experiments.md#option-6-name-experimental-apis-with-experimental-suffixprefix)

---


eladb (commented):
The diff includes a non trivial amount of formatting changes, which makes it almost impossible to review the actual content. Any chance these can be reverted? J

text/0249-v2-experiments.md
```diff
@@ -358,17 +370,288 @@ Disadvantages:
...
+version line - `2.x.x-experimental` are not viable solutions. They might prevent
+users from declaring a fixed dependency on the `aws-cdk-lib` mainline release,
+but they mandate it in other release line we own and maintain.
+
+To allow the CDK team to keep experimenting while still adhering to
+[Semantic Versioning](https://semver.org/#semantic-versioning-200), we suggest
+experimental features will be a part of `aws-cdk-lib`, and maintain backward
+compatibility. Experimental features will be add to `aws-cdk-lib` using a
+specific naming scheme. For example, the first experimental version of the
+`grantWrite` API will be introduced under the name `grantWriteExp_v1`, when it's
```
eladb, line 434: @RomainMuller how would the underscore work with jsii name translations?

eladb, line 434: I think the postfix should somehow indicate this is not the final version. In k8s they use 'v1alpha1', 'v1beta2'.

How about `_pre1` or `_pre2` or something like that?

text/0249-v2-experiments.md
```diff
@@ -358,17 +370,288 @@ Disadvantages:
...
+time to introduce a breaking change to the API, `grantWriteExp_v1` will be
+deprecated and replaced with a newer version - `grantWriteExp_v2`. When the API
+is finalize it will be introduce as `grantWrite`. Deprecated experimental
+versions will only be removed between major versions release.
+
+#### Implementation
+
+Experimental APIs specification in the AWS CDK docs:
```
eladb, line 442: From a positioning (and also tooling) perspective, I'd frame this like this:

"Experimental APIs will not longer be allowed in aws-cdk-lib. All modules and experimental members will become stable starting 2.0.0. This means that from this release, we are committed to never introduce breaking changes in a non-major bump".

"In parallel, we introduce a new model called "api previews" [or some other name we come with]. This model will explicitly indicate if APIs are not final by including a postfix......."




I'd recommend to write this section as a "working backwards" so it's very easy to understand the new user experience.

text/0249-v2-experiments.md
```diff
@@ -6,10 +6,12 @@ rfc pr: https://github.com/aws/aws-cdk-rfcs/pull/250
```
eladb, line 6: Like all RFCs, Can we add a "working backwards" section as the first section of the RFC. Write it as 1-2 paragraphs which will be included as part of the v2.0 announcement. 

See my comment below for proposed phrasing.

eladb (changes requested):

text/0249-v2-experiments.md
NetaNir: This is the section we will add to our general documentation, not the v2 release notes. If it will make it clearer I can replace it

NetaNir:
> The diff includes a non trivial amount of formatting changes, which makes it almost impossible to review the actual content. Any chance these can be reverted? J

Sure, I can revert it.

text/0249-v2-experiments.md
NetaNir: > "In parallel, we introduce a new model called "api previews" [or some other name we come with]. This model will explicitly indicate if APIs are not final by including a postfix......."
> 


I'm not sure I understand what you mean here. What module are you referring to in "introduce a module"? The way I think about it, an experimental APIs can be in every module. As in the example I listed, S3 can have an experimental API on the bucket construct.

eladb: _model_ not _module_..


What I am trying to say is that the model and mechanism we have to say for experimentation which is supported by jsii should stay untouched, only that we will no longer allow it in the CDK. So from the users perspective _there is no more "experimentation"_.

There will be a "new thing" with a new name and perhaps a set of new mechanisms to support it and that's the "api previews" or "api stage" or whatever we decide to call it.

text/0249-v2-experiments.md
NetaNir: Today, only static readonly (constants) are allowed to have `_` in the name. [JSII8002](https://github.com/aws/jsii/blob/f3163e130b3907288584fbe31d1b43550876482f/packages/jsii/lib/jsii-diagnostic.ts#L644): "Method and property (unless they are static readonly) names must use camelCase"


I'll change it and remove the underscore

skinny85:
> > The diff includes a non trivial amount of formatting changes, which makes it almost impossible to review the actual content. Any chance these can be reverted? J
> 
> Sure, I can revert it.

Would love to see the revert too  🙂.

text/0249-v2-experiments.md
NetaNir: I actually think that `alpha`/`beta` might make more sense as it universally known to be a prerelease version. I know it is not exactly our intention with experimental APIs naming, but could save us from inventing a standard and having to explain it?

text/0249-v2-experiments.md
NetaNir: Just to clarify. We want to add a working backward section to the general RFC or specifically to option 6 adde in this PR?

text/0249-v2-experiments.md
NetaNir: I think I get what you are saying, you want to completely remove the use of the term "experimental" in v2. v2 is completely stable, and we are adding a new concept "api lifecycle" (trying something), which will allow us to get new APIs early in front of developers so we can receive feedback before we finalize them.

@eladb  is that correct?

NetaNir: Replaced the implementation section with the section from v2 release notes.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+> /**
+> * This methods grants awesome powers
+> * @deprecated use `grantAwesomePowerPreV2`
+> */
+> grantAwesomePowerPreV1()
+> ```
+
+> When we decide its time to graduate the API, the latest preview version will
+> be deprecated and the final version - `grantAwesomePower` will be added.
+
```
NetaNir, line 485: Maybe there is a better way than code samples to add it, I wanted to add something that will be visually easy to understand. Maybe a snippet from our docs? Open for suggestions.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+  will result in a less aesthetic user code.
+- Graduating a module will require a lot of code changes from our users to
+  remove the prefix/suffix.
+- A lot of deprecated code in aws-cdk-lib, possibly blowing up `aws-cdk-lib`.
+  This might not be a real concern as we can reuse a lot the code between
+  different version of an API.
+
+**How can we encourage users to upgrade to a newer version?**
+
+When a new minor version introduces a breaking change to an experimental API,
```
eladb, line 575: Search & replace "experimental" with "preview" in this section, so we can "feel" if this semantics works:

```suggestion
When a new minor version introduces a breaking change to an preview API,
```

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+users have a clear motivation to upgrade to the new version of the API - they
+must do so in order to upgrade to a newer version of the AWS CDK. If the API is
+only deprecated, users have no motivation to upgrade to a newer version of the
+API, even worse, they might not be aware that a newer version exists. One way to
+encourage users to upgrade to a newer version of an API, is to supply tools that
+will inform users. For example, we can add a capability to the `cdk doctor`
```
eladb, line 581: I think the simplest and most effective tool will be runtime warnings when deprecated APIs are used.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+Newer version of experimental APIs your application is using are available. You should consider upgrading.
+To see which experimental APIs can be upgraded, execute `cdk --show-deprecated-api-usage`
+```
+
+**If there are no breaking changes, are the APIs really experimental?**
+
+Given that experimental APIs are safe to use, and no breaking changes will be
+introduced, we might consider not referring to them as experimental, and simply
+add versioning in the API name. The first version of `grantWrite` will be named
+`grantWriteV1`, the second version will be named `grantWriteV2` and so on. While
```
eladb, line 599:
```suggestion
`grantWritePre1`, the second version will be named `grantWritePre2` and so on. While
```

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+the API will not break, we should still add `experimental` for the following
+reasons:
+
+1. **Real-estate**: `grantWrite` is a much better name than `grantWriteV3` for
```
eladb, line 603:
```suggestion
1. **Real-estate**: `grantWrite` is a much better name than `grantWritePre3` for
```

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+To see which experimental APIs can be upgraded, execute `cdk --show-deprecated-api-usage`
+```
+
+**If there are no breaking changes, are the APIs really experimental?**
+
+Given that experimental APIs are safe to use, and no breaking changes will be
+introduced, we might consider not referring to them as experimental, and simply
+add versioning in the API name. The first version of `grantWrite` will be named
+`grantWriteV1`, the second version will be named `grantWriteV2` and so on. While
+the API will not break, we should still add `experimental` for the following
```
eladb, line 600: We should not use the `@experimental` for this. We should completely forbid (and enforce) using it in the AWS CDK across the board (but leave this as a jsii mechanism for others to use if they want).

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+3. **Setting the right expectations**: When an API is marked as experimental
+   users are aware that this is not the final version of the API, and its
+   deprecation is expected.
+
+### Open questions
+
+If we decide to implement option 6, we will need to formalize the experimental
+naming strategy.
+
+- Should experimental modules be named with `experimental`? e.g `aws-synthetics`
```
eladb, line 616: Module names and structure may also change as part of the development iterations, so I would also allow marking the module itself as a preview.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+  be named `aws-synthetics-pre-v1`, or is it sufficient that its types be named
```
eladb, line 617: Again, don't use `v1` here because it will be confusing with the major version of the AWS CDK itself:

```suggestion
  be named `aws-synthetics-pre1`, or is it sufficient that its types be named
```

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+  as experimental, e.g `CanaryPreV1`.
+- If a type that contains other types is experimental, is it sufficient that the
+  containing type is named as `experimental` or should its properties/methods
```
eladb, line 620:
```suggestion
  containing type is named as a preview or should its properties/methods
```

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+  should be named as experimental as well. e.g if the `Canary` construct is
```
eladb, line 621: I think that if a type is named `XxxPre1` then members don't need to also have the postfix. This will also reduce the amount of changes required when an API is finalized.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+  ```ts
+  const canary = new CanaryPreV1(this, 'MyCanary', {
+  ...
+  });
+  canary.createAlarmPreV1(...)
+  // or
+  canary.createAlarm(...)
+  ```
+
 ## 3. Extra unstable precautions
```
eladb, line 634: Missing: what kind of tools/mechanisms we need to support this (if any?). Do we need some awslint/jsii support?

eladb (commented):
Yes!! I think we are converging here. Option 6 FTW

Comments are mostly minor/syntactic.

We need to brainstorm a bit and finalize the actual terminology and naming conventions.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+
+  ```ts
+  const canary = new CanaryPreV1(this, 'MyCanary', {
+  ...
+  });
+  canary.createAlarmPreV1(...)
+  // or
+  canary.createAlarm(...)
+  ```
+
```
skinny85, line 633: My general feedback on Option 6 is that it's not baked enough yet.

It gives an example of evolving a method, but the interesting part here are the questions that have not been answered:

1. How are modules organized? I.e., I create a new L2. What module does it live in? `aws-cdk-lib`? If so, what's the name of the namespace for that module? If not `aws-cdk-lib`, where does it go?
2. Continuing the example from above ("I create a new L2"): how are the classes and interfaces named? What's the name of the resource interface (i.e., `IBucket`). What's the name of the Construct class with the new L2? What about other, non-Construct classes introduced as part of that L2?

etc.

Right now, this is simply a convention ("I'm not sure about `someMethod()`, let me call it `someMethodVer1()` to hedge my bets"), not a mechanism for doing experiments.

text/0249-v2-experiments.md
```diff
@@ -418,23 +675,6 @@ Disadvantages:
   other instances of using unstable code, working against stated goal #1.
 - Requires changes in JSII.
 
-### Force a naming convention for unstable code
```
skinny85, line 421: In my opinion, this is where proposal 6 should live, as it's not a full-fledged solution to the problem like the other solutions are, just a convention (and actually, it can be used with all 5 proposed options).

skinny85 (dismissed):
My general feedback is that this PR is a little too big 🙂. It does 2 things:

1. Rejects option 5 (this should be done by moving it to Appendix C).
2. Introduces a new option.

Can we please split the PR into those 2 things? I think it will be much easier to review this way.

As for option 6, with how it's written now, it's just a convention for method naming. It's not an actual solution to the problem the way the other 5 options are. Given that, I would prefer it moved to the `## 3. Extra unstable precautions` section.

Thanks for the effort on moving this conversation forward @NetaNir !

text/0249-v2-experiments.md
NetaNir: Yup. All in `aws-cdk-lib`. 

It is a convention, we need to formalize it for all types. 

If by "a mechanism for doing experiment" you mean a mechanism that allows us to release breaking changes, this PR suggest that any such solution is not viable.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+declare a range of supported versions which include an experimental version.
+Sticking to the above example, a CDK construct library declaring a peer
+dependency on version `^1.60.0` of `aws-cdk-lib`, can not be used in a CDK
+application that declares a dependency on `aws-cdk-lib` version
+`1.61.0-experimental`. This is because the patch part in `1.60.0` and
+`1.61.0-experimental` is not equal, which means `1.61.0-experimental` does not
+satisfy `^1.60.0`. In npm versions prior to npm-v7, if `peerDependencies`
+requirements are not met, executing `npm install` would only issue a warning. In
+npm-v7, which automatically tries to install `peerDependencies`, executing
+`npm install` will throw an error.
```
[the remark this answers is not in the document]
NetaNir: Mainly because there is no easy way for libraries to express they support in two major versions.

text/0249-v2-experiments.md
```diff
@@ -67,7 +67,13 @@ levels of abstraction. Changing to vending the Construct Library as a monolithic
 package is one part of making that possible; we should make sure our approach to
 unstable code also takes this into account.
 
-## 3. The CDK team can still perform API experiments
+## 3. We want to encourage customers to not declare a fixed version on packages we own, and are intended to be widely used.
```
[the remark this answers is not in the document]
NetaNir: I disagree, I think that this should be a goal.

> and in fact, with mechanisms like package-lock.json and similar ones, the customer effectively depends on a fixed version anyway

But they have tools such as dependebot that will update their lock file according to the declared version in the `package.json` file. It is a very different approach than using fixed version in a `package.json` files, which basically means "I only ever want to upgrade to a newer version of this package manually". 

This goal is the motivation behind option 6, so I think we should agree on it first.

text/0249-v2-experiments.md
```diff
@@ -67,7 +67,13 @@ levels of abstraction. Changing to vending the Construct Library as a monolithic
...
+
+Fixed versions are a bad practice, they prevent users from getting updates that
+includes critical fixes, security patches and more. The mechanism we choose to
+implement experimental features should adhere to the same standard.
+
+## 4. The CDK team can still perform API experiments
```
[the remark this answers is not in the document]
eladb: The goal is to continue to be able to use this method

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
    although something is to be said for it being the same with `private` fields
    today.
 
+#### POC results
+
+Option 5 was rejected since it relies on the use of prerelease qualifiers in a
```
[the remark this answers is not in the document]
eladb: I don't think this is necessary right now and will make this PR harder to read. All options that were not selected are rejected so the distinction is moot

I'd leave this here in context.

text/0249-v2-experiments.md
eladb: I agree with Neta. The goal of this RFC is to allow us to release APIs that are not "final" so that we can get feedback from real use before we commit to the final API. Our analysis shows that there is no viable way to do that and continue to respect semantic versioning. Option 6 proposes viable a model that allows us to achieve this goal.

It is also the exact approach taken by Kubernetes to achieve the same goal, which validates this approach further.

text/0249-v2-experiments.md
eladb: Not sure what you mean how option 6 it can be used with all five solutions. It stands on its own, and the only viable approach as I read it. 

I'd also suggest that at this point in the game, after a few months of deliberations, we should focus on finalizing the decision and less on the doc structure and organization.

skinny85: > Not sure what you mean how option 6 it can be used with all five solutions. It stands on its own, and the only viable approach as I read it.

Because you can use the `VerX` suffix for non-stable methods regardless of which package the class the method is on lives in.

> I'd also suggest that at this point in the game, after a few months of deliberations, we should focus on finalizing the decision and less on the doc structure and organization.

This is not nitpicking about the structure of the document. This is about the fact that we still have decisions to make. If choosing option 6 still forces us to make a decision: "which package does the non-stable code live in", as it does currently with the way it's written, then we need to have the available options listed in section "Proposed changes", chapter 2, with the options that we rejected moved to Appendix C. Otherwise, we'll keep chasing our tail, as we did so many times on this issue before.

I want to have a clear choice: "These are the 2 options we're left with (with 4 options rejected, listed in appendix C), let's decide between the 2".

Also, notice that the various options are in a chapter titled "2. Separate "breakable" code from mono-CDK main modules". If option 6 does not allow shipping code that we can break, then it _definitely_ does not belong in that chapter.

text/0249-v2-experiments.md
skinny85: > If by "a mechanism for doing experiment" you mean a mechanism that allows us to release breaking changes, this PR suggest that any such solution is not viable.

Not possible to release breaking changes _in `aws-cdk-lib`_.

If we have a separate module with breakable code, like the other 5 options beside option 5 specify, released with version `0.x`, then it definitely _is_ viable (see options 2. and 3. for details).

skinny85: > Yup. All in `aws-cdk-lib`.

Great. Which namespace? Let's say I'm adding the first L2 for a module `aws-xyz` that's currently `cnf-only`. What namespace does the new L2 land in? `aws-cdk-lib/aws-xyz`? Or somewhere else?

> It is a convention, we need to formalize it for all types.

Yes. That's kind of the purpose of this RFC - to propose a solution that covers all of the cases we know we need to deal with. It's hard to judge the merits of option 6 without seeing the entire proposal - how does it treat constructs, props, classes, interface properties, interface methods (will we have `bindVer1()` now, for example?), etc.

Right now, the proposal is a convention we already use, in a slightly different way. For example, in the SecretsManger module, which is stable, we screwed up the `fromSecretName()` method; so now, we have
`fromSecretNameV2()`.

text/0249-v2-experiments.md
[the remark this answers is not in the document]
skinny85: I think this is not a goal, but simply a consequence of the observation that "we can't release breaking changes in new versions of `aws-cdk-lib`".

But I honestly don't care that much, do what you feel is right 🙂 (if you're going to leave it, maybe use the phrasing I suggested above, because right now it reads pretty weird 😬).

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+**patch** are equal. This means that there is no way for a constructs library to
+declare a range of supported versions which include an experimental version.
+Sticking to the above example, a CDK construct library declaring a peer
+dependency on version `^1.60.0` of `aws-cdk-lib`, can not be used in a CDK
+application that declares a dependency on `aws-cdk-lib` version
+`1.61.0-experimental`. This is because the patch part in `1.60.0` and
+`1.61.0-experimental` is not equal, which means `1.61.0-experimental` does not
+satisfy `^1.60.0`. In npm versions prior to npm-v7, if `peerDependencies`
+requirements are not met, executing `npm install` would only issue a warning. In
+npm-v7, which automatically tries to install `peerDependencies`, executing
```
skinny85, line 398: Can we actually update [this section](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0249-v2-experiments.md#prerelease-tags) with these findings about `npm-v7`?

text/0249-v2-experiments.md
eladb: I'd like us to focus on the content and less on the structure at this point. 

As I read this, it's evident at this point that option 6 is the only viable approach to address our need for publishing non-final APIs. 

As I see it, the purpose of this PR is to make sure we didn't overlook anything before we sign off.

I agree that we need a few more details to cover the various types of APIs (modules, classes, methods, etc), but other then that the solution seems solid.

May I suggest that we leave the section reordering for a subsequent PR after this is finalized. I really rather not to have to review this once more if this is refactored again.

text/0249-v2-experiments.md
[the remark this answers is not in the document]
skinny85: Hence why I said, in my opinion, this PR should be split into 2 🙂.

text/0249-v2-experiments.md
skinny85: And one more aspect that this proposal has to include: what happens with all of the Construct Library modules that are `Experimental` or `DevPreview` in version `1.x` today?

eladb: I am not sure we need to discuss each api in this RFC, but the general approach for moving from v1 to v2 should be included.

I think that once this decision is made, we need to create concrete plan for implementing it as part of v2. We will need to make an api by api decision for what to do.

text/0249-v2-experiments.md
NetaNir: I can change the structure - sure. If it's just cosmetics, I prefer we do it after we agree on the best path forward. If it's something else, and you think this option needs more discussion,  I can edit it.

skinny85: I just want to make 100% sure we discard all 3 vending variants of option 5. (see https://github.com/aws/aws-cdk-rfcs/pull/279#discussion_r549891589).

text/0249-v2-experiments.md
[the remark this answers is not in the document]
NetaNir: Removed, I left it in the advantages section for option 6.

text/0249-v2-experiments.md
```diff
@@ -67,7 +67,13 @@ levels of abstraction. Changing to vending the Construct Library as a monolithic
 package is one part of making that possible; we should make sure our approach to
 unstable code also takes this into account.
 
-## 3. The CDK team can still perform API experiments
+## 3. We want to encourage customers to not declare a fixed version on packages we own, and are intended to be widely used.
+
+Fixed versions are a bad practice, they prevent users from getting updates that
+includes critical fixes, security patches and more. The mechanism we choose to
+implement experimental features should adhere to the same standard.
```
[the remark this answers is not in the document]
NetaNir: I will add it in the discussion section of option 6

text/0249-v2-experiments.md
NetaNir: > which package does the non-stable code live in

There is no more "unstable" code, all APIs are stable and they live in `aws-cdk-lib`. It is described in the release notes and in the advantages. Where else should it be mentioned?

NetaNir: > Also, notice that the various options are in a chapter titled 2. Separate "breakable" code from mono-CDK main modules. If option 6 does not allow shipping code that we can break, then it definitely does not belong in that chapter.

I can move it under the chapter titled 1 "1. No more breaking changes in the main mono-CDK modules".

text/0249-v2-experiments.md
rix0rrr: I know they do it this way in Kubernetes, but we should have a plan for how this makes sense for us before simply copying it.

I personally would hate the visual noise these add and would prefer something lighter, like an `X`, `X2`, `X3`, ... at the end.

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+Users can work around this by executing `npm install --force` but this is not a
+behavior we want to recommend.
+
+### Option 6: API Previews
+
+AWS CDK v2 release notes:
+
+> Starting with version 2.0.0 of the AWS CDK, all modules and members will
+> become stable. This means that from this release, we are committed to never
+> introduce breaking changes in a non-major bump.
```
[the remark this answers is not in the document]
rix0rrr: > the library will no longer include experimental APIs. 

I don't think this is true, right? If we have an `newStackNamingv1alpha1` property, that still counts as experimental

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,266 @@ Disadvantages:
...
+> other most common feedback we hear from customers - breaking changes are not
+> ok.
+>
+> #### Introducing API Previews
+>
+> To make sure we can keep adding features fast, while keeping our commitment to
+> not release breaking changes, we are introducing a new model - API Previews.
+> APIs that we want to get in front of developers early, and are not yet
+> finalized, will be added to the AWS CDK with a specific suffix: `PreX`. APIs
+> with the preview suffix will never be removed, instead they will be deprecated
```
rix0rrr, line 476: Maybe additionally stress here: "so it's safe to use them".

text/0249-v2-experiments.md
```diff
@@ -374,6 +380,257 @@ Disadvantages:
...
+   However, most IDEs will list methods that contains "grant" in any part will
+   be listed by autocomplete when a user types "grant".
+
+#### Alternatives:
+
+Using `v1` for visibility:
+
+- RcV1: for release candidate.
+- AlphaV1
+
```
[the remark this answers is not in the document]
rix0rrr: >  if a module is marked as a preview

What does that mean? There's a module with the `_pre1` suffix? Which does NOT have the CFN resources? Or does it? Which means different namespaces in Java/C#?

text/0249-v2-experiments.md
eladb: I like `X1`, `X2`, ...

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,266 @@ Disadvantages:
...
+>
+> /**
+> * This methods grants awesome powers
+> * @deprecated use grantAwesomePowerPre2
+> */
+> grantAwesomePowerPre1()
+> ```
+>
+> When we decide its time to graduate the API, the latest preview version will
+> be deprecated and the final version - `grantAwesomePower` will be added.
```
njlynch, line 509: Do we want to define a standard convention when we realize `grantAwesomePower` has a flaw and needs to be deprecated as well? `grantAwesomePowerPost1`? `grantAwesomePower2`?

rix0rrr: I don't care about the name (and suffixes are ugly), but I would like a tool-supported way to make sure that for the `vNEXT` we can re-use the "prime real estate" name for the *correct* function.

So even though in `v2`:

* `grantAwesomePower() -> @deprecated`
* `grantAwesomePowerBetter()  -> recommended`

I'd like in `v3`:

* `grantAwesomePower() ->` has the implementation of `grantAwesomePowerBetter()` from the previous version

Although maybe standardized suffixes and some compilation strategy in `jsii` are the easiest way to achieve that...

text/0249-v2-experiments.md
NetaNir: We will discuss the strategy for migrating dev preview/experimental API from v1 to v2 in a separate doc, after we agree on the mechanism for adding non final APIs.

text/0249-v2-experiments.md
NetaNir: To avoid adding a scary diff Im going to leave the section where it is to ease the review process, I will move it once we finalize the RFC

skinny85 (dismissed):
The document itself looks fine to me (with the provisions of discussing the details of Option 6 in a follow-up doc/PR).

As for the decision itself, the disadvantages of Option 6 are pretty scary to me, compared to 2 and 3, and considering this is something completely new we're doing here (and by that, I don't mean the method names, which are the easy part; I'm talking about the namespace/class/property names, which are meant to be discussed in that separate doc), I wonder what disadvantages we missed... (Simple example off the top of my head: will this make us more conservative in accepting community PRs, because everything is now stable, and the number of contributions to the CDK will diminish because of that....?)

In my opinion, Option 6 is much riskier, compared to either 2 or 3. I also hate the fact that the current AppMesh module, just to take a random example, will now have to be maintained for all of the life of CDK major version 2, even though we _know_ the code inside of it models the AppMesh service incorrectly, and we want to change it in a backwards-incompatible way. Same for EFS, and many other modules.

However, it seems like others don't share my worries, and the decision has already been made. If that's the case, I don't see any reason to waste time word-smithing this doc further (all of the difficult things will come out when we start implementing this anyway).

text/0249-v2-experiments.md
skinny85: > @skinny85 I was under the impression we are going to flip all v1 modules/APIs to stable in v2 in order to minimize the migration pain for customers, is that not the case?

No, that is not the case. If we adopted a different option than 6, all of the modules that were not "Stable" or "CFN-only" at the time of releasing a GA version of `aws-cdk-lib` version `2.x` would be placed into that solution's designated spot for code that is allowed to be broken (for example, that would be `aws-cdk-lib-experiments` if we chose option 2).

NetaNir:
> As for the decision itself, the disadvantages of Option 6 are pretty scary to me, compared to 2 and 3,

@skinny85 

This is exactly what this PR is for. Can you add your concerns as comment under "disadvantages" for option 6? 

As for options 2 and 3, they are more scary for me due to the following:

1. Encourage users to declare a fix dependency on a module/modules we own, this is a liability for us. This is true for using  `0.x` versioning and prerelease qualifiers. 
2. They don't solve the problem of "non final" APIs in stable modules.
3.  Potentially dependency hell.

NetaNir:
@skinny85 I see your concern about existing experimental modules migration to v2, I will add a section that address this in the doc. I think we can have a nice compromise.

rix0rrr:
Adam, I agree with you that it would be "better" FOR US to have those janky modules not stinking up our code base, but is it also better FOR USERS?

I have the feeling we're fundamentally optimizing for two different things:

* You seem to be optimizing for code base health
* The current proposal is optimizing for user convenience ("batteries included"), and trying to support an ecosystem where libraries that depend on different aspects of the CDK libraries in different stages of maturity can interact in the same program.

Both are valid things to optimize for. We just have to decide which one is worth more.

I think the proposal is written with the "user is more important" angle in mind, and I think I agree with that assessment.

skinny85:
It's not obvious to me that having a proliferation of modules like (this is just for illustrative purposes): `aws-cdk-lib/aws-appmesh-aplha-1`, `aws-cdk-lib/aws-appmesh-alpha-2`, ..., `aws-cdk-lib/aws-appmesh-beta-1`, `aws-cdk-lib/aws-appmesh-beta-2`, ..., and finally `aws-cdk-lib/aws-appmesh`, is actually better for users (will people realize which one to use? Will they know which one to contribute their feature/bugfix to? How will this affect the size of the `aws-cdk-lib` package when we're lugging around all of this junk? etc.).

MrArnoldPalmer (commented):
> The current proposal is optimizing for user convenience ("one library has everything and it won't break"), and trying to support an ecosystem where libraries that depend on different aspects of the CDK libraries in different stages of maturity can interact in the same program.

My disagreement with this statement stems from thinking about the ecosystem over the long term. As users start writing and distributing more construct libraries, having aws-cdk-lib be "one library that has everything" doesn't seem like an advantage, it seems like an unnecessary special case.

When thinking about an app that relies on a number of construct libraries, that may rely on other construct libs that then rely on aws-cdk-lib, having to know about this versioned naming in the case of aws-cdk-lib, but instead know about "^x.x.x" range resolution in dependencies' package.json doesn't feel easy. I'm not sure if this will actually be an issue because its hard to know how prevalent usage of these versioned "non-final" apis will be used by app authors vs libraries.

I agree separate libraries may be optimizing for "code base health" but I also see it as optimizing for "ecosystem" health. Experimenting with new constructs (or anything that technically CAN be decoupled into a separate library) being done in a separate repo/package/etc feels like it should be the default because that IS the default for non-core team members.

NetaNir:
> I agree separate libraries may be optimizing for "code base health" but I also see it as optimizing for "ecosystem" health.

@MrArnoldPalmer If we do release separate modules, will they be released as experimental (using semver prerelease qualifier)? If so, how will libraries express range of supported versions? Libraries will have to declare a fixed dependency on experimental modules, forcing their consumers to the same version, which will have a negative impact on  the ecosystem health.

MrArnoldPalmer:
I suppose libraries would start as 0.1.0 and could only bump the "minor" version when they actually break so users could express a range. If libraries are relying on a v0 construct and presenting a stable version to the user, they likely would want to take a fixed version either way to avoid breakage.

I think multiple third party libraries relying on different versions of a v0 construct (or the user wanting to take a non-compatible version) would hopefully be rare, but I have no data to back this. Hopefully we could mitigate this by limiting the number of v0 packages, keeping them very small and targeted, and maturing them relatively quickly.

This problem exists when doing construct library development independent of aws-cdk-lib though. Generally I think the community recommendation is "libraries that depend on v0 packages should be prepared to update/break". But yeah this is not ideal and TS types complicate this further than just JS node package concerns.

text/0249-v2-experiments.md
```diff
@@ -638,8 +638,71 @@ use a naming scheme that convey its non-final nature for the following reasons:
...
+A means it will be added to `aws-cdk-lib`, since module A depends on module B,
+it means that adding module A to `aws-cdk-lib` breaks condition 1 and creates a
+cyclic dependency. To prevent the cycle we will have to graduate B as well,
+which is not necessarily what we want. Additionally, if module B is an alpha
+module, module A will have to declare a fixed dependency on it, locking its
+consumers to the same version of module B - defeating the purpose of
+peerDependencies.
+
+> \*\* _There might be exceptions that will be considered on a per case basis_
+
```
eladb, line 670: So if an external contributor wants to contribute L2 for a module that we only have CFN classes for, what is their process? It seems like they won't be able to just submit a PR against the CDK repo. Someone will need to create a repo for them, etc. Worth mentioning at least, and acknowledging that this is not as easy as today.

text/0249-v2-experiments.md
```diff
@@ -638,8 +638,71 @@ use a naming scheme that convey its non-final nature for the following reasons:
...
+This stage is intended for modules that are in early development stage, they
+change rapidly and may incurs many breaking changes. These modules will be
+released separately from `aws-cdk-lib` and will have their own version, which
+allows for breaking changes in accordance to semver, e.g `0.x` or prerelease
+qualifiers.
+
+In order to release a module as alpha, the following conditions must be met:
+
+1. Module from `aws-cdk-lib` **can not** depend on it.
```
eladb, line 655: It's worth discussing the "integrations use case" (such as lambda-event-sources, sns-subscriptions). If those modules wish to include an integration with resource that is still in alpha, we will need to do something like only add the minimal bear interface to the stable module (i.e. `appmesh.IFoo` with only `fooArn` for example).

Worth mentioning this pattern because I think we will need to use it quite a lot in our transition from V1 to V2.

eladb (commented):
LGTM

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,329 @@ Disadvantages:
...
+clear motivation to upgrade to the new version of the API - they must do so in
+order to upgrade to a newer version of the AWS CDK. If the API is only
+deprecated, users have no motivation to upgrade to a newer version of the API,
+even worse, they might not be aware that a newer version exists. One way to
+encourage users to upgrade to a newer version of an API, is to supply tools that
+will inform users. For example, we can add a capability to the `cdk doctor`
+command that will notify users that their CDK application is using deprecated
+experimental APIs.
+
+Executing `cdk doctor` will print:
```
nija-at, line 615: Good idea but probably not `cdk doctor`. This is run only when there's a problem, but often not even then.
Maybe as part of `synth` or `deploy` itself.

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,329 @@ Disadvantages:
...
+> be deprecated and the final version - `grantAwesomePower` will be added.
+
+### Documentation
+
+How will different versions of an API be reflected in the AWS CDK
+documentations? One option is to only show the latest version of the API, and
+remove all deprecated versions. While this might reduce potential clutter in the
+docs, it might be frustrating for users who encounter an older version of an
+experimental API. A compromise can be to collapse all older version of an API,
+making them still accessible but somewhat hidden:
```
nija-at, line 518: or moving all preview APIs into a separate section?

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,329 @@ Disadvantages:
...
+   community to supply feedback.
+3. **Setting the right expectations**: When an API in preview users are aware
+   that this is not the final version of the API, and its deprecation is
+   expected.
+
+### Modules lifecycle
+
+This section discuss adding a new modules to `aws-cdk-lib`.
+
+#### Alpha (optional)
```
skinny85, line 645: How about we use the terminology we have in place today? So your "Alpha" is "Experimental", and "Beta" is "Developer Preview"?

text/0249-v2-experiments.md
```diff
@@ -373,6 +379,329 @@ Disadvantages:
...
+
+This stage is intended for modules that are in early development stage, they
+change rapidly and may incurs many breaking changes. These modules will be
+released separately from `aws-cdk-lib` and will have their own version, which
+allows for breaking changes in accordance to semver, e.g `0.x` or prerelease
+qualifiers.
```
skinny85, line 651: Should we note here that the decision on how to release those modules is a decision between options 2 and 3?

Seems like this paragraph is suggesting we just chose option 3, but it's not obvious to me why.

Maybe we can defer that decision to later? At the time of reviewing the doc that contains all of the items marked in this PR as ToDo?

skinny85 (commented):
LGTM. My only comments are:
* Let's use the wording we already have ("Experimental" and "Developer Preview").
* Let's not decide between options 2 and 3 for the unstable modules in this PR (there are too many items marked "ToDo" here to make that decision with enough confidence, at least for me).

text/0249-v2-experiments.md
NetaNir: We need to decide, I'll add a section detailing why I think option 6 is better

NetaNir: Add a section which discuss this, I have a strong preference to releasing each module separately. See [multiple-alpha-modules-vs-single-module](https://github.com/aws/aws-cdk-rfcs/blob/neta/option-6/text/0249-v2-experiments.md#multiple-alpha-modules-vs-single-module)

text/0249-v2-experiments.md
```diff
@@ -373,6 +367,358 @@ Disadvantages:
...
+
+> \*\* _The deprecation process will be discussed in the API previews
+> specification_
+
+#### Stable
+
+In this stage modules are added to `aws-cdk-lib` under their final name, e.g
+`aws-batch`. Addition of new APIs should follow the API previews specification.
+
+### Multiple alpha modules VS. single module
```
skinny85, line 666: Here is my on position in the "one unstable module vs many unstable modules" debate.

I am in favor of having a single module. I have just one reason for that position: it's simply less risky. We are taking on a huge amount of risk already by "betting the farm" on Option 6 that does something we've never done in the CDK before. Therefore, whichever option we can choose here that is less risky, I'm in favor of.

Why am I saying that Option 2 is less risky? Because it follows the same scheme as MonoCDK. With MonoCDK, we're basically saying that "L2s are special", and we need to release them together. If that is the case, I don't see why unstable L2s should be treated any differently than stable L2s.

You say that Option 3 has no downsides compared to Option 2, because unstable packages can easily depend on each other using `peerDependencies`. Well, not everyone uses `npm 7`. In fact, `npm 7` today ships only with unstable versions of Node (15). The vast, vast majority of our users are on `npm 6`. Which means they will have to manually fulfill the peer dependencies of unstable modules - a terrible experience.

But let's assume everyone uses `npm 7`. In my opinion, there are still potential pitfalls in this solution. Let's consider the following scenario. We have 2 unstable modules, `@aws-cdk-lib-experiments/aws-A` and `@aws-cdk-lib-experiments/aws-B`, and they peer depend on `0.10.0` of another unstable package, `@aws-cdk-lib-experiments/aws-X`. Our customer (who writes a CDK application, not a library) uses `A` and `B` in their `package.json`, both in version `0.10.0`.

Some time passes, and more versions are released. Our customer upgrades `A` to version `0.20.0`, because they need some new features of it. That newer version of `A` depends on version `0.20.0` of `X`, which had some breaking changes made to it compared to version `0.10.0` of `X`.

What happens when a customer upgrades `A` in their `package.json` to `0.20.0`, while leaving `B` unchanged, on `0.10.0`? Will they get an error during `npm install`? Will they have two versions of `X` in their dependency closure now, risking getting the dreaded "Construct is not an instance of Construct" error? Will they get a runtime error from `B`, because version `0.10.0` of `B` does not work with version `0.20.0` of `X` (because of the breaking changes in `X`)? Will they be forced to upgrade `B` to `0.20.0` too, negating your stated benefit of this approach? What if there's no `0.20.0` of `B`, because it's released from a completely different repository, on a different schedule, than the other modules?

This is just one potential scenario. I'm sure there are many other ones that we haven't even thought of. And that to me is the definition of risk: the fact that we have these blind spots that may surprise us after we make this decision.

We already have experience that many single modules cause dependency problems in the NPM ecosystem - even for stable modules, because of the "Construct is not an instance of Construct" TypeScript error. Why would we go back on all that experience from `v1`, and decide to keep releasing multiple modules in `v2`? It seems incongruent to me to insist on MonoCDK for stable modules, but separate packages for unstable ones. Either we want MonoCDK (and if  so, Option 2 is less risky), or we want Option 3, and then we don't need MonoCDK (because of `npm 7` and peer depedencies).

We are making a huge bet with v2 being shipped as MonoCDK. We are saying that we fixed the dependency problem for the CDK ecosystem with this change. It has some downsides, like the huge package size, but we're claiming that the downsides are worth it, because the dependency issues will be addressed once and for all. If we ship MonoCDK v2, but still have the dependency hell problem, just this time restricted to unstable modules, I think it will be a disastrous outcome.

> If we release all modules under a single package, in which every release may include breaking changes to any numbers of its modules, users are force to accept the breaking changes in all modules, even if they only want to upgrade a single module.

I agree that is a downside to Option 2. All of the options have downsides, none of them are perfect, including Option 3. _However_, I'm willing to swallow this known inconvenience of Option 2, if the alternative is to risk all of the unknown consequences of Option 3.

MrArnoldPalmer: Version mismatch when depending on unstable modules will happen with a single or multiple packages, its just a matter of "where". If a use relies on a third party package that relies on aws-cdk-experimental v0.1 and additionally needs aws-cdk-experimental v0.2 themselves for some newer feature, this presents a mismatch. I think part of my inclination for leaning towards separate packages is that the process of resolving these is the same in both. Any packages that rely on unstable construct libraries (third party or otherwise) will need to consistently update the versions of these dependencies in order for usage of the "experimental" ecosystem to be viable. Especially across third party construct libraries. Making aws-cdk-lib never break is what allows monocdk to circumvent this not the bundling itself.

Making everything peer dependencies all the way down also affects this because there only ever should be a single version in the resolved node_modules of any app consuming this chain of packages. Conflicts ultimately have to be resolved by the user if they exist. NPM 7 makes this easier by default but the end product is essentially the same, the user can declare the ultimate version of any peer dependencies.

skinny85: Libraries cannot depend on unstable modules - that is a problem that is simply impossible to solve. We have learned this from `v1`.

Note that my scenario happens for an **application** customer, not a **library** customer.

NetaNir: You are ignoring a few major differences:

1. **In v2 we will only be using peerDependencies:** The majority of problems in v1 were rooted in the fact that we declared peerDependencies **and** direct dependencies between modules. We did it because we had hundreds of modules, and adding them as peerDependencies would have forced consumers to install all of these modules themselves. **In v2 we will only use peerDependencies,** so even if a consumer will not use npm7, they will control the installed version. 
2. Due to the fact that experimental modules can introduce a breaking change in every release, libraries, including other experimental modules shouldn't depend on them.
3. From the 35 experimental modules we currently have - 24 (!) have no dependents and only stable dependencies. This means that there is no reason to treat them the same as third party libraries. If we can't make it work for us, we will never be able to grow the ecosystem.
 

Having a single package that includes all modules will also hurt the adoption of new experimental modules. If a new module is added in a specific version, which also includes breaking changes to other modules, customers might not being willing to upgrade if it means it will impact already deployed infrastructure. TThe fact that a breaking change in a cdk module may be a change to infrastructure that has already been deployed (or even downtime), means that customers are more hesitant to update their version. Having separate module mean we will reduce the blast radius of every update.

MrArnoldPalmer: >Note that my scenario happens for an application customer, not a library customer.

Totally. In your scenario with npm <7, the user will be prompted with an error saying that one or the other of their dependencies is missing a peer dependency. If both A and B libraries happen to be compatible with the version of X that is installed, this could be ignored and the app will work. If not, they will either need to explicitly install a version of X that they know to be compatible or also update the version of B as you suggested. This, as you mentioned, negates the positive effect of only needing to upgrade specific experimental packages when needed. Maybe we can do some analysis to get an idea of how often we could expect this to occur

I guess I'm unsatisfied with "Libraries cannot depend on unstable modules" because I believe they undoubtedly will and should be able to provided they are willing to take on that maintenance cost of keeping dependencies up to date. With that in mind separate experimental modules feel a lot like any other library that will rely on aws-cdk-lib during the 0.x phase of development. Some things you depend on may be stable, others may have unstable features that are of high value that you decide to use but with the known cost that 0.x comes with.

text/0249-v2-experiments.md
MrArnoldPalmer: We may want to consider third party ownership here. Allowing contributors to make and maintain their own repositories, provided we give them the templates/tools that make it appropriate for eventual inclusion in aws-cdk-lib, may be beneficial. This would standardize a path for contributing wholly new functionality that can be leveraged by users without the core team getting in the way. If something is ultimately deemed to not be relevant for inclusion in aws-cdk-lib it can just continue to be released as a separate module.

This has the downside of not being published under the aws-cdk namespaces in package managers and discoverability may be an issue. Additionally even with tools to make this as easy as possible contributing to an existing repository will always be easier.

text/0249-v2-experiments.md
rix0rrr: Some of these points have already been addressed in comments by Mitch and Neta, but I'm going to add my own view on them.

## Dependencies between experimental packages, what if a user upgrades one but forgets the other?

First of all, I have a suspicion this will be **rare**¹, so I'm sure it's worth optimizing for. But even then:

> Will they have two versions of X in their dependency closure now,

No! I get you've been traumatized by this, but again:

**The problem of v1 was not modularity, the problem of v2 was using `dependencies` where we should have been using `peerDependencies`**

`dependencies` will lead to multiple copies of a library, `peerDependencies` will not. You will get an installation error if you match incorrectly.

If we had used `peerDependencies` in v1, we would NOT have been in **technical** problems. The reason we're choosing `monocdk` over modular packaging is a **usability** issue (user needing to specify all transitive deps), which has been solved in NPM7. Had NPM7 come out earlier, we probably wouldn't have done this at all (or maybe we would have, but we definitely would have had fewer complaints).

But NPM moves at the pace it does, and c'est la vie.

But conclusion:

**Modular packaging is not a technical risk, it's only a usability issue in case of [many] transitive dependencies for NPM <7.**

> *¹) I'm not saying they won't happen, I'm saying they will be rare. A single counterexample does not invalidate this argument.*

## Backwards compatibility and ecosystem

> We also know from v1 experience that you cannot build an ecosystem of packages if those packages do not maintain backwards compatibility.

It is true, but this is completely independent from packaging. In fact, I would put it like this:

**If we decide to break APIs between releases, then monopackaging is worse than granular packaging**

That's because you don't get to pick and choose versions anymore.

Let's introduce some notation so that we can be precise about this.We have APIs `A` and `B` and they have backwards incompatible revisions (`A1`, `A2`, `A3`, ...). Furthermore, let's say that if we leave old copies of the API in place in new releases of a library (but `@deprecated`), we're going to write that as `(A1) + A2` (in this case, `A1` is deprecated but still in the library, and `A2` is also in there²). 

Let's say these are the releases of `cdk-experiments-change`:

| `cdk-experiments-change@0.1.0` | `cdk-experiments-change@0.2.0` | `cdk-experiments-change@0.3.0` |
|--------|-----------|-----------|
| `A1 + B1` | `A2 + B1` | `A2 + B2` |

First we change `A`, then we change `B`.

Now, assume that we have an application or library, doesn't really matter which, that's happily consuming `A1`. They are on library version `0.1.0`, and all is swell.

Now `B2` just got a feature that they want to use while staying on `A1` for the `A` API but... that's impossible! There's no package version that supplies `A1 + B2`.

And this is *potentially* fine, you can just update your code to migrate from `A1` -> `A2` instead as well, right? No big deal.

It becomes more hairy once we start mixing in libraries that you may not have control over. Different scenario:

Maybe you are an application that uses a library that hard-locked you to `0.1.0` because it uses `B1` (and you can't know whether `0.2.0` will still have `B1` or not, so you have to minor-version-lock). But now the app wants to use `A2`. We're in the scenario that there's there *is* a version that supplies `A2 + B1`, and yet you're **still not able to take it** because your library dependency locked you into a version of the CDK lib.

```
app —[ uses L1 ]⟶ lib@1.0.0 —[ uses B1 ]⟶ cdk-experiments-change@0.1.0   
app —[ uses B2 ]⟶ cdk-experiments-change@0.2.0                    **** CONFLICT ****
```

Bottom line:

**Building ecosystems around in-place broken APIs get *worse* if we tie those APIs together in releases.**

The only way around this is to do one of two things:

### Don't in-place break APIs

| `cdk-experiments-add@0.1.0` | `cdk-experiments-add@0.2.0` | `cdk-experiments-add@0.3.0` |
|--------|-----------|-----------|
| `A1 + B1` | `(A1) + A2 + B1` | `(A1) + A2 + (B1) + B2` |

So now everyone can take upwards unbounded (caret) dependency ranges and everyone can move forward. This is option 6 and what we're doing for the **stable** core lib.

### Don't tie breaking changes together together

| `a-experiment-change@0.1.0` | `a-experiment-change@0.2.0` |
|--------|-----------|-----------|
| `A1` | `A2` |

| `b-experiment-change@0.1.0` | `b-experiment-change@0.2.0` |
|--------|-----------|-----------|
| `B1` | `B2` |

You still cannot  take range dependencies, but at least now you can mix and match dependencies on `a-experiment-change` and `b-experiment-change` at will without them stepping on each other.

> *² Let's disregard naming for a moment, but writing `(A1) + A2` will obviously have to mean that we'll have to have renamed some members in order to let those incompatible APIs exist in the same library side-by-side. The methods/classes of `A2` will have to have new names.*

## Unknown risk

> the risk all of the unknown consequences of Option 3.

See... this might be the crux of the issue. I don't see any unknown consequences that we haven't already encountered here, and I also don't see any avenue for "unknown unknowns" to crop up.

We've been vending granular libraries for ages now, and we know exactly what's up. There's three big issues with them:

* Multiple copies of a library ⟶ fixed by using `peerDependencies`
* Many dependencies is a chore for users ⟶ fixed by using NPM7 or not having many dependencies (likely!)
* Adding peerDependencies is a breaking change ⟶ fixed by using NPM7 or bumping your own MV (we do that anyway!)

They all seem perfectly well addressed, or not a big deal, and definitely only impacting a minority of users (so honestly I feel these are issue are not really worth all the air time we're giving them).

Between the two solutions: monoexperiments and granular experiments, the granular experiments option seems the "technically correct" alternative of the two. It also has the advantage that it will be easier for 3rd parties to contribute modules, which we can first incubate outside our repo, then inside our repo, then graduate to beta or higher.

rix0rrr: > The reason we're choosing monocdk over modular packaging is a usability issue

A logical follow-up question at this point would be: if mono-packaging has better usability, why don't we do the same for mono-experiments then?

We could. But in order to get the touted benefits of mono-packaging, we must have **additive** (non-breaking) APIs.

The whole point of having our experiments separately is so that we wouldn't have to do that and we could freely break APIs. If we want to break APIs, then mono-packaging actually produces (technical) downsides.

text/0249-v2-experiments.md
```diff
@@ -373,6 +367,363 @@ Disadvantages:
...
+allowed to introduce any breaking changes to their API. Any non backward
+compatible change will be introduced via deprecation\*\*.
+
+> \*\* _The deprecation process will be discussed in the API previews
+> specification_
+
+#### Stable
+
+In this stage modules are added to `aws-cdk-lib` under their final name, e.g
+`aws-batch`. Addition of new APIs should follow the API previews specification.
```
nija-at, line 664: I don't see a difference between beta and stable. API breakages are not allowed in either case. can we just drop beta?

text/0249-v2-experiments.md
```diff
@@ -373,6 +367,363 @@ Disadvantages:
...
+> change, we will add `grantAwesomePowerPre2()` and deprecate
+> `grantAwesomePowerPre1`:
+>
+> ```ts
+> /**
+> * This methods grants awesome powers to the given principal
+> *
+> * @param grantee The principal to grant powers to
+> */
+> grantAwesomePowerPre2(grantee: iam.IGrantable)
```
nija-at, line 487: I would like to see how a (new) experimental construct in a stable library would look like in each of our major languages. Could we add something like this as an appendix item?

Let's say for instance, that a new experimental `Function` construct is being introduced into a stable construct library `aws-lambda` (hypothetically). Would the API look something like this?

```ts
const handler = new FunctionPre1({
  handlerPre1: 'index.handler',
  runtimePre3: RuntimePre2.NODEJS_12_X_PRE2,
  code: CodePre1.fromAssetPre5(...),
  ...
});
```

Similarly, how would this be in Java?

text/0249-v2-experiments.md
skinny85: So, to sum up the very long discussion (I'm guilty if this too):

Options 2 and 3 are very similar. All of the options involve some tradeoff, none of them are perfect. The tradeoff between 2 and 3 is the following:
* Option 2 allows the unstable modules to depend on each other (because it bundles them together), which option 3 does not allow.
* _However_, that bundling has the side-effect that all unstable modules are versioned together. So, if a customer wants to use 2 unstable modules, `A` and `B`, they always have to share a version, they can't use `A` in `0.10`, and `B` in `0.20`. This is the advantage that Option 3 has.

Now, it's a question of "which of the 2: unstable-to-unstable dependencies vs. separately-versioned unstables is more important?". I think reasonable people can disagree here. Neta and Rico (and probably Mitch) consider that "separate unstable versions" are more important. I consider "unstable-to-unstable" dependencies more important.

My particular reason (I don't want to speak for anyone else) for preferring "unstable-to-unstable" dependencies is that we know we need them with the current modules, because we have "unstable-to-unstable" dependencies _right now_. Which also means we will need this in the future, as AWS releases more services and more integrations between them.

As for the "separate versions of unstable", this is the situation today with `v1` and MonoCDK. And I've never heard customers saying that this particular limitation is causing them problems. So for me, it's "actual benefit we know we need right now" vs "possible downside that has not been reportd". Hence why I prefer Option 2.

But again, I think reasonable and smart people can disagree on this 🙂.

Here's the current snapshot of "unstable-to-unstable" dependencies (note that I removed a bunch of them relating to the fact that `cxapi`, `region-info`, `s3-assets`, etc. are still unstable - I assume those will become stable with V2). I'll let everyone be the judge if this is rare enough that it negates the downside of Option 3.

```
ℹ️️  Unstable package '@aws-cdk/aws-apigatewayv2-integrations' depends on unstable package '@aws-cdk/aws-apigatewayv2'
ℹ️️  Unstable package '@aws-cdk/aws-appmesh' depends on unstable package '@aws-cdk/aws-acmpca'
ℹ️️  Unstable package '@aws-cdk/aws-backup' depends on unstable package '@aws-cdk/aws-efs'
ℹ️️  Unstable package '@aws-cdk/aws-docdb' depends on unstable package '@aws-cdk/aws-efs'
ℹ️️  Unstable package '@aws-cdk/aws-eks' depends on unstable package '@aws-cdk/lambda-layer-awscli'
ℹ️️  Unstable package '@aws-cdk/aws-eks' depends on unstable package '@aws-cdk/lambda-layer-kubectl'
ℹ️️  Unstable package '@aws-cdk/aws-s3-deployment' depends on unstable package '@aws-cdk/lambda-layer-awscli'
ℹ️️  Unstable package '@aws-cdk/aws-ses-actions' depends on unstable package '@aws-cdk/aws-ses'
```

NetaNir: We have solid arguments arguments for each option, but its getting really hard to follow the conversation in this PR.

I'm going to summarize the major pro and cons of every approach **when compared to each other** in a short doc, we will review it on Monday and make a decision.

NetaNir: @skinny85 option 3 does allow unstable module to depend on other unstable modules (with the same graduation limitation as option 2), see Rico's response to your  "Will they have two versions of X in their dependency closure now" comment.

text/0249-v2-experiments.md
NetaNir: > Someone will need to create a repo for them, etc. Worth mentioning at least, and acknowledging that this is not as easy as today.

We will have a single monorepo with all experimental modules versioned independently

text/0249-v2-experiments.md
NetaNir: Im going to add all this in the specification doc, it will be easier to review it outside of this RFC.

text/0249-v2-experiments.md
NetaNir: Stable modules will have a stable name e.g  `aws-s3` , non final APIs within these modules will be add with the preview suffix, e.g `grantReadPre1`. Beta modules will have a suffix in their name, e.g `aws-s3-beta-1` but most** of their APIs will not have preview names.

The stable stage is intended for well baked modules with limited number of preview APIs. The beta stage is mainly intended for adding new modules. New modules are likely to change a lot, and as result will have many deprecated APIs.  Creating the module under a preview name allows us to deprecate the entire module when we are happy with the result, starting with a clean slate. This will effect the in IDE autocomplete experience, how the documentation looks and more. 

The ** explained:
We still need to finalize how will individual APIs be added inside the beta module. If we add them with a non preview name as suggested above, it will make the migration from beta to stable easy for customers - less code change. 

Adding an API to the  `aws-s3-beta-1` module:

```
grantWrite()
```

When we graduate the module, the name of the method will remain `grantWrite`, for modules with a large API surface area, this will have a substantial effect on the migration in terms of required code change. However, this assume only one breaking change for each API - the switch between beta to stable, which might not be true. It could be that during the beta phase of a module we will need to make a non backward compatible change to one of its API. Im not yet sure what's the right answer here, my current tendency is to assume that the common use case for beta development will be satisfied by a single non compatible change per API, and in cases where it is not, we can change the API name, e.g add `V2`. 

It will also make it easier to create a new L2, not having to add the preview suffix to all properties, class and methods.  But as I said, Im still thinking about it, feedback is welcome.

NetaNir: @nija-at

MrArnoldPalmer (approved):

text/0249-v2-experiments.md
nija-at: > The beta stage is mainly intended for adding new modules. New modules are likely to change a lot, and as result will have many deprecated APIs.

Wouldn't it be simpler if we just have all new modules start off as individual packages in the `aws-cdk-previews` (or whatever name we decide) repo?

It's not clear when one decides to make a beta package vs a package in `aws-cdk-previews` repo. Maybe I missed it?
It would be simpler to just have one.

text/0249-v2-experiments.md
nija-at: Sounds good.

eladb (approved):

text/0249-v2-experiments.md
NetaNir: It will not be possible for all new modules, for example, modules with stable dependents must be added to `aws-cdk-lib`. 
The same is true for v1 experimental modules that must be add to `aws-cdk-lib` for the same reason, here is a list of all experimental modules that have stable dependent: 

```
@aws-cdk/aws-s3-assets		
@aws-cdk/aws-ecr-assets		
@aws-cdk/cx-api	
@aws-cdk/region-info		
@aws-cdk/aws-efs	
@aws-cdk/aws-batch	
@aws-cdk/aws-glue	
@aws-cdk/aws-ses	
@aws-cdk/aws-autoscaling-common
```

Previous analysis of these modules done by the construct library squad determined that graduating some of these modules will require a lot of work. The beta stage allows us to add these modules to `aws-cdk-lib` without having to graduate them.

nija-at: Let's re-visit these and see if we can make these stable. 
Some of these have been around for a lonnngg time (eg. s3-assets, ecr-assets, region-info, autoscaling-common) and I don't see why we shouldn't just call them 'stable'. Others like efs, glue and ses will need to be looked at.

Another option is to make these a one time 'beta' package, and never any other package. I'd rather not have yet another option available unless there's a really good motivation.

eladb: Agreed. I think many of these can easily be stabilized without much additional effort given they've been baked for a very long time.

skinny85: I think these:

```
@aws-cdk/aws-s3-assets		
@aws-cdk/aws-ecr-assets		
@aws-cdk/cx-api	
@aws-cdk/region-info	
@aws-cdk/aws-autoscaling-common
```

are slam-dunks, and should be stabilized with v2.

These:

```
@aws-cdk/aws-efs	
@aws-cdk/aws-batch	
@aws-cdk/aws-glue	
@aws-cdk/aws-ses
```

are more tricky, and will probably require a separate decision for each one.

NetaNir: There are two separate discussions here:

1. Do we need the beta phase in general.
2. How should v1 experimental modules should be introduce in v2.

Lets keep the discussion about 2 to our planned API review meeting. For 1, happy to hear feedback and refine the RFC

nija-at: I don't think we need a general beta phase. I've not heard a good argument for having such a mechanism for future modules.

As for the existing modules, if the dependent modules don't expose any of the interfaces from their experimental dependencies, we can safely remove the dependency and use the L1 (or escape hatch).
Another options to consider could be to deprecate these APIs (in favour of APIs that don't expose the underlying experimental module) and remove them for v2.

We can make a one-time exception for a couple of these packages if none of these options are simple.

NetaNir: Good arguments for having a beta phase:

1. New modules that have stable dependencies and thus must be in `aws-cdk-lib`

skinny85: > Good arguments for having a beta phase:
> 
> 1. New modules that have stable dependencies and thus must be in `aws-cdk-lib`

Did you perhaps mean "dependents", not "dependencies"? Because stable dependencies are not a problem I don't think (all of the experimental modules will have a peer dependency on `aws-cdk-lib`, and so have access to all of the stable dependencies).

NetaNir: I did - edited

eladb: Did I miss something? I believe "beta" is equivalent to our current "developer preview" (or "release candidate"). My understanding is that modules start as "alpha" outside of aws-cdk-lib and when they are ready, they are added to the main library as "beta" to bake. Once finalized, the "beta" marker is removed and we call them stable. No?

nija-at: Per my discussion with @NetaNir on Friday, that's not the case. Modules can start either as alpha in the alpha repo or as beta in aws-cdk-lib. I'll let her expand on this.

@eladb - when they move to 'beta' within aws-cdk-lib, would the APIs be suffixed with the preview marker or stay the same? how would we change the APIs if we want to change something during the bake period? It's not clear to how we'll manage this 'beta' stage.

eladb: beta1 beta2...

Issue #287, opened 2021-02-17
Deprecated API Warnings


skinny85:
## Description

If an element in the CDK Construct Library (class, interface, property, function, etc.) is deprecated, using it inside a CDK application should produce a warning in the CDK CLI for all commands that perform synthesis (`cdk synth`, `cdk diff`, `cdk deploy`, etc.).

This will aid in the migration from CDK `V1` to `V2`.

## Roles

| Role                | User
|---------------------|------------------------------
| Author(s)           | @skinny85 
| API Bar Raiser      | @eladb 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [x] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

eladb:
@skinny85 has this been implemented already?

skinny85:
@eladb no, it has not - the design has been approved in the RFC, but the implementation work has not started.

PR #290, opened 2021-02-18
RFC 287: CLI deprecated warnings

skinny85:
* Tracking issue: #287
* [Rendered version](https://github.com/skinny85/aws-cdk-rfcs/blob/feat/cli-deprecation-warnings/text/287-cli-deprecation-warnings.md)

---

### DIFF
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -79,7 +79,7 @@ We are doing this to help customers have a smoother migration from `V1` to `V2` 
 
 ### Why should we _not_ do this?
 
-I see three reasons why we might not want to implement this feature:
+I see four reasons why we might not want to implement this feature:
 
 1. Deprecated elements are already highlighted by virtually any editor/IDE.
   In my experience, customers are diligent in moving away from deprecated APIs,
@@ -93,6 +93,14 @@ I see three reasons why we might not want to implement this feature:
   and from our own libraries using deprecated APIs.
   Which means a user might get warnings that they will be unable to get rid of
   (because they're coming from code that they don't control).
+4. Depending on the exact solution chosen
+  (see discussion below),
+  there might be edge cases where something is deprecated,
+  but we won't be able to report a particular pattern of its usage.
+  Some examples include:
+    * Using deprecated types only in type declarations.
+    * Static constants (which includes enums).
+    * Deprecated elements declared in the code, but not actually executed during synthesis.
 
 ### What changes are required to enable this change?
 
@@ -104,7 +112,7 @@ We will add a feature to JSII that adds the warning code to the emitted JavaScri
 It will be off by default, and will have to be activated explicitly to affect the emitted JavaScript code.
 We will also add an option to the JSII CLI that allows passing the name of the environment variable used for
 silencing the warnings, and turning the warnings into errors
-(the `CDK_DEPRECATED` variable mentioned above -
+(the `CDK_DEPRECATED` environment variable mentioned above -
 it can't be just `CDK_DEPRECATED` by default, as JSII does not have any knowledge of CDK!).
 
 #### 2. Using the changed JSII
@@ -114,9 +122,11 @@ we will need to use it in the CDK.
 We will have to start compiling CDK with the new option turned on,
 and also set the new JSII CLI option mentioned above to `CDK_DEPRECATED`.
 
-We should also modify our CDK test infrastructure to run with the `CDK_DEPRECATED_ERROR` option turned on by default,
-unless a test is explicitly checking a deprecated API as a regression test,
-in which case we should add an API for turning that option off for those tests.
+We should also modify our CDK test infrastructure to run with the `CDK_DEPRECATED`
+environment variable set to `error` by default,
+to prevent our own code from using deprecated APIs which would cause warnings to be shown to users.
+We should come up with a nice API that allows individual tests that are explicitly checking deprecated API(s)
+as regression tests to make the value of that environment variable `ignore` easily.
 
 ### Is this a breaking change?
 
@@ -127,7 +137,7 @@ No.
 1. Requires changes to JSII.
 2. Does not allow distinguishing between the customer using deprecated APIs,
   and the library code using deprecated APIs.
-  We can alleviate this problem by turning on the `CDK_DEPRECATED_ERROR` environment variable by default in our tests,
+  We can alleviate this problem by making the value of the `CDK_DEPRECATED` environment variable `error` by default in our tests,
   which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
 3. Adding emitting warnings for accessing deprecated static constants
   (including enums) will require huge changes to the emitted JavaScript
@@ -198,7 +208,7 @@ Disadvantages of this solution:
    and there is an issue about it on the TypeScript bug tracker,
    we cannot rely on them fixing this in a reasonable time frame.
 3. There is nothing out of the box for this in Python
-  (we would have to change JSII and write our own decorator
+  (we would have to change JSII and write our own decorator,
   which would have the same disadvantages as the TypeScript decorators solution).
 
 #### 4. Parse the customer's code
@@ -220,7 +230,7 @@ Disadvantages of this solution:
 
 1. We would have to write a separate parser and type-checker for each language supported by the CDK.
 2. This additional parsing could have an adverse impact on the performance of CDK commands.
-3. It's not obvious this analysis can even be performed at all in the case of dynamically-typed languages like JavaScript or Python
+3. It's not obvious this analysis can even be performed at all in the case of dynamically-typed languages like JavaScript or Python.
 
 ### What is the high level implementation plan?
 
@@ -232,7 +242,7 @@ The high-level implementation plan is:
   Effort estimate (development + code review): 2 weeks.
 
 2. Set the new option during building CDK,
-  make sure the tests use `CDK_DEPRECATED_ERROR`,
+  make sure the tests use `CDK_DEPRECATED` set to `error`,
   and provide an API for tests to opt-out of that on a case-by-case basis.
   Effort estimate (development + code review): 1 week.
 
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -155,11 +155,9 @@ No.
   and the library code using deprecated APIs.
   We can alleviate this problem by making the value of the `DEPRECATED` environment variable `error` by default in our tests,
   which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
-3. Adding emitting warnings for accessing deprecated static constants
-  (including enums) will require huge changes to the emitted JavaScript
-  (changing from a field to a `static` getter).
-4. We only report APIs actually hit during a particular execution of the code
-  (while missing statically declared but not actually invoked deprecated elements).
+3. We only report APIs actually hit during a particular execution of the code
+  (while missing statically declared but not actually invoked deprecated elements -
+  note though that IDEs/editors still warn about them).
 
 ### What alternative solutions did you consider?
 
@@ -216,8 +214,10 @@ Advantages of this solution:
 
 Disadvantages of this solution:
 
-1. Just deprecated warnings in the IDE can be missed by developers
-  (especially as more APIs get deprecated).
+1. TypeScript in VSCode (our most popular language/IDE combination)
+   does not by default show a complete list of deprecated API usages:
+   it will only show strikethroughs on identifiers you happen to be looking at.
+   Getting a complete list of deprecated APIs requires installing and configuring `eslint`.
 2. TypeScript currently does not properly detect deprecation of properties in object literals,
    and deprecation of properties (props) is very important for our project.
    Even though it will properly detect them while autocompleting,
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -1,5 +1,5 @@
 ---
-rfc pr: [#xxx](https://github.com/aws/aws-cdk-rfcs/pull/xxx) <-- fill this after you've already created the PR
+rfc pr: [#290](https://github.com/aws/aws-cdk-rfcs/pull/290)
 tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/287
 ---
 
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -30,14 +30,35 @@ you will receive a warning when performing any CLI operation that performs synth
   This API will be removed in the next major release
 ```
 
-There is also an environment variable `CDK_DEPRECATED` that you can use to change the behavior of this feature:
+There is also an environment variable `DEPRECATED` that you can use to change the behavior of this feature:
 
-* Setting that environment variable to the value `ignore` will silence these warnings
+* Setting that environment variable to the value `quiet` will silence these warnings
   (they will no longer be printed to the stderr stream).
 * Setting that environment variable to the value `error` will instead fail with an exception when any deprecated element is used.
 * Setting that environment variable to the value `warn` is the same as the default behavior
   (the warnings will be printed to the stderr stream,
   but will not cause errors).
+
+#### When exactly will this warning be shown?
+
+In general, this warning will be shown when executing any member
+(by which I mean function (static or standalone), constructor, method,
+or field (static or instance))
+from the AWS CDK Construct Library that is either:
+
+* Deprecated itself;
+* Belongs to a class or interface that is deprecated;
+* Has an argument that is of an interface type with a deprecated property,
+  and that property has been passed when calling the member.
+
+In particular, this means the warnings will not be printed for:
+
+* Any elements in your own code which are either explicitly deprecated,
+  or override any deprecated members from the Construct Library.
+* Any code that invokes deprecated members,
+  but that is itself not actually invoked when running CDK synthesis.
+* Using any deprecated members without actually invoking them
+  (for example, using a deprecated type only in a type declaration).
 
 ### Contributing guide
 
@@ -66,7 +87,7 @@ and thus no longer available to be used by your code.
 
 If you want to make sure your code does not use any deprecated APIs,
 and thus is ready for migrating to CDK `V2`,
-you can set the `CDK_DEPRECATED` environment variable to the value `error`,
+you can set the `DEPRECATED` environment variable to the value `error`,
 which will make any CDK command that invokes synthesis
 (`cdk synth`, `cdk deploy`, `cdk diff`, etc.)
 fail with an exception if any deprecated element is used.
@@ -110,23 +131,18 @@ There are two high-level components of this change:
 
 We will add a feature to JSII that adds the warning code to the emitted JavaScript for deprecated elements.
 It will be off by default, and will have to be activated explicitly to affect the emitted JavaScript code.
-We will also add an option to the JSII CLI that allows passing the name of the environment variable used for
-silencing the warnings, and turning the warnings into errors
-(the `CDK_DEPRECATED` environment variable mentioned above -
-it can't be just `CDK_DEPRECATED` by default, as JSII does not have any knowledge of CDK!).
 
 #### 2. Using the changed JSII
 
 Once we have modified JSII and released a new version of it,
-we will need to use it in the CDK.
-We will have to start compiling CDK with the new option turned on,
-and also set the new JSII CLI option mentioned above to `CDK_DEPRECATED`.
+we will need to use it in the CDK,
+and start compiling CDK with the new option turned on.
 
-We should also modify our CDK test infrastructure to run with the `CDK_DEPRECATED`
+We should also modify our CDK test infrastructure to run with the `DEPRECATED`
 environment variable set to `error` by default,
 to prevent our own code from using deprecated APIs which would cause warnings to be shown to users.
 We should come up with a nice API that allows individual tests that are explicitly checking deprecated API(s)
-as regression tests to make the value of that environment variable `ignore` easily.
+as regression tests to make the value of that environment variable `quiet` easily.
 
 ### Is this a breaking change?
 
@@ -137,7 +153,7 @@ No.
 1. Requires changes to JSII.
 2. Does not allow distinguishing between the customer using deprecated APIs,
   and the library code using deprecated APIs.
-  We can alleviate this problem by making the value of the `CDK_DEPRECATED` environment variable `error` by default in our tests,
+  We can alleviate this problem by making the value of the `DEPRECATED` environment variable `error` by default in our tests,
   which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
 3. Adding emitting warnings for accessing deprecated static constants
   (including enums) will require huge changes to the emitted JavaScript
@@ -242,7 +258,7 @@ The high-level implementation plan is:
   Effort estimate (development + code review): 2 weeks.
 
 2. Set the new option during building CDK,
-  make sure the tests use `CDK_DEPRECATED` set to `error`,
+  make sure the tests use the `DEPRECATED` environment variable set to `error`,
   and provide an API for tests to opt-out of that on a case-by-case basis.
   Effort estimate (development + code review): 1 week.
 
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -1,0 +1,190 @@
+---
+rfc pr: [#xxx](https://github.com/aws/aws-cdk-rfcs/pull/xxx) <-- fill this after you've already created the PR
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/287
+---
+
+# CDK CLI deprecation warnings
+
+If an element in the CDK Construct Library (class, interface, property, function) is deprecated,
+using it inside a CDK application should produce a warning in the CDK CLI for all commands that perform synthesis
+(`cdk synth`, `cdk diff`, `cdk deploy`, etc.).
+
+This is done in order to aid the migration from `V1` of the CDK to `V2`
+(where all deprecated elements from `V1` will be removed).
+
+## Working Backwards
+
+### CHANGELOG
+
+feat(cli): warn about usages of deprecated Construct Library elements
+
+### README
+
+When using any element (class, interface, property, function) that is deprecated in the Construct Library,
+you will receive a warning when performing any CLI operation that performs synthesis
+(`cdk synth`, `cdk diff`, `cdk deploy`, etc.) that looks similar to:
+
+```
+[Warning at /TestStack/IncludeTemplate] The API @aws-cdk/core.CfnInclude is deprecated:
+  please use the CfnInclude class from the cloudformation-include module instead.
+  This API will be removed in the next major release
+```
+
+## FAQ
+
+### What are we launching today?
+
+In the newest release of the AWS CDK command-line interface,
+any usage of a deprecated element  (class, interface, property, function)
+in your CDK code will result in a warning being printed in the console output.
+
+### Does this mean my code will stop working after this change?
+
+No! This feature only adds warnings to the CLI,
+it does not result in any code that was previously working to now fail.
+
+Note that if you want to migrate to `V2` of the CDK,
+which we strongly recommend,
+you will have to handle all the warnings that this feature emits,
+as all deprecated elements in `V1` will be removed from `V2`,
+and thus no longer available to be used by your code.
+
+## Internal FAQ
+
+### Why are we doing this?
+
+We are doing this to help customers have a smoother migration from `V1` to `V2` of the CDK.
+
+### Why should we _not_ do this?
+
+I see two reasons why we might not want to implement this feature:
+
+1. Deprecated elements are already highlighted by virtually any editor/IDE.
+  In my experience, customers are diligent in moving away from deprecated APIs,
+  so this feature might not provide much value.
+2. We risk overloading customers with warnings,
+  training them to ignore our output if there's too much of it.
+  A warning must be read in order to be effective, after all.
+
+### What changes are required to enable this change?
+
+One of the core tenets behind this RFC is that we want to avoid the need to add any code to the deprecated elements manually.
+Just deprecating the element should automatically trigger the warnings,
+without requiring any other code changes to the element being deprecated.
+
+There are two high-level components of this change:
+
+#### 1. Communication between the framework and the CLI
+
+Since we want to show the warnings when the CLI invokes synthesis,
+we need to communicate between the framework (where those deprecated usages happen)
+and the CLI (where the warnings will be shown).
+The natural way to do that is through the Cloud Assembly.
+
+We already have a facility for adding messages to the construct tree's metadata
+(through the Node class in the `constructs` library),
+which are then [collected into the Stack Cloud Assembly artifact type](https://github.com/aws/aws-cdk/blob/34a921b9667402b6d90731f1fd9e3de1ef27f8bf/packages/%40aws-cdk/core/lib/stack-synthesizers/_shared.ts#L63-L101),
+which the [CLI later renders](https://github.com/aws/aws-cdk/blob/34a921b9667402b6d90731f1fd9e3de1ef27f8bf/packages/aws-cdk/lib/api/cxapp/cloud-assembly.ts#L175-L216).
+We should in all likelihood re-use that mechanism.
+
+Note, however, that that requires access to the construct tree,
+which we don't necessarily have in a trivial way from all contexts
+(for example, for an access to a deprecated property of a struct interface).
+Which means we will need to add some mechanism to `core`
+that allows elements to register warnings with without necessarily having access to a construct,
+and which then will be gathered by synthesis and added to the resulting Cloud Assembly
+(parented below a certain construct, perhaps the `App`?).
+This also means we most likely won't be able to link the usage of the deprecated element to a particular `Stack`
+that it was used in.
+
+#### 2. Adding metadata when using deprecated elements
+
+The second element of this change is to actually add the appropriate metadata when any deprecated element is used.
+
+Before we dive into the possible solutions,
+I think it's worth it to talk about what solutions we _don't_ want,
+as that will strongly inform how we approach the problem.
+
+I think we can rule out any solution that requires parsing the customer's source code in search for usages of deprecated elements.
+The downsides of this approach include:
+
+* We would have to write a separate parser for each language supported by the CDK.
+* This additional parsing could have an adverse impact on the performance of CDK commands.
+
+If we agree that we don't want the above solution,
+then the only avenue we have left is injecting some extra code in the deprecated elements during compilation of the TypeScript code into JavaScript.
+There are two ways we can do that: through JSII, or purely through the TypeScript compiler.
+
+##### 2.1. Modifying JSII
+
+We could modify JSII to add a capability to it that allows injecting some code when compiling every deprecated element.
+It would involve writing a TypeScript transform,
+that adds some code whenever the deprecated element is accessed.
+Nick did a similar change in PR [#2348](https://github.com/aws/jsii/pull/2348)
+to add runtime version information to all JSII-compiled classes.
+
+Advantages:
+
+* Doing it in JSII wil make it easy to record the module the given element is in.
+
+Disadvantages:
+
+* Because the code adding the warning for the deprecated elements will be CDK-specific,
+  it will be pretty awkward to correctly pass what exactly that code should be from CDK to JSII
+  (and, additionally, that code will most likely be slightly different for each type --
+  class, method, property, etc. --
+  of deprecated element we support).
+* Requires coordination between the JSII and CDK projects
+  (merging a JSII PR, waiting for a release, updating CDK to use the new version, etc.),
+  which typically lengthens the total development time.
+
+##### 2.2. TypeScript decorators
+
+We can use [TypeScript decorators](https://www.typescriptlang.org/docs/handbook/decorators.html)
+to enhance the runtime JavaScript code with our custom logic.
+We can add an `awslint` rule that enforces that every element with the `@deprecated`
+JSDoc tag also needs to have the `@deprecated` decorator present.
+While decorators are still considered an experimental feature,
+our init templates and JSII enable them in `tsconfig.json`,
+which means the vast majority of our customers will have them enabled as well.
+
+Advantages:
+
+* Changes are needed only in the CDK project.
+
+Disadvantages:
+
+* It won't be possible to reliably get the module the deprecated element belongs to
+  (only its name).
+
+**My recommendation**: TypeScript decorators.
+
+### Is this a breaking change?
+
+No.
+
+### What are the drawbacks of this solution?
+
+The drawback of this solution is that we will overwhelm our customers with warnings,
+and train them to ignore the output of `cdk` commands because of that.
+
+### What alternative solutions did you consider?
+
+The alternative is to do nothing,
+and rely on editors/IDEs to warn customers of using deprecated features.
+
+### What is the high level implementation plan?
+
+The high-level implementation plan depends on the method chosen:
+
+1. If we decide to modify JSII,
+  we will have to start with a PR to that project first,
+  then update CDK to the latest JSII after it's released,
+  and then work on the CDK side after that.
+
+2. If we decide to go with TypeScript decorators,
+  all of the changes should fit comfortably in a single PR to the CDK project.
+
+### Are there any open issues that need to be addressed later?
+
+No.
--- a/text/287-cli-deprecation-warnings.md
+++ b/text/287-cli-deprecation-warnings.md
@@ -25,27 +25,29 @@ you will receive a warning when performing any CLI operation that performs synth
 (`cdk synth`, `cdk diff`, `cdk deploy`, etc.) that looks similar to:
 
 ```
-[Warning at /TestStack/IncludeTemplate] The API @aws-cdk/core.CfnInclude is deprecated:
+[WARNING] The API @aws-cdk/core.CfnInclude is deprecated:
   please use the CfnInclude class from the cloudformation-include module instead.
   This API will be removed in the next major release
 ```
 
+There are also two environment variables you can set to change the behavior of this feature:
+
+* Setting the environment variable `CDK_DEPRECATED_IGNORE` will silence these warnings
+  (they will no longer be printed to the console).
+
+* Setting the environment variable `CDK_DEPRECATED_ERROR` will instead fail with an exception when any deprecated element is used.
+
 ### Contributing guide
 
-To deprecate any element (class, interface, property, function, etc.)
-in the CDK, you need to make two changes:
-
-1. Add a `@deprecated` tag to the documentation block of the element.
-2. Add a `@deprecated` decorator (from the `@aws-cdk/core` module) to the element itself.
-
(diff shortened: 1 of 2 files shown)


text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+as that will strongly inform how we approach the problem.
+
+I think we can rule out any solution that requires parsing the customer's source code in search for usages of deprecated elements.
+The downsides of this approach include:
+
+* We would have to write a separate parser for each language supported by the CDK.
+* This additional parsing could have an adverse impact on the performance of CDK commands.
+
+If we agree that we don't want the above solution,
+then the only avenue we have left is injecting some extra code in the deprecated elements during compilation of the TypeScript code into JavaScript.
```
eladb, line 115: There is also the alternative of manual annotation, which I understand we dismissed because it's hard to enforce, but worth adding

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+
+We could modify JSII to add a capability to it that allows injecting some code when compiling every deprecated element.
+It would involve writing a TypeScript transform,
+that adds some code whenever the deprecated element is accessed.
+Nick did a similar change in PR [#2348](https://github.com/aws/jsii/pull/2348)
+to add runtime version information to all JSII-compiled classes.
+
+Advantages:
+
+* Doing it in JSII wil make it easy to record the module the given element is in.
```
eladb, line 128: Another advantage is that it will apply to other projects that use jsii (like other CDKs)

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+our init templates and JSII enable them in `tsconfig.json`,
+which means the vast majority of our customers will have them enabled as well.
+
+Advantages:
+
+* Changes are needed only in the CDK project.
+
+Disadvantages:
+
+* It won't be possible to reliably get the module the deprecated element belongs to
```
eladb, line 157: Can you elaborate on this?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+JSDoc tag also needs to have the `@deprecated` decorator present.
+While decorators are still considered an experimental feature,
+our init templates and JSII enable them in `tsconfig.json`,
+which means the vast majority of our customers will have them enabled as well.
+
+Advantages:
+
+* Changes are needed only in the CDK project.
+
+Disadvantages:
```
eladb, line 155: Would that work for all types of depreciations? For example, How would the decorator know about usage of an interface property in a construct?

Maybe a Prototype?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+* It won't be possible to reliably get the module the deprecated element belongs to
+  (only its name).
+
+**My recommendation**: TypeScript decorators.
+
+### Is this a breaking change?
+
+No.
+
+### What are the drawbacks of this solution?
```
eladb, line 166: I think this question is asking about the solution proposed, not the feature itself

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+
+The drawback of this solution is that we will overwhelm our customers with warnings,
+and train them to ignore the output of `cdk` commands because of that.
+
+### What alternative solutions did you consider?
```
eladb, line 171: The alternative solution are described above. Move them here and only leave the proposed solution at the top

eladb (changes requested):
Might be worth doing a prototype

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+# CDK CLI deprecation warnings
+
+If an element in the CDK Construct Library (class, interface, property, function) is deprecated,
+using it inside a CDK application should produce a warning in the CDK CLI for all commands that perform synthesis
+(`cdk synth`, `cdk diff`, `cdk deploy`, etc.).
+
+This is done in order to aid the migration from `V1` of the CDK to `V2`
+(where all deprecated elements from `V1` will be removed).
+
+## Working Backwards
```
eladb, line 15: Can you add a working backwards section from the "contribution guide" which explains the contributor's experience (I.e you add @deprecated, and then add a decorator etc)?

text/287-cli-deprecation-warnings.md
eladb, line 166: What's the best way to make this capability available to other CDKs?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+I think it's worth it to talk about what solutions we _don't_ want,
+as that will strongly inform how we approach the problem.
+
+I think we can rule out any solution that requires parsing the customer's source code in search for usages of deprecated elements.
+The downsides of this approach include:
+
+* We would have to write a separate parser for each language supported by the CDK.
+* This additional parsing could have an adverse impact on the performance of CDK commands.
+
+If we agree that we don't want the above solution,
```
rix0rrr, line 114: Let's think of things we wouldn't be able to analyze using dynamic analysis, then we can decide.

------

If someone puts this in their code:

```ts
new Blabla(this, 'bla', {
  bla: undefined,
});
```

we'd have a hard time detecting that. `if (props.bla)` wouldn't detect it. I think `if ('bla' in props)` would, but the latter wouldn't work for jsii languages.

------

If we internally call a deprecated method, we wouldn't be able to distinguish that from a deprecated method called by the user. *Ideally* we'd only warn about the latter.

I guess it shouldn't happen too much in practice, so it wouldn't matter... ?

------

We wouldn't be able to detect usage of deprecated enum values (or I suppose we could--it would be a bunch of additional work though).

------

We wouldn't be able to detect the use of deprecated type declarations (any of enum type, interface type, class type). For example:

```ts
DeprecatedClass obj = new NonDeprecatedClass();
// or
DeprecatedClass obj = construct.giveMeSomething();
```

----

Now this might all be acceptable (because we can argue they'd be rare, and a *correct* solution is fantastically expensive and might not be worth it). But let's trade this off with eyes open.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+It would involve writing a TypeScript transform,
+that adds some code whenever the deprecated element is accessed.
+Nick did a similar change in PR [#2348](https://github.com/aws/jsii/pull/2348)
+to add runtime version information to all JSII-compiled classes.
+
+Advantages:
+
+* Doing it in JSII wil make it easy to record the module the given element is in.
+
+Disadvantages:
```
rix0rrr, line 130: How would it work for struct properties?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,190 @@
...
+
+The high-level implementation plan depends on the method chosen:
+
+1. If we decide to modify JSII,
+  we will have to start with a PR to that project first,
+  then update CDK to the latest JSII after it's released,
+  and then work on the CDK side after that.
+
+2. If we decide to go with TypeScript decorators,
+  all of the changes should fit comfortably in a single PR to the CDK project.
```
rix0rrr, line 186: For some other person's definition of "comfortably" 🤣

skinny85: Says the guy with a [PR that changes 118 files](https://github.com/aws/aws-cdk/pull/12920) 😃.

text/287-cli-deprecation-warnings.md
skinny85: I don't think I know of this alternative. Do you care to expand on how would that work Elad?

text/287-cli-deprecation-warnings.md
skinny85: Great point. Added.

text/287-cli-deprecation-warnings.md
skinny85: Added some more explanation. Let me know if it's clear now.

text/287-cli-deprecation-warnings.md
skinny85: Good point. I checked before, and TypeScript allows for [property decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#property-decorators), which I thought meant we can have them on structs. But I did some more research, and turns out [you can't have decorators on interfaces](https://stackoverflow.com/questions/42699737/use-decorator-to-add-meta-data-to-interface-properties), as they get erased at runtime.

This definitely adds a very serious downside to the decorator route. I've added it to the list.

text/287-cli-deprecation-warnings.md
skinny85: Ah, OK. I misunderstood how this works then. I'll move the disadvantages of the decorators to here.

text/287-cli-deprecation-warnings.md
skinny85: Done.

text/287-cli-deprecation-warnings.md
skinny85: Done!

text/287-cli-deprecation-warnings.md
skinny85: I feel like it'll be hard to share this solution between CDKs. Even if we did the changes in JSII, the part that gathers the metadata into the Stack, and the part that renders it in the CLI, is CDK-specific. It would be hard to move that to a common place like `constructs` (because it uses the knowledge of the output format of each CDK - in the case of the AWS CDK, the Cloud Assembly, for instance).

text/287-cli-deprecation-warnings.md
skinny85: Fair enough. Added these considerations.

text/287-cli-deprecation-warnings.md
skinny85: I guess for struct properties, we would have to detect their usage statically in JSII 😕.

skinny85:
Thanks for the comments @eladb @rix0rrr ! I've incorporated them in the newest revision.

text/287-cli-deprecation-warnings.md
eladb: Just Hand-Writing code for each deprecated case.

text/287-cli-deprecation-warnings.md
eladb: Also, if you decorate properties of structs, how would you know they are actually used?

text/287-cli-deprecation-warnings.md
eladb: Can you also please Open an issue to add some clarification to the rfc template ?

skinny85:
@eladb I talked about this with @rix0rrr offline, and I think I'd want to re-structure the RFC a little bit.

Since the RFC format has a "working backwards" structure, it heavily encourages proposing a concrete solution to a problem, with a given customer experience.

I think maybe we should step a little, and re-phrase this RFC in terms of the problem we're trying to solve. In this case, it is:

### We want to help our customers migrate their applications from CDK V1 to V2

Given that goal, there are a few ways we can solve this issue.

The ideal experience I can imagine for this would be some tool, let's hypothetically call it `cdk upgrade`, that actually changes your code to be V2-compliant:

1. Changes the packages you depend on in `package.json` / `pom.xml` / `requirements.txt`/ etc.
2. Changes the imports your code uses, if needed.
3. Changes the usages of deprecated elements to their non-deprecated equivalents.

While this is the ideal experience, I think we can agree that building it would be fantastically expensive.

Given that, the second best thing we can do is warn the customer when they're using something that will not be available in V2, like deprecated elements (I can definitely see extending it to things like setting no longer supported feature flags, etc.). While I originally said in the RFC that it should happen for anything that does `cdk synth`, Rico convinced me that this is locking in to a particular solution too strongly.

What are the possible solution we can use to emit these warnings?

#### 1. Parse the customer's code directly

Advantages:

1. Allows us to discover some tricky deprecated elements,
  like enum values, or deprecated types used only in type declarations,
  that will be difficult to do with only runtime reporting.
2. Would prevent reporting warnings for deprecated usages outside the customer's code
  (for example, in our own libraries).

Disadvantages:

1. We would have to write a separate parser and type checker for each language supported by the CDK.
2. This additional parsing could have an adverse impact on the performance of CDK commands. This downside can be mitigated by moving this to a separate, non-`synth` command (`cdk doctor`, perhaps?), although that has its own downsides (customers might not know about a separate command).
3. It's not obvious this analysis can even be performed at all in the case of dynamically-typed languages like JavaScript or Python, and even TypeScript has some edge-cases making it pretty difficult.

#### 2. Use TypeScript decorators to add warnings at runtime

Advantages:

1. Changes only needed in CDK code.

Disadvantages:

1. It won't be possible to reliably get the module the deprecated element belongs to
  (only the name of the element) --
  the decorator is simply a JavaScript function defined in `core`,
  and doesn't have (easy) access to the information on what module a given object it's called on belongs to.
2. TypeScript does not allow decorators on interfaces
  (neither on the interface directly, nor on any of its properties).
  This means we will not be able to register warnings for structs, or struct properties.
3. We will not be able to distinguish between the customer's and our's usages of deprecated elements, which means the customer will be getting warnings they cannot fix.

#### 3. Modify JSII to change the emitted JavaScript for deprecated elements

Advantages:

1. Doing it in JSII will make it easy to record the module the given element is in.
2. Doing it in JSII allows this functionality to be used by other JSII projects,
  like `cdk8s`, or `cdktf`.

Disadvantages:

1. Because the code adding the warning for the deprecated elements will be CDK-specific,
  it will be pretty awkward to correctly pass what exactly that code should be from CDK to JSII
  (and, additionally, that code will most likely be slightly different for each type --
  class, method, property, etc. --
  of deprecated element we support).
2. Requires coordination between the JSII and CDK projects
  (merging a JSII PR, waiting for a release, updating CDK to use the new version, etc.),
  which typically lengthens the total development time.
3. We will not be able to distinguish between the customer's and our's usages of deprecated elements, which means the customer will be getting warnings they cannot fix. We can mitigate that warning by creating the API for silencing warnings, and then using that new API in all sites of deprecated code in our libraries, but that will be tedious, manual work, which will be hard (but probably not impossible) to verify in an automated way that it was done.

#### 4. Rely on linters

We can modify JSII emit deprecated elements into the source code of the languages used, which is then picked up by a linter for that language.

Advantages:

1. A simpler solution than nr 3 in terms of the number of JSII changes required.
2. Allows us to distinguish between the customer's and our own usages of deprecated elements, reducing the number of false positives.

Disadvantages:

1. Will require the customer to run a separate command on their source code, language-specific.

---------------------

@eladb @rix0rrr - what are your thoughts on re-writing the RFC to be more in that format?

eladb:
@skinny85, @rix0rrr, I agree that we need to devise a holistic story for migration from V1 to V2. Incidentally, @nija-at and I discussed this last week, and said we will likely need a dedicated RFC for this that will work backwards from the migration process. I think that for dev-preview we just need a doc that tells users what they need to do (we even have some snippets of it in the [monocdk RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0006-monolothic-packaging.md#module-names) and the [constructs RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0192-remove-constructs-compat.md#release-notes)).

Having said that, as I see it, warnings about usage of deprecated APIs has value outside of the migration point. We don't want users to use deprecated APIs in general, and we also want them to approach new major versions with the minimum amount of required changes. This implies that we want to inform them about usage of deprecated APIs as soon as possible. The shorter the list of changes they need to make (and the more mechanical it is), its more likely they will migrate. Additionally, some of the alternative solutions described above will not work *after* migration, only *before*, because the 2.0 codebase does not include any of the deprecated APIs (at least per our current plan).

As for solutions, I still think we should deeply look into the straightforward alternative of **manually annotating usage of deprecated APIs**. I have a feeling we ditched it too soon and I think from a cost/benefit perspective it's actually the most appealing one. I understand it might not be fool-proof and/or automatic as the other options, but I think it can pragmatically address the problem we are trying to solve (i.e. "warn users of 1.0 about usage of deprecated APIs").

I will articulate the solution so it's clear what I mean: we will manually go, search for "@deprecated" and add `Annotations.addDeprecation()` calls everywhere in our codebase when deprecated APIs are used (method calls, classes, props, etc). Simply write code. This is a one-time, mechanical effort across our codebase, which I don't think is going to be very expensive. As soon as we do that, we've already addressed the major customer need: v1.0 users will get warnings during synth/deploy if they use deprecated APIs (yeah they will see usage of deprecated APIs in dependencies but we can attach stack traces to construct warnings). Even if this is where we stop and maybe forget a few instances, I would argue that's probably "good enough" (esp. given IDEs also show usage of deprecated APIs). 

We can _also_ add a **PR linter rule** (@iliapolo can help) to identify introduction of new `@deprecated` APIs and require the PR diff will also include an `addDeprecation()` call. Again, this is not fool-proof and may require overrides in some occasions, but I'd argue it's completely sufficient for the need.

The main benefit of this approach is that we know it will work for all use cases and deprecation types (I still don't understand how some of the proposed solutions will identify usage of deprecated __props_ in constructs). It will also not require us to go into an adventure of building sophisticated tools and introduce more uncertainty into the 2.0 plan.

Given our current state with 2.0 timelines (and generally), it's important that we will be critical of how much risk/uncertainty a certain approach adds (that's missing in the analysis of the alternatives). In my mind all the solutions that involve per-language code analysis introduce too much uncertainty for this stage in the project, and any non-trivial solution requires a prototype to prove that it can address all the types of deprecations.

rix0rrr:
> building sophisticated tools

This does not need to be the case. We can just integrate with existing language tooling, for even less effort than writing our own dynamic deprecation checking (which, as shown, cannot check all usages of deprecations).

* TypeScript: eslint plugin https://github.com/gund/eslint-plugin-deprecation 
* Java: we're already covered because we already emit the `@Deprecated` annotation (compiler already produces warnings)
* .NET: we're already covered because we already emit the `[System.Obsolete()]` annotation (compiler already produces warnings)
* Python -- the only one that's not covered right now.

Is adding a `@jsii.deprecated` attribute to the Python generator and writing a documentation page on how to set up eslint for TypeScript projects is not less effort than manually annotating all code everywhere? Not to mention it produces better results too.

Also, don't forget that we're going to have to remove all that `Annotations.addDeprecation()` code again in the `v2` branch, where it's going to lead to forward merge conflicts which take time to clean up... so the TCO of this code is more than just a Ctrl-F and copy/pasting some lines in (which is probably already going to be quite sizeable, honestly). And we'd have to do the same again for deprecations added later in v2.

The ROI on manually writing a bunch of checking code seems to be strictly lower than taking a couple minutes to look around and reuse what's already there... and even doing the latter doesn't seem to be all that expensive when compared to the costs of the naive solutions.

I really don't get why you are insisting on it so much?

eladb:
It's an interesting analysis. The fact the TypeScript, Java and .NET are technically already covered is interesting (and I'd argue that now that VSCode strikethroughs deprecated APIs we don't even need the eslint rule). If we have a solution for python then we basically don't need to do _anything_ to address the main concern.

I like this direction: focus on finding a way to warn Python users and basically that's it. No need to do any code analysis or write new tools. If we know users get warnings (either compile time or IDE) about usage of deprecated APIs, I think we've covered the need this RFC attempts to address.

----

> don't forget that we're going to have to remove all that Annotations.addDeprecation() code again in the v2 branch

I was under the impression that we are just going to hide deprecated APIs, not delete the code. Did I miss anything?

rix0rrr:
> I'd argue that now that VSCode strikethroughs deprecated APIs we don't even need the eslint rule

Minor disagreement on that. What most people are going to use this for, is going to be to answer the question "am I using any deprecated APIs in this project? yes/no" (as a subquestion of the ACTUAL question "can I safely upgrade this project?")

Having to open the project in an IDE and scrolling through every file to see if any identifiers are struck through is not a good way to answer that question. You would want something that runs in bulk across all files and will identify the locations you need to change. eslint could do that (and a quick Google doesn't turn up any other tool that would do that, which surprises me but there we go).

rix0rrr:
> I was under the impression that we are just going to hide deprecated APIs, not delete the code. Did I miss anything?

Hmm. You are probably right. Disregard the point about merge conflicts then (but the point that we'll have to remove everything again when we go to actually strip the code remains).

rix0rrr:
Oh my god of course Python's solution has been overengineered again: https://docs.python.org/3/library/warnings.html

This looks promising, we can probably borrow something from this: https://deprecation.readthedocs.io/en/latest/

eladb:
If we take the approach of not adding a CDK/jsii-level mechanism to warn about deprecations, we should evaluate the potential developer experience in each language.

I think we want deprecation warnings to be surfaced _without_ requiring that users explicitly run some tool or change the configuration of their project. This implies that either this has built-in support in the language toolchain (Java/.NET/TypeScript*) _or_ that we utilize the CDK CLI somehow to surface those warnings via language-specific tools or through construct warnings (the original approach). *We should decide if IDE-only (TypeScript) is good enough or not. If it eventually turns out that doing more than IDE warnings is very expensive, we might be able to compromise here.

What about JavaScript? Do we want to address it? It is an officially supported CDK language, but we don't know how many people use JavaScript directly.

NetaNir:
+1 for leveraging existing deprecation mechanism in every target language.  It seems like the only language we don't have covered is TypeScript. I don't think we can rely on eslint, we don't even add it as a dependency in the CDK init template.

> We should decide if IDE-only (TypeScript) is good enough or not. If it eventually turns out that doing more than IDE warnings is very expensive, we might be able to compromise here.

IDE visual warnings are only helpful for parts of the code that are actively been worked on. The majority of code in a code base is not reviewed in the IDE as a part of a usual development flow (_side note: are TS strike-through available in all IDEs? Webstorm/Intellij?_).

Since the support in other languages for deprecated API warnings seems to be good enough, I would argue that we should decide if we want to invest in a TypeScript only solution. To clarify,  I mean: "If we do build something, it should be TS specific". 

A note about a solution that prints API deprecation warnings on every CDK CLI command. If we do add end up adding something like that, we must add a `--silence` flag. We hear often from users that adding warnings to CDK commands output breaks automation/testings.

eladb:
I am still not certain if the trivial manual construct annotations might be the most cost/effective approach given our multi-language story.


I am not sure what's the full solution for Python and JavaScript (and TypeScript*), so we need to evaluate this the overall effort and experience in all languages before we determine that simple warnings are "harder".

rix0rrr:
I would argue that for TypeScript, the combination of:

* IDE hints while working; and
* Running [check-typescript-deprecations](https://www.npmjs.com/package/check-typescript-deprecations) on-demand for a bulk notification:

```
$ npx check-typescript-deprecations

/Users/me/Project/bin/app.ts
  9:9  warning  'add' is deprecated. use `Tags.of(scope).add()`  deprecation/deprecation

✖ 1 problem (0 errors, 1 warning)
```

...are good enough user affordances, no?

eladb:
Who is going to run the typescript utility? Our CLI?

rix0rrr:
> Who is going to run the typescript utility? Our CLI?

My thought was: the user, if they definitely need to know the answer to their question "am I using deprecated APIs?" *definitively* and *right now*, in preparation of an upgrade (or just because they're curious).

To nudge users in the right direction in the course of normal development, I would say that the strikethrough is probably motivation enough.

eladb:
I tend to agree that IDE strikethroughs are probably good enough as an implicit signal. What about Python?

rix0rrr:
> What about Python?

The only mechanism Python has is to print warnings when the program is run. There is no static analysis (barring MyPy but it does not have deprecation support).

Result would be ~ the same as if we added CDK warnings.

nija-at:
Getting language IDEs and tooling to notify of deprecated API usage sounds like a good plan and the right overall direction.

For v2, however, we need something more proactive to nudge users the right way. It needs to be mildly annoying. They should have the option to opt-out but it should opt-in by default (especially for existing CDK apps). 
Asking users to set up a new tool is the opposite experience of that - an opt-in experience. This could result in far less users actually moving to the non-deprecated APIs, causing a lot more effort in migrating to v2 and finally resulting in reduced adoption of CDKv2.

If we want to stop supporting v1 and the two branch mode, we need to get more users onto v2.

rix0rrr:
# IntelliJ

image

image

image

rix0rrr:
# Microsoft Visual Studio (for Mac)

image

image

rix0rrr:
# VSCode

Apparently VSCode will strikethrough member accesses always, but properties only in the autocomplete list (not in running code). This is probably a bug...

JavaScript:

image

TypeScript

image

rix0rrr:
# Python (no IDE in particular)

image

image

image

skinny85:
Thanks for the comments @eladb @rix0rrr @nija-at @NetaNir!

I've submitted a new revision, where the preferred solution has been switched to changing JSII. I've also greatly expanded the "Alternative solutions" section to synthesize (hopefully) the discussion we've been having.

Please comment, and let me know what you think!

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+  (because they're coming from code that they don't control).
+
+### What changes are required to enable this change?
+
+There are two high-level components of this change:
+
+#### 1. Changes to JSII
+
+We will add a feature to JSII that adds the warning code to the emitted JavaScript for deprecated elements.
+It will be off by default, and will have to be activated explicitly to take effect.
```
rix0rrr, line 102: What does that mean?

The default behavior of the application will be `ignore`, and the CLI will in turn flip that to `warn` by default unless ignored?

Motivation: unit tests etc I suppose?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+No.
+
+### What are the drawbacks of this solution?
+
+1. Requires changes to JSII.
+2. Does not allow distinguishing between the customer using deprecated APIs,
+  and the library code using deprecated APIs.
+  We can alleviate this problem by turning on the `CDK_DEPRECATED_ERROR` environment variable by default in our tests,
+  which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
+3. We won't be able to register warnings for accessing deprecated static constants
```
rix0rrr, line 129: Not actually true I don't think.

We can turn:

```ts
class Foo {
  /** @deprecated */
  public static readonly CONSTANT = 5;
}
// into
class Foo {
  public static get CONSTANT(): number {
    console.error('[WARNING] oh no');
    return 5;
  }
}
```

And something similar for enums.

Honestly, I feel like something like:

```ts
// Would work fine and would work for enums too
Foo.prototype.CONSTANT = deprecatedAccessor(Foo.prototype.CONSTANT, 'Use Foo.Bloop instead');
```

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+We could simply piggyback on those,
+instead of writing our own.
+
+Advantages of this solution:
+
+1. Minimal development effort on our side.
+
+Disadvantages of this solution:
+
+1. Just deprecated warnings in the IDE can be missed by developers
```
rix0rrr, line 183: I don't find this convincing, and I'm not convinced that printing to stderr is less likely to be missed.

We're probably talking about different contingents of programmers here, and printing to stderr is optimizing more for developers "like us".

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+#### 3. Rely on language-specific deprecation warning mechanisms
+
+JSII emits deprecated elements with the correct language-specific mechanisms in place
+(for example, the `@Deprecated` annotation for Java).
+We could simply piggyback on those,
+instead of writing our own.
+
+Advantages of this solution:
+
+1. Minimal development effort on our side.
```
rix0rrr, line 179: Advantage missed: static analysis instead of runtime analysis allows listing of all deprecated API usages in the code base, **not the just ones that HAPPEN to be executed during a particular run**

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+### What are the drawbacks of this solution?
+
+1. Requires changes to JSII.
+2. Does not allow distinguishing between the customer using deprecated APIs,
+  and the library code using deprecated APIs.
+  We can alleviate this problem by turning on the `CDK_DEPRECATED_ERROR` environment variable by default in our tests,
+  which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
+3. We won't be able to register warnings for accessing deprecated static constants
+  (this includes enums).
+
```
rix0rrr, line 131: Drawback missed: can only report on APIs actually hit during a particular run.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+Advantages of this solution:
+
+1. Minimal development effort on our side.
+
+Disadvantages of this solution:
+
+1. Just deprecated warnings in the IDE can be missed by developers
+  (especially as more APIs get deprecated).
+2. The experience with TypeScript properties is currently pretty bad
+  (the property is struck through when selecting it,
```
rix0rrr, line 186: This description only makes sense if you've been following the discussion closely, and will fly right over the head of anyone coming into this clean.

Try rewriting to something like: 

"TypeScript currently does not properly detect deprecation of properties in object literals, and deprecation of properties (props) is very important for our project. Even though it will properly detect them while autocompleting, and there is an issue about it on the TypeScript bug tracker, we cannot rely on them fixing this in a reasonable time frame. "

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+
+The high-level implementation plan is:
+
+1. Make the changes in JSII
+  (will probably be a single PR),
+  and release a new version once those are merged in.
+
+2. Set the new option during building CDK,
+  make sure the tests use `CDK_DEPRECATED_ERROR`,
+  and provide an API for tests to opt-out of that on a case-by-case basis.
```
rix0rrr, line 220: 3. Rewrite all code to stop using deprecated APIs (estimated effort: 2 months)

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Are there any open issues that need to be addressed later?
+
+No.
```
rix0rrr, line 224:
```
### Are there any things that need to be paid attention to during implementation

<things from email chain>
```

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+We will also add two options to the JSII CLI that allow passing the names of the environment variables used for,
+respectively, silencing the warnings, and turning the warnings into errors.
+
+#### 2. Using the changed JSII
+
+Once we have modified JSII and released a new version of it,
+we will need to use it in the CDK.
+We will have to start compiling CDK with the new option turned on,
+and also set the two additional options mentioned above to be
+`CDK_DEPRECATED_IGNORE` and `CDK_DEPRECATED_ERROR`, respectively.
```
nija-at, line 112: Not sure I follow this sentence. Can you re-state?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+
+We should also modify our CDK test infrastructure to run with the `CDK_DEPRECATED_ERROR` option turned on by default,
+unless a test is explicitly checking a deprecated API as a regression test,
+in which case we should add an API for turning that option off for those tests.
```
nija-at, line 116: > add an API for turning that option off for those tests

Not following this. Add an API where? can you expand?

nija-at: We should discuss this f2f. 

I'm working on a similar area and have encountered unforeseen issues, and has generally been harder than I thought it will be. It's because of the different ways we deprecated elements.

Maybe you'll have the same problems, maybe not. Worth discussing.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+2. Would prevent reporting warnings for deprecated usages outside the customer's code
+  (for example, in our own libraries).
+
+Disadvantages of this solution:
+
+1. We would have to write a separate parser and type-checker for each language supported by the CDK.
+2. This additional parsing could have an adverse impact on the performance of CDK commands.
+3. It's not obvious this analysis can even be performed at all in the case of dynamically-typed languages like JavaScript or Python
+
+### What is the high level implementation plan?
```
nija-at, line 210: For the sake of planning, can you also attach high level effort estimates for each item.

text/287-cli-deprecation-warnings.md
skinny85: It means the code adding the warnings will not be added to the emitted JavaScript by default at all.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,224 @@
...
+
+```
+[WARNING] The API @aws-cdk/core.CfnInclude is deprecated:
+  please use the CfnInclude class from the cloudformation-include module instead.
+  This API will be removed in the next major release
+```
+
+There are also two environment variables you can set to change the behavior of this feature:
+
+* Setting the environment variable `CDK_DEPRECATED_IGNORE` will silence these warnings
```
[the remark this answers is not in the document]
skinny85: Done!

text/287-cli-deprecation-warnings.md
skinny85: Fair enough. That's a huge intrusion to the generated code though. I've re-phrased the downside to reflect that.

text/287-cli-deprecation-warnings.md
skinny85: I mean, we're showing the warnings in the IDE anyway - we're not taking this away. So if you don't agree they will be missed, then I guess you're in favor of a different option than the JSII one, no?

text/287-cli-deprecation-warnings.md
skinny85: Added.

text/287-cli-deprecation-warnings.md
skinny85: Added.

text/287-cli-deprecation-warnings.md
skinny85: Copied your wording verbatim.

text/287-cli-deprecation-warnings.md
skinny85: I don't think it will take that long... see https://github.com/aws/aws-cdk/pull/13500.

text/287-cli-deprecation-warnings.md
skinny85: Added!

text/287-cli-deprecation-warnings.md
skinny85: Re-worded. Let me know if this is clear now.

text/287-cli-deprecation-warnings.md
skinny85: I meant add the API to our testing libraries. Similarly like you added an API for tests using feature flags.

Sure, I'd love to discuss this F2F.

text/287-cli-deprecation-warnings.md
skinny85: Done.

skinny85:
@rix0rrr @nija-at included all of your comments, thanks! Would appreciate if you could take another pass.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,264 @@
...
+you will receive a warning when performing any CLI operation that performs synthesis
+(`cdk synth`, `cdk diff`, `cdk deploy`, etc.) that looks similar to:
+
+```
+[WARNING] The API @aws-cdk/core.CfnInclude is deprecated:
+  please use the CfnInclude class from the cloudformation-include module instead.
+  This API will be removed in the next major release
+```
+
+There is also an environment variable `CDK_DEPRECATED` that you can use to change the behavior of this feature:
```
eladb, line 33: Since this is implemented at the JSII level it doesn't make sense to call this `CDK_`, either `JSII_DEPRECATED` or just `DEPRECATED`

```suggestion
There is also an environment variable `DEPRECATED` that you can use to change the behavior of this feature:
```

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Working Backwards
+
+### CHANGELOG
+
+feat(cli): warn about usages of deprecated Construct Library elements
+
+### README
+
+When using any element (class, interface, property, function) that is deprecated in the Construct Library,
```
eladb, line 23: Can we describe when this will _not_ work? For example, what does it mean to "use" an interface in this context?

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,264 @@
...
+
+```
+[WARNING] The API @aws-cdk/core.CfnInclude is deprecated:
+  please use the CfnInclude class from the cloudformation-include module instead.
+  This API will be removed in the next major release
+```
+
+There is also an environment variable `CDK_DEPRECATED` that you can use to change the behavior of this feature:
+
+* Setting that environment variable to the value `ignore` will silence these warnings
```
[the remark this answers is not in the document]
skinny85: Really? I feel like `ignore` is a much more standard term... I've never seen `hide` used in this context.

text/287-cli-deprecation-warnings.md
skinny85: > Since this is implemented at the JSII level it doesn't make sense to call this `CDK_`,

Yes. That's why, a little below, it says:

> #### 1. Changes to JSII
>
> (...)
> We will also add an option to the JSII CLI that allows passing the name of the environment variable used for silencing the warnings, and turning the warnings into errors
> (the `CDK_DEPRECATED` environment variable mentioned above -
> it can't be just `CDK_DEPRECATED` by default, as JSII does not have any knowledge of CDK!).

I feel like that's a better solution than `JSII_DEPRECATED`, which is too low-level (I don't think many customers are aware of JSII, and TS/JS customers don't use it at all), and `DEPRECATED`, which is too generic.

text/287-cli-deprecation-warnings.md
skinny85: Since that depends on the exact solution chosen, I've omitted it from this part.

Do we need to get into such details in the user-facing documentation?

text/287-cli-deprecation-warnings.md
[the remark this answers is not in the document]
eladb: `ignore` doesn't reflect really reflect the user experience (it describes the what the mechanism is doing). I would use either `hide` or `silence` or `quiet` or something like that to make it clear that if this is set users won't see anything.

text/287-cli-deprecation-warnings.md
eladb: I feel like adding a CLI switch that says what the name of this environment variable is unneeded complexity. Let's keep things simple just call this `DEPRECATED`. It's not a one-way door and if we end up deciding that we want to add support for "controlling the name of the deprecation environment variable in jsii", we can always add that later (but I doubt that will ever happen).

text/287-cli-deprecation-warnings.md
eladb: This document should propose a solution and the "working backwards" section should be written as if this the proposed solution is the one we chose. If we end up changing the solution for whatever reason, we will need to evaluate the implications on the user experience and update this section (which is a good thing :-)).

text/287-cli-deprecation-warnings.md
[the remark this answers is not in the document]
skinny85: OK. Changed.

text/287-cli-deprecation-warnings.md
skinny85: Ok. Changed.

text/287-cli-deprecation-warnings.md
skinny85: OK. Added. Let me know if this works.

skinny85:
@eladb thanks for the review! Included all of your comments, would appreciate another pass.

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,280 @@
...
+
+1. Requires changes to JSII.
+2. Does not allow distinguishing between the customer using deprecated APIs,
+  and the library code using deprecated APIs.
+  We can alleviate this problem by making the value of the `DEPRECATED` environment variable `error` by default in our tests,
+  which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
+3. Adding emitting warnings for accessing deprecated static constants
+  (including enums) will require huge changes to the emitted JavaScript
+  (changing from a field to a `static` getter).
+4. We only report APIs actually hit during a particular execution of the code
```
eladb, line 161: Good call out (worth mentioning that this is displayed In the IdE of most languages)

eladb (dismissed):
Great write up

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,280 @@
...
+
+### What are the drawbacks of this solution?
+
+1. Requires changes to JSII.
+2. Does not allow distinguishing between the customer using deprecated APIs,
+  and the library code using deprecated APIs.
+  We can alleviate this problem by making the value of the `DEPRECATED` environment variable `error` by default in our tests,
+  which will be a forcing function for us to stop using deprecated APIs in the Construct Library.
+3. Adding emitting warnings for accessing deprecated static constants
+  (including enums) will require huge changes to the emitted JavaScript
```
rix0rrr, line 159: If we don't emit complex JavaScript, but generate this at the end of the file, it shouldn't be too bad:

```ts
function deprecate(obj, property, warning) {
   const oldValue = obj[property];
   Object.defineProperty(obj, property, {
     get: () => { console.error(warning); return oldValue; }
   });
}
```

```
> bar = { y: 42 }
{ y: 42 }
> deprecate(bar, 'y', 'You cant touch this');
undefined
> bar.y
You cant touch this
42
```

text/287-cli-deprecation-warnings.md
```diff
@@ -0,0 +1,280 @@
...
+Advantages of this solution:
+
+1. Minimal development effort on our side.
+2. Compared to runtime analysis, allows listing of all deprecated API usages in the code base,
+  not the just ones that happen to be executed during a particular run.
+
+Disadvantages of this solution:
+
+1. Just deprecated warnings in the IDE can be missed by developers
+  (especially as more APIs get deprecated).
```
rix0rrr, line 220: I think we're trying to say something more specific here.

```suggestion
1.  TypeScript in VSCode (our most popular language/IDE combination) does not by default show a complete
     list of deprecated API usages: it will only show strikethroughs on identifiers you happen to be looking at.
     Getting a complete list of deprecated APIs requires installing and configuring `eslint`.
```

(This objection is not true for other IDE/language combinations)

text/287-cli-deprecation-warnings.md
rix0rrr: Hah! :D

text/287-cli-deprecation-warnings.md
skinny85: Done.

text/287-cli-deprecation-warnings.md
skinny85: Removed this downside.

text/287-cli-deprecation-warnings.md
skinny85: Done.

eladb (approved):
🥇

eladb:
Added `do-not-merge` for final comments period.

PR #292, opened 2021-02-23
refactor(go): updates to RFC204 - golang bindings

RomainMuller:
This updates RFC204 based on the latest developments for the
implementation thereof. In particular, the following elements have been
changed:

- Updates to *jsii enums*
  - Preserve capitalisation of enum constant names, and separate from type name with `_`

- Updates to *jsii interfaces* (aka: behavioral interfaces)
  - Added a go struct for the concrete jsii proxy object

- Updates to *jsii structs* (aka: datatype interfaces)
  - Replaced `Iface` suffix in go interface names with an `I` prefix
  - Added *Approach 3* for handling jsii struct inheritance

- Updates to *jsii classes*
  - Removed `Iface` suffix from go interfaces generated for classes
  - Un-exported the go struct concrete type for the jsii proxy objects
  - Removed unused properties from concrete types for jsii proxy objects
  - Used mock jsii runtime calls to clarify activities that are proxied
    to the `@jsii/kernel` process
  - Using `_` to separate the type name & static method or property name

- Updates to *properties*:
  - Where possible (i.e: everywhere, except in `structs` approaches 1, 2 and 3,
    removed the `Get` prefix from getters (keeping `Set` prefix for setters), to
    match the recommendations of [effective go](https://golang.org/doc/effective_go#Getters)

---

### DIFF
--- a/text/204-golang-bindings.md
+++ b/text/204-golang-bindings.md
@@ -171,8 +171,8 @@ type ISecurityGroup interface {
     IResource
     IPeer
 
-    GetSecurityGroupId() string
-    GetAllowAllOutbound() bool
+    SecurityGroupId() string
+    AllowAllOutbound() bool
     AddIngressRule(peer: IPeer, connect: Port, description: *string, remoteRule: *boolean)
     AddEgressRule(peer: IPeer, connect: Port, description: *string, remoteRule: *boolean)
 }
@@ -790,8 +790,8 @@ import "jsii"
 
 // The interface represents the class in the API.
 type Greeter interface {
-    Greet()       string
-    GetGreeting() string
+    Greet()    string
+    Greeting() string
 }
 
 // The struct is the concrete implementation for the type. This is a JS object
@@ -811,7 +811,7 @@ func NewGreeter(message string) Greeter {
     return g
 }
 
-func (g *greeter) GetGreeting() (result string) {
+func (g *greeter) Greeting() (result string) {
     // Getting the property from the JS process
     jsii.Get(g, "greeting", &result)
     return
@@ -830,7 +830,7 @@ func Greeter_Foo() (result string) {
 }
 
 // static property getter
-func Greeter_GetHello() string {
+func Greeter_Hello() string {
     jsii.StaticGet("example.Greeter", "hello", &result)
     return
 }
@@ -883,9 +883,11 @@ In Go:
 ```go
 package animal
 
+import "jsii"
+
 // Base class as interface
 type Animal interface {
-    GetName() string
+    Name() string
 
     Move(distance int64)
     isAnimal() // private method saftey check
@@ -893,19 +895,23 @@ type Animal interface {
 
 // Base class implementation
 type animal struct {
-    Name string
+    // So this is not 0-wodth
+    _ byte // padding
 }
 
 func NewAnimal(name string) Animal {
-    return &animal{name}
+    a := &animal{}
+    jsii.Create(a, "example.Animal", []interface{}{name})
+    return a
 }
 
-func (a *animal) GetName() string {
-    return a.Name
+func (a *animal) Name() (result string) {
+    jsii.Get(a, "name", &result)
+    return
 }
 
 func (a *animal) Move(distance int64) {
-    fmt.Printf("%s moved %vm.\n", a.Name(), distance)
+    jsii.InvokeVoid(a, "move", []interface{}{distance})
 }
 
 // Child class
@@ -919,12 +925,8 @@ type snake struct {
 }
 
 func NewSnake(name string) Snake {
-    a := NewAnimal{name}  // or ExtendAnimal, to avoid introspection for super calls later
+    a := NewAnimal(name)  // or ExtendAnimal, to avoid introspection for super calls later
     return &snake{a}
-}
-
-func (s *snake) Name() string {
-    return s.name  // inherits `name` property from `Animal`
 }
 
 func (s *snake) Move(distance int64) {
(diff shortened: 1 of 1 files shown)


text/204-golang-bindings.md
```diff
@@ -66,8 +66,8 @@ package launchtype
 type LaunchType string
 
 const (
-    LaunchTypeEc2     LaunchType = "EC2"
-    LaunchTypeFargate LaunchType = "FARGATE"
+    LaunchTypeEC2     LaunchType = "EC2"
+    LaunchTypeFARGATE LaunchType = "FARGATE"
```
RomainMuller, line 70: Counter-argument here (just occurred to me) - is this going to look very awkward when the enum constant name includes `_` characters? This looks kinda "meh":

```go
LaunchTypePRETEND_THIS_EXISTS LaunchType = "PRETEND_THIS_EXISTS"
```

But on the other hand, we'd get into problems if we have the following (as the enum constant `LaunchTypePretendThisExists` and the class would end up getting the exact same name):

```ts
export enum LaunchType {
  PRETEND_THIS_EXISTS
}

export class LaunchTypePretendThisExists implements ILaunchType {
  //...
}
```

(Not that this problem doesn't also exist with the `SHOUT_CASED` names without `_`, too, e.g: `LaunchTypeEC2` class and `LaunchType.EC2` enum constant)

eladb: Is there a naming convention for Go consts? We should follow it if there is one.

I am wondering if perhaps we can just block something like `LaunchTypePretendThisExists` at the `jsii` compiler level/

As for underscores, many of our enums have underscores in them, so I think this is actually a pretty major reason not to do that. Alternatively, maybe we can add an underscore between the enum name and the member (something like `LaunchType_PRETENT_THIS_EXISTS`).

text/204-golang-bindings.md
```diff
@@ -443,34 +453,180 @@ wrappers, e.g.
...
+The advantage of this approach is that adding a new optional property to a super struct no longer results in a breaking change (the field will be
+absent from the subtype until it is re-generated against the new parent), as the embedding technique guarantees methods are promoted from the
+supertype. It also guarantees immutability of built instances, as the actual implementation is not exported.
+
+The inconvenient is that this adds relatively heavy boilerplate around use of structs, where the name of the struct is repeated twice in sequence.
+This could lead to some user confusion (at least until the suer becomes familiar with this idiom). A typical call would look like so:
+
+```go
+// Pretending one can instantiate the BaseService construct directly
+NewBaseService(scope, "ID", NewBaseServiceProps(BaseServiceProps{
```
eladb, line 606: Not sure I fully understand why `NewBaseService` cannot just accept a `BaseServiceProps` struct directly? What's the use case for passing in `IBaseServiceProps`?

text/204-golang-bindings.md
```diff
@@ -443,34 +453,180 @@ wrappers, e.g.
...
-wrapper type for optional values.
-* Generated Go interfaces corresponding to a datatype interface would need a suffix, e.g. Iface, in order to disambiguate it from the struct. This is
-  a bit verbose, and it may be worth considering switching the naming (i.e. adding a suffix to the struct instead) if the interface name is what will
-primarily be used by the customer.
+  wrapper type for optional values.
+* Generated Go interfaces corresponding to a datatype interface would need a name addendum, e.g. I prefix of Iface suffix, in order to disambiguate
+  it from the struct name. This can be confusing (I prefix makes structs and behavioral interfaces similarly named on APIs) or a bit verbose (Iface
+  suffix), and it may be worth considering switching the naming (i.e. adding a suffix to the struct instead), as the interface name is what will
+  primarily be used by the customer.
+* As the `@jsii/kernel` may decide to pass a data type by-reference (and not by-value), either a concrete struct needs to be implemented that does
```
eladb, line 621: what's the use case for the kernel to pass data by-ref, isn't the whole point of data structs that they are always passed by value?

text/204-golang-bindings.md
```diff
@@ -524,39 +680,57 @@ In Go: ([Go playground example](https://play.golang.org/p/T20xlddRo6A))
...
+func (g *greeter) Greet() (result string) {
+    // Invoking the method in the JS process
+    jsii.Invoke(g, "greet", []interface{}{}, &result)
+    return
 }
 
 // static method
-func GreeterFoo() string {
-    return fmt.Sprintf("foo");
+func GreeterFoo() (result string) {
```
eladb, line 721: How do we disambiguate this from a class named `GreeterFoo` (same problem we have with enums)?

eladb (commented):
Overall looks much better from the DevX point of view. I am not sure I fully understand the convoluted setup around data types. I'll follow up face to face

text/204-golang-bindings.md
RomainMuller: I guess we could name it `Greeter_Foo` here instead of `GreeterFoo`... This isn't super pretty but is probably the safest bet.

I actually haven't spent all that much time thinking about this case as part of this change.

text/204-golang-bindings.md
RomainMuller: The by-ref behavior is actually expected/used in Java & C# in the current state of things, IIRC to work around serialisation issues that otherwise arised. The issue down the line is that changing this now would likely be risky, and likely a large-is effort. It'll likely be much easier to deal with this quirk, at least in the short term.

text/204-golang-bindings.md
```diff
@@ -443,34 +458,266 @@ wrappers, e.g.
...
+    PropagateTags          PropagatedTagSource
+    EnableECSManagedTags   *bool
+    DeploymentController   DeploymentController
+
+    // Fields introduced by BaseServiceProps
+    LaunchType          LaunchType
+}
+
+// Optional - we can reduce boilerplate for conversion of BaseServiceProps -> BaseServiceOptions:
+func (b BaseServiceProps) ToBaseServiceOptions() BaseServiceOptions {
```
eladb, line 669: Nice touch

text/204-golang-bindings.md
RomainMuller: This is either *Approach 1* or *Approach 2* you're referring to. Both have caveats that cause non-breaking changes in TS to become breaking in Go.

But as we discussed, I've introduced *Approach 4*, which does not have these problems and will likely be more go-ey. It simply rids of the interfaces altogether, and uses the structs strictly by-value.

eladb (commented):
👍

text/204-golang-bindings.md
RomainMuller: Could not find a convention for constant naming - and [effective go](https://golang.org/doc/effective_go#constants) actually has examples in varying case formats... 

I've updated the PR with an `_` including example, and standardised on `TypeName_MEMBER_NAME`.

eladb: Sounds good. I think we can also standardize static methods/fields with an underscore between the type name and the member because we want them to be "enum-like". It's common to implement enums using public static fields.

text/204-golang-bindings.md
eladb: See my comment about enums. I think we want statics to look like enum members so underscore is the way to go.

MrArnoldPalmer (approved):

PR #297, opened 2021-03-11
chore: go optional values codegen

MrArnoldPalmer:
Adds section on optionality for generated go code.

---

### DIFF
--- a/text/204-golang-bindings.md
+++ b/text/204-golang-bindings.md
@@ -128,8 +128,8 @@ how typescript interfaces use extends.
 
 Interfaces can be used as fields on structs as way to encapsulate behavior and facilitate testing.
 
-See [optional values and pointer types](#optional-values-and-pointer-types) for more information about the usage of pointer to literals the usage of related helper
-functions.
+See [optional values and pointer types](#optional-values-and-pointer-types) for more information about the usage of pointer to literals the usage of
+related helper functions.
 
 ### Typescript Interfaces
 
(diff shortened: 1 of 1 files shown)


text/204-golang-bindings.md
```diff
@@ -1027,6 +1030,187 @@ func (c *cluster) HasEc2Capacity() (result bool) {
...
+```
+
+Essentially every type that the user passes to or receives from the JSII runtime will have to be a pointer type. We can't only make
+optional values pointers because then when a required value became optional in TS, this causes a type change in the generated go code which requires a code
+change. This means many non-breaking changes to the TS source would require major version bumps to the generated go modules.
+
+With these downsides, it still seems like this is the most idiomatic route to take. Go users are familiar with using pointer types in this way from other
+popular modules. Additionally, it is the only strategy we can identify that allows us to maintain JSII's versioning compatibility strategy.
+
+### Explored Alternatives:
```
eladb, line 1158: Is there any well-known established Go module that handles null primitives? Maybe leaning on something like that makes sense and then we don't need explicit pointers everywhere, no?

I found this one which looks pretty solid: https://github.com/guregu/null

It's ergonomics are not that bad.

eladb (dismissed):

text/204-golang-bindings.md
MrArnoldPalmer: Synced offline with the team about this and agreed that the ergonomics look very similarly with wrapper types, via a library or our own, as they do with pointers. Sticking with pointers as the recommendation as this seems to be the precedent in other libraries (gh api and aws-sdk notably) and is more what users expect. Leaving the option open to re-evaluate this in the future based on user feedback. Especially should be re-visited once generics are available and we can provide generalized wrapper types.

eladb (approved):

PR #307, opened 2021-03-25
RFC 6: Record decisions about shared packages

iliapolo:
Update v2 RFC with the approach we decided on for handling packages that are shared between the CLI and the Framework.

---


iliapolo:
@rix0rrr It seems like all these decisions have already been implemented right?`ubergen` already bundles all of these, and `yaml-cfn` has been [converted](https://github.com/aws/aws-cdk/pull/13699).

eladb (approved):
Looks good to me!

Issue #309, opened 2021-04-06
Parameter Store for cross stack references


corrjo:
## Description

(see:  https://github.com/aws/aws-cdk/discussions/13184, https://github.com/aws/aws-cdk/pull/13804)

Cloudformation exports work really well for consuming properties of long lived resources, like vpc or subnet ids, however when you have a cdk app with stacks and resources that have variable lifecycles, using exports can lead to a rough time with stacks or resources being locked because something is consuming its export leading to having to do something like https://www.endoflineblog.com/cdk-tips-03-how-to-unblock-cross-stack-references

How I handled this in CFN
Before using cdk, this was an issue that would pop up on occasion and this led me and my team to make broader use of nested stacks, and parameter store.

Nested stacks worked well enough, but it ended up requiring a lot of planning and overhead since passing values between nested stacks at the same level or more than 1 level deep becomes a bit convoluted, hence parameter store to the rescue.

Creating standardized parameter store paths for values based on the app or service we were deploying and consuming those cross stack using the native cloudformation parameter store parameter type gave us a lot of flexibility, allowed us to update resources as we needed and run the stack update on the consuming stacks to pick up new values

I believe using parameter store as the default method for sharing values across different stacks in the same app makes makes the most sense in cdk.

## Working Backwards

When passing objects or values between stacks, instead of the producer stack creating a cfn export, it creates a ssm parameter store parameter. The consuming stack then consumes that parameter and the value is resolved with a !Ref 

## Roles

* **Driver** (drives the proposal to completion): @corrjo
* **Approver(s)**: (assigned by CDK team)

jstag711:
Is this just bypassing the safety mechanism that CFN offers? If I have a Lambda in Stack A that imports an IAM Role from Stack B and sticks it in some env variable, even if you dynamically update the reference in the stack defnition, my active lambdas will still be using the old value and will fail assuming you deleted the old role.

Looking at https://www.endoflineblog.com/cdk-tips-03-how-to-unblock-cross-stack-references, isn't another option to cdk deploy --exclusively {consuming stack} to remove the reference, then proceed with the update of the rest of the stacks?

mrgrain:
Duplicate of https://github.com/aws/aws-cdk-rfcs/issues/82

PR #311, opened 2021-04-08
RFC 309: Loosely Coupled Cross Stack Ref

corrjo:
* Tracking issue: #309 
* [Rendered version](https://github.com/corrjo/aws-cdk-rfcs/blob/master/text/0309-parameter-cross-stack-ref.md)

---

### DIFF
--- a/text/0309-parameter-cross-stack-ref.md
+++ b/text/0309-parameter-cross-stack-ref.md
@@ -50,9 +50,9 @@ Prevent stacks being export locked and unable to update
 
 Exports lock resources for a reason. In some cases exports may be preferred.
 
-For example: 
-- Resources may be temporarily unavailable while being replaced causing 
-errors for downstream consumers. 
+For example:
+
+- Resources may be temporarily unavailable while being replaced causing errors for downstream consumers.
 
 - If a stack outside of the app consumes the parameter, there is no tracking, so if the provider
 stack is torn down, other stacks may unexpectedly break.
--- a/text/0309-parameter-cross-stack-ref.md
+++ b/text/0309-parameter-cross-stack-ref.md
@@ -1,0 +1,82 @@
+---
+rfc pr: [#xxx](https://github.com/aws/aws-cdk-rfcs/pull/xxx) <-- fill this after you've already created the PR
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/309
+---
+
+# [Loosely Coupled Cross Stack Ref]
+
+Use Parameter Store parameters for cross stack references
+
+## Working Backwards
+
+When passing object or class properties between stacks, instead of creating a CFN Export,
+Create a SSM Parameter in parameter store, then have the consuming stack create a parameter
+to consume the value from parameter store
+
+### CHANGELOG
+
+feat(core): loosely coupled cross stack references
+
+BREAKING CHANGE: Stacks will need to decouple their exports with dummy values
+
+### README
+
+#### Loosely Coupled Stack References
+
+By default when you pass a Stack property to another Stack cdk uses cloudformation exports and Fn::ImportValue to share
+the value cross stack. Adding `'@aws-cdk/core:looseCrossStackRefs'` to your cdk.json changes this behavior
+so that parameter store is used for storing and retrieving values across stacks
+
+## FAQ
+
+### What are we launching today?
+
+A new feature in core which enables stacks to be loosely coupled but maintain high cohesion by moving to parameters
+
+### Why should I use this feature?
+
+Creating standardized parameter store paths for values based on the stack we are deploying then consuming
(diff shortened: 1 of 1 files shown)


corrjo:
Re: 
https://github.com/aws/aws-cdk/pull/13804 
https://github.com/aws/aws-cdk/discussions/13184

text/0309-parameter-cross-stack-ref.md
```diff
@@ -0,0 +1,82 @@
...
+## Internal FAQ
+
+
+### Why are we doing this?
+
+Prevent stacks being export locked and unable to update
+
+### Why should we _not_ do this?
+
+Exports lock resources for a reason. In some cases exports may be preferred
```
michaelfecher, line 51: What are those cases? Examples make it better understandable.

text/0309-parameter-cross-stack-ref.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Is this a breaking change?
+
+Yes
+
+
+### What are the drawbacks of this solution?
+
+Tight coupling means the resources you are dependent on wont go away, in some instances
+this behavior prevents breaking consumers.
```
michaelfecher, line 64: How do you handle updates? You would need to notify the consumer stacks, right?

corrjo: I'm currently creating a dependency on the provider stack https://github.com/aws/aws-cdk/pull/13804/files#diff-d21961a629629b87c50d664aaf41594608e31d6ae693a2fc586062cdb3403eeeR175

The idea is when the provider is updated, it'll publish the new value and then cdk triggers a stack update on the dependent stacks and thats where they'll resolve the new parameter value.

But perhaps that dependency isn't enough to force an update of the consumer stack after the provide has been updated? I'll have to verify that it works as expected.

henrist: I've put some comments in https://github.com/aws/aws-cdk/discussions/13184#discussioncomment-611215 that can be related to this

text/0309-parameter-cross-stack-ref.md
```diff
@@ -0,0 +1,87 @@
...
+
+### CHANGELOG
+
+feat(core): loosely coupled cross stack references
+
+BREAKING CHANGE: Stacks will need to decouple their exports with dummy values
+
+### README
+
+#### Loosely Coupled Stack References
```
rix0rrr, line 24: I think I'd prefer the term "**weak** cross-stack references"

rix0rrr (commented):
Thanks for writing up a proposal!

Some things I'd like to see this RFC address:

* What the use cases we're unlocking? What are the problems we're addressing? (Not hand-wavey "things could happen", bus describe real, concrete cases please)
* Why are the current solutions not good enough? (the `stack.exportValue()` method currently)
* WHY is it okay to drop the hard-reference guarantees that exports provide? Either, convince the reader it's okay to take downtime while resources are being replaced, or otherwise, what are the assumptions you're making which will cause that to not occur in practice?
* Do we ever want to support having both types of references in one stack, but pin-pointed for different types of resources? I think yes, what would that look like? Honestly, we can reuse context, but we don't check context at the stack level but at the *producer construct*, and that would probably do it...?
* What does the migration path from hard references to weak references look like? You can't just switch over the flag because CFN won't let you deploy that. I think the flag needs to be tri-state: `'hard' | 'weak' | 'both'`, where you use `both` in the cutover period to still produce the required exports, but only consume the weak references.

text/0309-parameter-cross-stack-ref.md
ahammond: @rix0rrr https://www.endoflineblog.com/cdk-tips-03-how-to-unblock-cross-stack-references describes the workaround necessary to delete a resource that's exported from one stack and consumed by another using the tightly-coupled approach. Moving to loosely-coupled puts the responsibility on the shoulders of the user, which, I think is reasonable, or at least it's an option I'd use if it were available.

I can't think of any case where the strong guarantees of export are enough of a benefit that I'd actually pick them over params. I would absolutely LOVE to be _able_ to delete a stack which other stacks depend on. Because while it may sound like the wrong decisions and may even be the wrong decision most of the time, there are times when I could imagine it would be the best of a bunch of bad options.

I really like your observation about `both` mode though. I thing that the export model really needs to find a way to "still produce the required exports" in the case where resources are being deleted whether they're being replaced by params or just refactored out of the code in general.

nija-at:
RFC Bar Raiser: @nija-at

nija-at (changes requested):
Setting this to 'request changes' for comments to be addressed.

nija-at:
Closing this RFC for now. Please re-open or create a new one if anyone is interested in picking this up.

PR #314, opened 2021-04-13
RFC 324: Construct Hub

RomainMuller:
* Tracking issue: #324 
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/rmuller/cdk-construct-hub/text/0314-cdk-construct-hub.md)

---


text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+# CDK Construct Hub
+
+A community-driven hub for discovering and sharing CDK constructs.
+
+## Working Backwards
+
+### PRESS RELEASE
+
+Today, Amazon Web Services Inc., an Amazon company, announced the CDK Construct
+Hub, a service that makes it easy to find AWS Cloud Development Kit (CDK)
```
eladb, line 15: General note: one of the goals of the construct hub is to be home to constructs for all CDK domains

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+following prerequisites:
+
+1. Cross-language support has been configured correctly using [jsii]
+1. The `cdk` keyword is referenced (in the `keywords` key of `package.json`)
+
+[jsii]: https://aws.github.io/jsii
+
+### Can I opt packages out of the public instance of CDK Construct Hub?
+
+There is currently no self-service way to un-list a package from the CDK
```
eladb, line 92: Why? We should be able to easily support this through an additional label

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+
+> [Example rendering](#rendering-package-detail-page)
+
+Clicking on a given package's tile in the landing page or search result page
+brings the user to a package detail page that represents the following
+information:
+
+- Package name, tags, and description
+- Overview of publishing information (author, version, license, etc...)
+  - Where possible, links to the source repository, issue tracker, etc...
```
eladb, line 134: Links to all package managers

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+- Install instructions
+- Available languages
+- Package's documentation
+  - Rendered content of the package's `README.md`
+  - API reference documentation, in all supported languages
```
eladb, line 139: Language selection should not only impact API reference documentation but also the README, installation instructions, etc. So we wanted this to be a "global" language switcher.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+
+The list of available languages is also a selector: upon clicking a language,
+the installation instructions and documentation automatically switch to the
+relevant form for the currently selected language.
+
+### Support
+
+The CDK Construct Hub offers ways for customers to engage maintainers about
```
eladb, line 147: How are we planning to implement this? Is this based on the jsii author metadata?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+indexed and presented on the application particularly quickly. A consistency
+window between a couple of hours and a day is probably acceptable.
+
+A simple data pipeline will ingest package publication events created by a
+source-specific adapter (e.g: by polling on the CouchDB instance replica for the
+[npmjs.com registry](https://skimdb.npmjs.com/registry)), and produce static
+artifacts the website is based off of. This pipeline is in particular
+responsible for extracting the `.jsii` assembly file from the npm packages, and
+to prefetch all information necessary to back the webiste's search and package
+detail pages.
```
eladb, line 176: ~One thing I am not completely clear about is when is rosetta executed to translate samples in README/reference?~

Answered below.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+#### Back-End
+
+1. The back-end for CDK Construct Hub starts with a custom-built event source,
+   which implements the logic necessary to detect new packages that are relevant
+   to the CDK Construct Hub, and sends messages to an SQS queue for further
+   processing. Notifications have the following attributes:
+
+   Name        | Description
+   ------------|----------------------------------------------------------------
+   `assembly`  | The `.jsii` assembly included in the package
```
eladb, line 187: The assembly can be quite big. Will this be an issue?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+   This transformation is backed by the `jsii-rosetta` tool, which is part of
+   the [jsii project][jsii].
+
+   [jsii]: https://aws.github.io/jsii
+
+1. A Lambda function keeps the `latest.json` object updated with the last 20
+   package versions indexed (according to the `time` field reported by the
+   ingestion component).
+
+1. A Lambda function keeps the `catalog.json` object updated with the latest
```
eladb, line 223: How are we planning to handle concurrency? Are these updates serialized through a queue?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+
+   Name        | Description
+   ------------|----------------------------------------------------------------
+   `assembly`  | The `.jsii` assembly included in the package
+   `time`      | The timestamp at which the version was created
+   `integrity` | An integirity check for the complete record
+
+1. A Lambda function then picks up messages from the SQS queue and prepares the
+   artifacts consumed by the front-end application, stored in a dedicated S3
+   bucket using the following key format:
```
eladb, line 193: Is this bucket also fronted by CloudFront? I believe the prototype does that.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+
+   - If a different language (than JavaScript) is selected, the
+     `assembly-${lang}.json` object is fetched instead, providing the relevant
+     documentation elements.
+
+[react]: https://reactjs.org
+
+### Are there any open issues that need to be addressed later?
+
+- When the catalog becomes too large to be convenient to process in-memory on
```
eladb, line 254: Should we show some data based on the current catalog size? See [packages.json](https://awscdk.io/index/packages.json)

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+   Name        | Description
+   ------------|----------------------------------------------------------------
+   `v${major}` | All versions part of this major version stream, and their status
+
+   [SemVer]: https://semver.org/spec/v2.0.0.html
+
+1. A series of Lambda functions prepare language-specific assembly files for
+   each configured language, with adjusted naming conventions, and updated
+   sample code fragments, and proceeds to store those at:
+   `assemblies/${assembly.name}/v${assembly.version}/assembly-${lang}.json`.
```
eladb, line 213: How are we handling eventual consistency here? Will a package be available in the hub only after all its languages have been rendered?

text/0314-cdk-construct-hub.md
RomainMuller: But you cannot add a label to an _already_ published version (those are immutable)... The corollary is if _any_ version of a package has the label, then _all_ versions of the package are opted out, forever... This does not allow library authors to change their minds in the future...

text/0314-cdk-construct-hub.md
RomainMuller: Yeah based on information included in the jsii assembly (bug tracker URL, repository URL, homepage, etc...)

text/0314-cdk-construct-hub.md
RomainMuller: I guess it may be necessary to stage it into S3 in certain extreme cases... Or compression? Both?

text/0314-cdk-construct-hub.md
RomainMuller: Yes - the app would be 100% served by CloudFront.

text/0314-cdk-construct-hub.md
RomainMuller: I was thinking about showing a "please come again in a few minutes" kind of message in this case.

text/0314-cdk-construct-hub.md
eladb: We can decide that the labels on the latest version will determine if the module appears or not but we can leave this detail for later.

text/0314-cdk-construct-hub.md
eladb: Since we need the .jsii file in S3 anyway (this is the primary source of the UI), I think a reference to it should be fine.

eladb (commented):
I was missing the discussion about Construct Hub as a construct.

Specifically we should discuss in advance:
- Monitoring
- Input streams
- Showing both public & internal modules

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,278 @@
...
+
+# CDK Construct Hub
+
+A community-driven hub for discovering and sharing CDK constructs.
+
+## Working Backwards
+
+### PRESS RELEASE
+
+Today, Amazon Web Services Inc., an Amazon company, announced the CDK Construct
```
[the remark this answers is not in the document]
pgollucci: Would this also serve things like projen constructs or eventually Armkit ?

text/0314-cdk-construct-hub.md
pgollucci: yes please this is already fractured

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,310 @@
...
+constructs into their application.
+
+Customers can also deploy a private instance of Construct Hub to facilitate
+sharing infrastructure abstractions published to a private package registry
+within the organization.
+
+### Where can I find the Construct Hub?
+
+The public instance of Construct Hub is at: `${TBD}`.
+
```
pgollucci, line 78: duplicated above

text/0314-cdk-construct-hub.md
pgollucci: define configured correctly? like its in package.json or it actually working? Or its on ALL languages or 1

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,310 @@
...
+## Specification
+
+### Landing Page
+
+> [Example rendering](#rendering-landing-page)
+
+The landing page for the Construct Hub includes a search bar that allows
+developers to search for constructs that solve their use-case using relevant
+keywords, and optionally filtering on a specific programming language and CDK
+flavor (AWS CDK / CDK8s / CDK-Tf).
```
pgollucci, line 113: CDKtf not CFKtf and projen?

text/0314-cdk-construct-hub.md
pgollucci: That's going to get nosy as time goes on....

text/0314-cdk-construct-hub.md
pgollucci: I like the reference idea but I don't really think it will matter much.  Even Oracle can do JSON well now.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,310 @@
...
+   ingestion component).
+
+1. A Lambda function keeps the `catalog.json` object updated with the latest
+   versions of each package's major version lines, which backs the websites'
+   search page. The object may be sharded if it becomes too large.
+
+#### Front-End
+
+The front-end consists in a simple [React] application hosted in an S3 bucket.
+A CloudFront distribution fronts this application, ensuring it is served from
```
pgollucci, line 234: Nextjs, Tailwind, TS or JS ?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,310 @@
...
+necessary components of the application. The public instance of CDK Construct
+Hub will simply be an AWS-Managed instance of that, which feeds from packages
+published to the [npm registry](https://npmjs.com).
+
+It will expose a simple API: a `ConstructHub` class will be the main entry point
+to the application, with the following configuration properties:
+
+Name         | Description
+-------------|------------------------------------------------------------------
+`dsnName`    | The DNS name to use for hosting the Construct Hub instance
```
pgollucci, line 269: dsnName -> dnsName

text/0314-cdk-construct-hub.md
RomainMuller: Right up, but this is actually a relatively standard thing to put in the FAQ for some reason.

text/0314-cdk-construct-hub.md
RomainMuller: I'm getting into *some* aspects of technical decisions, but trying to avoid closing too many doors this early. I am here to describe what will be done, not how it will be done (at least - implementors get to have some freedom of movement).

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
RomainMuller: It could for sure. Maybe we wanna rephrase to avoid CDK and use just Constructs? I don't know... it's an open question for now I guess...

text/0314-cdk-construct-hub.md
RomainMuller: I tried to rephrase most of it to "distance" from CDK (and related) to open up to "general" constructs as much as possible.

text/0314-cdk-construct-hub.md
RomainMuller: It's mostly an implementation detail, I think.

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: Let's just use constructs for now

text/0314-cdk-construct-hub.md
eladb: Sounds reasonable

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+If IP addresses are fed into the analyitics service, these will not be
+persisted, but may instead be mapped to a coarse-grained location (e.g:
+country).
+
+Legal review will be necessary to determine if the dataset implies compliance
+requirements for GDPR, CCPA, and similar consumer privacy regulations.
+
+#### Construct Packaging
+
+The Construct Hub is to be created as a reusable construct, including all
```
NetaNir, line 325: ## How much does it cost to operate a ConstructHub?

Since we are going to offer ConstructHub as a reusable construct for customers to operate their own instance, we need to have a mechanism to measure the cost of operating a ConstructHub.  Before committing to operate their own version of ConstructHub, customers will want to know it costs as a dependency of different factors such as number of packages ingested, the rate in which packages are ingested and more.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+
+   Name        | Description
+   ------------|----------------------------------------------------------------
+   `assembly`  | The `.jsii` assembly included in the package
+   `time`      | The timestamp at which the version was created
+   `integrity` | An integirity check for the complete record
+
+   In cases where the `assembly` object is too large, it may be staged in S3
+   instead of being sent directly as part of the message payload. In such cases,
+   the `assembly` field contains an S3 URI instead of a JSON object.
```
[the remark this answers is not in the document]
RomainMuller: Yeah this is a good point. I was talking w/ Addi about it, and it is preferable to have only 1 code path -- the one that stages in S3. I'll update the doc to reflect that.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+   [SemVer]: https://semver.org/spec/v2.0.0.html
+
+   In order to ensure consistency of the contents of the `versions.json` object,
+   its content will always be generated from the live inventory of objects in S3
+   (and not incrementally using the update trigger event), and the last event
+   timestamp will be written in an object metadata attribute. The S3 Bucket will
+   have object versioning enabled, and once a new version of the `versions.json`
+   object has been created, its metadata will be checked against any previous
+   versions of the object, and the "current" version will be set to the one with
+   the latest event timestamp if needed.
```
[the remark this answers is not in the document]
RomainMuller: At some point I want to allow users to view any version of the package they want, and I'd rather avoid having to resort to an S3 List operation if possible. This file can also serve as a tool to diagnose system operations (what versions have been processed, etc....).

It is however not a priority for the MVP/DevPreview, so it could be considered a "nice-to-have". I'd kinda like to keep it in there however... Unless we all agree we don't want to design/scope this until we know we require it.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+
+A simple data pipeline will ingest package publication events created by a
+source-specific adapter (e.g: by polling on the CouchDB instance replica for the
+[npmjs.com registry](https://skimdb.npmjs.com/registry)), and produce static
+artifacts the website is based off of. This pipeline is in particular
+responsible for extracting the `.jsii` assembly file from the npm packages, and
+to prefetch all information necessary to back the webiste's search and package
+detail pages.
+
+#### Back-End
```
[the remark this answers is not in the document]
RomainMuller: This is actually a very interesting suggestion! I had not considered it this way because of my past experience with a similar data pipeline, at a time when StepFunctions lacked native integrations for most of these touch points... But this is no longer the case and it could definitely be a significant improvement in operability.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+   repositories, etc.
+
+1. **Ingestion:** A Lambda function then picks up messages from the SQS queue
+   and prepares the artifacts consumed by the front-end application, stored in a
+   dedicated S3 bucket using the following key format:
+   `assemblies/${assembly.name}/v${assembly.version}/assembly.json`
+
+   This function validates the contents of the `assembly` to ensure the message
+   was well-formed. In case the message is found to be incoherent, it is sent to
+   a dead-letter queue and an alarm is triggered.
```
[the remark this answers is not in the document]
RomainMuller: The goal of this safeguard is to ensure the discovery function did not send a malformed payload into the system. The discovery function can be seen as an integration component (an operator for a private instance could provide their own, and they could make mistakes).

We want to ensure that whenever data enters the system, it has been validated.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+   have object versioning enabled, and once a new version of the `versions.json`
+   object has been created, its metadata will be checked against any previous
+   versions of the object, and the "current" version will be set to the one with
+   the latest event timestamp if needed.
+
+   The operator of private instances can provide an optional SNS topic which
+   will be notified once the information for a new package version is ready to
+   be browsed.
+
+1. **Doc-Gen:** A series of Lambda functions prepare language-specific assembly
```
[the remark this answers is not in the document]
RomainMuller: Regarding the 15 minutes limit, this is indeed a risk. This is largely why I've specified that one lambda execution will happen per language, so we hedge that risk. There are several ways around this however:

- Move this to ECS
- Optimise the `jsii-rosetta` codebase for speed (some of it is slower than it needs to be)
- Actually perform this step at package build time, through a `jsii` change
- ...

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+counts, etc. No cross-site tracking will be performed.
+
+If IP addresses are fed into the analyitics service, these will not be
+persisted, but may instead be mapped to a coarse-grained location (e.g:
+country).
+
+Legal review will be necessary to determine if the dataset implies compliance
+requirements for GDPR, CCPA, and similar consumer privacy regulations.
+
+#### Construct Packaging
```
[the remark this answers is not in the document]
RomainMuller: Yes - this is largely for illustration purposes, and to give a sense of the general direction. This isn't meant to be prescriptive at this stage (mostly it can serve as a way to identify minimal requirements).

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+It exposes the following attributes, to allow integrations to operate correctly:
+
+Name                | Description
+--------------------|-------------------------------------------------------------------------------
+`ingestionQueueUrl` | The URL of the SQS queue where ingestion messages should be sent
+`ingestionRoleArn`  | The IAM role used to process input payloads (and read S3 staged objects)
+`ingestionDlqArn`   | The ARN of the ingestion pipeline's dead letter queue
+
+#### Monitoring & Operations
+
```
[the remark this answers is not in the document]
RomainMuller: That's a good idea. It'll definitely help ensure we build a platform that is immediately designed for operability.

text/0314-cdk-construct-hub.md
RomainMuller: Interesting! There are different aspects to cost though (there is infrastructure cost, operational overhead, etc...). It may be difficult to get a sense of these costs before we actually operate our own hub for a short while. I'll nevertheless add a mention to this in the document, because this will eventually be important information!

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+ownership of the package(s) you are requesting un-listing for.
+
+### Can I trust all constructs listed on Construct Hub?
+
+Construct Hub does not perform a comprehensive audit of packages it indexes.
+You should not trust packages found via the Construct Hub any differently
+from packages found via other package search tools or registries, such as
+npmjs.com, Maven Central, PyPI, NuGet, etc.
+
+## Specification
```
eladb, line 104: @RomainMuller, I think we need to expand this section with an exact list of features we plan to deliver for each page (and also ideally some form of a roadmap for future versions). @addihorowitz has most of this documented internally, but it should be part of this document.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+
+### What is the high level implementation plan?
+
+#### Overview
+
+The Construct Hub is a relatively simple static web application. While the
+dataset it exposes (all Construct Libraries) may be relatively fast moving,
+there is no necessity for newly published packages (or versions thereof) to be
+indexed and presented on the application particularly quickly. A consistency
+window between a couple of hours and a day is probably acceptable.
```
eladb, line 177: I think we need a tighter SLA here. From our experience with the construct catalog prototype, publishers are eager to see their packages "in lights" appear in the construct hub. Especially if there is a great documentation experience involved.

Let's target a 10min SLA (from time the package is published to npm to the time it appears in the hub). We should make sure we have some provisions to measure this SLA (npm includes information about publish time).

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+artifacts the website is based off of. This pipeline is in particular
+responsible for extracting the `.jsii` assembly file from the npm packages, and
+to prefetch all information necessary to back the webiste's search and package
+detail pages.
+
+#### Back-End
+
+1. **Package Discovery:** The back-end for Construct Hub starts with a custom
+   built event source, which implements the logic necessary to detect new
+   packages that are relevant to the Construct Hub, and sends messages to an SQS
```
eladb, line 191: I think we should add some form of early filtering at the ingestion point. For example, we should skip any versions that we have already processed at this stage instead of queueing them, right?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+   ------------|----------------------------------------------------------------
+   `assembly`  | The `.jsii` assembly included in the package
+   `time`      | The timestamp at which the version was created
+   `integrity` | An integirity check for the complete record
+
+   In cases where the `assembly` object is too large, it may be staged in S3
+   instead of being sent directly as part of the message payload. In such cases,
+   the `assembly` field contains an S3 URI instead of a JSON object.
+
+   Construct Hub comes with a built-in integration function that listents to
```
eladb, line 207: I am assuming we are planning to simply query npm and filter versions we've already seen, right?

```suggestion
   Construct Hub comes with a built-in integration function that listens to
```

eladb: What's the protocol of the package source? In other words, what would I need to do in order to implement my own source? Do I just need to push messages to a queue? What do these messages contain?

text/0314-cdk-construct-hub.md
eladb: I think it's safe to assume that some .jsii manifests cannot fit into an SQS message (how big is the aws-cdk-lib manifest?).

Would it make more sense for this stage to just enqueue a message that contains the URL of the tarball and have the rest of the pipeline handle downloading the tarball and extracting the .jsii manifest from it?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,626 @@
...
+- Package name, tags, and description
+- Overview of publishing information (author, version, license, etc...)
+  - Where possible, links to the source repository, issue tracker, etc... based
+    on information registered in the `.jsii` assembly file.
+- Install instructions
+- Available languages
+- Package's documentation
+  - Rendered content of the package's `README.md`
+  - API reference documentation, in all supported languages (only the version
+    for the currently selected language is displayed)
```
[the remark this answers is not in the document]
eladb: Another roadmap idea: "virtual construct libraries" for L1s. For example, one idea that came up is to automatically generate a virtual construct library for each Terraform provider from the Terraform registry and publish them to the hub with instructions on how to use `cdktf import` in order to generate their constructs. Same for k8s CRDs and CFN registry resources. This might be a great way to seed the hub with tons of great and useful content.

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: Worth considering, but also might be a bit of an overkill.

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: +1 on making sure we don't expose ourselves to foreign code execution. @NetaNir I am assuming this is covered in the threat analysis, but let's make sure we have some tickets to ensure that we never execute foreign code on the backend.

As for aws-cdk-lib size, we should be able to find data about this, no? I am comfortable not optimizing for this too early (e.g. use Lambda initially) as long as we did _some_ sanity check. We need to support aws-cdk-lib in day one, so if it already know that we have an issue, then a different compute solution might be in order.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+     version will be set to the one with the latest event timestamp if needed.
+
+   The operator of private instances can provide an optional SNS topic which
+   will be notified once the information for a new package version is ready to
+   be browsed.
+
+1. **Doc-Gen:** A series of Lambda functions prepare language-specific assembly
+   files for each configured language, with adjusted naming conventions, and
+   updated sample code fragments, and proceeds to store those at:
+   `assemblies/${assembly.name}/v${assembly.version}/assembly-${lang}.json`.
```
eladb, line 250: I don't like the `assemblies` term here. From an external perspective. This is just the metadata of a library/module/package in the construct hub.

```suggestion
   `assemblies/${assembly.name}/v${assembly.version}/assembly-${lang}.json`.
```

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+   and performing the necessary corrections if needed.
+
+1. **Catalog Builder:** A Lambda function keeps the `catalog.json` object
+   updated with the latest versions of each package's major version lines, which
+   backs the websites' search page. The object may be sharded if it becomes too
+   large. S3 Object Versioning will be used to ensure concurrent updates to the
+   content result in a consistent end state, by listing object versions before
+   and after updating the `catalog.json` object and performing the necessary
+   corrections if needed.
+
```
eladb, line 275: Please include the public URL schema here (basically that's our backend API).

If I recall we discussed that the backend URLs will generally derive from the frontend URLs with an `api.` subdomain or something like that:

Get the JSII manifest of the latest version of a module:

```
GET https://api.constructs.dev/packages/packages[/@<scope>]/<name>
```

Get the JSII manifest of a specific version:

```
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>
```

Go do languages come into play here?

```
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/python
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/java
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/go
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/typescript
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/javascript
GET https://api.constructs.dev/packages/@<scope>]/<name>/v/<version>/csharp
```

(not 100% sure, but something along those lines).

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+   URL                                                       | Description
+   ----------------------------------------------------------|----------------------------------------------------------
+   `<root>/packages[/@<scope>]/<name>`                       | Redirects to the latest known version of `[@scope/]name`
+   `<root>/packages[/@<scope>]/<name>/v/<version>`           | The overview of `[@scope/]name` at `version`
+   `<root>/packages[/@<scope>]/<name>/v/<version>/docs`      | The documentation (`README.md`) of `[@scope/]name` at `version`
+   `<root>/packages[/@<scope>]/<name>/v/<version>/docs/<FQN>`| The documentation for type `FQN` in `[@scope/]name` at `version`
+
+   - When a `<version>` parameter is only a prefix of a SemVer version number,
+     the URL redirects to the latest known version in that line (e.g: `/v/1`
+     would redirect to `/v/1.2.3`, and `/v/1.1` would redirect to `/v/1.1.9`)
```
eladb, line 309: Sounds like a nice feature but not sure that's needed for v1

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+   - A query parameter (e.g: `?lang=go`) may be added to anchor a documentation
```
eladb, line 310: What happens if the language parameter is not specified?

text/0314-cdk-construct-hub.md
eladb: Super important and I think this is an opportunity to innovate and come up with some general model/tools for calculating costs of constructs

text/0314-cdk-construct-hub.md
eladb: The reason it's important is that this will help us make sure the design actually covers all the required features

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+   Construct Hub comes with a built-in integration function that listents to
+   changes on the [npmjs.com registry](https://skimdb.npmjs.com/registry)
+   database. Operators of private instances are able to interface with other
+   package registries (including private ones), such as CodeArtifact
+   repositories, etc.
+
+1. **Ingestion:** A Lambda function then picks up messages from the SQS queue
+   and prepares the artifacts consumed by the front-end application, stored in a
+   dedicated S3 bucket using the following key format:
+   `assemblies/${assembly.name}/v${assembly.version}/assembly.json`
```
eladb, line 216: I am wondering if perhaps we should provision for additional annotations that the backend may want to add. For example, the classification which CDK a module works with might be some logic that runs during package processing and we would want to add that to the response to the client.

So maybe this file should look like this:

```json
{
  "name": ...,
  "version": ...,
  "manifest": { .jsii manifest },
  ... // additional fields that backend adds if needed
}
```

Basically what I am proposing is to store the .jsii manifest inside an "envelope" that will allow us to add additional fields by the backend without having to tweak the jsii manifest itself.

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
RomainMuller: > A feature we should add to the readme is an "Edit" button in a package overview page and jump directly to the GitHub repository with that README.md file to propose an edit (bonus points if we can do something similar in reference docs, but that's borderline rocket science).

I don't think those are V1/MVP features. They also sound like wee UI elements that can be easily done later on.

---

> Another roadmap idea: "virtual construct libraries" for L1s. For example, one idea that came up is to automatically generate a virtual construct library for each Terraform provider from the Terraform registry and publish them to the hub with instructions on how to use `cdktf import` in order to generate their constructs. Same for k8s CRDs and CFN registry resources. This might be a great way to seed the hub with tons of great and useful content.

That sounds like a whole new set of features, with a very different source and data flow than what is currently discussed here. While that sounds like an interesting idea, I don't think it can fit in our aspirational timeline.

text/0314-cdk-construct-hub.md
RomainMuller: > I am assuming we are planning to simply query npm and filter versions we've already seen, right?

No, we are going to query the CouchDB replica's change stream, to discover newly published package versions (and only those).

---

> In other words, what would I need to do in order to implement my own source? Do I just need to push messages to a queue? What do these messages contain?

Yes, you publish to SQS, and the contents of the messages is what is described/specified in the relevant section of this document.

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
RomainMuller: The **Discovery** function is an integration point - it may be implemented external to the hub. If we're integrating with a 3P artefact repository (CodeArtifact, Artifactory, GH Packages, etc...), authentication may be needed to download the tarball. If the download was to happen in the **Ingestion** function (which is strictly internal to the hub), then we'd need to allow configuring authentication somehow... and I'd rather not.

The alternative is that the Discovery function creates a "hub-readable" copy of the tarball somewhere (e.g: S3) if the package isn't public... But then why are we duplicating that effort in the system?

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,651 @@
...
+
+   This function validates the contents of the `assembly` to ensure the message
+   was well-formed. In case the message is found to be incoherent, it is sent to
+   a dead-letter queue and an alarm is triggered.
+
+   - Optionally (this information may not be necessary, in particular for the
```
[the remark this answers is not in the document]
RomainMuller: Not sure this is a requirement for the MVP (worth discussing)

text/0314-cdk-construct-hub.md
RomainMuller: Those annotations can be stored in S3 object metadata if they are small enough. These can be manipulated without ever touching the manifest itself. Now if we foresee we'd possibly need more metadata than fits in S3 object metadata... then yeah, envelope is a better solution.

eladb: How will the client get the object metadata? Sounds like the wrong tool. All I am proposing is to return an object that contains the manifest as a field instead of the manifest directly, so we can later add additional fields to that object if we want. It's a common practice in protocol design...

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: Both of these ideas are roadmap (I.e future ideas). I don't expect us to support them in v1 but it's sometimes useful to list some future directions in the initial spec to inform the design.

text/0314-cdk-construct-hub.md
eladb: I am just looking for a way to identify filter construct libraries early, and wondered if a dependency on  CDK is a good signal

text/0314-cdk-construct-hub.md
eladb: > we are going to query the CouchDB replica's change stream, to discover newly published package versions (and only those).

Can you describe the protocol in the doc?  Does this mean we need to store our cursor state somewhere? What's the plan for backfilling existing data?

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: I understand your point about authentication but my concern is about resiliency and timeouts of the ingestion function. A single lambda invocation won't always be able to handle an entire batch of updates, so we will need some form of queuing or recovery or re-execution based on which items have been processed (downloaded, + jsii extracted + queued).

How do you see this addressed? I guess we can split the ingestion into two stages separated by a queue...

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
eladb: Feels like it should be there given we have taken a goal to support the aws CDK in the hub, and we have many users that don't use the latest version.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,665 @@
...
+
+## Technical Design
+
+### Overview
+
+![Architecture Diagram](../images/cdk-construct-hub/architecture-diagram.png)
+
+### What is the high level implementation plan?
+
+#### Overview
```
eladb, line 184: Worth adding some design tenets. One important one is that v1 will be solely based on cloudfront as the backend so there are no "write" routes on the server (not sure how to articulate it but it's a really important one).

Analytics will be implemented through a 3rd party (Adobe analytics)

text/0314-cdk-construct-hub.md
RomainMuller: Object metadata is part of the headers you get with the object data.

eladb:
@RomainMuller what's the status here? Can we announce final comments period?

text/0314-cdk-construct-hub.md
NetaNir: Can we add a place holder for this? I think its something we need to keep in mind as we develop, so it will be good to have a reminder and also for soliciting feedback from others.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,714 @@
...
+for, in the programming language they want to use.
+
+Construct Hub automatically indexes all construct libraries published to the
+various general purpose package managers, so that developers have a single place
+to search for constructs, regardless of preferred programming language or
+provisioning engine. Developers will find constructs published by AWS, partners,
+and individual developers. Constuct Hub also generates unified documentation for
+the constructs in all supported programming languages.
+
+Construct Hub is also available as a deploymable CDK construct library, so
```
ericzbeard, line 54: deployable

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,714 @@
...
+following prerequisites:
+
+1. Cross-language support has been configured correctly using [jsii]
+1. The `cdk` keyword is referenced (in the `keywords` key of `package.json`)
+
+[jsii]: https://aws.github.io/jsii
+
+### Can I opt packages out of the public instance of Construct Hub?
+
+There is currently no self-service way to un-list a package from the Construct
```
ericzbeard, line 92: Can we do this somehow with project metadata, like another keyword in package.json? That would guarantee that it was the project owner who decided to not list the package, or de-list it.

text/0314-cdk-construct-hub.md
```diff
@@ -0,0 +1,714 @@
...
+tracker URL, repository URL, and homepage URL.
+
+## Technical Design
+
+### Design Tenets
+
+- The initial version of Construct Hub is a static website
+  - It is served entirely by CloudFront
+  - All content served is strictly read-only
+- Search happens client-side, based off a single catalog object
```
ericzbeard, line 183: Have we calculated how big the catalog has to be to start causing issues with doing everything on the client? I'm not saying it shouldn't be static in v1, but we should at least know exactly when we need to convert it to API calls to a back end. Also do we know exactly how big the catalog will be on day 1? How many constructs are there?

text/0314-cdk-construct-hub.md
RomainMuller: Package metadata cannot be updated on already published versions, making it impossible to un-publish existing versions this way...

text/0314-cdk-construct-hub.md
RomainMuller: This is answered in open issues.

text/0314-cdk-construct-hub.md
RomainMuller: There is one already...

text/0314-cdk-construct-hub.md
RomainMuller: It might be. But then also - what are we trying to protect against?

text/0314-cdk-construct-hub.md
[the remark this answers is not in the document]
RomainMuller: >an entire batch of updates

The batch size is 1 for this function. It triggers off of an SQS queue (so queuing is built right in there).

iliapolo (approved):

PR #315, opened 2021-04-21
RFC 0192: update Construct.construct with Construct.node throughout

nija-at:
Following up on c8802395, update the rest of the RFC with the change not
to rename 'node' to 'construct'.

---


eladb (approved):
:-(

The most unfortunate thing in the world

Issue #328, opened 2021-06-02
polyglot assert library


nija-at:
## Description

The CDK assert library, `@aws-cdk/assert`, is currently available only to javascript users.
Make this available in all CDK supported languages

## Roles

* **Driver** (drives the proposal to completion): @nija-at 
* **Approver(s)**: (assigned by CDK team)

nija-at:
We've released the module `@aws-cdk/assertions`. This is now complete.

PR #330, opened 2021-06-02
rfc 328 - polyglot assert

nija-at:
* Tracking issue: #328
* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/nija-at/polyglot-assert/text/0328-polyglot-assert.md)

---

### DIFF
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -226,6 +226,22 @@ This is not a breaking change.
    assertions.assertResource('Foo::Bar', { ... });
    ```
 
+### What are the mitigations and/or long term plans to address these concerns?
+
+Once this module reaches general availability, we will deprecate the old 'assert'
+module. At that point, all of its source code can now be made as part of the new
+'assert' module.
+
+The dependencies on 'cloudformation-diff' and 'cfnspec' will need to be carefully
+evaluated and a decision made on whether to lift-and-shift the relevant code, or
+to move these out of the monorepo as independent modules.
+
+Friendly interfaces for popular testing frameworks such as jest, nunit, junit, etc.
+can be built as separate modules specific to those languages (not jsii modules)
+on top of the new 'assert' library.
+If necessary, this can be owned and maintained by the CDK team, but probably more
+suited for the community to pick up and build as third party modules.
+
 ### What alternative solutions did you consider?
 
 See [Appendix B](#appendix-b---alternatives) for alternatives.
@@ -239,9 +255,7 @@ The implementation plan is fairly simple.
 
 ### Are there any open issues that need to be addressed later?
 
-> Describe any major open issues that this RFC did not take into account. Once
-> the RFC is approved, create GitHub issues for these issues and update this RFC
-> of the project board with these issue IDs.
+None.
 
 ## Appendix A - Detailed Design
 
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -18,40 +18,24 @@ functionalities to users in all CDK supported languages.
 
 ### README
 
-<!--BEGIN STABILITY BANNER-->
-
----
-
-![cdk-constructs: Stable](https://img.shields.io/badge/cdk--constructs-stable-success.svg?style=for-the-badge)
-
----
-
-<!--END STABILITY BANNER-->
-
-> NOTE: This module contains *beta APIs*.
->
-> Some of the symbols in the APIs are suffixed with the `Beta<n>`.
-> When we have backwards incompatible change, we will create a new
-> symbol with a `Beta<n+1>` suffix and deprecate the `Beta<n>` symbol.
----
 This module allows asserting the contents of CloudFormation templates.
 
 To run assertions based on a CDK `Stack`, start off with -
 
 ```ts
 import { Stack } from '@aws-cdk/core';
-import { TemplateAssertionsBeta1 } from '@aws-cdk/assertions';
+import { TemplateAssertions } from '@aws-cdk/assertions';
 
 const stack = new cdk.Stack(...)
 ...
-const assert = TemplateAssertionsBeta1.fromStack(stack);
+const assert = TemplateAssertions.fromStack(stack);
 ```
 
 Alternatively, assertions can be run on an existing CloudFormation template -
 
 ```ts
 const template = fs.readFileSync('/path/to/template/file');
-const assert = TemplateAssertionsBeta1.fromTemplate(template);
+const assert = TemplateAssertions.fromTemplate(template);
 ```
 
 #### Full Template Match
@@ -171,7 +155,7 @@ assert.assertHasResource('Foo::Bar', {
   Properties: { Foo: 'Bar' },
   DependsOn: [ 'Waldo', 'Fred' ],
 }, {
-  part: ResourcePartBeta1.COMPLETE,
+  part: ResourcePart.COMPLETE,
 });
 ```
 
@@ -187,7 +171,7 @@ String json = """
 
 Map expected = new Gson().fromJson(json, Map.class);
 assert.assertHasResource("Foo::Bar", expected,
-  new AssertResourceOptionsBeta1.Builder().part(ResourcePartBeta1.COMPLETE).build());
+  new AssertResourceOptions.Builder().part(ResourcePart.COMPLETE).build());
 ```
 
 ```py
@@ -201,7 +185,7 @@ expected = """
   } """;
 
 assertion.assert_has_resource('Foo::Bar', json.loads(expected),
-  assertion.AssertResourceOptionsBeta1(part=ResourcePartBeta1.COMPLETE))
+  assertion.AssertResourceOptions(part=ResourcePart.COMPLETE))
 ```
 
 ## FAQ
@@ -239,33 +223,24 @@ This module is available in the languages as per the table below -
 
 ### Is this feature available in the AWS CDK v2?
 
-This module is available in both the AWS CDK v1 and v2, just like all of our other
-modules.
+This module is available in both the AWS CDK v2, just like the other CDK modules.
 
-In CDKv2, this module will be available as part of `aws-cdk-lib`.
-The import statement you will need to use is -
+Initially, this will be an experimental module. During this period, this module
+will be available under the package name `@aws-cdk/assertions` (or as identified
+in the table above) with the versioning scheme - `2.0.0-alpha.X` (or `0.x.y`,
+decided in a different RFC).
+
+Once this module is generally available, it will be available as part of `aws-cdk-lib`,
+and the import statement you will need to use is -
 
 ```ts
 import { assertions } from 'aws-cdk-lib';
 ```
 
 Read more about `aws-cdk-lib` at
-<https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#readme>.
-
-### Why are the APIs suffixed with the `BetaX` suffix?
-
-We have promised that there will be no backwards incompatible or breaking changes
-to our APIs in `aws-cdk-lib` as part of the AWS CDK v2.
-
-However, this module is still under development and is a developer preview release,
-where we are collecting feedback from users and performing necessary improvements.
-
-When an API needs to be modified in backwards incompatible ways, we will create a
-new API with a new suffix (`Beta2`, etc.). The old APIs will continue to be work but
-will be marked as deprecated.
-
-When the module is ready for prime time, we will publish the APIs without the 'Beta'
-suffix and mark all 'Beta' APIs as deprecated.
+<https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#readme> and about the
+CDK module lifecycle at
+<https://github.com/aws/aws-cdk-rfcs/blob/master/text/0107-construct-library-module-lifecycle.md>.
 
 ## Internal FAQ
 
@@ -325,6 +300,12 @@ This is not a breaking change.
    const assertions = TemplateAssertions.fromStack(stack);
    assertions.assertResource('Foo::Bar', { ... });
    ```
+
+4. With the current solution, the `@aws-cdk/assertions` module must be part
+   of the monocdk. It cannot be reorganized and released as-is into a
+   separate repository.
+
+   See [Appendix A](#appendix-a---detailed-design) for more details.
 
 ### What are the mitigations and/or long term plans to address these concerns?
 
@@ -402,8 +383,10 @@ are published as a separate module. There is a strict requirement imposed by the
 `aws-cdk-lib` packaging mechanism that it cannot depend on any of the experimental
 modules.
 
-To adhere to these constraints, the new assert module is marked as 'stable' and
-we will use the experimental API feature, i.e., suffix all APIs with `Beta<n>`.
+The 'assertion' module will be marked as 'experimental' and released as any other
+experimental API.
+This 'assertion' module would only be a `devDependencies` from other CDK modules,
+and hence will not be recognized as a dependenecy of `aws-cdk-lib`.
 
 ## Appendix B - Alternatives
 
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -41,14 +41,14 @@ To run assertions based on a CDK `Stack`, start off with -
 ```ts
 const stack = new cdk.Stack(...)
 ...
-const assert = TemplateAssertionsBeta1.fromStackBeta1(stack);
+const assert = TemplateAssertionsBeta1.fromStack(stack);
 ```
 
 Alternatively, assertions can be run on an existing CloudFormation template -
 
 ```ts
 const template = fs.readFileSync('/path/to/template/file');
-const assert = TemplateAssertionsBeta1.fromTemplateBeta1(template);
+const assert = TemplateAssertionsBeta1.fromTemplate(template);
 ```
 
 #### Full Template Match
@@ -101,7 +101,7 @@ assert.assertResourceBeta1('Foo::Bar', {
   Properties: { Foo: 'Bar' },
   DependsOn: [ 'Waldo', 'Fred' ],
 }, {
-  part: ResourcePartBeta1.COMPLETE_BETA1,
+  part: ResourcePartBeta1.COMPLETE,
 });
 ```
 
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -33,7 +33,7 @@ to users in all CDK supported languages.
 > All beta symbols are suffixed with the `Beta<n>`. When we have backwards
 > incompatible change, we will create a new symbol with a `Beta<n+1>` suffix
 > and deprecate the `Beta<n>` symbol.
-----
+---
 This module allows asserting the contents of CloudFormation templates.
 
 To run assertions based on a CDK `Stack`, start off with -
@@ -151,7 +151,7 @@ import { assertv2 as assert } from 'aws-cdk-lib';
 ```
 
 Read more about `aws-cdk-lib` at
-https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#readme.
+<https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#readme>.
 
 ### Why are all the APIs suffixed with the `Beta1` suffix?
 
@@ -190,9 +190,9 @@ The implementation of this module would simply be a scaffold around the existing
 To build this scaffold, we will 'vendor in' the libraries `assert-internal`,
 `cloudformation-diff` and `cfnspec`.
 
-See [Appendix A](#appendix-a) for the design and rationale.
+See [Appendix A](#appendix-a---detailed-design) for the design and rationale.
 
-Prototype / Implementation: https://github.com/aws/aws-cdk/pull/14952.
+Prototype / Implementation: <https://github.com/aws/aws-cdk/pull/14952>.
 
 ### Is this a breaking change?
 
@@ -228,7 +228,7 @@ This is not a breaking change.
 
 ### What alternative solutions did you consider?
 
-See [Appendix B](#appendix-b) for alternatives.
+See [Appendix B](#appendix-b---alternatives) for alternatives.
 
 ### What is the high level implementation plan?
 
@@ -245,7 +245,7 @@ The implementation plan is fairly simple.
 
 ## Appendix A - Detailed Design
 
-### Key Requirement
+### Key Requirement
 
 The **key requirement** for the design choices is *dogfooding*.
 The new assert APIs must be usable from within the AWS CDK, and over time,
@@ -266,7 +266,7 @@ must be part of the AWS CDK monorepo.
 ### Dependencies
 
 The main dependencies of the new assert module are the existing `assert` module,
-which in turn mainly depend on `cloudformation-diff` and `cfnspec`. 
+which in turn mainly depend on `cloudformation-diff` and `cfnspec`.
 The latter two are required to display human readable diffs, around what resources
 are different and how IAM permissions have changed.
 
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -39,16 +39,56 @@ This module allows asserting the contents of CloudFormation templates.
 To run assertions based on a CDK `Stack`, start off with -
 
 ```ts
+// In typescript
+import { Stack } from '@aws-cdk/core';
+import { TemplateAssertionsBeta1 } from '@aws-cdk/assertions';
+
 const stack = new cdk.Stack(...)
 ...
 const assert = TemplateAssertionsBeta1.fromStack(stack);
 ```
 
+```java
+// In Java
+import software.amazon.awscdk.core.Stack;
+import software.amazon.awscdk.assertions.TemplateAssertionsBeta1;
+
+Stack stack = new Stack(...)
+...
+TemplateAssertionsBeta1 assert = TemplateAssertionsBeta1.fromStack(stack);
+```
+
+```py
+# In Python
+from aws_cdk import (core, assertions)
+
+stack = core.Stack(...)
+...
+assertion = assertions.TemplateAssertionsBeta1.fromStack(stack)
+```
+
 Alternatively, assertions can be run on an existing CloudFormation template -
 
 ```ts
+// In typescript
 const template = fs.readFileSync('/path/to/template/file');
 const assert = TemplateAssertionsBeta1.fromTemplate(template);
+```
+
+```java
+// In Java
+import com.google.commons.io.Files; // using Google Guava. Use any utility of your choice
+
+String template = Files.readString("/path/to/template/file");
+TemplateAssertionsBeta1 assert = TemplateAssertionsBeta1.fromTemplate(template);
+```
+
+```py
+# In Python
+with open('/path/to/template/file', 'r') as file:
+  template = file.readlines().join('')
+...
+assertion = assertions.TemplateAssertionsBeta1.fromTemplate(template)
 ```
 
 #### Full Template Match
@@ -57,6 +97,7 @@ The simplest assertion would be to assert that the template matches a given
 template.
 
 ```ts
+// In typescript
 assert.assertTemplateMatches({
   Resources: {
     Type: 'Foo::Bar',
@@ -67,13 +108,59 @@ assert.assertTemplateMatches({
 });
 ```
 
+```java
+// In Java, using text blocks and Gson
+import com.google.gson.Gson;
+
+String json = """
+  {
+    "Resources": {
+      "Type": "Foo::Bar",
+      "Properties": {
+        "Baz": false
+      }
+    }
+  } """;
+
+Map expected = new Gson().fromJson(json, Map.class);
+assert.assertTemplateMatches(expected);
+```
+
+```py
+# In Python
+import json
+
+expected = """
+  {
+    "Resources": {
+      "Type": "Foo::Bar",
+      "Properties": {
+        "Baz": false
+      }
+    }
+  } """
+
+assertion.assertTemplateMatches(json.loads(expected))
+```
+
 #### Counting Resources
 
 This module allows asserting the number of resources of a specific type found
 in a template.
 
 ```ts
+// In typescript
 assert.assertResourceCountIs('Foo::Bar', 2);
+```
+
+```java
+// In Java
+assert.assertResourceCountIs("Foo::Bar", 2);
+```
+
+```py
+# In Python
+assertion.assertResourceCountIs('Foo::Bar', 2)
 ```
 
 #### Resource Matching
@@ -85,6 +172,7 @@ The following code asserts that the `Properties` section of a resource of type
 `Foo::Bar` contains the specified properties -
 
 ```ts
+// In typescript
 assert.assertHasResource('Foo::Bar', {
   Foo: 'Bar',
   Baz: 5,
@@ -92,17 +180,76 @@ assert.assertHasResource('Foo::Bar', {
 });
 ```
 
+```java
+// In Java, using text blocks and Gson
+import com.google.gson.Gson;
+
+String json = """
+  {
+    "Foo": "Bar",
+    "Baz": 5,
+    "Qux": [ "Waldo", "Fred" ],
+  } """;
+
+Map expected = new Gson().fromJson(json, Map.class);
+assert.assertHasResource("Foo::Bar", expected);
+```
+
+```py
+# In Python
+import json
+
+expected = """
+  {
+    "Foo": "Bar",
+    "Baz": 5,
+    "Qux": [ "Waldo", "Fred" ],
+  } """;
+
+assertion.assertHasResource('Foo::Bar', json.loads(expected))
+```
+
 The same method allows asserting the complete definition of the 'Resource'
 which can be used to verify things other sections like `DependsOn`, `Metadata`,
 `DeletionProperty`, etc.
 
 ```ts
+// In typescript
 assert.assertHasResource('Foo::Bar', {
   Properties: { Foo: 'Bar' },
   DependsOn: [ 'Waldo', 'Fred' ],
 }, {
   part: ResourcePartBeta1.COMPLETE,
 });
+```
+
+```java
+// In Java, using text blocks and Gson
+import com.google.gson.Gson;
+
+String json = """
+  {
+    "Properties": { "Foo": "Bar" },
+    "DependsOn": [ "Waldo", "Fred" ],
+  } """;
+
+Map expected = new Gson().fromJson(json, Map.class);
+assert.assertHasResource("Foo::Bar", expected,
+  new AssertResourceOptionsBeta1.Builder().part(ResourcePartBeta1.COMPLETE).build());
+```
+
+```py
+# In Python
+import json
+
+expected = """
+  {
+    "Properties": { "Foo": "Bar" },
+    "DependsOn": [ "Waldo", "Fred" ],
+  } """;
+
+assertion.assertHasResource('Foo::Bar', json.loads(expected),
+  assertion.AssertResourceOptionsBeta1(part=ResourcePartBeta1.COMPLETE))
 ```
 
 ## FAQ
--- a/text/0328-polyglot-assert.md
+++ b/text/0328-polyglot-assert.md
@@ -7,14 +7,14 @@ tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/328
 
 The `@aws-cdk/assert` module is only available to javascript users.
 
-The new polyglot assert module - `@aws-cdk/assertv2` - provides the same set of functionalities
-to users in all CDK supported languages.
+The new polyglot assert module - `@aws-cdk/assertions` - provides the same set of
+functionalities to users in all CDK supported languages.
 
 ## Working Backwards
 
 ### CHANGELOG
 
-**feat:** announcing assertv2: the assert library is now available in all CDK supported languages.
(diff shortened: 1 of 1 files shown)


text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+The same method allows asserting the complete definition of the 'Resource'
+which can be used to verify things other sections like `DependsOn`, `Metadata`,
+`DeletionProperty`, etc.
+
+```ts
+inspect.assertResourceBeta1('Foo::Bar', {
+  Properties: { Foo: 'Bar' },
+  DependsOn: [ 'Waldo', 'Fred' ],
+}, {
+  part: ResourcePartBeta1.COMPLETE_BETA1,
```
eladb, line 104: Again, no need to add _BETA1 everywhere. It's sufficient that the top level type has that prefix...

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+});
+```
+
```
eladb, line 107: What other APIs are we planning to offer? Or is that it?

I was also wondering if we should include some of the utils under `SynthUtils` (bad name) to make snapshot tests easier:

```ts
import { Template } from 'assertv2';

expect(Template.fromStack(stack)).toMatchSnapshot(); // <-- jest
```

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+const inspect = StackAssertionsBeta1.fromTemplateFileBeta1(file);
+```
+
+#### Full Template Match
+
+The simplest assertion would be to assert that the template matches a given
+template.
+
+```ts
+inspect.assertMatchTemplateBeta1({
```
rix0rrr, line 60: The `assert` everywhere seems a little redundant? Especially if we rename the `inspect` variable `assert`. How about:

```suggestion
assert.matchTemplateBeta1({
```

(And same for `assert.hasResourceBeta1` etc)

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+### Is this feature available in the AWS CDK v2?
+
+This module is available in both the AWS CDK v1 and v2, just like all of our other
+modules.
+
+In CDKv2, this module will be available as part of `aws-cdk-lib`.
+The import statement you will need to use is -
+
+```ts
+import { assertv2 as assert } from 'aws-cdk-lib';
```
rix0rrr, line 150: Why the rename?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+specific properties are present.
+
+The following code asserts that the `Properties` section of a resource of type
+`Foo::Bar` contains the specified properties -
+
+```ts
+inspect.assertResourceBeta1('Foo::Bar', {
+  Foo: 'Bar',
+  Baz: 5,
+  Qux: [ 'Waldo', 'Fred' ],
```
rix0rrr, line 91: What about complex matcher assertions? `objectLike`, `arrayWith` etc?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+The import statement you will need to use is -
+
+```ts
+import { assertv2 as assert } from 'aws-cdk-lib';
+```
+
+Read more about `aws-cdk-lib` at
+<https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#readme>.
+
+### Why are all the APIs suffixed with the `Beta1` suffix?
```
rix0rrr, line 156: Can we at least simplify to `Beta` instead of `Beta1` ?

text/0328-polyglot-assert.md
eladb, line 150: Why do we need to call this `assertv2` in `aws-cdk-lib`? There isn't an `assert` submodule there AFAIK. Let's clean that up in v2...

text/0328-polyglot-assert.md
eladb, line 156: See comment above - we just need the entrypoint class

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+   Users previously using -
+
+   ```ts
+   expect(stack).toHaveResource('Foo::Bar', { ... });
+   ```
+
+   will now have to use
+
+   ```ts
+   const assertions = StackAssertions.fromStack(stack);
```
rix0rrr, line 225:
```suggestion
   const assertions = StackAssertionsBeta1.fromStackBeta1(stack);
```

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
+---
+rfc pr: https://github.com/aws/aws-cdk-rfcs/pull/330
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/328
+---
+
+# Polyglot Assert
+
+The `@aws-cdk/assert` module is only available to javascript users.
+
+The new polyglot assert module - `@aws-cdk/assertv2` - provides the same set of functionalities
```
rix0rrr, line 10: My vote would be for `@aws-cdk/assertions`.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+
+   ```ts
+   expect(stack).toHaveResource('Foo::Bar', { ... });
+   ```
+
+   will now have to use
+
+   ```ts
+   const assertions = StackAssertions.fromStack(stack);
+   assertions.assertResource('Foo::Bar', { ... });
```
eladb, line 226: IIUC, this will throw an exception if the assertion fails. Is that the desired behavior? Perhaps we need to offer an API that just returns `true`/`false` so it can be used with testing frameworks:

```ts
const assert = StackAssertions.fromStack(stack);
expect(assert.hasResource('Foo::Bar', { ... }).toBeTruthy();
```

Otherwise, we basically halt test execution for each failure, no?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+All three of these modules are pure typescript, and not jsii modules, whereas the
+new assert module is a jsii module.
+jsii only allows other jsii modules as regular dependencies, and any non jsii module
+should be [bundled](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bundleddependencies).
+
+However, all three modules are within the AWS CDK monorepo and the tooling of `yarn`,
+`lerna` and `npm` do not support bundling modules that are within the monorepo.
+
+For this reason, we have gone with the option to 'vendor in' these modules by copying
+the source directly into the new assert module with some minor adjustments.
```
eladb, line 296: I am assuming copying will be done during build, correct? Or are we planning to copy & paste?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+> and deprecate the `Beta<n>` symbol.
+---
+This module allows asserting the contents of CloudFormation templates.
+
+To run assertions based on a CDK `Stack`, start off with -
+
+```ts
+const stack = new cdk.Stack(...)
+...
+const inspect = StackAssertionsBeta1.fromStackBeta1(stack);
```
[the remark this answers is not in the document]
nija-at: Not codified as canon. We can make that change if we all agree. 

cc @skinny85 @NetaNir

text/0328-polyglot-assert.md
nija-at: Adding the `assert` prefix makes it clear that these methods will throw if the conditions are not met.

I've left space, for the future, to add a `matchTemplate()` that returns true/false. Maybe premature(?).

text/0328-polyglot-assert.md
nija-at: silly me. changed.

text/0328-polyglot-assert.md
nija-at: The focus of this RFC was to discuss the basic premise of the APIs (and the beta suffix) and the design/implementation complexity.

It's in no way capturing all features of the assert that we will support. This will be added later.

text/0328-polyglot-assert.md
nija-at: I don't think that's what we decided.

On the next iteration, this will be `Beta2` which kind of makes the `Beta` or `Beta1` moot, no?

cc @NetaNir @skinny85

text/0328-polyglot-assert.md
nija-at: Like it.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
nija-at: > Let's write this README file as if the API was final (without the pre-release postfixes), so we can sense the ergonomics without the "pollution".

I've intentionally taken the opposite approach to get a sense of the ergonomics with the pollution. This is the APIs our users would encounter initially. 
I assumed it's not too hard to imagine the API without the pollution compared to how it would be with the "pollution".

> As for the name of the entrypoint, StackAssertions is okay, but I am wondering if this is more of TemplateAssertions. It "speaks" in CFN template terminology, not in CDK stack terms. WDTY?

I like it.

> I am also wondering if we should use Preview instead of Beta to avoid confusion

I believe we're already starting to use this - https://github.com/aws/aws-cdk/pull/13780
Happy to change, of course. cc @NetaNir @skinny85

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+```
+
+Alternatively, assertions can be run on an existing CloudFormation template -
+
+```ts
+const file = '/path/to/file';
+const inspect = StackAssertionsBeta1.fromTemplateFileBeta1(file);
```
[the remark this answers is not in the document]
nija-at: Sure. Changed.

Let's get into these nuances in the implementation PR. The focus for this RFC should be on the general approach and implementation choices.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
nija-at: see https://github.com/aws/aws-cdk-rfcs/pull/330#discussion_r644000892

text/0328-polyglot-assert.md
nija-at: Let's say I wanted to change the behaviour of this enum value. With your suggestion I would need to create a new enum `ResourcePartBeta2` and then create new methods with `Beta2` (or whatever) suffix everywhere this enum is used.

More pragmatic to keep the suffix, no?

text/0328-polyglot-assert.md
nija-at: The goal is to offer all functionality that makes sense. This is just the first iteration.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+You should use this feature if you would like to ensure that your CDK constructs
+and CDK applications render expected CloudFormation resources and properties.
+
+### How is this different from the existing `assert` module?
+
+The two modules provide the same set of features. However, the existing module
+was only available to javascript users.
+The new version launched today is available to users of all CDK supported
+languages available today and any that we will support in the future.
+
```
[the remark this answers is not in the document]
nija-at: None that I'm aware of, eventually. We will, of course, start with the most used and iteratively add more. 

Do you have some in mind that we shouldn't offer?

text/0328-polyglot-assert.md
nija-at: Per Rico's suggestion, I'm going to call this `assertions` - https://github.com/aws/aws-cdk-rfcs/pull/330#discussion_r643943776. So that's what it will be in v2.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,335 @@
...
+### What are the drawbacks of this solution?
+
+1. Ideally, the dependencies, `assert`, `cloudformation-diff` and `cfnspec`
+   are all available to be consumed as regular dependencies but are not,
+   since they are part of the same monorepo.
+
+   Instead, the design 'vendors in' these dependencies, i.e., copies over
+   the source code and rewrites imports, thereby increasing tech debt.
+
+2. Any new feature or bug fix to the new assert module needs to be performed
```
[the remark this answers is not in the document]
nija-at: My statement is true for features (it will have to be added in both libraries); and not true for most bug fixes.

I'll clean up the language.

text/0328-polyglot-assert.md
nija-at: Automatically done during build.

text/0328-polyglot-assert.md
nija-at: > Otherwise, we basically halt test execution for each failure, no?

I don't *think* so. AFAIK, most testing frameworks will continue to execute other test cases in the suite.

It will stop execution of that particular test case, same as when `.toBeTruthy()` is not met.

text/0328-polyglot-assert.md
nija-at: Intentional. I was showing this without the beta suffix for comparison with the snippet above.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
eladb: I am okay with `beta1` (just on the top class of course)

text/0328-polyglot-assert.md
eladb: I am not sure I can think of a reason you'd want to change the value of the enum member. It's normally not even needed to specify a value.

We should be consistent about these suffixes. If the entire type is "Beta", I don't see any value in actually polluting all the members with that suffix.

text/0328-polyglot-assert.md
eladb: Love it

text/0328-polyglot-assert.md
eladb: Yes, I think you are right. So I am fine with throwing and only supporting assert methods in the initial iteration

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
skinny85: I agree there's no reason to name the properties/methods of a class suffixed with `Beta1` also with a suffix. Example: https://github.com/aws/aws-cdk/pull/13780 (the class is `ServiceCatalogDeployActionBeta1`, but its properties do not have the `Beta1` suffix).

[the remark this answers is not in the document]
skinny85: Yes, `Preview` was "voted off the island" in https://github.com/aws/aws-cdk/pull/13780#pullrequestreview-629595262.

text/0328-polyglot-assert.md
skinny85: I think it's not horrible to suffix the first iteration with `Beta`. If we need a second iteration, we can still use `Beta2`, but if there is never a need for a second iteration, then the preview API looks nicer with just `Beta`.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+const assert = TemplateAssertionsBeta1.fromTemplate(template);
+```
+
+#### Full Template Match
+
+The simplest assertion would be to assert that the template matches a given
+template.
+
+```ts
+assert.assertMatchTemplateBeta1({
```
skinny85, line 60: My vote is in line with what Elad is saying - I wouldn't have the `Beta` preview suffix on the methods (having it one the class is enough).

Also, I would suggest changing the method names to read a little more fluently.

```suggestion
assert.assertTemplateMatches({
```

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+});
+```
+
+#### Counting Resources
+
+This module allows asserting the number of resources of a specific type found
+in a template.
+
+```ts
+assert.assertResourceCountBeta1('Foo::Bar', 2);
```
skinny85, line 76:
```suggestion
assert.assertResourceCountIs('Foo::Bar', 2);
```

text/0328-polyglot-assert.md
skinny85: I would add `objectLike()` and `arrayWith()` as well at this point - it's interesting how to represent them in non-TypeScript languages.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+#### Resource Matching
+
+Beyond resource counting, the module also allows asserting that a resource with
+specific properties are present.
+
+The following code asserts that the `Properties` section of a resource of type
+`Foo::Bar` contains the specified properties -
+
+```ts
+assert.assertResourceBeta1('Foo::Bar', {
```
skinny85, line 88:
```suggestion
assert.assertHasResource('Foo::Bar', {
```

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+which can be used to verify things other sections like `DependsOn`, `Metadata`,
+`DeletionProperty`, etc.
+
+```ts
+assert.assertResourceBeta1('Foo::Bar', {
+  Properties: { Foo: 'Bar' },
+  DependsOn: [ 'Waldo', 'Fred' ],
+}, {
+  part: ResourcePartBeta1.COMPLETE,
+});
```
skinny85, line 105: Can you show how will this look in non-TypeScript languages?

text/0328-polyglot-assert.md
skinny85: `SynthUtils` seems like a great addition (for example, to test that a validation exception is thrown).

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+### Is this feature available in the AWS CDK v2?
+
+This module is available in both the AWS CDK v1 and v2, just like all of our other
+modules.
+
+In CDKv2, this module will be available as part of `aws-cdk-lib`.
+The import statement you will need to use is -
+
+```ts
+import { assertions } from 'aws-cdk-lib';
```
skinny85, line 150: Really? We're adding a testing library to `aws-cdk-lib`? This seems like a mistake to me...

Is this decision final?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+
+   Instead, the design 'vendors in' these dependencies, i.e., copies over
+   the source code and rewrites imports, thereby increasing tech debt.
+
+2. Any new feature or bug fix to the new assert module that require changes
+   to the API needs to be performed both in the old assert module and new.
+   This can lead to weird/complicated implementation, since they both follow
+   different design patterns.
+
+3. The 'jest' friendly interfaces that were previously available are no longer
```
skinny85, line 215: How about we also plan to create framework-specific libraries in addition to the core library?

So, we would have `@aws-cdk/assertions-jest`, `@aws-cdk/assertions-nodeunit`, `aws-cdk-assertions-junit`, etc. Each of them would depend on the core assertions library.

Thoughts on this?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+
+The dependencies on 'cloudformation-diff' and 'cfnspec' will need to be carefully
+evaluated and a decision made on whether to lift-and-shift the relevant code, or
+to move these out of the monorepo as independent modules.
+
+Friendly interfaces for popular testing frameworks such as jest, nunit, junit, etc.
+can be built as separate modules specific to those languages (not jsii modules)
+on top of the new 'assert' library.
+If necessary, this can be owned and maintained by the CDK team, but probably more
+suited for the community to pick up and build as third party modules.
```
skinny85, line 244: Ah, I see you already thought about it.

In my opinion, this is something that should be owned by our team. I would add them to the RFC plan.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+This is reinforced by the old assert library which, over time, has evolved with
+new features added to it.
+
+### Ship Vehicle
+
+The assert library works on `Stack` which is a construct available in `@aws-cdk/core`
+(or `aws-cdk-lib` in v2). Hence, takes a dependency on this construct.
+
+Given this and working backwards from the key requirement, the new assert module
+must be part of the AWS CDK monorepo.
```
skinny85, line 279: This logic doesn't convince me.

Why can't this be a separate module that depends on `aws-cdk-lib`?

skinny85 (commented):
My two cents. In general, looks really good.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+
+![cdk-constructs: Stable](https://img.shields.io/badge/cdk--constructs-stable-success.svg?style=for-the-badge)
+
+---
+
+<!--END STABILITY BANNER-->
+
+> NOTE: This module contains *beta APIs*.
+>
+> All beta symbols are suffixed with the `Beta<n>`. When we have backwards
```
ericzbeard, line 33: Why not vend this as a separate experimental package, and you wouldn't have to suffix anything.

ericzbeard: Ah now that I read to the end I see why. Maybe naming the top level module as beta would allow you to omit the suffixes.

text/0328-polyglot-assert.md
ericzbeard: Agree, I think adding examples for other languages, particularly Python, would be valuable for the target audience.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
NetaNir: Yup `betaX` it is. Adam and I will figure out a way to announce it and document it asap

text/0328-polyglot-assert.md
nija-at: Answered in the FAQ.

text/0328-polyglot-assert.md
nija-at: This RFC is about making this decision.

Why is it a mistake to add a testing library to `aws-cdk-lib`?

text/0328-polyglot-assert.md
nija-at: We can always add this later. Let's first assess demand by just shipping what we have.

text/0328-polyglot-assert.md
nija-at: How would you satisfy the requirement I've identified (in the above section) if this is a separate module?
Do you disagree with that requirement?

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,336 @@
...
+```ts
+const stack = new cdk.Stack(...)
+...
+const assert = TemplateAssertionsBeta1.fromStack(stack);
+```
+
+Alternatively, assertions can be run on an existing CloudFormation template -
+
+```ts
+const template = fs.readFileSync('/path/to/template/file');
```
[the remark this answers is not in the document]
nija-at: > I have to say, I don't understand why would anyone use this. What's the use cases for this?

use case: Run any of our assertions on a CloudFormation template not generated by the CDK. 
I don't see why this should be used only on template generated from a CDK `Stack`. There's literally no additional effort in supporting this.

> Should we not have a way to create an Assertions object from the Cloud Assembly that's returned from `App.synth()`?

Reading between the lines, it seems you're assuming the use case is to pass a CDK synthesized template generated in the cloud assembly. It's not.

text/0328-polyglot-assert.md
nija-at: Nice! I like the suggestion around fluency.

text/0328-polyglot-assert.md
nija-at: Not the focus of this RFC. We can discuss the ergonomics around this in a different avenue.

text/0328-polyglot-assert.md
nija-at: I'd like to keep this `Beta1` unless someone wants to fight me on this. 

Leaving it as `Beta` makes it unclear to the maintainer if the next iteration is `Beta0`, `Beta1`, ...

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
skinny85: > > Should we not have a way to create an Assertions object from the Cloud Assembly that's returned from `App.synth()`?
> 
> Reading between the lines, it seems you're assuming the use case is to pass a CDK synthesized template generated in the cloud assembly. It's not.

Not sure what you mean exactly. Can we please add an API based on the Cloud Assembly? We need it ourselves, as we use it today ([one of many examples](https://github.com/aws/aws-cdk/blob/093ebcb942cd789b2516b4b3f577db3c9c1766c4/packages/%40aws-cdk/core/test/cfn-resource.test.ts#L63-L72)).

text/0328-polyglot-assert.md
skinny85: > I'd like to keep this `Beta1` unless someone wants to fight me on this.
> 
> Leaving it as `Beta` makes it unclear to the maintainer if the next iteration is `Beta0`, `Beta1`, ...

I don't think that's a very serious problem, but I don't care enough to fight 🙂.

text/0328-polyglot-assert.md
skinny85: We can ship a separate experimental assert library from a module in the monorepo, re-writing it as needed at build time.

Same way as we re-write the stable modules today to package them as part of monocdk, and the same way as we plan to ship the existing experimental modules like `@aws-cdk/aws-appmesh` (which has version `1.107.0`) as new modules `@aws-cdk-lib-experiments/aws-appmesh` (with version `0.1.0`).

text/0328-polyglot-assert.md
NetaNir: I agree with @nija-at, we want to avoid any ambiguity here. `beta1` < `beta2` in every context, but there are frameworks/products/environments  in which `beta2` < `beta`, does it make sense? Not sure, but let's avoid having to explain it to our users.

text/0328-polyglot-assert.md
NetaNir: To @nija-at point, if we only use `betaX` on the containing type (e.g class), what do we do when we want to deprecate an internal member. Fro example, lets say the `autoScalingGroup` class is experimental, and we only want to change the behavior of `formAutScalingGroupName` method, do we need to deprecate the entire `autoScalingGroup` class for that?


```
export class AutoScalingGroupBeta  {

   public static fromAutoScalingGroupName(scope: Construct, id: string, autoScalingGroupName: string): 
    .....
  }
}
```  


My  gut feeling is that  it will be  less  ideal from a  user experience  prespecitve. 

there  are a lot  of nuance here, I think we need to flush out the preview specification so we can be consistent across the code base.

eladb: I would just deprecate `AutoScalingGroupBeta`, and introduce `AutoScalingGroupBeta2` with this new method, while proxying all the deprecated methods from the deprecated type to the new type.

NetaNir: How will this work for users consuming `fromAutoScalingGroupName`? If you proxy it, you will be introducing a breaking change

eladb: Good callout. Of course this specific method will not be proxied as-is in order to preserve the legacy behavior in the deprecated class.

text/0328-polyglot-assert.md
skinny85: Since `beta` will be deprecated by the time `beta2` happens, I'm not too concerned about that ambiguity.

But like I said, I don't care too much about this particular aspect.

text/0328-polyglot-assert.md
skinny85: Agree with Elad here. Having to call `fromAutoScalingGroupName()` `AutoScalingGroupBeta1.fromAutoScalingGroupNameBeta1()` on the off-chance we will want to deprecate it seems like the wrong tradeoff to me.

Also, let's assume for a second we do go down this route, and call it `AutoScalingGroupBeta1.fromAutoScalingGroupNameBeta1()`. What do we call the replacement method then? `AutoScalingGroupBeta1.fromAutoScalingGroupNameBeta2()`? That looks pretty bad...

NetaNir: I think we are over indexing on how things look. Our main objective is to make sure it is clear that an API is in preview. The second objective is to make is easy for users to use preview APIs, will introducing a whole new class will make it harder for users to upgrade? It can be, as the class usage is more prevalent than a single method in a user application (more references to the `AutoScalingGroup` class in the code base than to `fromAutoScalingGroupName`).

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
nija-at: In the example provided, why do we do this instead of simply `assertMatchTemplate(stack, { ... })`?

text/0328-polyglot-assert.md
nija-at: I'll point you again towards the requirements section and quote it here, because your reply seems to ignore it -

> The **key requirement** for the design choices is *dogfooding*.
> The new assert APIs must be usable from within the AWS CDK, and over time,
> all CDK modules that use the old assert module should be migrated to the new one.

If we ship this as an experimental module, I believe we cannot take declare a dependency on it from any of our stable modules.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
nija-at: Oh I get it now. Our current assert library depends on core and hence cannot be used in the unit tests for core.
The same restrictions will apply to the new assert library too.

Do you have an example that's not in core?

text/0328-polyglot-assert.md
rix0rrr: I do, actually.

I hate `Beta1` with a passion. I will fight :)

text/0328-polyglot-assert.md
skinny85: > If we ship this as an experimental module, I believe we cannot take declare a dependency on it from any of our stable modules.

Pretty sure we can, because it will be a `devDependency`.

skinny85: Wait, actually. Why do we need to declare a dependency on this module at all? `aws-cdk-lib` does not ship with any tests...?

nija-at: Maybe but why. We'll still get into weird situations that we don't foresee.

For instance:
let's say, `@aws-cdk/aws-lambda` -> `aws-cdk-assertions` as a dev dependency. What will `aws-cdk-assertions` depend on? `@aws-cdk/core` or `aws-cdk-lib`?
It'll have to be the former, which means v2 customers can't use the new library. Or we'll need to do some re-packaging and re-publishing, etc.

Why deal with all this complexity? If we just publish this as any CDK construct library, it keeps things simple and gets the job done. We can deal with it the same way we handle everything else.
Why do you not like shipping this as any other construct library? The only difference I see is that these constructs does not participate in synthesis, which is acceptable.

I'd rather not create another special module that operates differently from the rest.

skinny85: > Why deal with all this complexity? If we just publish this as any CDK construct library, it keeps things simple and gets the job done. We can deal with it the same way we handle everything else.
Why do you not like shipping this as any other module? The only difference I see is that these does not participate in synthesis, which is acceptable.

Because it bloats the `aws-cdk-lib` module with test code, and forces you to use the ugly `Beta1` names.

I honestly don't see the problem here. We will have a module in the monorepo, `@aws-cdk/assertions` or something, which the single (v1) modules will depend on. This takes care of dogfooding. Then we re-package the module to be released with whatever name we want, `@aws-cdk-lib/assertions` or something else, in version `0.x.y`. It will `peerDepend` on `aws-cdk-lib`.

This is exactly the process we want to have for other experimental modules.

nija-at: Ok. Thanks for the feedback.

I'll take that into consideration before my final decision.

text/0328-polyglot-assert.md
eladb: @skinny85 what's your arguments against making the assertion library part of the AWS CDK. It's a common practice to include assertion utilities in standard libraries (see node.js's 'assert' for example).

We want to standardize test assertions to be an intrinsic part of the AWS CDK so it can be used as a foundation for the ecosystem.

The _important_ side effect is that it including it in the AWS CDK itself will allow us to use it from within the AWS CDK codebase, even if/when we aws-cdk-lib becomes a truly monolithic codebase (after we stop supporting v1).

text/0328-polyglot-assert.md
eladb: @skinny85 I am not sure I understand your arguments against including this is the standard library (besides module size, which is not a very strong argument given this is likely to add a few KiBs).

eladb:
LGTM approved as far as I am concerned

text/0328-polyglot-assert.md
otaviomacedo: Do we have any concerns around the size of the CDK package? If so, this could be an argument against including assertions in the main package, which would force users to deploy a testing library to a production environment, where it's not used. I know with Javascript we can use things like webpack, but I'm not sure whether this can be done in other languages. Then again, depending on the additional size, this may not even be relevant.

eladb: We do have concerns around the size but the assertion library is negligible comparing to the two huge lambda layers we bundle into the library, which is where we should optimize (see [here](https://github.com/aws/aws-cdk/projects/10#card-54949219)).

text/0328-polyglot-assert.md
skinny85: It also allows us to skip the ugly `Beta1` suffixes.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
skinny85: > > I have to say, I don't understand why would anyone use this. What's the use cases for this?
> 
> use case: Run any of our assertions on a CloudFormation template not generated by the CDK.
> I don't see why this should be used only on template generated from a CDK `Stack`. There's literally no additional effort in supporting this.

I think that, if you want to support that, the input should be the path to the file, not the contents of the file. Something like this:

```ts
const assert = TemplateAssertionsBeta1.fromFile('/path/to/template/file');
```

The reason for this is that parsing CloudFormation-flavored YAML is very non-trivial ([see here](https://github.com/aws/aws-cdk/blob/252dfa2f84f24ef57ab632e8ee5092544c850a5f/packages/%40aws-cdk/yaml-cfn/lib/yaml.ts#L21-L30)), and you don't want your customers to have to reinvent that.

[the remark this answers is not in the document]
skinny85: > Oh I get it now. Our current assert library depends on core and hence cannot be used in the unit tests for core.
> The same restrictions will apply to the new assert library too.
> 
> Do you have an example that's not in core?

Sure. [Here's one](https://github.com/aws/aws-cdk/blob/252dfa2f84f24ef57ab632e8ee5092544c850a5f/packages/%40aws-cdk/aws-s3-assets/test/asset.test.ts#L50-L72).

We have many, many tests like these.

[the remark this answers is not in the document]
nija-at: Sure, we can add it as and when we need it, just like any other library. 
The plan is to migrate (some) modules to start using these assertion library and eventually all modules will use it. This will force us to build this up iteratively.

That's not the focus of this RFC.

text/0328-polyglot-assert.md
nija-at: That's a fair argument. 
If we can release this library as an experimental module without adding any further complexity, we can do so.

However, after graduating to stable, the module will be included in `aws-cdk-lib` (just like any other experimental module).

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,483 @@
...
+    "Properties": { "Foo": "Bar" },
+    "DependsOn": [ "Waldo", "Fred" ],
+  } """;
+
+Map expected = new Gson().fromJson(json, Map.class);
+assert.assertHasResource("Foo::Bar", expected,
+  new AssertResourceOptionsBeta1.Builder().part(ResourcePartBeta1.COMPLETE).build());
+```
+
+```py
```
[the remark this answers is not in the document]
nija-at: Generally, I agree with you.

The part I want to emphasize in the RFC is that, for the best ergonomics in non-javascript languages, users will have to use a separate 3p library to convert to `dict` in Python and `Map` in Java, etc. 
My examples show the use in `Gson` (for Java) and `json` (for Python).

Calling out that experience is important.

text/0328-polyglot-assert.md
```diff
@@ -0,0 +1,483 @@
...
+languages available today and any that we will support in the future.
+
+### How do I get started?
+
+This module is available in the languages as per the table below -
+
+| Language              | Pkg Manager | Package                                       |
+| --------------------- | ------------|---------------------------------------------- |
+| typescript/javascript | npm         | `@aws-cdk/assertions`                         |
+| Java                  | Maven       | `software.amazon.awscdk.assertions`           |
```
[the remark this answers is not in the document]
nija-at: why? 

Isn't it a valid FAQ that users would expect to see? especially given that, the primary value prop of this RFC is that the module is available in all languages.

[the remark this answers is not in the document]
eladb: Of course it is but just feels like it's a detail we don't have to expand on in the RFC since it's derived from the submodule name and there isn't any control here.

text/0328-polyglot-assert.md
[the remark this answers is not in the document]
eladb: Feels like we can add that later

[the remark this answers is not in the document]
nija-at: Definitely a nice idea that we can add as the library evolves.

skinny85 (dismissed):
LGTM.

PR #335, opened 2021-06-11
RFC 1: Faster Development Iterations (aka "CDK Watch")

offbyone:
* Tracking issue: #1
* [Rendered version](https://github.com/offbyone/aws-cdk-rfcs/blob/0001-cdk-update/text/0001-cdk-update.md)

---

### DIFF
--- a/text/0001-cdk-update.md
+++ b/text/0001-cdk-update.md
@@ -3,7 +3,7 @@ rfc pr: [#335](https://github.com/aws/aws-cdk-rfcs/pull/335)
 tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/1
 ---
 
-# Faster personal CDK deployments
+# CDK Accelerate
 
 Shorten the development iteration speed -- the "edit-compile-test" loop -- for
 CDK applications.
@@ -12,16 +12,17 @@ CDK applications.
 
 ### CHANGELOG
 
-- feat(cli): implement `cdk update` and `cdk update --watch` for asset updates
+- feat(cli): implement `cdk deploy --accelerated` and
+  `cdk deploy --accelerated --watch`
 
 ### README
 
-The `cdk update` command accelerates the edit-compile-test loop for your CDK
-application during development by inspecting the assets and stack resources in
-your application, identifying those that can be updated in-place without a full
-CloudFormation stack udpate, and doing so using AWS service APIs directly. It
-can either do this as an interactive command or by running continuously and
-monitoring the input files.
+The `cdk deploy --accelerated` ("CDK accelerate") command accelerates the
+edit-compile-test loop for your CDK application during development by inspecting
+the assets and stack resources in your application, identifying those that can
+be updated in-place without a full CloudFormation stack udpate, and doing so
+using AWS service APIs directly. It can either do this as an interactive command
+or by running continuously and monitoring the input files.
 
 For supported construct types (see list, below) the update command will identify
 assets whose content has changed since the last time the CDK application was
@@ -35,35 +36,19 @@ contains resources supported by the update process. At this time, only Lambda
 Functions and ECS Services are supported, however more resource types will be
 added to this list in the future.
 
-At a high level, `update` operates by reading the current state of the stack,
-performing a stack diff on it, and if the stack is different only in resources
-supported by update plugins, running those in lieu of a full deployment.
-
-#### Preparing Your Stack
-
-In order to ensure that update is not invoked against production stacks, you
-must explicitly designate your stack as being updateable in your application.
-This is a stack-level property, set in the `StackProps` parameter to the stack
-constructor:
-
-```typescript
-const myDevelopmentStack = new ApplicationStack(app, "StackNameForDev", {
-    ...
-    interactiveUpdate: true,
-});
-```
-
-The `interactiveUpdate` property is optional, and when present and `true` the
-`cdk update` functionality will be executed on the stack.
+At a high level, `accelerate` operates by reading the current state of the
+stack, performing a stack diff on it, and if the stack is different only in
+resources supported by update plugins, running those in lieu of a full
+deployment.
 
 #### Running Update
 
-To run a one-shot update, you can invoke the `cdk update` command on some or all
-of your stacks:
+To run a one-shot update, you can invoke the `cdk deploy --accelerated` command
+on some or all of your stacks:
 
 ```
-$ cdk update ApplicationStack
-Checking stack ApplicationStack for possible asset update
+$ cdk deploy --accelerated ApplicationStack
+Checking stack ApplicationStack for possible accelerated update
  * LambdaFunction[AppFunction]: File asset changed
 ApplicationStack can be updated
 1/1 stacks can be updated:
@@ -76,10 +61,10 @@ If the update is to an attribute of the stack that cannot be updated, the
 command will offer to perform a full deployment:
 
 ```
-$ cdk update ApplicationStack
-Checking stack ApplicationStack for possible asset update
+$ cdk deploy --accelerated ApplicationStack
+Checking stack ApplicationStack for possible accelerated update
  * LambdaFunction[Appfunction]: Function data changed
-ApplicationStack can *not* be rapidly updated.
+ApplicationStack has changes that can not be rapidly updated.
 Perform a full deployment of ApplicationStack? [Y/n]: Y
 Deploying ApplicationStack
 ...
@@ -90,11 +75,11 @@ When multiple stacks are provided, update only occurs if all pending changes are
 suitable for update. Otherwise, a full deployment is done:
 
 ```
-$ cdk update ApplicationStack DataStack
-Checking stack ApplicationStack for possible asset update
+$ cdk deploy --accelerated ApplicationStack DataStack
+Checking stack ApplicationStack for possible accelerated update
  * LambdaFunction[Appfunction]: Function data changed
-ApplicationStack does not support update.
-Checking stack DataStack for possible asset update
+ApplicationStack has changes that can not be rapidly updated.
+Checking stack DataStack for possible accelerated update
  * LambdaFunction[StreamFunction]: File asset changed
 DataStack can be updated
 1/2 stacks can be updated: Perform a full deployment? [Y/n]? Y
@@ -105,16 +90,16 @@ Deploying DataStack
 Done!
 ```
 
-In addition to running in a one shot mode, the `cdk update` command also has a
-`--watch` command line option that enable it to monitor the assets on disk and
-perform an update when they change.
+In addition to running in a one shot mode, the `cdk deploy --accelerated`
+command also has a `--watch` command line option that enable it to monitor the
+assets on disk and perform an update when they change.
 
 ```
-$ cdk update --watch ApplicationStack DataStack
-Checking stack ApplicationStack for possible asset update
+$ cdk deploy --accelerated --watch ApplicationStack DataStack
+Checking stack ApplicationStack for possible accelerated update
  * LambdaFunction[Appfunction]: No Changes
 ApplicationStack can *not* be updated
-Checking stack DataStack for possible asset update
+Checking stack DataStack for possible accelerated update
  * LambdaFunction[StreamFunction]: No Changes
 
 Watching stack inputs for changes:
@@ -145,10 +130,11 @@ images.
 
 ### PRESS RELEASE
 
-DATE - AWS announces the `cdk update` command for the AWS CDK toolkit. The
-`update` command allows CDK application developers to rapidly update code within
-their CDK application when no other AWS resources are changing, bypassing the
-more correct, but time consuming, stack update procedure.
+DATE - AWS announces the `cdk deploy --accelerated` command for the AWS CDK
+toolkit. The `accelerate` command allows CDK application developers to rapidly
+update code within their CDK application when no other AWS resources are
+changing, bypassing the more correct, but time consuming, stack update
+procedure.
 
 The CDK toolkit uses CloudFormation under the hood to manage changes to AWS
 resources, which safely updates resources. This update process is slow, however,
@@ -156,13 +142,13 @@ and during development it performs safety checks that add overhead to developer
 workflows. Those delays are interposed between every change a developer makes,
 and over the course of a product may add hours to the total development time.
 
-The `cdk update` command provides a way to bypass those checks when working in
-development, directly updating the code in Lambda functions and ECS services
-without a full CDK deployment. When the `update` command can identify that all
-changes in the stack have shortcut support, it will directly publish the assets
-to AWS and then modify the underlying CDK resources directly.
+The `cdk deploy --accelerated` command provides a way to bypass those checks
+when working in development, directly updating the code in Lambda functions and
+ECS services without a full CDK deployment. When the `accelerate` command can
+identify that all changes in the stack have shortcut support, it will directly
+publish the assets to AWS and then modify the underlying CDK resources directly.
 
-> Using cdk update dramatically improves our development iteration speed for
+> Using CDK accelerate dramatically improves our development iteration speed for
 > Lambda functions. It cuts down the change-deploy-test cycle to seconds, where
 > it was minutes before with a full cdk deploy each time. Unlike some other
 > solutions, it actually updates our function running on AWS, which means there
@@ -171,14 +157,15 @@ to AWS and then modify the underlying CDK resources directly.
 
 Any CDK users who are deploying code to Lambda functions or containers on ECS or
 Fargate can take advantage of this tool today, by configuring their stacks for
-interactive update, and then using `cdk update` instead of `cdk deploy`.
+interactive update, and then using `cdk deploy --accelerated` instead of
+`cdk deploy`.
 
 ## FAQ
 
 ### What are we launching today?
 
-The `cdk update` and `cdk update --watch` commands, with support for rapid
-update of Lambda function code and Fargate images.
+The `cdk deploy --accelerated` and `cdk deploy --accelerated --watch` commands,
+with support for rapid update of Lambda function code and Fargate images.
 
 ### Why should I use this feature?
 
@@ -208,7 +195,7 @@ since each code change currently requires a CloudFormation stack update to apply
 new code, _or_ manually introducing drift to the application by inspecting the
 stack and manipulating the resources directly.
 
-The CDK update tool will allow the CDK to handle the resource changes rather
+The CDK accelerate tool will allow the CDK to handle the resource changes rather
 than this manual process, introducing some implicit safety and reducing the
 manual labor of the update.
 
@@ -219,25 +206,17 @@ to shortcircuit the safe CloudFormation deployment process. If a user runs CDK
 update on their production stack, it can perform updates without adequate stack
 update safety checks. Releasing a public tool with instant update capability
 into the CDK may not be the right way to make this functionality public. To
-mitigate this, `update` only runs on stacks that have explicitly opted in to the
-capability.
+mitigate this, `accelerate` only runs on stacks that have explicitly opted in to
+the capability.
 
 ### What changes are required to enable this change?
 
-This RFC requires three changes:
+This RFC requires two changes:
 
-1. An implementation of the `cdk update` command in the CDK toolkit that can
-   examine the Application stacks for updates that can be applied and apply
-   them.
-2. Adding the `interactiveUpdate` flag to the `StackProps` object, allowing a
-   stack to be configured for interactive update. This property will default to
-   `false`
-
-   **note:** I chose in-stack markup instead of a `cdk.json` flag so that if
-   there is an L3 construct for a personal development stack (or a library that
-   creates them) then the attribute can be set there instead.
-
-3. An expansion of the CDK asset model to include optional construction process
+1. An implementation of the `cdk deploy --accelerated` command in the CDK
+   toolkit that can examine the Application stacks for updates that can be
+   applied and apply them.
+2. An expansion of the CDK asset model to include optional construction process
    metadata and construction input metadata. This is required to allow CDK to
    identify input changes to the asset, reconstruct it from the defined
    processes, and then publish it to the necessary AWS service.
@@ -269,6 +248,10 @@ address the whole set of them.
 We also considered introspecting the CDK model ourselves, but concluded that
 there was little value in reinventing the wheel, so to speak, when CDK already
 had all of the information we'd need to deliver this.
+
+For identifying stacks that are subject to accelerated deployment, we are
+considering defining a full "Personal Development Stack" model, possibly based
+off of information in the CDK `App` context.
 
 ### What is the high level implementation plan?
 
--- a/text/0001-cdk-update.md
+++ b/text/0001-cdk-update.md
@@ -269,10 +269,27 @@ off of information in the CDK `App` context.
 ### What is the high level implementation plan?
 
 We will start from the prototype CDK update command that identifies the Lambda
-resources and then publishes them using the CDK toolkit.
+resources and then publishes them using the CDK toolkit, and extend that to
+implement support for ECR images associated with ECS and Fargate tasks, API
+Gateway definitions, and Step Function workflows. Those will be implemented
+directly in the toolkit code as part of the launch for the feature. The toolkit
+implementation will be designed to conform to an interface that provides:
 
-We will propose updates to the initial CDK constructs that are capable of
-accelerated deployment and make those changes to the CDK.
+- Watchable filesystem resources
+- A way of updating the watchable resource list
+- A method for determining whether the construct can be updated in place
+- A method for updating the construct in place
+
+If changes to the CDK constructs are necessary to implement the accelerated
+development process, we will make those changes as well. In the longer term we
+must lay the groundwork for moving the logic defining the update process into
+the Construct library, which implies a design for passing these values by way of
+the Cloud Assembly.
+
+We will implement a filesystem watcher for the CDK toolkit that works on one or
+more directory trees, watching for changes. It will base its watch list on the
+set of files indicated by the toolkit, and update them when those responses
+change.
 
 Additionally, a `--watch` flag and a file watcher will be added to support
 monitoring the inputs of stack resources for changes.
--- a/text/0001-cdk-update.md
+++ b/text/0001-cdk-update.md
@@ -1,0 +1,293 @@
+---
+rfc pr: [#335](https://github.com/aws/aws-cdk-rfcs/pull/335)
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/1
+---
+
+# CDK Update
+
+Rapid and automatic updates for code-only changes to CDK application stacks.
+
+## Working Backwards
+
+### CHANGELOG
+
+- feat(cli): implement `cdk update` and `cdk update --watch` for asset updates
+
+### README
+
+The `cdk update` command accelerates the asset deployments for your CDK
+application during development by inspecting the assets in your application,
+identifying those that can be updated in-place without a full CDK deployment,
+and doing so. It can either do this as an interactive command or by running
+continuously and monitoring the input files.
+
+For supported construct types (see list, below) the update command will identify
+assets whose content has changed since the publication of the CDK, assemble them
+from sources if necessary, and then publish them to their asset storage medium.
+It will then use AWS SDK commands to directly modify the CDK resources to use
+the new asset paths. The update command acts on one or more stacks in the
+application, and will only perform the shortcut if all changes are updateable.
+
+A stack is "updateable" if the stack is explicitly marked as updateable, and it
+contains any assets supported by the update process. At this time, only Lambda
+and ECS images are supported, and so any stack with one of those is updateable.
+
+At a high level, `update` operates by reading the current state of the stack,
+performing a stack diff on it, and if the stack is different only in resources
+supported by update plugins, running those in lieu of a full deployment.
+
+#### Preparing Your Stack
+
+In order to ensure that update is not invoked against production stacks, you
+must explicitly designate your stack as being updateable in your application.
+This is a stack-level property, set in the `StackProps` parameter to the stack
+constructor:
+
+```typescript
+const myDevelopmentStack = new ApplicationStack(app, "StackNameForDev", {
+    ...
+    interactiveUpdate: true,
+});
+```
+
+The `interactiveUpdate` property is optional, and when present and `true` the
+`cdk update` functionality will be executed on the stack.
+
+#### Running Update
+
+To run a one-shot update, you can invoke the `cdk update` command on some or all
+of your stacks:
+
+```
+$ cdk update ApplicationStack
+Checking stack ApplicationStack for possible asset update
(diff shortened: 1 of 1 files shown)


text/0001-cdk-update.md
```diff
@@ -0,0 +1,293 @@
+---
+rfc pr: [#335](https://github.com/aws/aws-cdk-rfcs/pull/335)
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/1
+---
+
+# CDK Update
```
eladb, line 6: We call this feature "CDK Watch" (see tracking issue) to signal to users that this feature _must not_ be used in production context

text/0001-cdk-update.md
```diff
@@ -0,0 +1,293 @@
...
+
+Rapid and automatic updates for code-only changes to CDK application stacks.
```
eladb, line 8: The goal of this feature is to shorten the development iteration speed (also known as "inner loop"), so the scope of should be expanded. The feature is not just for code-only changes. We want CDK Watch to support things like API Gateway models, Step Functions, etc.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Working Backwards
+
+### CHANGELOG
+
+- feat(cli): implement `cdk update` and `cdk update --watch` for asset updates
+
+### README
+
+The `cdk update` command accelerates the asset deployments for your CDK
```
eladb, line 18: As mentioned above - this is not just about assets - we need to make sure we design this to allow adding "shortcut deployments" for any type of resource. Lambda handlers backed by assets are just one example.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### Preparing Your Stack
+
+In order to ensure that update is not invoked against production stacks, you
+must explicitly designate your stack as being updateable in your application.
+This is a stack-level property, set in the `StackProps` parameter to the stack
+constructor:
+
+```typescript
+const myDevelopmentStack = new ApplicationStack(app, "StackNameForDev", {
```
eladb, line 47: Since I don't think we are going to offer a dedicated command for this, but rather support this only via watch, this is not required.

text/0001-cdk-update.md
eladb, line 18: I think this needs to be part of the `deploy` command and not a separate command. Eventually. this is about deployment of CDK apps. It's not a separate operation.

If I work backwards from the user experience, I don't think users need to "care" if they are short-circuiting or not. They just want to develop their app and have the CLI deploy it continuously and as fast as possible.

So, primarily this feature should be supported via a `--watch` or (`-w`) switch to `cdk deploy`:

```shell
cdk deploy --watch
```

I expect this command to synth & deploy your app and then watch for changes, synthesize your `cdk.out` directory and perform deployments as needed (with or without short-circuiting). In case the deployment only includes resources that can be short-circuited, it would run the short-circuiting logic. Otherwise, it will perform a normal deployment.

We can add a switch called `--fast` or `--short-circuit` (or something else) which will control the behavior of this feature. It has three modes: `auto`, `never` and `only`.

1. `--fast=auto` - if the update includes only short-circuitable resources, it will apply them. If the update includes non-short-circuitable resources, it will perform a normal  deployment. This is the default if `--watch` is enabled.
2. `--fast=never` - this is the default for `cdk deploy` - never perform a short-circuit deployment.
3. `--fast=only` - only perform short-circuit deployments - fails if there are other updates.

Here are some examples:

```shell
cdk deploy --fast=never  # equivalent to "cdk deploy"
cdk deploy --fast=auto  # one-off, allows both normal and short-circuit updates
cdk deploy --fast=only   # one-off, only allows short-cirtcuitable updates
cdk deploy --watch --fast=auto    # continuous, equivalent to "cdk deploy --watch"
cdk deploy --watch --fast=only     # continuous, only allows short-circuitable updates
cdk deploy --watch --fast=never   # continuous, will perform normal CFN updates
```

text/0001-cdk-update.md
```diff
@@ -0,0 +1,293 @@
...
+
+```
+
+The watcher can only watch for file changes. For Lambda code, that means that
+directories will be watched for changes and zipped if they change. If the file
+asset is a zip file, then the update will fire whenever that changes. The
+existing docker asset builder will be used to watch for changes in local docker
+images.
+
+#### Resource Support
```
eladb, line 129: I'd like to make sure we also plan to support other things like step function models

eladb (changes requested):
Let's make sure we are aligned on the scope and UX of the feature. Happy to spend some time in a meeting if you need any clarifications.

text/0001-cdk-update.md
offbyone: Here's where we differ; I think it's important to separate the deploy operation from this one, because they are _wildly_ different in what they require. `deploy` requires, generally, _CloudFormation_ access and a role that can interact with that set of tools. _`update`_ requires administrative level access to the account, and directly modifies the resources therein. That's different enough that I'm not comfortable, from a UX point of view, trying to bundle them into the same command. It would be confusing to have `deploy` actually need _different credentials_ depending on what flags you are passing it, whereas here we can document a clear boundary: When you are using `deploy`, you are acting in the role of the production deployment of the system, and when you are using `update` (name tbd :) ) you are acting in the role of the developer.

I would rather see more commands than fewer, and with a more clearly delineated set of requirements, capabilities, and responsibilities.

I think that the ability to perform a fast update vs a full deploy is useful -- the `--fast=auto` you mention above.

text/0001-cdk-update.md
offbyone: I am not 100% on board with that change of name. "watch" implies a persistent process, whereas this _may_ do that, but may also be a one-shot.

text/0001-cdk-update.md
offbyone: Agreed. I didn't plan to support them in an initial release, but SFN is a very obvious place where this functionality would be fantastic to have.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,300 @@
...
+
+The `cdk update` command accelerates the edit-compile-test loop for your CDK
+application during development by inspecting the assets and stack resources in
+your application, identifying those that can be updated in-place without a full
+CloudFormation stack udpate, and doing so using AWS service APIs directly. It
+can either do this as an interactive command or by running continuously and
+monitoring the input files.
+
+For supported construct types (see list, below) the update command will identify
+assets whose content has changed since the publication of the CDK, assemble them
```
ericzbeard, line 27: Should "since the publication of the CDK" be "since the last time the CDK application was deployed" ?

text/0001-cdk-update.md
ericzbeard: We also discussed the possibility of `cdk accelerate` to align with similar features planned in the SAM Accelerate project.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,300 @@
...
+from sources if necessary, and then publish them to their asset storage medium.
+It will then use AWS SDK commands to directly modify the CDK resources to use
+the new asset paths. The update command acts on one or more stacks in the
+application, and will only perform the shortcut if all changes are updateable.
+
+A stack is "updateable" if the stack is explicitly marked as updateable, and it
+contains any assets supported by the update process. At this time, only Lambda
+and ECS images are supported, and so any stack with one of those is updateable.
```
skinny85, line 35: This paragraph is not what we want, I think - the support is based on the types of resources, not on types of Assets.

I would re-word it to be something like:

```suggestion
A stack is "updateable" if the stack is explicitly marked as updateable, and it
contains resources supported by the update process. At this time, only Lambda
Functions and ECS Services are supported, however more resource types will be added to this list in the future.
```

text/0001-cdk-update.md
```diff
@@ -0,0 +1,300 @@
...
+  to support asset construction directly.
+- Direct support for monitoring container repositories for changes (possibly via
+  polling) instead of only supporting local rebuild.
+
+## Appendix
+
+Feel free to add any number of appendices as you see fit. Appendices are
+expected to allow readers to dive deeper to certain sections if they like. For
+example, you can include an appendix which describes the detailed design of an
+algorithm and reference it from the FAQ.
```
skinny85, line 300: As long as there are no appendices, let's get rid of this section:

```suggestion
```

text/0001-cdk-update.md
```diff
@@ -0,0 +1,300 @@
...
+
+In order to ensure that update is not invoked against production stacks, you
+must explicitly designate your stack as being updateable in your application.
+This is a stack-level property, set in the `StackProps` parameter to the stack
+constructor:
+
+```typescript
+const myDevelopmentStack = new ApplicationStack(app, "StackNameForDev", {
+    ...
+    interactiveUpdate: true,
```
skinny85, line 51: Whether or not we want to make this change, I think the name `interactiveUpdate` doesn't represent the functionality in this RFC very well.

Maybe something like `allowsShortCircuitDeployment`?

text/0001-cdk-update.md
```diff
@@ -0,0 +1,300 @@
...
+The `interactiveUpdate` property is optional, and when present and `true` the
+`cdk update` functionality will be executed on the stack.
+
+#### Running Update
+
+To run a one-shot update, you can invoke the `cdk update` command on some or all
+of your stacks:
+
+```
+$ cdk update ApplicationStack
```
skinny85, line 64: I don't like the fact that you have to specify the Stack name on the command line, _and_ mark it as `interactiveUpdate: true`. Feels like irritating double-booking to me; why do I have to tell you twice that I want to update this stack with `cdk update`?

I would either have the property on `Stack`, _or_ specifying the names through CLI arguments. Not both.

skinny85 (changes requested):
Looks great! My 2 cents.

skinny85:
> I think this needs to be part of the `deploy` command and not a separate command. Eventually. this is about deployment of CDK apps. It's not a separate operation.
>
> If I work backwards from the user experience, I don't think users need to "care" if they are short-circuiting or not. They just want to develop their app and have the CLI deploy it continuously and as fast as possible.
>
> So, primarily this feature should be supported via a `--watch` or (`-w`) switch to `cdk deploy`:
>
> ```shell
> cdk deploy --watch
>```
>
> I expect this command to synth & deploy your app and then watch for changes, synthesize your `cdk.out` directory and perform deployments as needed (with or without short-circuiting). In case the deployment only includes resources that can be short-circuited, it would run the short-circuiting logic. Otherwise, it will perform a normal deployment.
>
> We can add a switch called `--fast` or `--short-circuit` (or something else) which will control the behavior of this feature. It has three modes: `auto`, `never` and `only`.
>
> 1. `--fast=auto` - if the update includes only short-circuitable resources, it will apply them. If the update includes non-short-circuitable resources, it will perform a normal  deployment. This is the default if `--watch` is enabled.
> 2. `--fast=never` - this is the default for `cdk deploy` - never perform a short-circuit deployment.
> 3. `--fast=only` - only perform short-circuit deployments - fails if there are other updates.
>
> Here are some examples:
>
> ```shell
> cdk deploy --fast=never           # equivalent to "cdk deploy"
> cdk deploy --fast=auto            # one-off, allows both normal and short-circuit updates
> cdk deploy --fast=only            # one-off, only allows short-cirtcuitable updates
> cdk deploy --watch --fast=auto    # continuous, equivalent to "cdk deploy --watch"
> cdk deploy --watch --fast=only    # continuous, only allows short-circuitable updates
> cdk deploy --watch --fast=never   # continuous, will perform normal CFN updates
> ```

My reasoning for having it as a separate command (I'm not married to the `update` name, BTW) is to make sure this is not used for production deployments. I think that, if this is a separate command, it will be easier to explain in our docs this should not be used in production, and we can also print a warning in the CLI each time the `update` command is executed.

In your example, I don't like the complexity that `--fast` changes its default based on whether `--watch` was passed or not. I think that will be difficult to explain, remember, and might be error-prone for customers.

I also don't see the value of `--fast=only`. I agree with your initial statement:

>  I don't think users need to "care" if they are short-circuiting or not. They just want to develop their app and have the CLI deploy it continuously and as fast as possible.

I think giving customers the `--fast` switch where they can select `--fast=only` goes against this statement.

text/0001-cdk-update.md
eladb: I am okay with @skinny85's suggestion for the title.

We will make a decision about the "marketing name" later.

text/0001-cdk-update.md
eladb: The CDK CLI commands are designed around user intent, not around their implementation. The "deploy" command includes asset publishing, diff checks, change set creation, etc. From a user standpoint, they just want to deploy.

Same here, the intent of the developer using this feature is to deploy their development stack and therefore this feature is a sub-feature of the "deploy" command and not a separate command.

> I would rather see more commands than fewer, and with a more clearly delineated set of requirements, capabilities, and responsibilities.

I don't think "more" or "less" is the guiding design principle. Our primary concern when we design APIs is the mental model of the user and their intent, and we want to do the "right thing" with the minimum amount of required cognitive load or leakage of the implementation.

I'll frame it differently - if CFN updates would have been very fast and we didn't need to implement these short-circuits (i.e. `cdk deploy` would always take 1s), do you think we would need a separate command to deploy stacks? The answer is _no_. Our users "don't care" that CFN updates take a long time - they just want to deploy their apps and go on with their lives. This is the experience we are striving for (and perhaps at some point CFN will be fast enough that we wouldn't need those short circuits).

offbyone: I think your counterfactual is understandable, but I think we need to put the user's expectations front and centre here, and one of the things that having a separate command does is put a very bright line around the behaviour, which is _wildly_ different from the `deploy` capability. Doing it in the `deploy` command is _possible_, sure; I'm not going to die on that hill, but I can tell you in my experience using CLIs, when you have a command that dramatically changes its mode of operation when some flags are changed -- not just by altering details, but a completely new mode! -- it is surprising and hard to reason about. "Are there combinations of flags that can cause this? What if there are other commands that have similar flags? Do those do the same thing?" A flag introduces the possibility of a cross-cutting function.

I don't believe this is the same thing as a deployment; I don't think we should put it in the same place.

text/0001-cdk-update.md
offbyone: Let's wordsmith the flag as a last detail, but I think there's room for improvement once we agree on the command name.

text/0001-cdk-update.md
offbyone: The stack is specified because it's possible to have multiple stacks and only want to update one of them, or to watch one, etc. This, and it matches the usage of stacks elsewhere in the CDK cli. I don't like specifying the stacks either, but it _is_ necessary if you have "stages" of stacks.

skinny85: Then there's no reason for having the `interactiveUpdate` flag at all, and I would remove it.

offbyone: There is, though! Using an at-work example, for a single application that deploys many stacks, only a very small subset of those stacks should be allowed to be updated using this mechanism, and the others need to be denied by policy somewhere. One way that can be done is to explicitly identify which stacks in an app are suitable for it, another is to assume that the user is going to "know" the ones that are permitted in practice. How, otherwise, would a "production" stack prevent a rapid update, if your app consists of multiple stacks, each corresponding to a deployment stage?

ericzbeard: I'm wondering why we are concerned with which ones are permitted, when we are targeting local developer accounts?

text/0001-cdk-update.md
```diff
@@ -105,16 +90,16 @@ Deploying DataStack
 Done!
 ```
 
-In addition to running in a one shot mode, the `cdk update` command also has a
-`--watch` command line option that enable it to monitor the assets on disk and
-perform an update when they change.
+In addition to running in a one shot mode, the `cdk deploy --accelerated`
```
ericzbeard, line 93: Will we also have a short form, like `cdk deploy -a -w` ?

offbyone: That's a good question; the space of single-letter flags is much smaller than the long option ones, so that depends on if we think it's worth claiming one for this feature. I think "no" for `--watch` -- that's not going to be run as often, by its very nature -- and I lean towards "no" for `--accelerate` if we use that, because I can't see an unambiguous short name for it (`-a` would be better as a shortcut for `--all` for example)

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Why should I use this feature?
+
+If you are developing a CDK application and want to publish your code changes
+without waiting for CloudFormation to perform a full, safe stack update, you
+should use this feature.
+
+### Does it work when my Lambda is in Go or Java?
+
+To an extent, yes, but there are caveats. In order to update the lambda code you
```
corymhall, line 178: What about if the user is using the native bundling functionality (like with [aws-lambda-go](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-lambda-go-readme.html)? Will it be able to use the builtin asset bundler? Seems like it would be similar to how it would work for docker images.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+CDK applications.
+
+## Working Backwards
+
+### CHANGELOG
+
+- feat(cli): implement `cdk deploy --accelerated` and
+  `cdk deploy --accelerated --watch`
+
+### README
```
eladb, line 18: Please add a section here that shows how the `cdk deploy --help` will describe this feature. Something like this:

```
cdk deploy --help

    --accelerated Perform accelerated deployment (`--accelerated=only` will fail if there are non-accelerated updates)
    -w --watch   Watch for file changes and deploy any updates. Implies --accelerated=auto


Examples:
    cdk deploy -w
    cdk deploy --accelerated
    cdk deploy --accelerated=only --watch
...    
```

Something like that.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+Fargate can take advantage of this tool today, by configuring their stacks for
+interactive update, and then using `cdk deploy --accelerated` instead of
+`cdk deploy`.
+
+## FAQ
+
+### What are we launching today?
+
+The `cdk deploy --accelerated` and `cdk deploy --accelerated --watch` commands,
+with support for rapid update of Lambda function code and Fargate images.
```
eladb, line 168: This is not just about lambda functions and fargate images and it's important that this design will take that into account.

text/0001-cdk-update.md
eladb: Good point, we should add a FAQ

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+resources and then publishes them using the CDK toolkit. We will expand it to
+generalize it to use a simple plugin model to define the update mechanism, which
+will allow for ECS support in addition to Lambda support.
+
+Additionally, a `--watch` flag and a file watcher will be added to support
+monitoring the inputs of assets for changes.
+
+### Are there any open issues that need to be addressed later?
+
+- This RFC requires changes that may affect
```
eladb, line 266: Which changes?

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+  [rfc-0092 - cdk assets](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0092-asset-publishing.md),
+  as it entails changes to the asset model to support asset transformation and
+  sources.
+- This RFC can be extended to add support for further pluggable asset and update
```
eladb, line 270: What's the design of the extensibility model? What does it take to add a new resource type? Can this be done outside of the CLI codebase or does it require a change to the CLI? (I am okay with both models).

eladb (changes requested):
Another review iteration

The RFC needs to dive deeper into the architecture and design (totally fine to do that as we implement/prototype). 

For example, I expected to find some information on how changes are identified and associated with a specific update type and what it means to add new resource types.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+or by running continuously and monitoring the input files.
+
+For supported construct types (see list, below) the update command will identify
+assets whose content has changed since the last time the CDK application was
+deployed, and then publish them to their asset storage medium. It will then use
+AWS SDK commands to directly modify the CDK resources to use the new asset
+paths. The update command acts on one or more stacks in the application, and
+will only perform the shortcut if all changes are updateable.
+
+A stack is "updateable" if the stack is explicitly marked as updateable, and it
```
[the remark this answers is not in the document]
offbyone: I deleted that paragraph per your subsequent comment, so we can pretend that these words never happened ;)

text/0001-cdk-update.md
offbyone: It no longer does, since I pulled the asset changes out. Deleting.

text/0001-cdk-update.md
offbyone: I don't actually know; typescript is new enough to me that I'm not sure what affordances the language has for extensibility that way. I _suspect_ that the best design here will be to actually embed the extensions in the constructs themselves, so that we can allow for this in constructs we haven't yet thought of, or even in higher level constructs that are amenable to this approach (think SAM applications, for example). I'll put that down in the RFC and we can discuss it.

eladb: I don't think this is about typescript. Adding this to constructs sounds appealing but I am not sure how it will work.

I would start with articulating the straightforward solution and evolve from there.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+Checking stack DataStack for possible accelerated update
+ * LambdaFunction[StreamFunction]: No Changes
+
+Watching stack inputs for changes:
+ * LambdaFunction[Appfunction]: <spinner>
+ * LambdaFunction[StreamFunction]: <spinner>
+
+```
+
+The watcher can only watch for file changes. For Lambda code, that means that
```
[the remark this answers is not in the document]
mbonig: This could catch someone by surprise. 

let's say I'm wailing away on lambda code and I add a new env variable, my IDE probably auto-saves files and when I flip back to my lambda code to continue changing code in the handler a deploy starts going out. 

Could a warning with prompt be provided in this case ?

text/0001-cdk-update.md
```diff
@@ -0,0 +1,299 @@
...
+   applied and apply them.
+2. The CDK toolkit must be able to query a CDK resource for the set of
+   filesystem resources it must monitor for the `--watch` operation, and run a
+   filesystem monitor for that.
+
+### Is this a breaking change?
+
+No, it is not.
+
+### What are the drawbacks of this solution?
```
wong-a, line 241: Isn't stack drift another drawback since mutations are happening outside of CloudFormation? Drift isn't a big concern for development stacks, but there's nothing stopping anyone from doing it in prod stack. 

The main concern would be situations where the regular `cdk deploy` using CloudFormation fails due to out of stack mutations. Perhaps that's mitigated by restricting accelerated deployments to certain constructs and preventing resource deletion?

text/0001-cdk-update.md
```diff
@@ -0,0 +1,299 @@
...
+- AWS Lambda `Function`
+  - file and directory assets
+- AWS Fargate
+  - image assets
+- ECS
+  - image assets
+
+#### Future Support Plans
+
+- StepFunctions
```
wong-a, line 136: Might be out of scope but it may be worth considering partial synthesis too. I'm not sure if other constructs could benefit from this besides state machines.

When writing [StateMachines](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-stepfunctions.StateMachine.html) in CDK, you lose the workflow visualization and it's harder to understand the flow without seeing the graph. The only way to see it is by deploying the stack and viewing the created resource in the console. AWS Toolkit for Visual Studio Code supports state machine visualization from JSON or YAML which can be generated from `cdk synth`. But we don't need the full CFN template or reference resolution, just the state machine structure. Being able to view the graph even before deploying to get the JSON definition would shorten the dev cycle even further.

skinny85: That's a good suggestion @wong-a, but this is tricky. It's really difficult to perform partial synthesis in CDK, because the various constructs can reference each other, including Stacks. So, performing partial synthesis can result in very weird errors because of generating incorrect CloudFormation templates.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+# CDK Accelerate
+
+Shorten the development iteration speed -- the "edit-compile-test" loop -- for
+CDK applications.
+
+## Working Backwards
+
+### CHANGELOG
+
+- feat(cli): implement `cdk deploy --accelerated` and
```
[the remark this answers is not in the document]
rix0rrr: Henrique had a great name for this: "Speedy-K". Don't know how to work that in. `cdk deploy --speedy`? ;)

text/0001-cdk-update.md
```diff
@@ -0,0 +1,299 @@
...
+manual labor of the update.
+
+### Why should we _not_ do this?
+
+This solution has a risk of introducing a deployment tool that users might use
+to shortcircuit the safe CloudFormation deployment process. If a user runs CDK
+update on their production stack, it can perform updates without adequate stack
+update safety checks. Releasing a public tool with instant update capability
+into the CDK may not be the right way to make this functionality public. To
+mitigate this, `accelerate` only runs on stacks that have explicitly opted in to
```
rix0rrr, line 225: Interesting. This isn't mentioned anywhere else, and no code sample is shown... or have I missed it?

text/0001-cdk-update.md
```diff
@@ -0,0 +1,299 @@
...
+### Is this a breaking change?
+
+No, it is not.
+
+### What are the drawbacks of this solution?
+
+- Updating like this still entails a fair amount of preprocesing time, since for
+  complex projects (Golang and Java lambdas, for example) there remains a
+  compilation step.
+- The update model requires the AWS account used to publish the stack to also
```
rix0rrr, line 246: Oh shit good point. This is not going to fare well with the current `deploy-action`...

text/0001-cdk-update.md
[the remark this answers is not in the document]
eladb: I feel like the name needs to reflect the non-production nature of this behavior. How about `--bypass` - could work.. Then we have `--bypass=always`, `--bypass=never` and `--bypass=auto`.

text/0001-cdk-update.md
```diff
@@ -269,10 +269,27 @@ off of information in the CDK `App` context.
 ### What is the high level implementation plan?
 
 We will start from the prototype CDK update command that identifies the Lambda
-resources and then publishes them using the CDK toolkit.
-
-We will propose updates to the initial CDK constructs that are capable of
-accelerated deployment and make those changes to the CDK.
+resources and then publishes them using the CDK toolkit, and extend that to
```
eladb, line 272: Rename "toolkit" with "CLI" please (we no longer use this term)

text/0001-cdk-update.md
```diff
@@ -269,10 +269,27 @@ off of information in the CDK `App` context.
...
+- A method for determining whether the construct can be updated in place
+- A method for updating the construct in place
+
+If changes to the CDK constructs are necessary to implement the accelerated
+development process, we will make those changes as well. In the longer term we
+must lay the groundwork for moving the logic defining the update process into
+the Construct library, which implies a design for passing these values by way of
+the Cloud Assembly.
+
+We will implement a filesystem watcher for the CDK toolkit that works on one or
```
eladb, line 289: The watch list could be language-dependent so at least we should add support for specifying a set of watch-exclusion globs that will be applied to `cdk init` templates for each language. For example, for TypeScript, we will likely want to exclude `node_modules/**/*` from watch

eladb (changes requested):
There are still quite a lot of unresolved conversations. Please make sure to either address them or add a respond and resolve all conversations before merging.

text/0001-cdk-update.md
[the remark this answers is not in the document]
offbyone: It's not clear what we're `--bypass`-ing if we do it that way, so I'm not inclined on that. I'm going to stick with `--accelerated` during development, and see if we find a different name that makes more sense going forward.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+Checking stack ApplicationStack for possible accelerated update
+ * LambdaFunction[AppFunction]: File asset changed
+ApplicationStack can be updated
+1/1 stacks can be updated:
+Updating LambdaFunction in ApplicationStack
+...
+Update complete!
+```
+
+If the update is to an attribute of the stack that cannot be updated, the
```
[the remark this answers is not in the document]
offbyone: Will do

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+command will offer to perform a full deployment:
+
+```
+$ cdk deploy --accelerated ApplicationStack
+Checking stack ApplicationStack for possible accelerated update
+ * LambdaFunction[Appfunction]: Function data changed
+ApplicationStack has changes that can not be rapidly updated.
+Perform a full deployment of ApplicationStack? [Y/n]: Y
```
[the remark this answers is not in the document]
offbyone: That makes sense, and I'll implement it that way. Expect an update.

text/0001-cdk-update.md
offbyone: +1

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+
+The watcher can only watch for file changes. For Lambda code, that means that
+directories will be watched for changes and zipped if they change. If the file
+asset is a zip file, then the update will fire whenever that changes. The
+existing docker asset builder will be used to watch for changes in local docker
+images.
+
+#### Resource Support
+
+- AWS Lambda `Function`
```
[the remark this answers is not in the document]
offbyone: Per our discussion, will add this.

text/0001-cdk-update.md
```diff
@@ -0,0 +1,285 @@
...
+  - image assets
+- ECS
+  - image assets
+
+#### Future Support Plans
+
+- StepFunctions
+- API Gateway models
+
+### PRESS RELEASE
```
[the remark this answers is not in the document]
offbyone: Sure, can do. It's in here because it was in the RFC template when I made this, but I'm not attached to it.

text/0001-cdk-update.md
offbyone: I'll have to pull this out; it's not actually part of the final RFC.

text/0001-cdk-update.md
offbyone: It's both a drawback _and_ the primary mechanism of the tool, so it's a bit hard to say one way or the other. Bug and feature, if you will :)

text/0001-cdk-update.md
offbyone: Shall do

text/0001-cdk-update.md
offbyone: Good idea.

skinny85 (approved):
LGTM!

@eladb do you have any more comments?

eladb (approved):
Lgtm

PR #344, opened 2021-06-27
chore: construct hub - doc generation in the backend

iliapolo:
Introduce design modifications pertaining to generating package documentation artifacts in the backend as opposed to the front end. 

* [Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/epolon/backend-goc-generation/text/0314-cdk-construct-hub.md)

---

### DIFF
--- a/text/0314-cdk-construct-hub.md
+++ b/text/0314-cdk-construct-hub.md
@@ -400,7 +400,6 @@ Package documentation is created by the **Doc-Gen** function, which produces a J
 Since the front-end fetches and parses this file, we'd like for it to operate on type safe interfaces that describe the schema. This will ensure that breaking changes will manifest as compile time errors, and force us to handle them.
 
 To that end, we create a dedicated package, called `construct-hub-docgen`, which will define this schema, and provide an API to produce it.
-
 The version of the schema file will be derived from the major version of the package.
 Every time a change is made to the schema, we automatically detect what type of change has been made. (i.e breaking or not)
 
@@ -412,15 +411,60 @@ If a breaking change is detected, we enforce that the next version to be release
 
 Naturaly, every time a breaking change occurs, the front-end application will have to request and render a different schema file, while still maintaining support for all earlier versions.
 
-```ts
-const schemaVersion = parseMajor(require('node_modules/construct-hub-docgen/package.json').version)
-```
-
-> i.e: the latest schema version the front-end should support is the version of `construct-hub-docgen` it depends on.
+To achieve this, we implement a simple heristic in the front-end to fetch the appropriate version.
 
 ```ts
-switch (schema)
+// the latest version we support is derived from the docgen dependency
+const LATEST_SUPPORTED = parseMajor(require('node_modules/construct-hub-docgen/package.json'));
+
+// so we first try the latest
+let version = LATEST_SUPPORTED;
+let ok = false;
+
+while (!ok) {
+  ok = fetch(`packages/.../docs-${language}.${version}.json`);
+  if (!ok) {
+    // version isn't available yet (still backfilling...)
+    // so lets try earlier ones
+    version = version - 1;
+  }
+}
 ```
+
+One the correct version is determined, we pass it to the render function, which chooses the correct compoenent.
+
+```ts
+export function render(version: number) {
+  switch (version) {
+    case LATEST_SUPPRORTED:
+      return (<PackageDocs></PackageDocs>);
+    case 1:
+      return (<PackageDocsV1></PackageDocsV1>);
+    break:
+      throw new Error(`Unsupported docs version: ${version}`);
+  }
+}
+```
+
+In this scenario, the latest major version is `2`, and the front-end supports both `1` and `2`.
+When major version `3` is introduced, `<PackageDocs>` is copied over to `<PackageDocs2>`, and `<PackageDocs>` is changed with the code to support major version `3`.
+
+```ts
+export function render(version: number) {
+  switch (version) {
+    case LATEST_SUPPRORTED:
+      return (<PackageDocs></PackageDocs>);
+    case 2:
+      return (<PackageDocsV2></PackageDocsV2>);
+    case 1:
+      return (<PackageDocsV1></PackageDocsV1>);
+    break:
+      throw new Error(`Unsupported docs version: ${version}`);
+  }
+}
+```
+
+With this mechanism, we only have to actively maintain the latest version of the `<PackageDocs>` component.
 
 #### Website Analytics
 
--- a/text/0314-cdk-construct-hub.md
+++ b/text/0314-cdk-construct-hub.md
@@ -372,7 +372,8 @@ The details of URLs to be configrued in CloudFront is available in [appendix
 
 #### Backend X Frontend Compatibility
 
-As described earlier, the front-end application consumes artifacts that are stored on the backend. Since these artifacts are subject to change, we need to define a mechanism by which the front-end is able to detect and operate on various versions of them.
+As described earlier, the front-end application consumes artifacts that are stored on the backend. Since these artifacts are subject to change,
+we need to define a mechanism by which the front-end is able to detect and operate on various versions of them.
 
 As a general guideline, we *prefer to show an older version, than to show nothing at all*.
 To be more concrete, older versions of the artifacts should still be available for consumption by older consumers.
@@ -393,9 +394,12 @@ Specicially, we describe how to implement this compatibility model for each type
 
 ##### Package Documentation
 
-Package documentation is created by the **Doc-Gen** function, which produces a JSON definition file: `packages/${assembly.name}/v${assembly.version}/docs-${lang}.v1.json`
+Package documentation is created by the **Doc-Gen** function, which produces a JSON definition file:
 
-Since the front-end fetches and parses this file, we'd like for it to operate on type safe interfaces that describe the schema. This will ensure that breaking changes will manifest as compile time errors, and force us to handle them.
+`packages/${assembly.name}/v${assembly.version}/docs-${lang}.v1.json`
+
+Since the front-end fetches and parses this file, we'd like for it to operate on type safe interfaces that describe the schema.
+This will ensure that breaking changes will manifest as compile time errors, and force us to handle them.
 
 To that end, we create a dedicated package, called `construct-hub-docgen`, which will define this schema, and provide an API to produce it.
 The version of the schema file will be derived from the major version of the package.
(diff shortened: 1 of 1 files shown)


eladb (commented):
Looks pretty solid. Worth also discussing how the backend will automatically backfill new versions.

Also, what's the format of the .json file?

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +370,61 @@ pipeline.
...
+Each time a breaking change is performed, an additional file is stored. For example:
+
+- `<some-artifact.v1.json>`
+- `<some-artifact.v2.json>`
+
+This way, old consumers are free to operate on `v1`, while newer consumers can operate on `v2`.
+
+> Note that non breaking changes don't require consumer changes, so we don't have to encode minor/patch versions.
+
+Specicially, we describe how to implement this compatibility model for each type of artifact.
```
gabewomble, line 393:
```suggestion
Specifically, we describe how to implement this compatibility model for each type of artifact.
```

iliapolo:
> Looks pretty solid. Worth also discussing how the backend will automatically backfill new versions.
> 
> Also, what's the format of the .json file?

@eladb The format of the `.json` file is to be determined - @MrArnoldPalmer is working on it. 

I'll add a note about backfilling

iliapolo:
After a conversation with @RomainMuller about the current approach, he insightfully noted that while this works for existing packages, it breaks down for newly ingested packages.

Basically, once the schema version changes, newly ingested packages will only have the new version available to them, which means that if the front-end supporting it hasn't been deployed yet, this will break. 

One option is to always generate all available schema versions, which seems redundant and complicated. 
Another option is to enforce a deployment order, which is what the proposal now describes.

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+  "version": "v2",
+  ...
+}
+```
+
+Every time a change in the artifact occurs, the contents of the artifact is replaced with the new version.
+This means that at any given point in time, the artifact can either be in its new version, or the old one.
+
+As a general guideline, we *prefer displaying an older version of an artifact, than to display nothing at all.*.
+To that end, the front-end application will have to support displaying all versions of the artifact.
```
RomainMuller, line 398: It'll need to be able to render documents using a set of previous versions of the schema, but it does not have to be _all_ versions. Generally speaking, we could require that instances are upgraded through all interim major versions... So the front-end of any major may only ever need to render the current & previous version.

I guess this might be the spirit of what you were trying to say on L395, but the fact the statement does not specifically call out *major* versions, it's a little bit blurry.

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+##### Package Documentation
+
+Package documentation is created by the **Doc-Gen** function, which produces a JSON definition file:
+
+`packages/${assembly.name}/v${assembly.version}/docs-${lang}.json`
+
+Since the front-end fetches and parses this file, we'd like for it to operate on type safe interfaces that describe the schema.
+This will ensure that breaking changes will manifest as compile time errors, and force us to handle them.
+
+To that end, we create a dedicated package, called `construct-hub-docgen`, which will define this schema, and provide an API to produce it.
```
RomainMuller, line 411: `@construct-hub/package-description-schema`? It should be named after what it represents, not after what produces it.

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+The version of the schema file will be derived from the major version of the package.
```
RomainMuller, line 412: I'd argue the version of the schema _is_ the version of the package.

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+When a new major version of the schema is released, it is automatically picked up and
+deployed to the backend. Since the schema is overidden, this will break the front-end
+because the version supporting the new schema hasn't been deployed yet.
+We cannot afford this, which means we must enforce a deployment order in this scenario,
+by which the backend is deployed **only after** the front-end supporting it is deployed.
+
+We can enforce this by employing the following mechanism:
+
+- The front-end declares a `peerDepenedency` on `construct-hub-docgen`. (e.g `^1.0.0`)
+- The backend declares a `devDependency` on `construct-hub-docgen` (e.g `1.0.0`)
```
RomainMuller, line 431: Why pin?

```suggestion
- The backend declares a `devDependency` on `construct-hub-docgen` (e.g `^1.0.0`)
```

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+This mechanism enforces that when the backend picks up a new version major of the schema,
+it will have to pick up a new version of the front-end that supports it.
+
+From this point on, we just need to make sure we deploy the front-end first in the CDK application.
+
+In order for the front-end to support multiple versions of the schema, it will have to:
+
+1. Fetch the schema from `/../docs-{lang}.json`
+2. Extract the version field.
+3. Based on the version, render the appropriate component.
```
RomainMuller, line 456: This seems to imply the front-end will need to have the type definitions of both the previous and current versions at the same time.

This means a whole section is missing that would describe what to do when a breaking change happens on the package that contains the type definitions/schema - namely the `next` version needs to also contain the type for the previous major... This can be achieved using a discriminated type union (those are neat):

```ts
export type PackageDescription = PackageDescriptionV1 | PackageDescriptionV2;

export const enum SchemaVersion {
  V1 = 'v1',
  V2 = 'v2',
}

export interface PackageDescriptionV1 {
  readonly version: SchemaVersion.V1;

  // Other properties
}

export interface PackageDescriptionV2 {
  readonly version: SchemaVersion.V2;

  // Other properties
}
```

text/0314-cdk-construct-hub.md
```diff
@@ -362,6 +367,96 @@ pipeline.
...
+
+```json
+{
+  "version": "v2",
+  ...
+}
+```
+
+Every time a change in the artifact occurs, the contents of the artifact is replaced with the new version.
+This means that at any given point in time, the artifact can either be in its new version, or the old one.
```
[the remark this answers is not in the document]
iliapolo: Thanks, yes I indeed meant *an older one*. Changed to:

*Since artifacts are created in a long running background process, the new version may not be available immediately. This in turn means that at any given point in time, the artifact can either be in its new version, or an older one.*

text/0314-cdk-construct-hub.md
```diff
@@ -295,6 +295,11 @@ detail pages.
      pre-compute transliterated assemblies for their libraries, so that these
      are immediately available as they packages are indexed.
 
+  To ensure compatiblity between the schema of the definition, and the consuming
+  code (i.e the front-end application), the stored file will contain its version in the `version` field of the schema.
+
+  > For more information on this, see the [Backend X Frontend Compatiblity](#backend-x-frontend-compatibility) section.
```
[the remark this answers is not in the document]
iliapolo: 👍

text/0314-cdk-construct-hub.md
iliapolo: > Generally speaking, we could require that instances are upgraded through all interim major versions... So the front-end of any major may only ever need to render the current & previous version.

How would we do that? Imagine a breaking change is deployed, meaning the front-end now has to support two major versions. Now, before the backfilling finishes, another breaking change is deployed, so some packages may have major version 3, but some might still be at major version 2. The front-end now has to support 3 major versions.

text/0314-cdk-construct-hub.md
iliapolo: The idea was for the package to define both the schema, and expose the generator that produces that artifact based on that schema. So it is a docgenerator. 

Regarding the `@construct-hub` scope, I don't know that we need it...actually, what if we just push this as a new major version of [`jsii-docgen`](https://github.com/cdklabs/jsii-docgen)? Its essentially what it is, apart from the `constructs` specific code which we can maybe allow using a `constructsAware` option? @eladb what do you think? 99% of this code is going to be applicable outside of construct-hub.

text/0314-cdk-construct-hub.md
iliapolo: Yeap, fair enough

text/0314-cdk-construct-hub.md
iliapolo: I'm of the opinion that `devDependencies` should always be pinned. But we don't have to discuss this here, I think the spirit is clear, this is just an example. (both pinned and caret will work)

text/0314-cdk-construct-hub.md
iliapolo: > This seems to imply the front-end will need to have the type definitions of both the previous and current versions at the same time.

I kind of considered this an implementation detail we'll sort out later. 

Since the front-end will actively maintain only the latest major version, we could naively switch to `any` in the components that use an older major version (this code is supposed to locked, untouched once a new major version is introduced). This way we don't need to keep older schema definitions in the package, just the latest one.

Going with your approach, slightly deviated, we could also have each component version import the appropriate schema definition, exposed from the schema package:

```ts
import { spec } from 'jsii-docgen/spec/v1`

export class PackageDocsV1 {
}
```

```ts
import { spec } from 'jsii-docgen/spec/v2`

export class PackageDocsV2 {
}
```

Btw, with this approach, the backend can have full control over when to start generating new schemas.

Do you feel have to flush this out right now? I'd like to get the implementation going and than update the RFC with the approach we decide on.

text/0314-cdk-construct-hub.md
RomainMuller: My argument is that you could _require_ all objects be migrated to v2 before the v3 change can be deployed. There could be a deploy-time pre-condition check, for example.

text/0314-cdk-construct-hub.md
RomainMuller: I think the doc should probably mention there should be a provision around what will happen when we have to major-bump the schema, but I guess we can defer determining what the actual process is to a later point (when we actually need that - since we should be striving to avoid resorting to this as much as possible).

text/0314-cdk-construct-hub.md
RomainMuller: I'm a little skeptical about the `docgen` part of the name, because most people would likely expect this produces a docsite, not a bunch of JSON files. But the naming is not important here, as long as we agree on the responsibilities & outcomes.

text/0314-cdk-construct-hub.md
RomainMuller: I see. I guess the examples between parentheses are a little polarising/distracting. But I do get the sentiment.

text/0314-cdk-construct-hub.md
iliapolo: I see, but do we want that? I think if we see it becomes a burden for the front-end to support, maybe introduce this option in the future.

text/0314-cdk-construct-hub.md
iliapolo: So actually I believe that even when we introduce the schema, we will still retain the option for generating a markdown file. Its already there and can actually serve as useful feature outside the scope of construct-hub. For example for generating an API reference on the GitHub repo, the same thing `jsii-docgen` does now. 

Given that, lets push it to `jsii-docgen`? I agree `docgen` might have an implicit context of `docsite`, but we can live with that, especially since we already do.

text/0314-cdk-construct-hub.md
iliapolo: 👍

text/0314-cdk-construct-hub.md
eladb: Yes, I love it. Agree with @iliapolo that docgen should have an option to output markdown (i.e. render markdown from the doc model).

And we are slowly moving towards a future where JSII knows about constructs anyway (think c8), so this is definitely not a break in the layers. I am comfortable for jsii-docgen to know about constructs without any additional switches.

text/0314-cdk-construct-hub.md
iliapolo: > I think the doc should probably mention there should be a provision around what will happen when we have to major-bump the schema

Right now this is the provision :)

*Based on the version, render the appropriate component.*

I'll defer the actual process to a later point.

iliapolo:
@RomainMuller Made a few adjustments and decided to go with `jsii-docgen`.

RomainMuller (approved):

Issue #353, opened 2021-07-08
Constructs for all public CloudFormation resources and modules


eladb:
## Description

Proposal for publishing an AWS CDK construct library which includes strongly-typed constructs for all publish resources and modules in the CloudFormation registry.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @eladb 
| Author(s)           | @eladb 
| API Bar Raiser      | @alias
| Stakeholders        | @addihorowitz, @Chriscbr, @iliapolo

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [ ] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #356, opened 2021-07-08
RFC 353: Constructs for CloudFormation Extensions

eladb:
This is a proposal to publish construct libraries automatically generated for all public CloudFormation public extensions (resources/modules). See #353 for additional details.

Implementation is under: https://github.com/cdklabs/cdk-cloudformation-types

APIs are signed off by @rix0rrr 

---

### DIFF
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -194,6 +194,8 @@ export class Cluster extends CfnResource {
   }
 }
 
+// ---------------- output of json2jsii:
+
 /**
  * A resource that creates Amazon Elastic Kubernetes Service (Amazon EKS) clusters.
  *
@@ -459,5 +461,4 @@ export interface KubernetesApiAccessEntry {
   readonly groups?: string[];
 
 }
-```
 ```
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -16,7 +16,7 @@ makes it easier to use public CloudFormation extensions in your CDK apps.
 For example, let's say I want to define a GitHub repository using the `TF::GitHub::Repository` resource:
 
 ```ts
-import { Repository } from '@cdk-cloudformation-extensions/tf-github';
+import { Repository } from '@cdk-cloudformation-extensions/tf-github-repository';
 
 new Repository(this, 'MyRepo', {
   name: 'my-repo',
@@ -25,13 +25,14 @@ new Repository(this, 'MyRepo', {
 });
 ```
 
-For each namespace (e.g. `TF::GitHub` in the above example) in 
-the public CloudFormation Registry, a library is available under 
-the scope `@cdk-cloudformation-extensions/NAMESPACE`. This library
-includes construct libraries for all the types in that namespace.
+For each type (e.g. `TF::GitHub::Repository` in the above example) in 
+the public CloudFormation Registry, a module is available under 
+the scope `@cdk-cloudformation-extensions/NAMESPACE-TYPE`. This library
+includes a construct class and all the relevant data types for this
+type.
 
-New versions of all modules are released daily and include the 
-latest versions of all constructs.
+The module version corresponds to the version of the type schema
+in the public registry.
 
 ---
 
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -9,14 +9,14 @@ resources and modules published to the public CloudFormation Registry.
 
 ## README
 
-The `@cdk-cloudformation-extensions/xxx` scope includes construct libraries with generated strongly-typed "L1" constructs
+The `@cdk-cloudformation-types/xxx` scope includes construct libraries with generated strongly-typed "L1" constructs
 for all the public extensions (resources and modules) in the AWS CloudFormation public registry. This library
 makes it easier to use public CloudFormation extensions in your CDK apps.
 
 For example, let's say I want to define a GitHub repository using the `TF::GitHub::Repository` resource:
 
 ```ts
-import { CfnRepository } from '@cdk-cloudformation-extensions/tf-github-repository';
+import { CfnRepository } from '@cdk-cloudformation-types/tf-github-repository';
 
 new CfnRepository(this, 'MyRepo', {
   name: 'my-repo',
@@ -27,7 +27,7 @@ new CfnRepository(this, 'MyRepo', {
 
 For each type (e.g. `TF::GitHub::Repository` in the above example) in
 the public CloudFormation Registry, a module is available under
-the scope `@cdk-cloudformation-extensions/NAMESPACE-TYPE`. This library
+the scope `@cdk-cloudformation-types/<namespace-type>`. This library
 includes a construct class and all the relevant data types for this
 type.
 
@@ -70,13 +70,13 @@ new CfnResource(this, 'MyEksCluster', {
 });
 ```
 
-With `@cdk-cloudformation-extensions` all public resources and modules will be listed in the **Construct Hub** like any
+With `@cdk-cloudformation-types` all public resources and modules will be listed in the **Construct Hub** like any
 other construct and by importing the relevant `cdk-cloudformation-extensions` module into their projects, they will be able to
 use them via strongly-typed classes. IDEs will show type information and inline help derived from the
 extension schema.
 
 ```ts
-import { CfnCluster } from '@cdk-cloudformation-extensions/awsqs-eks-cluster';
+import { CfnCluster } from '@cdk-cloudformation-types/awsqs-eks-cluster';
 
 new CfnCluster(this, 'MyEksCluster', {
   name: 'my-new-cluster',
@@ -146,19 +146,24 @@ We will use the following naming scheme:
   * .NET Namespace: `Cdklabs.CdkCloudFormationTypes.<NamePascalCase>` (e.g. `Cdklabs.CdkCloudFormationTypes.MongodbAtlasProject`)
 
 Alternatives:
-* `@cdk-types/mongodb-atlas-project`
-* `@cdk-cfn/mongodb-atlas-project`
-* `@cdkcfn/mongodb-atlas-project`
-* `@awscdk/mongodb-atlas-project`
-* `@cfn/mongodb-atlas-project`
-* `@cfn/cdk-mongodb-atlas-project`
-* `@cloudformation/cdk-mongodb-atlas-project`
-* `@cfntypes/mongodb-atlas-project`
-* `@cloudformation-registry/mongodb-atlas-project`
-* `@cfn-registry/cdk-mongodb-atlas-project`
-* `@cfn-types/cdk-mongodb-atlas-project`
-
-
+1. `@cdk-types/mongodb-atlas-project`
+2. `@cdk-cfn/mongodb-atlas-project`
+3. `@cdkcfn/mongodb-atlas-project`
+4. `@awscdk/mongodb-atlas-project`
+5. `@cfn/mongodb-atlas-project`
+6. `@cfn/cdk-mongodb-atlas-project`
+7. `@cloudformation/cdk-mongodb-atlas-project`
+8. `@cfntypes/mongodb-atlas-project`
+9. `@cloudformation-registry/mongodb-atlas-project`
+10. `@cfn-registry/cdk-mongodb-atlas-project`
+11. `@cfn-types/cdk-mongodb-atlas-project`
+12. `@cfn-registry/cdk-mongodb-atlas-project`
+13. `@cloudformation-registry/cdk-mongodb-atlas-project`
+13. `@cdk-cloudformation-registry/cdk-mongodb-atlas-project`
+13. `@cdk-resources/cdk-mongodb-atlas-project`
+13. `@awscdk-resources/mongodb-atlas-project`
+13. `@cloudformation-resources/mongodb-atlas-project`
+13. `@cloudformation-types/mongodb-atlas-project`
 
 #### Versioning
 
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -137,8 +137,8 @@ that these modules include L1s which we also refer to as "CFN resources" in the 
   * Distribution name: `cdk-cloudformation-<name-kebab-case>` (e.g. `cdk-cloudformation-mongodb-atlas-project`)
   * Module: `cdk_cloudformation_<name_snake_case>` (e.g. `cdk_cloudformation_mongodb_atlas_project`)
 * **NuGet**:
-  * Package ID: `Cdklabs.CdkCloudFormation.<NamePascalCase>` (e.g. `Cdklabs.CdkCloudFormation.MongodbAtlasProject`)
-  * .NET Namespace: `Cdklabs.CdkCloudFormation.<NamePascalCase>` (e.g. `Cdklabs.CdkCloudFormation.MongodbAtlasProject`)
+  * Package ID: `CdkCloudFormation.<NamePascalCase>` (e.g. `CdkCloudFormation.MongodbAtlasProject`)
+  * .NET Namespace: `CdkCloudFormation.<NamePascalCase>` (e.g. `CdkCloudFormation.MongodbAtlasProject`)
 
 Alternatives considered:
 1. `@cdk-cloudformation-types/mongodb-atlas-project`
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -207,6 +207,6 @@ version of the resource they use, we determined that the best approach is to pub
 
 ### Are there any open issues that need to be addressed later?
 
-* __Property name conversion__: `json2jsii` converts field names to camelCase to adhere with naming typescript naming conventions.
+* ~__Property name conversion__~ (resolved by [cdklabs/json2jsii#480](https://github.com/cdklabs/json2jsii/pull/480)): `json2jsii` converts field names to camelCase to adhere with naming typescript naming conventions.
   This means that we need to map field names back when we define the `CfnResource`. I think we might want to add a
   feature to `json2jsii` that will generate conversion functions that can be used to convert back data types to the original schema.
--- a/text/353-cfn-registry-constructs.md
+++ b/text/353-cfn-registry-constructs.md
@@ -4,29 +4,29 @@
 * **Tracking Issue**: #353
 * **API Bar Raiser**: @rix0rrr
 
-A set of construct libraries which includes generated constructs for all CloudFormation 
+A set of construct libraries which includes generated constructs for all CloudFormation
 resources and modules published to the public CloudFormation Registry.
 
 ## README
 
-The `@cdk-cloudformation-extensions/xxx` construct libraries includes generated strongly-typed "L1" constructs 
+The `@cdk-cloudformation-extensions/xxx` scope includes construct libraries with generated strongly-typed "L1" constructs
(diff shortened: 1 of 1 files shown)


hoegertn:
I would love if every extension module would also contain a construct I can deploy to enable this type in my AWS account or org using the `AWS::CloudFormation::TypeActivation`

eladb:
> I would love if every extension module would also contain a construct I can deploy to enable this type in my AWS account or org using the `AWS::CloudFormation::TypeActivation`

I love this idea - I'll incorporate into the RFC.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+in the AWS CDK and simply publish JSII construct libraries which includes 
+statically-generated L1s for all the public registry extensions.
+
+#### Library per type
+
+CloudFormation extensions are semantically versioned. This means that new versions of an extension
+may include breaking changes. We also want to allow users to pick up a specific version of an extension.
+In light of these constraints, we determined that the best path forward is to publish each extension
+as a separate module, with a version that corresponds to the extension version in the CloudFormation
+registry. This way, we will have 1:1 alignment and users will have the most freedom.
```
Chriscbr, line 117: Does the version alignment include all levels (major, minor, and patch numbers)? What if we needed to apply some kind of security patch etc.?

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+> Describe any problems/risks that can be introduced if we implement this RFC.
+
+### What alternative solutions did you consider?
+
+#### `cdk import` versus library
+
+As mentioned above, an alternative approach would be to add support for `import` in the CDK CLI. The downsides of this approach (at this point) 
+are that (a) it is a bigger investment as it involves a change to the CDK CLI; and (b) the Construct Hub won't be able to surface these
+types automatically (we will need some custom support for "imported" types, which is something that we would like to add to the Construct Hub
+in the future for all CDK domains).
```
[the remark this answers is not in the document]
eladb: They are very different experiences from a developer perspective. I'll call that out.

text/353-cfn-registry-constructs.md
eladb: Good call. What do you think? I believe we can technically decide that the patch number is not coupled with the registry version to allow ourselves to deal with such cases.

nija-at:
RFC Bar Raiser: @rix0rrr

text/353-cfn-registry-constructs.md
Chriscbr: I think that would be acceptable.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+});
+```
+
+With `@cdk-cloudformation-extensions` all public resources and modules will be listed in the **Construct Hub** like any
+other construct and by importing the relevant `cdk-cloudformation-extensions` module into their projects, they will be able to
+use them via strongly-typed classes. IDEs will show type information and inline help derived from the
+extension schema.
+
+```ts
+import { Cluster } from '@cdk-cloudformation-extensions/awsqs-eks';
```
rix0rrr, line 79: This module name is inconsistent with the previous example. To be in line it would have to be

```ts
import { Cluster } from '@cdk-cloudformation-extensions/awsqs-eks-cluster';
```

Which one was intended?

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+As mentioned above, an alternative approach would be to add support for `import` in the CDK CLI. The downsides of this approach (at this point) 
+are that (a) it is a bigger investment as it involves a change to the CDK CLI; and (b) the Construct Hub won't be able to surface these
+types automatically (we will need some custom support for "imported" types, which is something that we would like to add to the Construct Hub
+in the future for all CDK domains).
+
+#### Monolithic versus module per resource
+
+As mentioned above, due to the fact that resource schemas are versioned and we want to allow users to select which
+version of the resource they use, we determined that the best approach is to publish a module for each resource.
+This has logicical/operational implications that we believe can mostly be mitigated using projen's capabilities to
```
rix0rrr, line 173: Can you be a bit more specific about the implications you're thinking about, because I don't really see them.

It seems to me we need a cron job either way, and publishing is mostly the same to what we're already doing in CDK; the only difference would really be whether we hand-write a single `package.json`, or generate many of them automatically... correct?

The projen path would expect us to commit the `package.json`s, which I'm not even sure we necessarily need to do... but I guess we can.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+### Are there any open issues that need to be addressed later?
+
+* __Property name conversion__: `json2jsii` converts field names to camelCase to adhere with naming typescript naming conventions. 
+  This means that we need to map field names back when we define the `CfnResource`. I think we might want to add a 
+  feature to `json2jsii` that will generate conversion functions that can be used to convert back data types to the original schema.
+
+## Appendix
+
+```ts
+export class Cluster extends CfnResource {
```
rix0rrr, line 191: Do we call this `CfnCluster` to indicate it's a raw CFN resource with no human flavoring?

What do we do for attributes? `attrClusterName` like L1s? And mutable properties like L1s? People are going to need these.

At the same time, having both `cluster.attrClusterName` and `cluster.clusterName` available has led a LOT of people to use the wrong property on our current collection of L1s. `cluster.propClusterName` instead maybe?

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+As mentioned above, due to the fact that resource schemas are versioned and we want to allow users to select which
+version of the resource they use, we determined that the best approach is to publish a module for each resource.
+This has logicical/operational implications that we believe can mostly be mitigated using projen's capabilities to
+reason about the project structure and publishing workflows using code.
+
+### What is the high level implementation plan?
+
+- [ ] Implement [`cdk-import`](https://github.com/cdklabs/cdk-import) as a command-line tool that generates L1 constructs for registry extensions.
+- [ ] Create a mono-repo-style project using projen that utilizes `cdk-import` to generate a module for each registry resource.
+
```
rix0rrr, line 180:
```
- [ ] Run the script on a schedule in us-east-1
```
?

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+#### Code Generator
+
+The code generator will be executed daily (e.g. through a GitHub workflow),
+query the CloudFormation Registry, and generate L1s for all the types based
+on their metadata and schema.
+
+We can use the `ListTypes` API to list all the types in the registry:
+
+```shell
+aws cloudformation list-types --visibility=PUBLIC
```
rix0rrr, line 128:
```
| grep -v AWS::
```

:)

rix0rrr (commented):
Sounds good!

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
+# Constructs for Public CloudFormation Extensions
+
+* **Original Author(s):**: @eladb
+* **Tracking Issue**: #353
+* **API Bar Raiser**: @rix0rrr
+
+A set of construct libraries which includes generated constructs for all CloudFormation 
+resources and modules published to the public CloudFormation Registry.
+
```
nija-at, line 9: (Feeds into the next comment about repo org)

Can you add a minimal working backwards info on how users would build L2s on top of these L1s?

Perhaps in the form of contribution guide or a section in our "design guidelines".

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+and to that end, if we simply publish a construct library that includes constructs for all the public,
+extensions, this library will naturally be discoverable through the Hub.
+
+### Why should we _not_ do this?
+
+We have a longer term plan ([RFC#77](https://github.com/aws/aws-cdk-rfcs/issues/77)) to add 
+support for a CLI command called `cdk import` which will allow users to generate L1s from any
+registry type just-in-type and add them to their project (this is similar to CDK8s and CDKtf).
+
+### What is the technical solution (design) of this feature?
```
nija-at, line 105: Can you also talk about where the code and gen scripts will live? A separate repo, I hope.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+
+And then for each type, we can use `DescribeType` to retrieve the type information and its schema. 
+
+For example, this command will return the description of the `AWSQS::EKS::Cluster` resource:
+
+```shell
+aws cloudformation describe-type --arn arn:aws:cloudformation:us-east-1::type/resource/408988dff9e863704bcc72e7e13f8d645cee8311/AWSQS-EKS-Cluster
+```
+
+For reference: the output of this command can be found [here](https://gist.github.com/eladb/bf417c07444027d6954360295df4ee37#file-awsqs-vpc-vpcqs-module-json).
```
nija-at, line 139: (minor)

Generally, avoid using gist. It's lifecycle and ownership model is different from RFC.
Prefer using appendices instead.

The output doesn't look too big to add these inline.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,467 @@
...
+support for a CLI command called `cdk import` which will allow users to generate L1s from any
+registry type just-in-type and add them to their project (this is similar to CDK8s and CDKtf).
+
+### What is the technical solution (design) of this feature?
+
+The idea is to do something similar to what we do with the CloudFormation L1s 
+in the AWS CDK and simply publish JSII construct libraries which includes 
+statically-generated L1s for all the public registry extensions.
+
+#### Library per type
```
nija-at, line 111: I like the simplicity of this solution, however we should be careful.

To get a sense of scale, how many public types are there today, in orders of magnitude?

Are there limits in the target pkg managers that we will hit here?

It could be useful to publish these under a separate account in these pkg managers to keep the blast radius contained if we hit a hard limit.
I actually don't know how we would wiggle out if we hit such a limit and aren't able to publish any more modules. Might be worth giving some more thought?

text/353-cfn-registry-constructs.md
eladb: The `awssqs-eks-cluster` one. Fixed.

text/353-cfn-registry-constructs.md
eladb: I think a gist is a good tool for this. I understand what you are saying about the lifecycle, but this is just a raw dump of an output of a command (300 lines), so I'll keep these in a gist. Seems to me like a pragmatic thing to do.

text/353-cfn-registry-constructs.md
eladb: Removed the sentence about projen and added "generated" to indicate that the modules will be generated. That's the important point. We will determine how to generate them later. That's not that important.

text/353-cfn-registry-constructs.md
eladb: Deleted the appendix - we can discuss the details over https://github.com/cdklabs/cdk-import/pull/2

text/353-cfn-registry-constructs.md
eladb: > To get a sense of scale, how many public types are there today, in orders of magnitude?

There are currently 44 public types (resources & modules), but we expect this list to grow.

> Are there limits in the target pkg managers that we will hit here?

Doesn't seem like there are any constraints on the number of packages in a scope in Npm. But if we will hit any limits, we can decide to split into multiple scopes/accounts/whatever. Since these modules will be listed in the construct hub their discoverability is not predicated on the scope/group-id or anything.

eladb: Added a note.

text/353-cfn-registry-constructs.md
eladb: Added https://github.com/cdklabs/cdk-import/

text/353-cfn-registry-constructs.md
eladb: Not sure I understand what you are after...

text/353-cfn-registry-constructs.md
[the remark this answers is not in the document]
eladb: Added some notes.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,212 @@
...
+
+* [ ] Implement [`cdk-import`](https://github.com/cdklabs/cdk-import) as a command-line tool that generates L1 constructs for registry extensions.
+* [ ] Create a mono-repo-style project using projen that utilizes `cdk-import` to generate a module for each registry resource.
+* [ ] Created a scheduled job which generates the modules periodically (should probably run against `us-east-1`)
+
+### Are there any open issues that need to be addressed later?
+
+* __Property name conversion__: `json2jsii` converts field names to camelCase to adhere with naming typescript naming conventions.
+  This means that we need to map field names back when we define the `CfnResource`. I think we might want to add a
+  feature to `json2jsii` that will generate conversion functions that can be used to convert back data types to the original schema.
```
Chriscbr, line 212: I think this can be removed since it has been resolved.

eladb: Done. Thanks!

text/353-cfn-registry-constructs.md
nija-at: We usually build L2s on top of L1s. Are we going to encourage building L2s for these or are they going to remain as L1s. 
If so, where are these L2s going to be contributed to? i.e., which repo and how is it organized.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,212 @@
...
+CloudFormation extensions are semantically versioned. This means that new versions of an extension
+may include breaking changes. We also want to allow users to pick up a specific version of an extension.
+In light of these constraints, we determined that the best path forward is to publish each extension (type)
+as a separate module, with a version that corresponds to the extension version in the CloudFormation
+registry. This way, we will have 1:1 alignment and users will have the most freedom.
+
+There are currently 44 public types (resources & modules), but we expect this
+list to substantially grow. If this dramatically scales (to thousands of
+packages) and we hit any limitations of package managers (e.g. account/scope
+limits), we can shard the solution (multiple accounts, multiple scopes, etc).
```
nija-at, line 122: It would be best to separate these already from the account used by the aws cdk packages. A separate account dedicated for cfn registry constructs seems reasonable.

If/when we hit the undocumented limit, we don't block the release of one from the other. (equivalent to our best practice where we use different aws accounts for different services/apps)

text/353-cfn-registry-constructs.md
nija-at: We usually build L2s on top of L1s. 
Are we going to encourage building L2s for these or are they going to remain as L1s?

If so, where are these L2s going to be contributed to? i.e., which repo and how is it organized. 
Especially given that each type is its own module, and L2s (and bind patterns) bring multiple types together, how is this going to work?

nija-at (commented):
Looks good. Just the two comments below.

text/353-cfn-registry-constructs.md
eladb: I don't believe people will build L2s for these. If they will, they can either use `cdk-import` and embed the imported L1s in the L2 library or just use these generated L1 libraries as a dependency of their L2s. At any case, this is out of scope here.

text/353-cfn-registry-constructs.md
eladb: I feel this would be a premature optimization. I am planning to use the cdklabs accounts for this. Every set of publishing accounts is a nightmare to maintain (especially given we need to rotate keys periodically).

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,257 @@
...
+
+#### Naming Scheme
+
+> *This is still not finalized. From a CDK user standpoint, I am not sure if they should really
+> care if these are CloudFormation registry extensions or just normal CDK libraries. Consider a user
+> that goes to Construct Hub and searches for "mongodb". They find this library, import it and use it.
+> As far as they are concerned, they just used a construct library. The fact that this is an L1 type
+> and not some L2/L3 is an implementation details, isn't it. So I am not sure that the semantics which
```
rix0rrr, line 130: While in principle I agree, there are practical differences.

People have gotten used to certain affordances in L2s that these constructs won't have. Typed arguments instead of ARNs (`bucket: IBucket` vs `bucketArn: string`), integration classes, enums, a README with a thorough explanation and examples, etc.

Since these classes will be generated at the L1 level, the `Cfn` prefix will indicate that all these expected affordances are absent.

text/353-cfn-registry-constructs.md
```diff
@@ -0,0 +1,257 @@
...
+> involves "CloudFormation" or "extensions" or "types" is not a leak in the abstraction*.
+
+We will use the following naming scheme:
+
+* **npm**: 
+  * Package name: `@cdk-cloudformation-types/<name-kebab-case>` (e.g. `cdk-cloudformation-types/mongodb-atlas-project`)
+* **Maven Central**:
+  * Group ID: `io.github.cdklabs.cdk_cloudformation_types`
```
rix0rrr, line 138: Why not `software.amazon` ?

I don't understand the branding of these packages as `cdklabs`:

* cdklabs is not a software domain, and it's not an organization or entity (Amazon is the entity), so I'm not sure what `cdklabs` refers to here.
* Does 'labs' in the name imply that these packages are not ready? That they're experimental? Will they ever graduate to production-ready? (Are questions that customers are going to be asking themselves if they see the word *labs*)
* If the branding of these packages is "cdklabs", why is that branding not reflected in the NPM name and the PyPI name?

text/353-cfn-registry-constructs.md
rix0rrr: On the other hand, the `Cfn` prefix might be taken to mean there *should* be a better, preferred version of the construct... which will never arrive.

rix0rrr: I am coming down on the side of NO `Cfn` prefix.

rix0rrr (approved):
We can quibble about names and naming for weeks, and I'm not sure we'd ever come up with a satisfying answer. I sure don't have a better idea.

Approved, with the note that I think we shouldn't be using "cdklabs" in the package names. Please reconsider that scheme. 

Other than that, let's do it!

text/353-cfn-registry-constructs.md
eladb: Cdklabs is a publishing entity, mostly aligned with Our GitHub org. All the of the libraries we publish from the cdklabs GitHub org are under these namespaces.

text/353-cfn-registry-constructs.md
eladb: Why not use The Cfn prefix?

eladb: @rix0rrr and I discussed this and decided we will keep the `Cfn` prefix in order to allow future publishing of 1st party L1s as part of this project. In which case we will likely have a library such as `@cdk-cloudformation/aws-lambda.CfnFunction`.

text/353-cfn-registry-constructs.md
eladb: Maven recommends using `io.github.<org>` and the group ID and since this project is part of the `cdklabs` github org, we will keep this group ID.

Removed the `Cdklabs` prefix for NuGet, so it's now just `CdkCloudFormation.Foo`.

rix0rrr (approved):

Issue #66, opened 2018-06-29
StackSets Support


eladb:
## Description

Allow users to define a template through AWS CDK constructs and deploy as a Self-Managed CloudFormation StackSet.

* User defines stack set the same way as normal cdk stack
* When user executes `cdk deploy` it would create a stack set if it does not exist or update an existing stack set + existing stack set instances.

Proposed Scope:
* Create New **Self-Managed** Stack Set ([docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStackSet.html )).
* Update Existing **Self-Managed** Stack Set:
  * Update includes updating existing Stack Set instances as well
  * https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateStackSet.html
  * https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateStackInstances.html
* Ability to set stack set configuration/deployment properties:
  * Administration Role, Execution Role, Fault Tolerance, Concurrency, etc. 

Out of Scope for v1 (May be iteratively added in future):
* Service-Managed Stack Set
* Add/Remove stack set instances by ID or OU

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @linsona
| Author(s)           | @linsona
| API Bar Raiser      | @skinny85
| Stakeholders        | @alias, @alias, @alias

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at
  [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

Related: https://github.com/aws/aws-cdk/issues/11896

[...]

PR #357, opened 2021-07-08
RFC 66: Self Managed Stack Set Support

linsona:
* Tracking issue: #66
* [Rendered version](https://github.com/linsona/aws-cdk-rfcs/blob/rfc-66/text/66-self-managed-stack-sets.md)

---


text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+  * Create Bootstrapless StackSet Synthesizer
+  * Update CLI to collect StackSet artifact types in addition to Stacks
+* StackSet with file asset support
+  * Create Process for bootstrapping to enable file asset support
+  * Create StackSet Synthesizer for stack sets that inclue file assets
+
+### Are there any open issues that need to be addressed later?
+* Docker asset management
+* Future support for service managed stack sets
+
```
eladb, line 106: CDK pipelines support

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+## Working Backwards
+
+### CHANGELOG
+- feat(core): New Stack Set core type 
+- feat(assembly): New stack set artifcat type and schema
+- feat(cli): ```cdk deploy``` collects and creates/updates stack set 
+- feat(lib): API Calls/Logic for create/update stack sets
+
+### README
+The ```cdk.StackSet``` construct in CDK defines AWS resources in a stack set and stack set configuration/deployment preferences. This is different from the ```cdk.CfnStackSet``` construct which creates a CFN stack that deploys a CFN stack set. The ```cdk.StackSet``` construct creates the stack set directly if it does not exist, or updates a stack set and existing stack set instances while monitoring the stack set operation.
```
eladb, line 22: Why the triple quotes?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+* Stack Set Instance Add/Remove through CDK
+* Stack Set deletion
+* Docker Assets
+
+## Internal FAQ
+### Why are we doing this?
+CDK does not support the creation and updating of CFN stack sets directly, only through an intermediate single stack. This would also enable customers with existing stack sets to utlize CDK to update in the future.
+
+### Why should we _not_ do this?
+CDK does provide the ```cdk.CFNStackSet``` construct, but requires an intermediate single stack to deploy the stack set which does not monitor stack set operations on update.
```
eladb, line 78: Can you elaborate on this? I am not sure I fully understand the limitations of the L1 and why this is the only way to solve them

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+### What changes are required to enable this change?
+Changes are required since CDK currently does not have the concept/logic of deploying stack sets directly.
```
eladb, line 81: The question is "what" changes

text/66-self-managed-stack-sets.md
linsona: I can change to single or bold, just wanted to highlight

text/66-self-managed-stack-sets.md
linsona: Will add

text/66-self-managed-stack-sets.md
linsona: Got it, will update

text/66-self-managed-stack-sets.md
linsona: Good question! Our original use case was to utilize CDK to deploy to existing stack sets, and eventually to create new stack sets in the future. We originally evaluated CFNStackSet, but the way it worked was by deploying a stack with "AWS::CloudFormation::StackSet" CFN resource that deployed the stack set. We found it a little odd where we found it  to require a stack as a wrapper around stack set.

Originally we thought that importing an existing stack set into a new stack was not supported, but I just did some more testing with CFN resource imports and it does work. The caveat is, we would need to pass parameters/tags from the stack to the stack set, migrate all our stack sets to provision from stack (for our team that is upwards of 5000 stack sets), and update all our existing systems to interact with stack sets through stacks vs directly through stack sets. This migration might not be viable for our team.

I looked through a bunch of related git issues regarding stack sets, do you know what the customer had in mind when asking for "native cdk stack set support"? Trying to gauge if this design aligns with those requests, or if CDK team is recommending CFNStackSet as the pattern for deploying stack sets.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+* **Tracking Issue**: #66
+* **API Bar Raiser**: @skinny85
+
+## Working Backwards
+
+### CHANGELOG
+- feat(core): New Stack Set core type 
+- feat(assembly): New stack set artifcat type and schema
+- feat(cli): ```cdk deploy``` collects and creates/updates stack set 
+- feat(lib): API Calls/Logic for create/update stack sets
```
skinny85, line 19: This is not really a customer-facing feature, so we wouldn't tag it with `feat` in the Changelog 😉. Let's remove it from the list.

text/66-self-managed-stack-sets.md
skinny85: Yes, let's just use single backticks for these. I don't think we need to bold them.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+### README
+The ```cdk.StackSet``` construct in CDK defines AWS resources in a stack set and stack set configuration/deployment preferences. This is different from the ```cdk.CfnStackSet``` construct which creates a CFN stack that deploys a CFN stack set. The ```cdk.StackSet``` construct creates the stack set directly if it does not exist, or updates a stack set and existing stack set instances while monitoring the stack set operation.
+
+#### Usage
+To define a stack set, extend the core construct `StackSet`. Then define AWS resources in the same way as defining resources for stacks. Compared to the `Stack` construct, the `StackSet` core construct has different props specific to StackSets. The props define stack set configurations and operation preferences for when updates occur.
+
+**Example:**
+```ts
+class POCStackSet extends cdk.StackSet {
```
skinny85, line 29: I feel like the name could be a little bit better here 😉. Anyway, in the CDK, we would call this a `PocStackSet` 😃.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+    constructor(scope: cdk.App, id: string, props: cdk.StackSetProps = {}) {
```
skinny85, line 30: `scope` should be `cdk.Construct`, and `props?: StackSetProps` saves a few characters over `props: cdk.StackSetProps = {}` 😜.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+new POCStackSet(app, 'poc-stack-set',
+    {
+        faultTolerancePercentage: 10,
+        maxConcurrentPercentage: 10,
+        regionConcurrencyType: 'PARALLEL'
+        .
+        .
+        .
+     }
+);
```
skinny85, line 47:
```suggestion
new POCStackSet(app, 'poc-stack-set', {
  faultTolerancePercentage: 10,
  maxConcurrentPercentage: 10,
  regionConcurrency: cdk.StackSetConcurrency.PARALLEL,
  // ...
});
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+**Example:**
+```ts
+class POCStackSet extends cdk.StackSet {
+    constructor(scope: cdk.App, id: string, props: cdk.StackSetProps = {}) {
+        super(scope, id, props);
+
+        new Role(this, 'ExampleRole', {
+          assumedBy: new ServicePrincipal('s3.amazonaws.com')
+        });
```
skinny85, line 35:
```suggestion
        new iam.Role(this, 'ExampleRole', {
          assumedBy: new ServicePrincipal('s3.amazonaws.com'),
        });
```

Some imports at the top of the file would be helpful.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+    }
+}
```
skinny85, line 37: Let's add an empty line after this line.

text/66-self-managed-stack-sets.md
skinny85, line 22: General comment on the entire document:

Please be cognizant of the line lengths in this file (it should be readable also in its plaintext form).
I like to break lines according to sentence boundaries - it makes any future diff to this file read better than if the newlines are strewn randomly across sentences.

So:

```suggestion
The `cdk.StackSet` construct in CDK defines AWS resources in a stack set and its configuration and deployment preferences.
This is different from the `cdk.CfnStackSet` construct which creates a CloudFormation stack that deploys a CloudFormation stack set.
The `cdk.StackSet` construct creates the stack set directly if it does not exist,
or updates a stack set and existing stack set instances if it does,
while monitoring the stack set operation.
```

Same comment everywhere in the document.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
+---
+rfc pr: [#347](https://github.com/aws/aws-cdk-rfcs/pull/347)
+tracking issue: https://github.com/aws/aws-cdk-rfcs/issues/66
+---
+
+# CDK Self Managed Stack Set Support
```
skinny85, line 6: General comment in the file: it's usual in Markdown to leave an empty line after a header. I would do it here everywhere.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+        .
+        .
+        .
+     }
+);
+```
+
+**Deployment:**
+
+To deploy the stack set, the CLI command remains the same: ```cdk deploy```. If the stack set **does not exist**, then CDK will create the stack set directly in CFN. If the stack set **does exist**, then the stack set itself will be updated as well as all **existing** stack set instances. Stack sets instances are added to the stack set outside of the CDK context.
```
skinny85, line 52: Let's avoid using "CFN" here. Let's just spell it out as "CloudFormation".

Same comment everywhere in the doc.

skinny85, line 52: Let's be explicit the command can take in the name of the StackSet, like it does for Stacks: `cdk deploy poc-stack-set`.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+## FAQ
+
+### What are we launching today?
+The core construct StackSet, which will enable users to define and deploy stack sets directly from CDK.
+
+### Why should I use this feature?
+This core constuct directly creates a CFN stack set directly vs. provisioning a single stack that deploys a stack set. The deployment also updates the stack set, updates existing stack set instances, and monitors the stack set operation.
```
skinny85, line 60: Repeated "directly".

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+### What is supported?
+* Self Manged Stack Sets
+  * Create, Update, Update Existing Stack Set Instances
+* File Assets Management
```
skinny85, line 65: I think this should be simply "CDK file Assets".

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+### What is not supported?
+* Service Managed Stack Sets
+* Stack Set Instance Add/Remove through CDK
+* Stack Set deletion
+* Docker Assets
```
skinny85, line 71:
```suggestion
* CDK Docker Assets
```

skinny85: I feel like this distinction between Docker and file Assets is not a great idea. I would probably say that either both of them are supported, or neither is, and I'm leaning toward "neither" right now, because of how complex this will be.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+CDK does not support the creation and updating of CFN stack sets directly, only through an intermediate single stack. This would also enable customers with existing stack sets to utlize CDK to update in the future.
+
+### Why should we _not_ do this?
+CDK does provide the ```cdk.CFNStackSet``` construct, but requires an intermediate single stack to deploy the stack set which does not monitor stack set operations on update.
+
+### What changes are required to enable this change?
+Changes are required since CDK currently does not have the concept/logic of deploying stack sets directly.
+
+### Is this a breaking change?
+No - this change does not require customers with existing apps to update code to consume the new CDK version.
```
skinny85, line 84: Just "no" is fine here 😉.

```suggestion
No.
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+The handling of file assets is more complex than a normal stack CDK app since stack set instances are deployed independently from CDK context and exist in different accounts and regions. This is dicussed more in the design details.
+
+### What alternative solutions did you consider?
+An alternative is the usage of ```cdk.CFNStackSet```, which is not ideal for customers with existing stack sets or need to monitor stack set operations during updates.
+
+### What is the high level implementation plan?
+At a high level, the implementation will be done in two phases:
+* Bootstrapless StackSet
+  * Create StackSet core construct
+  * Create StackSet artifact type for manifest
```
skinny85, line 96:
```
  * Create StackSet artifact type in the Cloud Assembly
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+## Out of Scope Features
+* Delete Self-Managed Stack Set
+* Service Managed Stack Set
+* Add/Remove stack set instances through CDK
+* Update to only specific stack set instances
+* Stack Set Instance Level Parameter Overrides
+
+## Assumptions
+* Management Account contains Stack Set administration role and child accounts contains Stack Set execution role for enabling stack set operations
```
skinny85, line 127: Not sure I understand this assumption... this should be done as part of `cdk bootstrap` of the child accounts, right?

If there are changes we need to make to `cdk bootstrap` to support this, they should be specified in this RFC.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+  * How file assets are defined
+  * How bootstrapping works with stack sets
+  * How file assets are made available to accounts “unmanaged” by CDK and in regions that the stack set deploys to
+
+## [1] CLI Integration - Stack/Stack Set Collection
+## High Level CLI/Deploy Flow
+![Architecture Diagram](../images/self-managed-stack-sets/high_level_cli_deploy_flow.jpg)
+1. CLI calls list/synth/deploy
+1. Subprocess invokes CDK App
+    1. StackSet construct generates template and artifact is added to manifest as a new artifact type
```
skinny85, line 162: ~~manifest~~ -> Cloud Assembly

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+1. CLI reads manifest
+1. CLI deploys to AWS environment
```
skinny85, line 164: Numbers got screwed up here.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+  * Under the hood, there will need to be some branching logic between types and changes to stacks collection since the current code is tightly tied to stack artifact types
+
+**Option 2:** Add Separate commands or flag to distinguish between stacks/stack sets
+
+* **Pros**:
+  * Isolates stack sets to its own flows -- less risk of breaking stacks flow
+* **Cons**:
+  * Odd user experience since the user will need to think about when to use commands for stack vs stack sets
+  * Some code duplication, since the stack set flow is nearly identical, with some differences in input parameters
+  * Commands will need to be ran twice if the CDK app contains both stacks/stack sets
```
skinny85, line 193: Let's go with Option 1, and remove Option 2.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+  topLevelStacks: T[],
+  extend: ExtendedStackSelection = ExtendedStackSelection.None): StackCollection<T> {
+  if (topLevelStacks.length > 0) {
+    return this.extendStacks(topLevelStacks, stacks, extend);
+  } else {
+    throw new Error('No stack found in the main cloud assembly. Use "list" to print manifest');
+  }
+}
+``` 
+**Option 2:** Duplicate cloud assembly functions explicitly for stack sets
```
skinny85, line 219: Let's go with Option 1 here, and remove Option 2.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+* **Pros**:
+  * Isolates code for stack vs stack sets, so less risk of breaking stack flows
+* **Cons**:
+  * Code duplication and more risk of diff between nearly identical functions
+
+## [2] Construct Changes (CDK app) - Stack Set Artifact Type Synthesis
+### [2.1] CDK Artifacts/Manifest
+CDK CLI reads the manifest file after a CDK app is invoked (or directly from a previously generated manifest), and utilizes those directives in the manifest to build Artifact objects for deployment actions. The manifest should include all assets to be deployed as well as configurations on how those assets should be deployed. A new artifact should be created and added to the manifest:
```
skinny85, line 228: Substitute "manifest" with "Cloud Assembly" everywhere in the document.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+    {
+        faultTolerancePercentage: 10,
+        maxConcurrentPercentage: 10,
+        regionConcurrencyType: 'PARALLEL'
+        .
+        .
+        .
+     }
+);
+```
```
skinny85, line 277: Same comments here for this code example as above.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+**Synthesis Flow:**
+
+![Architecture Diagram](../images/self-managed-stack-sets/cdk_stack_set_construct_synth_flow_option_1.jpg)
+* User extends the core StackSet construct, and defines infrastructure resources
+* A new synthesizer adds the new stack set artifact type to the manifest
```
skinny85, line 282: Not sure why do we need a new synthesizer? Why not modify the existing one(s) to handle this?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+* **Pros**:
+    * Unified user experience for stack/stack set that aligns with normal CDK patterns 
+    * No need to pass a stack into a stack set construct, users just extend core.StackSet
+
+* **Cons**:
+    * Potentially Large/Complex changes to overall system (Everything that utilizes Stack type may need to be converted)
+    * Potentially some areas where stacks/stack-sets do not share same properties, so branch logic required
+
+**Option 2:** Create a normal “core.Stack“ construct and pass to a new ”core.StackSet“ construct
```
skinny85, line 294: Let's go with Option 1 here, and remove Option 2.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+* Stacks and Stack Sets are collected from the manifest
+* Branch between deploying stack vs stack set
+* Prepare deployment and assets
+* Make API calls to deploy stack sets
+
+### [3.1] cdk.ts Changes
+This is the initial entry point for the CLI. Both stacks and stack sets will utilize this the same deploy commands without branching. 
+
+### [3.2] cdktoolkit.ts Changes
+This package is responsible for collecting the stacks and assets from the manifest. Changes:
```
skinny85, line 352: "package" used to mean "file" I'm pretty sure.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+* The deploy command will need to include a call described in section 1.2 to also retrieve stack sets
+  * const stacksSets = await this.selectStackSetssForDeploy(options.selector, options.exclusively);
+* From this point, the flow will branch between stacks and stack sets.
+
+### [3.2] cloudformation-deployments.ts Changes
+This package is responsible for preparing the deployment, orchestrating asset publishing, and infrastructure deployment. Changes:
```
skinny85, line 358: "package" used to mean "file" I'm pretty sure.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+* New function for orchestrating asset deployment and infrastructure deployment for stack sets. More on asset deployment describe in the Lambda assets section.
```
skinny85, line 360:
```suggestion
* New function for orchestrating asset deployment and infrastructure deployment for stack sets.
More details on asset deployment are in the file assets section below.
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+![Architecture Diagram](../images/self-managed-stack-sets/low_level_stack_set_deploy.jpg)
+* Check if the stack set exists
+  * If not, then create
+* If the stack set does exist, check if there are any stack set operations in progress and wait until complete
+* When the stack set is ready to accept operations, call update stack set
+  * When calling update stack set API, no accounts or regions will be supplied, which will trigger an update on the stack set itself and all stack set instances
+* Wait for the update stack set operation to complete
+
+## [4] File Assets Management
+In a normal stack deployment, artifacts are defined within specific AWS resource constructs such as lambda or just the assets construct in general. They are published to S3 and referenced in the generated template by parameters (LegacyStackSynthesizer) or direct injection (DefaultStackSynthesizer). For Stack Sets, the same behavior is ideal, but there are a few more caveats:
```
skinny85, line 378:
```suggestion
In a normal stack deployment, artifacts are defined within specific AWS resource constructs such as lambda or just the assets construct in general.
They are published to S3 and referenced in the generated template by parameters (`LegacyStackSynthesizer`) or direct injection (`DefaultStackSynthesizer`).
For Stack Sets, the same behavior is ideal, but there are a few more caveats:
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+
+### [4.1] Bootstrapping
+
+CDK apps that utilize assets require “Bootstrap” infrastructure for storing and publishing assets. This can either be done using the cdk bootstrap command which will deploy a default stack with all the required infrastructure or a customized template. 
+
+### Permissions
+Since stack set instances are deployed outside the CDK context and can be deployed to “external” accounts, child accounts need permissions to retrieving assets from the centralized staging bucket in the management account. This can be achieved by adding a bucket policy to the staging buckets:
+
+```json
+{
```
skinny85, line 403: Can you show the entire resource that you're adding to the bootstrap template here?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+  {synthesizer: new DefaultStackSynthesizer(
+    {fileAssetsBucketName: 'cdk-${Qualifier}-assets-000000000000-${AWS::Region}'})});
+const stackSet = new cdk.StackSet(app, 'stack-set-poc',
+  {
+    stack: stackDef,
+    faultTolerancePercentage: 10,
+    maxConcurrentPercentage: 10,
+    regionConcurrencyType: 'PARALLEL'
+  }
+);
```
skinny85, line 446: Let's move this example to the "nice" API.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,502 @@
...
+**Option 2:** Bucket replication to all regions
+Instead of publishing assets directly to every regional staging bucket, assets are replicated using S3 replication from the staging bucket in the region where the stack set is deployed to all the other regional staging buckets.
+
+* **Pros**:
+  * CLI does not have to do the work of copying files locally to multiple regions
+* **Cons**:
+  * Potential for replication lag since it is asynchronous. Most of the time it will replicate within seconds, but has an 99.99% 15 min SLA. A mitigation to this is making API calls to verify objects have been replicated before moving on.
+  * New regions will need to manually copy latest artifacts over. Bucket replication only works when new objects are added, and when deploy is ran, the asset will not be re-upload if it already exists
+
+# Implementation Plan
```
skinny85, line 490: I feel like this section is duplicated from above, and can be safely removed.

skinny85 (changes requested):
This is a great first attempt @linsona!

Some comments from me, but the general direction looks really promising!

text/66-self-managed-stack-sets.md
linsona: Ah gotcha (my ide was text wrapping), will fix

text/66-self-managed-stack-sets.md
linsona: Make sense, so I ended up moving this to not supported section. Thinking that this would be handled in a separate RFC?

text/66-self-managed-stack-sets.md
linsona: Yeah, so I'm on the fence. This is a resource that "belongs" to cloud formation stack set, vs something that is specific to cdk. Customer may already have the stack set administration role created, or could create through a stack in CDK.

As for the stack set execution role, this exists in the child account, which CDK would not have context about

text/66-self-managed-stack-sets.md
linsona: Remove -- yeah added originally to have an opposing alternative

text/66-self-managed-stack-sets.md
linsona: This was just removed since it was part of option 2

text/66-self-managed-stack-sets.md
linsona: So, I ended up moving file asset in general to not supported.

text/66-self-managed-stack-sets.md
linsona: So writing a POC, I ended up extending `DefaultStackSynthesizer` as `BootstraplessStackSetSynthesizer`

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+* **API Bar Raiser**: @skinny85
+
+## Working Backwards
+
+### CHANGELOG
+
+- core: New Stack Set core type 
+- assembly: New stack set artifcat type and schema
+- cli: `cdk deploy` collects and creates/updates stack set 
+- lib: API Calls/Logic for create/update stack sets
```
skinny85, line 21: Like I said before, this should be removed.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+* **Tracking Issue**: #66
+* **API Bar Raiser**: @skinny85
+
+## Working Backwards
+
+### CHANGELOG
+
+- core: New Stack Set core type 
+- assembly: New stack set artifcat type and schema
+- cli: `cdk deploy` collects and creates/updates stack set
```
skinny85, line 20: These should follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+- core: New Stack Set core type 
+- assembly: New stack set artifcat type and schema
+- cli: `cdk deploy` collects and creates/updates stack set 
+- lib: API Calls/Logic for create/update stack sets
+
+### README
+
+The `cdk.StackSet` construct in CDK defines AWS resources in a stack set and stack set
+ configuration/deployment preferences. 
+ This is different from the `cdk.CfnStackSet` construct which creates a CloudFormation stack that deploys a CloudFormation stack set.
```
skinny85, line 27: I think this is not precisely what happens.

```suggestion
 This is different from the `cdk.CfnStackSet` construct, which requires a CloudFormation stack that then deploys a CloudFormation stack set. 
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+The `cdk.StackSet` construct in CDK defines AWS resources in a stack set and stack set
+ configuration/deployment preferences. 
+ This is different from the `cdk.CfnStackSet` construct which creates a CloudFormation stack that deploys a CloudFormation stack set. 
+ The `cdk.StackSet` construct creates the stack set directly if it does not exist, or updates a stack set and existing stack set instances while monitoring the stack set operation.
+
+#### Usage
+
+To define a stack set, extend the core construct `StackSet`. 
+Then define AWS resources in the same way as defining resources for stacks. 
+Compared to the `Stack` construct, the `StackSet` core construct has different props specific to StackSets.
```
skinny85, line 34:
```suggestion
Compared to the `Stack` construct, the `StackSet` core construct has different properties specific to StackSets. 
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+ configuration/deployment preferences. 
+ This is different from the `cdk.CfnStackSet` construct which creates a CloudFormation stack that deploys a CloudFormation stack set. 
+ The `cdk.StackSet` construct creates the stack set directly if it does not exist, or updates a stack set and existing stack set instances while monitoring the stack set operation.
+
+#### Usage
+
+To define a stack set, extend the core construct `StackSet`. 
+Then define AWS resources in the same way as defining resources for stacks. 
+Compared to the `Stack` construct, the `StackSet` core construct has different props specific to StackSets. 
+The props define stack set configurations and operation preferences for when updates occur.
```
skinny85, line 35:
```suggestion
The properties define the configuration of the  stack set and the preferences used for update operations.
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+
+**Example:**
+```ts
```
skinny85, line 38: Empty line before this line would be nice.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+* CDK Assets (CDK File Assets/CDK Docker Assets)
+  * More information on CDK asset management and complexities described in the [File Assets Management](#4-file-assets-management) section.
+
+## Internal FAQ
+
+### Why are we doing this?
+
+Today, CDK does not support the creation and updating of CloudFormation stack sets directly. This solution would also enable deploying to existing stack sets with low effort.
+
+CDK does provide the `cdk.CFNStackSet` construct, which could accomplish similar behavior.
```
skinny85, line 96: It's `cdk.CfnStackSet`, not `cdk.CFNStackSet`.

Same typo is repeated in many places in the doc.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+  * More information on CDK asset management and complexities described in the [File Assets Management](#4-file-assets-management) section.
+
+## Internal FAQ
+
+### Why are we doing this?
+
+Today, CDK does not support the creation and updating of CloudFormation stack sets directly. This solution would also enable deploying to existing stack sets with low effort.
+
+CDK does provide the `cdk.CFNStackSet` construct, which could accomplish similar behavior. 
+The difference is that the `cdk.CFNStackSet` construct provisions a single CloudFormation stack with the resource type `AWS::CloudFormation::StackSet`, and then the stack provisions the actual stack set.
```
skinny85, line 97:
```suggestion
The difference is that the `cdk.CfnStackSet` construct provisions a single resource with the type `AWS::CloudFormation::StackSet`,
and then the stack that that resource belongs provisions the actual stack set.
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+Customer can deploy to existing stack sets, but requires some migration steps for importing an existing stack set into the resource `AWS::CloudFormation::StackSet` before before being able to use `cdk.CFNStackSet`.
+
+### Why should we _not_ do this?
+
+We might not want to do this since CDK does provide the `cdk.CFNStackSet` construct, which could accomplish similar behavior. 
+The intermdiate stack may not be ideal for some customers, and may require large migration effort for customers who manage many stacksets.
+
+### What changes are required to enable this change?
+
+To enable this feature, we would need to introduce a new StackSet construct for customers to extend, a new assembly/artifact type for the Cloud Assembly, stack set synthesizer, and deployment code that creates/updates stack sets using AWS CloudFormation APIs.
```
skinny85, line 108: This line is very long, let's break it up.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Is this a breaking change?
+
+No.
+
+### What are the drawbacks of this solution?
+
+The handling of file assets is more complex than a normal stack CDK app since stack set instances are deployed independently from CDK context and exist in different accounts and regions.
```
skinny85, line 116: Same here (very long line).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+### What alternative solutions did you consider?
+
+An alternative is the usage of `cdk.CFNStackSet`, which is not ideal for customers with existing stack sets since there is significant migration effort.
+
+### What is the high level implementation plan?
+* Create StackSet core construct
+* Create StackSet artifact type in the Cloud Assembly
+* Create Bootstrapless StackSet Synthesizer
+* Update CLI to collect StackSet artifact types in addition to Stacks
+
```
skinny85, line 128: I feel like we're missing the step where we add the calls to the CloudFormation APIs in the CLI to actually create/update the StackSets.

text/66-self-managed-stack-sets.md
skinny85: Sorry,  I still need more details here.

- How is the management account administration Role created? Manually? Through the CDK? How do we set in the CDK what Role is used?
- How is the execution Role created? Manually? Using CDK?
- Do we require that an account+region be `cdk bootstrap`ped before a StackSet instance of a StackSet managed through the CDK can be deployed to it? Why yes/no?
- How does the CloudFormation service know which Role to use when performing the StackSet instance deployment? Can that be set in the CDK?

All of these questions should be answered here.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+* Service Managed Stack Set
+* Add/Remove stack set instances through CDK
+* Update to only specific stack set instances
+* Stack Set Instance Level Parameter Overrides
+* CDK Assets (CDK File Assets/CDK Docker Assets)
+  * More information on CDK asset management and complexities described in the [File Assets Management](#4-file-assets-management) section.
+
+## Assumptions
+
+* Management Account contains Stack Set administration role and child accounts contains Stack Set execution role for enabling stack set operations
```
skinny85, line 160: Can we settle on the terminology "parent account" and "child account" here? " Management Account" just sounds weird (and is also weirdly capitalized).

text/66-self-managed-stack-sets.md
skinny85: Numbers are still screwed up here.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+At a high-level, the flow for stacks vs stack sets are nearly identical with a few different configuration parameters and end deployment logic.
+
+Commands can be ran as the are today, and under the hood logic will be branched based on artifact type.
+
+* **Pros**:
+  * Unified customer experience for deploying infrastructure
+  * Allows deployment of stacks/stack sets in the same execution
+* **Cons**:
+  * Risk of breaking current stacks flow
+  * Under the hood, there will need to be some branching logic between types and changes to stacks collection since the current code is tightly tied to stack artifact types
```
skinny85, line 222: Let's remove the pros & cons list (we've already made a decision).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+CLI commands retrieve stacks from the Cloud Assembly. The same logic is necessary for a new stack set artifact type, but they are typed for only for *CloudFormationStackArtifact* objects.
+
+A new abstract class **CloudFormationArtifact** would be created which **CloudFormationStackArtifact (current)** and **CloudFormationStackSetArtifact (new)** will inherit. 
+Then in the CloudAssembly convert to functions to generics based on the **CloudFormationArtifact** type.
+
+* **Pros**:
+    * Less code duplication
+    * Less risk of differences in retrieval if updates are made to CloudAssembly for stack/stack set retrieval
+* **Cons**:
+    * Risk of breaking of how stacks work today
```
skinny85, line 235: Again, no need for the list - the decision has been made.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+  }
+}
+```
+
+## [2] Construct Changes (CDK app) - Stack Set Artifact Type Synthesis
+
+### [2.1] CDK Artifacts/Cloud Assembly
+
+CDK CLI reads the Cloud Assembly after a CDK app is invoked (or directly from a previously generated Cloud Assembly), and utilizes those directives in the Cloud Assembly to build Artifact objects for deployment actions. 
+The Cloud Assembly should include all assets to be deployed as well as configurations on how those assets should be deployed. A new artifact should be created and added to the Cloud Assembly:
```
skinny85, line 255: Be mindful of the line length please.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+        new iam.Role(this, 'ExampleRole', {
+          assumedBy: new iam.ServicePrincipal('s3.amazonaws.com')
+        });
+    }
+}
+
+new PocStackSet(app, 'poc-stack-set', {
+  faultTolerancePercentage: 10,
+  maxConcurrentPercentage: 10,
+  regionConcurrencyType: 'PARALLEL'
```
skinny85, line 55: This should be an enum. Not a string.

```suggestion
  regionConcurrency: cdk.StackSetRegionConcurrency.PARALLEL,
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
```
skinny85, line 304: Same note as above.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,505 @@
...
+
+![Architecture Diagram](../images/self-managed-stack-sets/low_level_stack_set_deploy.jpg)
+* Check if the stack set exists
+  * If not, then create
+* If the stack set does exist, check if there are any stack set operations in progress and wait until complete
+* When the stack set is ready to accept operations, call update stack set
+  * When calling update stack set API, no accounts or regions will be supplied, which will trigger an update on the stack set itself and all stack set instances
+* Wait for the update stack set operation to complete
+
+## [4] File Assets Management
```
skinny85, line 369: Let's rename this section "CDK Assets handling" (Assets are not managed).

text/66-self-managed-stack-sets.md
skinny85: Still no resource here.

text/66-self-managed-stack-sets.md
skinny85: Still here, still not using the "nice" API (the indentation is also weird).

skinny85 (changes requested):
This is shaping up very nicely!

A few comments still.

text/66-self-managed-stack-sets.md
linsona: Talked offline, required for emitting additional artifact type and properties

skinny85: If you could put in the RFC that justification why is a new synthesizer needed, that would be great 🙂.

text/66-self-managed-stack-sets.md
linsona: Added

text/66-self-managed-stack-sets.md
linsona: Moved to "nice" API

text/66-self-managed-stack-sets.md
linsona: Reformatted

text/66-self-managed-stack-sets.md
linsona: Fixed in all occurences

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+* Stack Set Instance Level Parameter Overrides
+* CDK Assets (CDK File Assets/CDK Docker Assets)
+  * More information on CDK asset handling and complexities described in the [CDK Assets Handling](#4-cdk-assets-handling) section.
+
+## Assumptions
+
+* Parent account contains Stack Set administration role and child accounts contains Stack Set execution role for enabling stack set operations
+  * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacksets-prereqs-self-managed.html
+  * The Stack Set administration role in the parent account is used by CloudFormation to assume the execution role in the child account. This role can be created manually or may be created through a CDK stack. This `role arn` is passed into the stack set CDK construct properties 
+  * The Stack Set execution role in the child is the role assumed by the administration role in the parent account, and utilized to provision resources in the child account. This role should be provisioned manually, and the `role name` is passed into the stack set CDK construct properties
```
skinny85, line 170: I assume this was supposed to read "child account".

skinny85: Again, same as above, this is a gigantic bombshell that is not mentioned anywhere in the customer-facing docs of this RFC.

All of this information needs to be moved all the way up to the "working backwards" section of this RFC.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+* Stack Set Instances
+  * In this initial release, stack set instances should be added manually to stack set after it was initially created by CDK
+  * Since this initial iteration does not include CDK assets support such as files/docker images, bootstrapping multiple regions is not required. Bootstrapping a single region in the parent account may be required for templates larger than 50kb.
```
skinny85, line 173: Line length again.

text/66-self-managed-stack-sets.md
skinny85: Can you explain why wouldn't we add this Role to `cdk bootstrap`, and require a child account be `cdk bootstrap`ped before a StackSet instance of a StackSet managed by the CDK can be created inside of it?

That seems better than requiring a manually-provisioned Role, no?

text/66-self-managed-stack-sets.md
skinny85: Not sure I understand what you mean by "Bootstrapping a single region"...? You will need to `cdk bootstrap` in every region you're creating a StackSet with a template > 50 KB, right?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+
+* Implementation does not incur breaking changes for existing users or future feature implementation
+  * I.e Current CDK users should not have to convert any other CDK app implementations to accommodate for changes implemented in this design
+
+* Implementation is extensible for future features such as:
+  * Self-managed Stack Sets
+  * Stack Set Instance Add/Remove
+  * Additional asset handling
+
+# Design
```
skinny85, line 185: I'm wondering whether this section is useful here. I feel like it goes into a lot of detail that we generally don't go into in the RFC.

@linsona do you want our feedback on some specific part of this section? Because if not, I actually think we should remove it, and this should be just a shared document that's used inside of your team.

skinny85 (changes requested):
Next round of comments. We definitely need to iron out the Roles that were surfaced in this iteration - that seems to be a huge thing that's completely omitted from the current RFC. Also, let's expand the current "working backwards" section to include actually creating StackSet instances (as that's the final purpose of SackSets after all). That should illuminate what is the experience we want for the Roles here.

Also, make sure the linting pases on your document (you can run it locally to make sure it doesn't fail).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+  regionConcurrencyType: cdk.StackSetRegionConcurrency.PARALLEL,
+  //...
+});
+```
+
+**Deployment:**
+
+To deploy the stack set, the CLI command remains the same: `cdk deploy poc-stack-set`. 
+If the stack set **does not exist**, then CDK will create the stack set directly in CloudFormation. If the stack set **does exist**, then the stack set itself will be updated as well as all **existing** stack set instances. 
+Stack sets instances are added to the stack set outside of the CDK context.
```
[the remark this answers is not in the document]
linsona: Yeah -- added to assumptions, but would be a good idea to add here as well

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+### What is supported?
+
+* Self Manged Stack Sets
+  * Create, Update, and Update Existing Stack Set Instances directly through CDK
+
+### What is not supported?
+
+* Service Managed Stack Sets
+* Stack Set Instance Add/Remove through CDK
+* Stack Set deletion
```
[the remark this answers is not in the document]
linsona: Trying to think about the behavior here since a stack set must delete the stack set instances before deleting the stack set. I think we can include in this iteration, and treat failure to delete instances in the same pattern as a resource failing to delete for a single stack

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,537 @@
...
+* Update to only specific stack set instances
+* Stack Set Instance Level Parameter Overrides
+* CDK Assets (CDK File Assets/CDK Docker Assets)
+  * More information on CDK asset handling and complexities described in the [CDK Assets Handling](#4-cdk-assets-handling) section.
+
+## Assumptions
+
+* Parent account contains Stack Set administration role and child accounts contains Stack Set execution role for enabling stack set operations
+  * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacksets-prereqs-self-managed.html
+  * The Stack Set administration role in the parent account is used by CloudFormation to assume the execution role in the child account. This role can be created manually or may be created through a CDK stack. This `role arn` is passed into the stack set CDK construct properties
```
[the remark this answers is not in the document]
linsona: Yeah so the stack set administration role for CloudFormation API is optional property. Optional in a sense that CloudFormation assume the user has created the role with default name: `AWSCloudFormationStackSetAdministrationRole`

Otherwise you can specify in the request for a role with a customized role name. I can make this more explicit in the ReadMe section

text/66-self-managed-stack-sets.md
linsona: Yeah, my thought process for the stack set administration role in the parent account was it is a property of the CloudFormation stack set vs a property of CDK. That being said, I think customer could provision the stack set administration role through CDK as a single stack.

As for the stack set execution role, this one is kind of odd in a sense that it is managed outside of the CDK context. If a customer defines all the child accounts in their CDK app, then they could provision the execution role from CDK, but question might be what is the value of stack sets.

text/66-self-managed-stack-sets.md
linsona: No, so when a stack set is created/update with a template (either inline or through S3), the CloudFormation service itself stores the template and manages the templates. I.e. stack set instances in child accounts don't read from the bootstap bucket

skinny85: I'm not talking about the StackSet instances though. I'm talking about StackSets themselves (the parent account).

linsona: hmmm, if you deploy a stack set in different regions then yes. Generally, a user would only deploy a stack set in a single region which deploys stack set instances to different accounts/regions.

skinny85: Let's be precise then.

"The parent account needs to be bootstrapped with the `cdk bootstrap` command in each region that a StackSet is to be created in."

Or something like that. In general, the whole point of the RFC is to precisely describe the behavior and requirements of the new functionality.

nija-at:
RFC Bar Raiser: @eladb

text/66-self-managed-stack-sets.md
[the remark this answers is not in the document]
linsona: Talked offline, in this iteration, delete will not be supported  and will add reasoning of why

text/66-self-managed-stack-sets.md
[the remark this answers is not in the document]
linsona: Talked offline, will include more details and bootstrap workflow for parent and child accounts

text/66-self-managed-stack-sets.md
linsona: Talked offline, will include more details and bootstrap workflow for parent and child accounts

text/66-self-managed-stack-sets.md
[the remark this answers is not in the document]
linsona: Adding a prereqs section

text/66-self-managed-stack-sets.md
linsona: Removing the design section

text/66-self-managed-stack-sets.md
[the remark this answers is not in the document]
linsona: Talked offline, scoping out for this iteration since requires collecting all stack set instances and deleting stack set instances before deleting stack set. Add list reasons why its not being added

text/66-self-managed-stack-sets.md
linsona: Talked offline on how we can use bootstrap for default stack set roles

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,196 @@
...
+
+### README
+
+The `cdk.StackSet` construct in CDK defines AWS resources in a stack set and stack set
+ configuration/deployment preferences.
+ This is different from the `cdk.CfnStackSet` construct, which requires a CloudFormation stack that then deploys a CloudFormation stack set.
+ The `cdk.StackSet` construct creates the stack set directly if it does not exist,
+or updates a stack set and existing stack set instances while monitoring the stack set operation.
+
+#### Prerequisites
```
skinny85, line 32: Let's divide this section a little more finely. Let's say that there are a few ways you can prepare your parent and child accounts ford deploying StackSets.

1. You can use the `cdk bootstrap` command. Show the exact invocations you need in both the parent and child accounts. And show how it looks in the code.
    - Describe how you can pass the `--qualifier` option to `cdk bootstrap`, and how is that reflected in the code (perhaps there's a `qualifier` property when creating a `cdk.StackSet`?).
2. You can create the Roles manually. Describe what the Roles have to be (names, trust policies permission policies). Then, show how is that reflected in the code (probably properties like `administratorRoleName` and `executionRoleName` of `cdk.StackSet`).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,196 @@
...
+
+Using CDK Bootstrap to provision default cdk stack set roles:
+
+* Parent Account:
+  * Using the parent account credentials and region, run `cdk bootstrap`
+  * Specific stack sets resources in bootstrap:
+    * A staging bucket to store templates larger than 50kb
+    * A default stack set administration role
+      * `cdk-${Qualifier}-stack-set-admin-role-{AWS::AccountId}-${AWS::Region}`
+      * Ex. cdk-hnb659fds-stack-set-admin-role-111111111111-us-east-1
```
skinny85, line 50: Let's skip these details (remember, this is the customer-facing documentation - you don't have to explain all of the implementation details, you just need to provide instructions on how to use it).

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,196 @@
...
+* Child Accounts:
+  * Using the child account credentials and region, run `cdk bootstrap --trust <parent account id>`
+  * Specific stack sets resources in bootstrap:
+    * A default stack set execution role
+      * `cdk-${Qualifier}-stack-set-exec-role-{AWS::AccountId}-${AWS::Region}`
+      * Ex. cdk-hnb659fds-stack-set-exec-role-222222222222-us-east-1
```
skinny85, line 56: Same as above.

skinny85 (changes requested):

text/66-self-managed-stack-sets.md
linsona: Broke up the section into those two methods

text/66-self-managed-stack-sets.md
linsona: Removed -- added appendix section with the bootstrap resource additions

text/66-self-managed-stack-sets.md
linsona: Removed -- added appendix section with the bootstrap resource additions

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+          - Action:
+              - sts:AssumeRole
+            Resource:
+              - Fn::Sub: "arn:aws:iam::*:role/cdk-${Qualifier}-stack-set-exec-role-${AWS::Region}"
+            Effect: Allow
+        Version: '2012-10-17'
+      Roles:
+        - Ref: CloudFormationStackSetAdministrationRole
+      PolicyName:
+        Fn::Sub: cdk-${Qualifier}-stack-set-admin-role-policy-${AWS::Region}
```
skinny85, line 246: Can we switch this to an inline policy on the `CloudFormationStackSetAdministrationRole`?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+            Principal:
+              AWS:
+                Fn::GetAtt: CloudFormationStackSetAdministrationRole.Arn
+          - Fn::If:
+              - HasTrustedAccounts
+              - Action: sts:AssumeRole
+                Effect: Allow
+                Principal:
+                  AWS:
+                    Ref: TrustedAccounts
```
skinny85, line 263: Since this Role has Admin permissions (effectively), can you instead make it trust the `CloudFormationStackSetAdministrationRole` from the other accounts instead?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+          - HasCloudFormationExecutionPolicies
+          - Ref: CloudFormationExecutionPolicies
+          - Fn::If:
+            - HasTrustedAccounts
+            # The CLI will prevent this case from occurring
+            - Ref: AWS::NoValue
+            # The CLI will advertise that we picked this implicitly
+            - - Fn::Sub: "arn:${AWS::Partition}:iam::aws:policy/AdministratorAccess"
+      RoleName:
+        Fn::Sub: cdk-${Qualifier}-stack-set-exec-role-${AWS::Region}
```
skinny85, line 276: So, I see that `CloudFormationStackSetAdministrationRole` has `${AWS::AccountId}` in its name, but this one doesn't.

Why the discrepancy?

skinny85 (changes requested):
We're getting there! These are basically cosmetic issues, I think we've pretty much settled on the customer experience.

skinny85:
Also, the linter is still failing 😉.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
+---
```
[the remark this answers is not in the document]
linsona: Yep, good call out

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+  synthesizer: new cdk.StackSetSynthesizer({ qualifier: 'xyz' }),
+  faultTolerancePercentage: 10,
+  maxConcurrentPercentage: 10,
+  regionConcurrencyType: cdk.StackSetRegionConcurrency.PARALLEL,
+  //...
+});
+
+// Using customized stack set roles
+new PocStackSet(app, 'poc-stack-set-3', {
+  administrationRoleArn: 'arn:aws:iam::111111111111:role/AWSCloudFormationStackSetAdministrationRole',
```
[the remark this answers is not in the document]
linsona: yeah, so this follows the CloudFormation API for create stack set, but we could build the arn based on the CDK creds context.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+new PocStackSet(app, 'poc-stack-set-1', {
+  faultTolerancePercentage: 10,
+  maxConcurrentPercentage: 10,
+  regionConcurrencyType: cdk.StackSetRegionConcurrency.PARALLEL,
+  //...
+});
+
+// Using default cdk bootstrap stack set roles with custom qualifier
+new PocStackSet(app, 'poc-stack-set-2', {
+  synthesizer: new cdk.StackSetSynthesizer({ qualifier: 'xyz' }),
```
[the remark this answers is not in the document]
linsona: Yeah was on the fence on this once since this is the pattern for changing qualifier for normal single stack construct. But yeah, can take in qualifier directly

text/66-self-managed-stack-sets.md
linsona: Yep, will do -- was just following the pattern from the other IAM role resources

text/66-self-managed-stack-sets.md
linsona: So this is the stack set execution role in the child accounts, which must have the same name across all child accounts. (Adding account id would break that)

text/66-self-managed-stack-sets.md
linsona: yeah, I can test that out today

text/66-self-managed-stack-sets.md
skinny85: Ah, of course. This is probably worth a comment in the template body.

text/66-self-managed-stack-sets.md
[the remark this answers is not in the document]
skinny85: I think we have to (think of the case where the StackSet is env-agnostic).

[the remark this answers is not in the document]
skinny85: Given that, let's change this to also be a name.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,278 @@
...
+
+There are a few methods to creating the stack set roles:
+
+1. CDK Bootstrap with default stack set roles
+   1. Bootstrap Parent Account
+      1. `cdk bootstrap aws://11111111111/us-east-1`
+   1. Bootstrap Child Account
+      1. `cdk bootstrap aws://22222222222/us-east-1 --trust 11111111111 --cloudformation-execution-policies arn:aws:iam::aws:policy/AdministratorAccess`
+      1. **Note:** The child account should be bootstrapped to the same region where the parent account is bootstrapped.
+   1. Specifying custom qualifier:
```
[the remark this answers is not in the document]
linsona: Updated

text/66-self-managed-stack-sets.md
linsona: Talked offline -- we decided to update bootstrap CLI to to generate the ARNs. There is no way using native CloudFormation intrinsic functions to transform account ids to the specific role ARN format we require.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+### CHANGELOG
+
+* feat(assembly): add stack set artifact type and schema
+* feat(core): add `StackSet` construct
+* feat(cli): CLI commands like `deploy` now support stack sets
+* feat(bootstrap): add IAM Roles for performing CDK stack set deployments
+
+### README
+
+The `cdk.StackSet` construct in CDK defines AWS resources in a stack set and stack set
```
eladb, line 26: Please replace "stack sets" (or any variation thereof) with "StackSets" (one word) across this doc. I believe this is the how it is used by the service documentation.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+ configuration/deployment preferences.
+ This is different from the `cdk.CfnStackSet` construct, which requires a CloudFormation stack that then deploys a CloudFormation stack set.
+ The `cdk.StackSet` construct creates the stack set directly if it does not exist,
+or updates a stack set and existing stack set instances while monitoring the stack set operation.
+
```
eladb, line 31: Add some information about what StackSets are (from the official CFN docs) and a reference to the official docs entry point

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+#### Prerequisites
+
+To enable the use of self managed stack sets two roles must exist:
+
+* **Stack Set Administration Role:** Role in the parent account, which is used by CloudFormation to assume the stack set execution role in the child account.
+* **Stack Set Execution Role:** Role in the child accounts, which allows the stack set administration role in parent account to assume,
+and perform the stack set instance deployment.
```
eladb, line 38: These roles sounds very much like the ones we use for CDK pipelines. Can't we reuse them?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+Compared to the `Stack` construct, the `StackSet` core construct has different properties specific to StackSets.
+The properties define the configuration of the stack set and the preferences used for update operations.
+
+**Example:**
+
+```ts
+import * as cdk from '@aws-cdk/core';
+import * as iam from '@aws-cdk/aws-iam';
+
+class PocStackSet extends cdk.StackSet {
```
eladb, line 80: Our convention is not to use namespaced imports for core types:

```suggestion
class MyStackSet extends StackSet {
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+    constructor(scope: cdk.Construct, id: string, props?: cdk.StackSetProps) {
+        super(scope, id, props);
+
+        new iam.Role(this, 'ExampleRole', {
+          assumedBy: new iam.ServicePrincipal('s3.amazonaws.com')
+        });
```
eladb, line 86: Change to 2 space indentation please

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+
+```ts
+import * as cdk from '@aws-cdk/core';
+import * as iam from '@aws-cdk/aws-iam';
+
+class PocStackSet extends cdk.StackSet {
+    constructor(scope: cdk.Construct, id: string, props?: cdk.StackSetProps) {
+        super(scope, id, props);
+
+        new iam.Role(this, 'ExampleRole', {
```
eladb, line 84: Would be nice to include a bit more involved example (i.e. add a bucket or something)

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+// Using default cdk bootstrap stack set roles with custom qualifier
+new PocStackSet(app, 'poc-stack-set-2', {
+  qualifier: 'xyz',
+  faultTolerancePercentage: 10,
+  maxConcurrentPercentage: 10,
+  regionConcurrencyType: cdk.StackSetRegionConcurrency.PARALLEL,
+  //...
+});
+
+// Using customized stack set roles
```
eladb, line 107: To maintain a good flow for the document, I would put all the information about custom bootstrapping, roles and qualifiers to a separate section at the end "Customizing StackSet Bootstrapping".

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+  faultTolerancePercentage: 10,
+  maxConcurrentPercentage: 10,
+  regionConcurrencyType: cdk.StackSetRegionConcurrency.PARALLEL,
+  //...
+});
+```
+
+**Deploy Stack Set:**
+
+To deploy the stack set, the CLI command remains the same: `cdk deploy poc-stack-set`.
```
eladb, line 120: I understand this is mentioned just after, but when I read this first it wasn't clear to me that when I first deploy a stack set, it won't deploy any instances. Maybe merge the two sections about "deploy stackSets" and "creating StackSet instances" into a single section about deployment.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+      and the Stack Set Execution roles for the child accounts.
+   * The administration role and execution role name should be specified as part of the StackSet properties.
+
+#### Usage
+
+To define a stack set, extend the core construct `StackSet`.
+Then define AWS resources in the same way as defining resources for stacks.
+Compared to the `Stack` construct, the `StackSet` core construct has different properties specific to StackSets.
+The properties define the configuration of the stack set and the preferences used for update operations.
+
```
eladb, line 73: Are there any limitations on StackSets comparing to normal stacks? If there are, this is where they should be listed.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+### What is supported?
+
+* Self Manged Stack Sets
+  * Create, Update, and Update Existing Stack Set Instances directly through CDK
+
+### What is not supported?
+
+* Service Managed Stack Sets
+* Stack Set Instance Add/Remove through CDK
+* Stack Set Deletion
```
eladb, line 152: What would be the experience of running "cdk destroy"? Maybe we can support `cdk destroy` for StackSets that do not have instances?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+and import the stack set into the resource `AWS::CloudFormation::StackSet`.
+
+### Why should we _not_ do this?
+
+We might not want to do this since CDK does provide the `cdk.CfnStackSet` construct, which could accomplish similar behavior.
+The intermdiate stack may not be ideal for some customers, and may require large migration effort for customers who manage many stacksets.
+
+### What changes are required to enable this change?
+
+To enable this feature, we would need to introduce:
```
eladb, line 179: What about bootstrap stack changes?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+  * A new stack set synthesizer is required to emit new stack set artifact and properties to the CloudAssembly
+* Deployment code that creates/updates stack sets using AWS CloudFormation APIs
+
+### Is this a breaking change?
+
+No.
+
+### What are the drawbacks of this solution?
+
+The handling of CDK assets is more complex than a normal stack CDK app.
```
eladb, line 193: Sounds like assets are not going to be supported anyway

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+Stack set instances are deployed independently from CDK context, and exist in different accounts/regions.
+This is dicussed more in the design details.
```
eladb, line 195:
```suggestion
This is discussed more in the design details.
```

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+
+### What alternative solutions did you consider?
+
+An alternative is the usage of `cdk.CfnStackSet`, which is not ideal for customers with existing stack sets since there is significant migration effort.
```
eladb, line 199: I have to say that this does not sound like a good enough reason not to implement this. Migrating from existing infrastructure (either CFN-based or not) is hard in most cases...

What are other reasons for us to support this?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+* CDK asset handling
+* Future support for service managed stack sets
+* Future support for delete stack set
+* CDK Pipelines
+
+## Appendix
+
+### CDK Bootstrap CLI Changes
+
+To support stack sets,
```
eladb, line 221: Repeat from above: can we use/expand the CDK pipelines roles we already have in the bootstrap stack? Intuitively these should suffice since CDK Pipelines is basically doing the same thing as stack sets.

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+a stack set administration role should be created in the parent account and a stack set execution role should be created in the child accounts.
+To make this process easier, default roles have been added to the default bootstrap template. (Bootstrapping steps describe in the README section).
+
+**CLI Changes:**
+
+* Update bootstrap CLI to include new flag `--stack-set`
```
eladb, line 227:
```suggestion
* Update bootstrap CLI to include new flag `--stackset`
```

eladb, line 227: If possible, it would be nice if we didn't need any of this

eladb (changes requested):

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+
+To deploy the stack set, the CLI command remains the same: `cdk deploy poc-stack-set`.
+If the stack set **does not exist**, then CDK will create the stack set directly in CloudFormation.
+If the stack set **does exist**,
+then the stack set itself will be updated as well as all **existing** stack set instances.
+
+**Creating Stack Set Instances:**
+
+Before creating stack set instances, child accounts should be bootstrapped (or custom stack set roles created) and the initial stack set deployed.
+Stack set instances can be added to the stack set by console/cli/api by specifying the child account id and region.
```
nija-at, line 128: How exactly? Why not via the cdk?

text/66-self-managed-stack-sets.md
```diff
@@ -0,0 +1,329 @@
...
+Today, CDK does not support the creation and updating of CloudFormation stack sets directly.
+This solution would also enable deploying to existing stack sets with low effort.
+
+CDK does provide the `cdk.CfnStackSet` construct, which could accomplish similar behavior.
+The difference is that the `cdk.CfnStackSet` construct is a resource in a single CloudFormation stack with the resource type `AWS::CloudFormation::StackSet`,
+where stack provisions the actual stack set.
+This requires a 1:1 mapping of parameters and tags that need to be passed from the stack to the stack set.
+Customer can deploy to existing stack sets, but requires some migration steps.
+The customer would need to create a new stack,
+and import the stack set into the resource `AWS::CloudFormation::StackSet`.
```
nija-at, line 170: I'm mostly repeating myself here, but this doesn't cover why this approach is better than using `CfnStackSet`.

We should be talking about this in terms of what we are enabling for users here that didn't previously exist.

nija-at (changes requested):

text/66-self-managed-stack-sets.md
linsona: Yeah, so after discussing with Adam, we decided to go the route of introducing new roles:
* The `Stack Set Execution Role`  names should be identical in all accounts, which is a requirement for how stack sets operate. The current roles in the bootstrap template are suffixed with specific account ids
* The `Stack Set Execution Role` also must trust the `Stack Set Administration Role`, to assume and provision the stack set instance. 
  * In the current bootstrap, the trust flag would give any IAM entity in the account permissions to assume the execution role, which may have a large scope of permission. 
  * This is different from trust given to the ` DeploymentActionRole`, where it only has permissions to to create stack/changeset, and the actual role with large scope of permissions is `CloudFormationExecutionRole` which is not assumable to by the trusted account

Issue #359, opened 2021-07-11
Construct Hub Deny List


eladb:
## Description

To ensure the integrity of the website and prevent recurring abuse we need to have the ability to block specific packages from being ingested.

Note that it is not sufficient to not list offending package in the client side, we need to prevent the package from entering the processing pipeline.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @NetaNir (security requirement)
| Author(s)           | @eladb 
| API Bar Raiser      | @RomainMuller 
| Stakeholders        | @iliapolo

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #361, opened 2021-07-11
RFC 359: Construct Hub Deny List

eladb:
This is a request for comments about supporting a deny list in Construct Hub. See #359 for
additional details. 

APIs are signed off by @RomainMuller.

---

### DIFF
--- a/text/359-construct-hub-deny-list.md
+++ b/text/359-construct-hub-deny-list.md
@@ -33,10 +33,29 @@ new ConstructHub(this, 'ConstrucHub', {
 });
 ```
 
-The keys are package names and may also include specific version numbers. 
+The keys are package names and may also include specific version numbers.
 If a version is not specified (e.g.`boombam`), then all versions of a specific module are denied. 
 The value of each entry includes metadata which will be emitted in the ingestion log and in the future displayed in the 
 "where is my package" page for publisher diagnostics.
+
+```ts
+interface DenyListEntry {
+  /** The reason why this package/version is denied */
+  readonly reason?: string;
+  
+  /** The user name that signed-off on the deny list entry */
+  readonly user?: string;
+  
+  /** The time this package was added */
+  readonly date?: Date;
+  
+  /** An internal reference to an issue or ticket that tracks this deny list entry */
+  readonly ref: string;
+}
+```
+
+The only required field is `ref` as in some cases it won't be possible to publish information
+about a deny list entry, but we need some way to track it internally.
 
 ---
 
--- a/text/359-construct-hub-deny-list.md
+++ b/text/359-construct-hub-deny-list.md
@@ -16,46 +16,16 @@ To add a package to the deny list, add it to the `denyList` option when defining
 
 ```ts
 new ConstructHub(this, 'ConstrucHub', {
-  denyList: {
-    '@aws-cdk/aws-eks@1.44.0': {
-      reason: 'Security issue.',
-      user: 'eladb',
-      date: new Date('2021-07-11T12:57:48.693Z'),
-      ref: 'T123333322',
-    },
-    'boombam', {
-      reason: 'Copyright violation',
-      user: 'theadmin',
-      date: new Date('2033-07-11T12:57:48.693Z'),
-      ref: 'T137838738383',
-    },
+  denyList: [
+    { package: '@aws-cdk/aws-eks', version: '1.44.0', reason: 'Security issue.' },
+    { package: 'boombam', reason: 'Copyright violation' },
   }
 });
 ```
 
-The keys are package names and may also include specific version numbers.
-If a version is not specified (e.g.`boombam`), then all versions of a specific module are denied. 
-The value of each entry includes metadata which will be emitted in the ingestion log and in the future displayed in the 
-"where is my package" page for publisher diagnostics.
-
-```ts
-interface DenyListEntry {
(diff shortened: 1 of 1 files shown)


text/359-construct-hub-deny-list.md
```diff
@@ -0,0 +1,127 @@
...
+2. We will trigger a lambda function every time the deny list file is created/updated. This lambda function will 
+   iterate over the packages in the deny list and will delete any objects from the packages s3 bucket that match a 
+   denied package. The object key prefix will be based on the deny entry key (name + version or just name). 
+   We currently assume this process will fit into a single lambda execution given we will issue an S3 ListObjects 
+   with a prefix filter. Timeout alarms will let us know if this does not scale and requires an additional indirection.
+3. The "discovery" and "ingestion" lambda functions will retrieve this file for each request, and will consult the 
+   deny list for every package. If an incoming package is in the deny list, it will be skipped and a log entry will 
+   be emitted with the deny information.
+4. The "inventory" process will retrieve the deny list and will filter out all denied package from inventory. 
+   Statistics about denied package will also be included in the inventory metrics.
```
RomainMuller, line 84: I'm not sure I understand what this means... Can you elaborate / make it more concrete?

I suppose this means the inventory function would give out a metric for how many artefacts still exist for deny-listed packages? Or does it simply give a metric for how many entries are currently in the deny-list?

text/359-construct-hub-deny-list.md
```diff
@@ -0,0 +1,127 @@
...
+### What is the high level implementation plan?
+
+Project plan TBD.
+
+### Are there any open issues that need to be addressed later?
+
+* [ ] Packages denied in the public hub will still be allowed in private hubs until they are explicitly 
+      added to the deny list there. That might actually be the desired behavior (e.g. local deployment may 
+      want to show a package that is blocked in the public hub), but it should be communicated that if a 
+      package is reported and denied from the public hub, it will still appear in private hubs unless it is explicitly denied there.
```
RomainMuller, line 125: I think the documentation needs to mention that the deny-list is instance-specific. That has several implications:

- If someone files a request against one construct hub instance, only that instance may get the package deny-listed
- The "report abuse" UI should be configuration (i.e: the Construct Hub needs to show a deployment-specific email address to use for submitting reports)

text/359-construct-hub-deny-list.md
eladb: The latter - just emit the number of deny list entries. I'll make that clearer.

text/359-construct-hub-deny-list.md
eladb: Yes, I agree with both bullets. I'll elaborate.

text/359-construct-hub-deny-list.md
```diff
@@ -0,0 +1,137 @@
...
+- [ ] Filter denied packages in "ingestion"
+- [ ] Add deny list count to "inventory"
+- [ ] Create deny-list handler - triggered by updates to the deny list and deletes existing packages
+- [ ] Update catalog builder when objects are deleted
+- [ ] Move "report abuse" email address to `ConstructHub` API so it's configurable in private instances
+- [ ] Update documentation to indicate that deny list (and report abuse) is instance-specific
+
+### Are there any open issues that need to be addressed later?
+
+* [ ] @NetaNir can you elaborate on this please? "(Security-requirement) "Alert when a specific package owner
```
eladb, line 136: @NetaNir can you please let me know what this means?

text/359-construct-hub-deny-list.md
```diff
@@ -0,0 +1,156 @@
...
+});
+```
+
+The keys are package names and may also include specific version numbers.
+If a version is not specified (e.g.`boombam`), then all versions of a specific module are denied. 
+The value of each entry includes metadata which will be emitted in the ingestion log and in the future displayed in the 
+"where is my package" page for publisher diagnostics.
+
+```ts
+interface DenyListEntry {
```
eladb, line 42: @RomainMuller I added this

text/359-construct-hub-deny-list.md
NetaNir: Yes, package owner is a bad title here. The spirit of the security review feedback was that if there is a single entity abusing consturctHub we should have a way to alarm on it. So for example, in the context of a node package context, an entity can be package owner, an npm scope, a repo URL.

eladb: Can you provide a concrete example on when an alarm should be raised? walk me through it because I am not sure I understand the requirement in our context

text/359-construct-hub-deny-list.md
RomainMuller: Why are all fields optional?

RomainMuller (approved):

text/359-construct-hub-deny-list.md
NetaNir: Eva publish a packages under the scope `@abuse-chub`. All of these package are maliciously crafted to exploit a feature in `jsii-docgen`, could be a vulnerability or an edge case that causes it to run forever (which will cause the transliterator Lambda to time out without updating the marker, resulting in the same malicious packages being processed again and again, jamming our ingestion pipeline). In this case we will want to alert that the scope `@abuse-chub` is abusing the CDK. If we dont want to implement this logic on the scope/author/... we can alarm on an individual package.

PR #373, opened 2021-08-23
rfc(374): The jsii compiler to follow TypeScript versioning

RomainMuller:
This RFC #374 proposes to change the versioning scheme of the `jsii` compiler to stop
conforming to [semantic versioning][semver], and instead use the `major.minor`
version of the [TypeScript] compiler it is built on.

[semver]: https://semver.org/spec/v2.0.0.html
[TypeScript]: https://www.typescriptlang.org

---


text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+*jsii* tools they use. It allows developers to benefit from improvements to the
+*TypeScript* language (and compiler performance) faster, by allowing them to
+upgrade their project on their own timeline - in the same way they would be able
+to do for a "pure" *TypeScript* project.
+
+Since all parts of the *jsii* constellation are inter-operable with all major
+releases of the *jsii* packages, developers are able to opt into new and
+improved code-generation features of `jsii-pacmak` on their own schedule (i.e:
+when they release a new major version of their library) without requiring all
+their dependencies and dependents use the same version of `jsii-pacmak`.
```
rix0rrr, line 145: Like... exact same version? Do we not check assembly version? How do we even check or validate what version of `pacmak` a dependency was generated with?

Or is `assembly version == jsii version` ? If so, then another "Why Not Just X" that should be addressed is "why not decouple assembly version from jsii version" ?

text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+### Is this a breaking change?
+
+This is not a breaking change in the traditional sense: all existing code will
+continue to work with the current release lines of all packages in the
+ecosystem. However, the new versioning strategy of `jsii` is a departure from
+Semantic Versioning, which will be accompanied with a major version change (from
+`1.x` to `4.3.x` - matching the current `typescript` release line). This is to
+be noted in the release ntoes.
+
+### What alternative solutions did you consider?
```
rix0rrr, line 217: This section is discussing alternatives along different axes.

The first paragraph addresses a different "versioning" solution (so an altogether different approach to solving the problem) (fully decoupled versions vs linking major versions to CDK).

The second paragraph addresses, once we've picked fully decoupled versions, an **implementation** concern about how we will version the packages independently (monorepo vs multirepo).

Those concerns do not really belong at the same level. I would suggest making it more clear what aspect of the problem we're targeting when discussing these alternatives.

text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+## Internal FAQ
+
+### Why are we doing this?
+
+We are doing this to reduce the friction involved in making major changes to any
+of the *jsii* constellation of packages. This will allow breaking changes to be
+introduced in parts of the ecosystem without bumping the major version of the
+rest of packages.
+
+This will improve our pace of innovation, as we will hence be able to introduce
```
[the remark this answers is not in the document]
nija-at: I agree with @rix0rrr. I'm missing the motivation.

text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+### Release Notes entry
+
+```md
+Future releases of the *jsii* package constellation will no longer follow an
+all-at-once version increment strategy. Instead, each individual package will be
+released under their own version line. It is hence no longer necessary to ensure
+the version of all *jsii* packages in a product's dependency are aligned on the
+exact same release. For example, it is now possible to have a dependency on the
+`jsii` compiler at `4.3.0` while using `jsii-pacmak` at `1.35.0`, which
+previously would have led to a fatal error when executing `jsii-pacmak`.
```
nija-at, line 27: There must be some version order requirement between the producer and consumer sides, no?

Specifically when the feature expects properties produced by the compiler.

text/0374-jsii-modular.md
nija-at: I like to add these as several appendices as "alternatives for <design/choice>" and discuss pros and cons there, and link them off from here.

text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+represent a major burden on the community, meaning our ability to release
+improvements to *jsii* packages would be severely limited.
+
+It would also be possible to unlock versions in the mono-repository, in order to
+allow each packages to include breaking changes when necessary. However, this
+makes it all too easy to inadvertently break backwards compatibility, as it is
+so tempting to introduce the fix for a breaking change in the same PR as the
+breaking change itself. Using several repositories makes this much less likely.
+
+### What are the drawbacks of this solution?
```
nija-at, line 234: What about your current PR validation (and build?) process that uses GH actions? 
Any particular challenges in that area?

text/0374-jsii-modular.md
```diff
@@ -0,0 +1,256 @@
...
+For example, this allows the `jsii` compiler to be upgraded to a new release
+line of `typescript` without forcing every downstream consumer to also accept
+the syntax breaking changes at the same time. Instead, consumers will be free
+to take on a new release line of `jsii` independently from their dependencies as
+well as from their consumers.
+
+Separating the mono-repository into multiple repositories will help ensure we
+maintain backwards compatibility between individual packages, and reduce the
+risk of creating codependent packages (different repositories need to be
+released independenlty, so dependencies can only flow in an asyclic manner).
```
nija-at, line 169:
```suggestion
risk of creating co-dependent packages (different repositories need to be
released independently, so dependencies can only flow in an acyclic manner).
```

nija-at, line 169: Don't monorepos provide the same or some of these assurances? For eg, I believe you can't have cyclic dependencies within packages of a monorepo.

It's not clear to me (from the rfc) why coupling jsii versions and breaking off monorepo are related. 

By not breaking off the monorepo, some of the concerns you've listed in the next question can be avoided while still decoupling jsii release versions.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+One of the main rationales for this change is to enable the use of new
+TypeScript releases without forcing the enitre package ecosystem to make the
+change at the same time (due to language-breaking changes).
+
+An option to address this would be to allow customers to "bring their own"
+TypeScript compiler, by making it a `peerDependency`. This however comes with
+a significant challenge: the TypeScript compiler API is not stable between
+TypeScript releases, and it is hence not possible to author a `jsii` compiler
+that can leverage an arbitrary TypeScript version.
```
rix0rrr, line 275: But it is though. We could have different adapters for different TS versions:

```js
// package.json
{
  "devDependencies": {
     "typescript-39": "npm:typescript@~3.9.0",
     "typescript-47": "npm:typescript@~4.7.0"
  }
}
```

```ts
// typescript-39.ts
import type * as ts from 'typescript-39';

class Typescript39Compiler implements ICompiler {
  private readonly ts: typeof import 'typescript-39';

  constructor() {
    this.ts = requireFromUserScope('typescript');

    const program = new this.ts.Program(...);
  }
}
```

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+It would technically be possible to de-couple the `jsii` package from the rest
+of the toolchain's versioning, and to issue a new `jsii` major version each time
+we update the [TypeScript] compiler it internally uses. This would however make
+it difficult for customers to understand the relationship between `jsii` major
+releases and the [TypeScript] language level they support.
+
+Using the same version number prefix (`major.minor`) as the [TypeScript]
+compiler makes the relationship clear and removes the need to maintain separate
+documentation for customers to understand what they are upgrading to.
+
```
rix0rrr, line 288: Missing alternative: something like @iliapolo is doing for cdk8s: maintain N different releases in parallel on different branches. I'll let him weigh in on the pros and cons.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+## Working Backwards
+
+### Release Notes entry
+
+```md
+Starting with this release, the `jsii` compiler will no longer conform to
+[semantic versioning][semver]. Instead, its `major.minor` version will match
+that of the [TypeScript] compiler it uses (the [TypeScript] compiler does not
+conform to [semantic versioning][semver]).
```
rix0rrr, line 19: I don't think this is the most interesting thing about this decision. The most interesting thing is more directly framed as:

* jsii will now start tracking new TypeScript versions and will only release new features to the latest TypeScript version (not covered by the current wording)
* Users may have to change their code when they want to adopt new jsii versions (somewhat covered by implication by the current wording, but I think it's worthwhile to be explicit about it)

rix0rrr: We are looking at this, right?

image

* If you want a newer jsii feature, you must take a newer TypeScript version
* If you want a newer TypeScript feature, you must take a newer jsii version

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+were superceded by a new *current* line.
+
+## Internal FAQ
+
+### How will we maintain backwards compatibility?
+
+The change only affects the [TypeScript] compiler version used internally by
+`jsii` and the versioning scheme for the `jsii` package itself.
+
+The compiler will continue to emit `.jsii` assemblies that conform to the schema
```
rix0rrr, line 157: Are we committed to not breaking that?

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+### Why are we doing this?
+
+Before this change, `jsii` users had been stuck with [TypeScript] `3.9` for a
+very long time, due to the introduction of several breaking changes in the
+language specification: upgrading the [TypeScript] compiler that `jsii` builds
+on would cause existing code to break.
+
+This could not be addressed by issuing a new major release of `jsii`, as this
+would cause the ecosystem to be fragmented between packages that have migrated
+to the new major release, and those that haven't. Additionally, a package could
```
rix0rrr, line 178: Why is this? That would only be true if the runtime changed versions as well, right? Why don't we decouple the runtime version from the compiler version?

rix0rrr: And in fact, this proposal doesn't talk about it but it sure seems like an implication of this is that we should be decoupling the runtime version from the jsii compiler version even starting with this proposal -- otherwise jsii 3.9.x packages require runtime version 3.9.x and jsii 4.7.x reuqires runtime 4.7.x and we would be inducing the same fragmentation this paragraph warns about.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+Following [TypeScript] versions means releasing new `major.minor` releases (that
+may include breaking changes) more often than is currently done (the
+[TypeScript] maintainers typically declare a new `major.minor` line four
+times per year: in February, May, August and November respectively). This is a
+lot more release lines than other AWS products offer, and it would not be
+reasonable to uphold the [AWS SDKs and Tools maintenance policy][aws-policy],
+as it requires offering full support (features and bug fixes) for "Generally
+Available" releases for a minimum of 24 months. Diverging from this
+[policy][aws-policy] is a significant change that customers may not expect.
```
rix0rrr, line 204: Ultimately the digression from the policy is more about refusing to backport new features to jsii 3.9 than it is about changing the version number. Might be worth being more explicit about that.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+* Make the `jsii` compiler transparently prepare down-leveled declarations files
+  to ensure backwards-compatibility of compiler outputs with previous releases
+  of `jsii`. This can be done using the [`downlevel-dts`][downlevel-dts] package
+  to produce declarations files compatible with [TypeScript] `3.9`, and adding
+  a `typesVersions` key to the `package.json` file of packages.
+
+  - See proof-of-concept: [`aws/jsii#3501`][aws/jsii-3501]
+
+* Release the initial `jsii` release on the `4.7` line.
```
rix0rrr, line 238: So we are choosing to disregard all versions between 3.9 and 4.7, yeah?

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+### What are the drawbacks of this solution?
+
+This solution is a stark departure from versioning schemes and maintenance
+policies used by AWS on open-source products. It implies offering critical bug
+fixes and security updates for up to 5 different release lines (1 current, and
+3 to 4 previous), which might require significant effort.
+
+Following the [TypeScript] `major.minor` reduces our ability to introduce
+breaking changes to the `jsii` compiler, as these must be timed together with
```
rix0rrr, line 297: What kind of breaking changes would we be intending to release? Anything more than removal of deprecated features?

I think I would still feel better if we chose not to introduce any breaking changes on top of the potential syntactic breaking changes introduced by the underlying TypeScript (but I'm fine with deprecation+removal a good while later).

text/0374-jsii-ts-version.md
RomainMuller: The `cdk8s-plus` model involves releasing a different *package* per target version (here it'd be `jsii-47` for example), which I **strongly** dislike (and have recommended against multiple times over -- because this is a glaring supply-chain attack bait...).

I'd also want to passively encourage people to relatively closely track the latest release... not incitate/facilitate them staying on an ancient version for sustained periods of time.

text/0374-jsii-ts-version.md
RomainMuller: Yeah that's correct.

text/0374-jsii-ts-version.md
RomainMuller: For the foreseeable future, I don't think there is much of a way around that.

text/0374-jsii-ts-version.md
RomainMuller: Basically this proposal is entirely about decoupling the compiler version from the rest of the toolchain's versions...

In the current state of things, we MUST either major-bump nothing, or major-bump everything... and cause a schism.

---

The previous version of this RFC was going the extreme direction of suggestion we decouple the versions of **everything** and break the mono-repo out to individual repositories (possibly in a new GitHub organization, but that is beyond the point).

The same reasoning line was exposed there but literally every piece of feedback I received on that initial version included comments along the lines of "but why?", and I honestly don't have more things to say about "why" than this (and that feedback seemed to indicate this is not a sufficient justification?)

As a result, I decided to narrow the scope down to ONLY address the one problem that is currently a known problem: we've been stuck on TypeScript 3.9 for **way** too long, and need a clean way out.

text/0374-jsii-ts-version.md
RomainMuller: You seem to be under-estimating the maintenance burden here...

There is one new release of TypeScript per quarter (approximately), and the AWS SDKs and Tools maintenance policy mandates that everything that was deemed "GA" has full support for at least 24 months.

This does not mean "just" back-porting to jsii 3.9... Assuming we had followed this versioning strategy 2 years ago now, we'd have to back-port features & bug fixes to:
- 3.9 (at least until mid-November 2022)
- 4.0
- 4.1
- 4.2
- 4.3
- 4.4
- 4.5
- 4.6 (would be current, and 4.7 will arrive "imminently")

We'd have to build, test and release those 8 different versions.

One could argue that back-porting could be largely automated, but this fails to account for how the compiler API has significant changes in between these releases... That sounds like A LOT of thinks to maintain at the same time, and I'd rather not have to manage that many.

text/0374-jsii-ts-version.md
RomainMuller: Yes. I don't think there is much of a point in inflicting that additional work on us. That is, unless it turns out your "adapter" idea is feasible (then I guess why not).

text/0374-jsii-ts-version.md
RomainMuller: That's an interesting proposition... I'm going to be investigating that one a bit more...

Instinctively, I feel we might have an issue due to how the `ts.Node` types (and descendants) might not be compatible between versions, and the `jsii.Assembler` currently works by traversing the AST and gathering information... The feasibility of this approach hinges on our ability to design an abstraction that hides those differences effectively while not forcing us to effectively maintain a distinct assembler per compiler version.

text/0374-jsii-ts-version.md
Chriscbr: Agree 100% with sticking to a single package

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+Initial communication                     | 1 day      |
+**Optional:** Break mono-repository out   | 5 days     |
+Have `jsii` down-level `.d.ts`            | 5 days     | [`aws/jsii#3501`][aws/jsii-3501]
+Update release automation                 | 2 days     |
+Update [TypeScript] dependency            | 5 days     |
+Initial Release                           | 1 day      |
+Maintenance Announcement                  | 1 day      |
+Move to Maintenance                       | 1 day      |
+Move to end-of-life                       | 1 day      |
+
```
Chriscbr, line 315: Missing items: updating jsii website documentation, updating jsii contributor guide, updating projen with new defaults for `JsiiProject`, QA / manual validation on some of our larger libraries (construct-hub perhaps?)

text/0374-jsii-ts-version.md
RomainMuller: That'll be a rejected alternative. The TypeScript compiler fights against this idea pretty strongly: writing an adapter layer almost boils down to re-writing a unified front-end to the entire TypeScript compiler surface by hand (it's almost impossible to re-use common signatures, in parts due to how the TypeScript compiler exports a mix of values & pure types).

I'll see to add a paragraph describing the problem here...

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+  the maintenance policy that is applicable to newer releases.
+
+### Is this a breaking change?
+
+This is not a breaking change in the traditional sense, however this is a
+significant change in versioning schemes used for the `jsii` compiler, and a
+departure from the [AWS SDKs and Tools maintenance policy][aws-policy].
+
+As such, this warrants ample communication with our customer base to avoid they
+are caught by surprise, or have broken expectations.
```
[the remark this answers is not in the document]
RomainMuller: What I want to say here is that we are making a breaking change... to our versioning scheme (SemVer -> TypeScript-Ver).

Customers will have to update their code when they move from `jsii@1.x` to `jsii@4.7.x` due to language specification changes, but the real think they need to be very carful about is to update from a caret dependency `jsii@^1.0.0` to a tilde dependency `jsii@~4.7.0`.

The level/duration of maintenance for future versions will also be different, and we should definitely encourage users to closely track `latest` as much as possible.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+Developers are free to decide when it is appropriate for them to upgrade their
+`jsii` dependency to a new `major.minor` version line, bringing in new
+[TypeScript] language features as well as performance improvements and bug
+fixes.
+
+The `jsii` compiler makes the due dilligence to ensure the artifacts produced
+(in particular, the `.d.ts` declarations files and the `.jsii` assemblies)
+remain compatible with previous versions, so that consumers of libraries need
+not update at the same time as their dependencies.
```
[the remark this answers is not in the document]
RomainMuller: Actually introduced it as a new question.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+
+```md
+Starting with this release, the `jsii` compiler and `jsii-rosetta` will no
+longer conform to [semantic versioning][semver]. Instead, its `major.minor`
+version will match that of the [TypeScript] compiler it uses (the [TypeScript]
+compiler does not conform to [semantic versioning][semver]). New features will
+only be introduced on the `jsii` release line that targets the latest
+[TypeScript] compiler.
+
+Users may need to modify they code when upgrading from one `major.minor` release
```
mrgrain, line 28:
```suggestion
Users may need to modify their code when upgrading from one `major.minor` release
```

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+this proposal and will continue to conform to [semantic versioning][semver]. In
+particular, code generated by `jsii-pamcak` will continue to use the `1.x`
+release line of `@jsii/kernel`, `@jsii/runtime`, and the languages' runtime
+libraries, so as to remain completely backwards compatible.
+
+## Working Backwards
+
+### Release Notes entry
+
+```md
```
mrgrain, line 20:
```suggestion
```markdown
```

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+particular, code generated by `jsii-pamcak` will continue to use the `1.x`
+release line of `@jsii/kernel`, `@jsii/runtime`, and the languages' runtime
+libraries, so as to remain completely backwards compatible.
+
+## Working Backwards
+
+### Release Notes entry
+
+```markdown
+Starting with this release, the `jsii` compiler and `jsii-rosetta` will no
```
rix0rrr, line 21: As a final nit, the emphasis on "semantic versioning" in this document feels a little misplaced.

Ultimately the crux of it is that there will be two different *versioning lines* or *versioning strategies* used in the jsii ecosystem:

- One for the spec and runtime, which is our own versioning line
- One for the user-facing tools that process TypeScript, which follows `tsc` versioning line

It just so happens that our controlled versioning line will follow semver and the other one will not.

But even **more importantly**, I suppose we will also guarantee eternal backwards compatibility in our own versioning line? (Which makes the discussion about semver a bit moot, effectively it becomes "we will never break backwards compatibility AND we version according to semver ⇒ our major version will always remain 1")

But the different versioning lines and our backwards compatibility promise (if any) should be more prominent than whether we force some technical standard on our version number.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+the effort required for these updates, the release strategy allows developers to
+do so on their own schedule.
+
+This change is made to allow developers to benefit from the latest and greatest
+features of the [TypeScript] language without requiring the entire ecosystem to
+make the switch at the same time.
+
+Other packages in the jsii toolchain (such as `jsii-pacmak`, `@jsii/spec`, and
+others) are not affected by this change and will continue to follow [semantic
+versioning][semver] as they currently do. The bindings generated by
```
rix0rrr, line 40: Does this document say anything about the backwards compatibility guarantees of these tools?

For example, we could be MV-bumping on every release, and proudly claim "we are following semver", yet our customers would still hate us.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+delcared end-of-life.
+```
+
+### `README.md` for the `jsii` compiler
+
+````md
+In order to allow developers to access and leverage the latest and greatest of
+*TypeScript* language features, `jsii` compiler releases follow the `typescript`
+package releases. For example, `jsii` version `4.3.x` is built on top of version
+`4.3.x` of the `typescript` compiler.
```
rix0rrr, line 61: Yes! This is exactly what I was expecting to be reading!

rix0rrr (dismissed):

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+12 months, after which it will transition into the *End-of-Life* tier.
+
+### What is the support policy for these new releases?
+
+Starting with release `4.9.0`, new features will only be added to the *current*
+release line (the *latest* `major.minor` stream), which corresponds to the
+currently active [TypeScript] release line.
+
+Critical bug fixes and security patches will be provided for previous release
+lines for 12 months, which covers 3 or 4 previous lines. Features will not be
```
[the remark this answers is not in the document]
RomainMuller: In the current state of things, bug-and-security maintenance for a compiler line does not require a ton of work (few bug reports against the compiler), and so I am assuming 1 year of maintenance is not an issue.

That said, this is a line in the sand and we can absolutely discuss where it should be put.

text/0374-jsii-ts-version.md
RomainMuller: It does in the sense that it has provisions around how you shouldn't be forced to upgrade _your_ dependency on a jsii tool because one of your dependencies or dependents has upgraded already (so every new line must be able to operate on output of previous AND future versions).

text/0374-jsii-ts-version.md
RomainMuller: Thats a very interesting way to put it, and I like it. I was perhaps excessively focusing on the change of expectations users can have about the maintenance & compatibility of various versions of the tools (i.e: so far we claimed to do `semver`, and we may no longer do in some spaces). But that doesn't do a great job at expressing what is the new state of the world in a clear way.

I'll look into re-phrasing those bits...

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+In order to maximize compatibility between [TypeScript] compiler versions, and
+since [TypeScript] occasionally introduces backwards-incompatible syntax changes
+(additions, modifications) to the declarations files (`.d.ts`), the `jsii`
+compiler will proactively produce down-leveled declarations files targeting
+[TypeScript] compiler versions used by previous (not yet *end-of-life*) releases
+of `jsii`. This can be achieved using the [`downlevel-dts`][downlevel-dts]
+utility. Failure to do so may make it impossible to use a library compiled with
+a given release of `jsii` in a project that is still using an older release,
+which would be akin to forcing dependents to upgrade at the same pace as their
+dependencies (which is undersiable).
```
ansgarm, line 293:
```suggestion
dependencies (which is undesirable).
```

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+
+As such, this warrants ample communication with our customer base to avoid they
+are caught by surprise, or have broken expectations.
+
+### What alternative solutions did you consider?
+
+#### Bring-your-own TypeScript version
+
+One of the main rationales for this change is to enable the use of new
+TypeScript releases without forcing the enitre package ecosystem to make the
```
ansgarm, line 403:
```suggestion
TypeScript releases without forcing the entire package ecosystem to make the
```

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+a significant challenge: the TypeScript compiler API is not stable between
+TypeScript releases. Addressing this would require maintaining an adapter layer
+for each supported version of the TypeScript compiler, and this would require
+tremendous efforts, in particular as there appears to be no way in TypeScript
+to re-export a `namespace` including all the type declarations it contains, as
+`typeof ts` (assuming `ts` is the TypeScript compiler namespace) only represents
+entities with a run-time value, and none of the interfaces it contains. The
+consequence of this is that writing an adapter for a particular [TypeScript]
+version requires wrapping all useful functions of the compiler API and
+re-delcaring the types they use, which represents hundreds if not thousands of
```
ansgarm, line 417:
```suggestion
re-declaring the types they use, which represents hundreds if not thousands of
```

ansgarm (commented):
Found some tiny typos while reading through your RFC, @RomainMuller 🔍😄

I really like this proposal!
It was interesting to read as matching a TypeScript version with JSIIs `major.minor` sounds like a solution to a quite similar problem we're facing with pre-built provider bindings in CDKTF which depend on the CDKTF version and the underlying Terraform Provider version (also TIL about cdk8s approach here).

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,504 @@
...
+Developers hence need to also be able to control which [TypeScript] version is
+being used by `jsii-rosetta` if they want to use this.
+
+Using the same `major.minor` version as the [TypeScript] compiler makes it easy
+for developers to identify which [TypeScript] language features are available to
+them or not. As the [TypeScript] compiler does not conform to [semantic
+versioning][semver], this implies `jsii` and `jsii-rosetta` also need to stop
+conforming to it.
+
+### Why should we _not_ do this?
```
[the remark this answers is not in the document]
rix0rrr: > Only if they are currently on a compiler release that is end-of-life, which... I guess is the same with all end-of-life software.

I might have missed this, but are you backporting all bugfixes to all versions?

> > and features

> This is by design!

Well. Yes for TSC features, but not for jsii-related features, right? Let's say we add a "named type unions" feature to jsii... would that feature also be backported to all 4 current release lines? Or just be added to the latest?

That's quite a heavy maintenance burden you would be taking on if yes...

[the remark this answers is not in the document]
RomainMuller: I added a blurb in "why not" about this as you suggested. Not sure this is a deal breaker here, but certainly should inform our final decision on the maintenance/support policy.

rix0rrr (dismissed):
I'm still good with this plan if you are. Just making sure we're going into all consequences eyes open :).

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,656 @@
...
+
+BREAKING CHANGE: In order to allow developers to use the latest & greatest
+features of TypeScript, the `jsii` compiler and `jsii-rosetta` are now made
+in-line with those of the `typescript` compiler (e.g: `jsii@4.9.x` is built on
+TypeScript 4.9.x). Since the TypeScript compiler does not follow semantic
+versioning, we strongly recommend you upgrade your `devDependency` on `jsii` and
+`jsii-rosetta` to use a tilde range (e.g: `~4.9.0`) to be able to control when
+you migrate to future [TypeScript] language versions. New `jsii` and
+`jsii-rosetta` features will only be introduced in the _latest_ release line,
+while older release lines only receive bug fixes until they are delcared
```
comcalvi, line 59: minor:
```suggestion
while older release lines only receive bug fixes until they are declared
```

text/0374-jsii-ts-version.md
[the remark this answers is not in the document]
comcalvi: I'd prefer we didn't backport changes like that for a whole year. It was a ton of maintenance work for the oncall when we did it for CDK...it made the oncall's life much easier when CDK v1 finally hit EOL.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,656 @@
...
+
+### What is the technical solution (design) of this feature?
+
+The proposed delivery plan for this feature is as follows:
+
+* Communicate about the upcoming change in new releases on the `v1` release
+  line, including a planned timeline for the initial release of the `v4.9` line,
+  and language about the departure from [SemVer].
+
+* **Optional:** Separate the `jsii` compiler from the rest of packages in the
```
comcalvi, line 499: This is worthwhile. I think it's confusing if `jsii` follows different versioning than `jsii-pacmak` when they live in the same repo.

text/0374-jsii-ts-version.md
```diff
@@ -0,0 +1,656 @@
...
+  jsii toolchain into separate repositories. This would make it easier to
+  release the `jsii` compiler separately from other parts of the toolchain that
+  follow a different versioning scheme.
+
+  - Move other packages to a new mono-repository (or to several new individual
+    package repositories), e.g: `github.com/aws/jsii-toolchain`.
+
+  - Make the `github.com/aws/jsii` repository by a single-package repository.
+
+* Update the *release* automation to use the [TypeScript] compiler version's
```
comcalvi, line 509: does this mean that new jsii versions can only be made when there are new TSC releases? Should we add a new dot to the release for any critical regression fixes, so we have a 

`major.minor.patch.jsiiPatch` scheme?

RomainMuller: I'm pretty sure npm will not accept that version scheme.

Concretely, we are free to make new versions whenever we want under the new scheme, but they'll all be "patch" versions (introduction of new features, bug fixes, etc... all those as patch versions).

We can however only make BREAKING changes in a version when a new TypeScript compiler version is released (this happens once per quarter, which I believe is frequent enough).

mrgrain (approved):
This was accepted

mrgrain (approved):

Issue #31, opened 2019-12-09
Integration tests


eladb:
|PR|Champion|
|--|--------|
|# |        |

## Description

- [ ] Add a source hash to avoid manual updates
- [ ] Run periodically in a pipeline
- [ ] Support Assertions
- [ ] Support snapshotting the full cloud assembly
- [ ] Externalize the cdk-integ tool
- [ ] See #89 and #88 for more ideas

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

douglasnaphas:
I definitely want the ability to automate the execution of [these kinds of stack verification steps](https://github.com/aws/aws-cdk/blob/e8a8c1c4ce84df66ec16377ec4f2f62afa957f09/packages/%40aws-cdk/aws-apigateway/test/authorizers/integ.token-authorizer.lit.ts#L7-L12).

For example, the file linked above has:

```ts
/*
 * Stack verification steps:
 * * `curl -s -o /dev/null -w "%{http_code}" <url>` should return 401
 * * `curl -s -o /dev/null -w "%{http_code}" -H 'Authorization: deny' <url>` should return 403
 * * `curl -s -o /dev/null -w "%{http_code}" -H 'Authorization: allow' <url>` should return 200
 */
```

So if I run `yarn integ authorizers/integ.token-authorizer.lit.js`, I'd like for it to run the `curl` commands listed above prior to stack cleanup, and fail if it does not get the expected status codes.

I believe this included in the task _Support Assertions_, based on the first bullet point in [#88](https://github.com/aws/aws-cdk-rfcs/issues/88#issue-555196306).

eladb:
Thanks, copying @nija-at

robert-pitt-foodhub:
Hey, so I have just came across this PR and as I am currently implementation a solution for my team I wanted to just share my somewhat simplistic approach to performing post deploying integration testing on a stack.

1. Environment
In our company we use jest as our testing environment and framework for unit testing, this is common practice  for typescript based applications we so wanted to ensure a consistent and unified experience for developers when writing unit tests or integration tests.

[...]

Issue #300, opened 2018-08-18
Programmatic access of AWS CDK CLI


hassankhan:
## Description

Hi there, thanks for creating this package! I was just wondering if there's any way of triggering CDK CLI commands programatically? This would really help with respect to integrating with existing tools and workflows.

[~~RFC 300: Programmatic access of AWS CDK CLI~~](https://github.com/aws/aws-cdk-rfcs/pull/475)

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @hassankhan, @mrgrain 
| Author(s)           | @mrgrain 
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

thenikso:
I managed to do this following some @ryan-mars suggestions.

```
yarn add aws-cdk
yarn add @aws-cdk/core
# plus all the @aws-cdk/... you need
```

```js
import { App, Stack } from '@aws-cdk/core';
import { SdkProvider } from 'aws-cdk/lib/api/aws-auth';
import { CloudFormationDeployments } from 'aws-cdk/lib/api/cloudformation-deployments';

class ExampleStack extends Stack { /* your stack here */ }

const app = new App();
const stack = new ExampleStack(app, 'StackName', {});

const stackArtifact = app.synth().getStackByName(stack.stackName);

const sdkProvider = await SdkProvider.withAwsCliCompatibleDefaults({
  profile: 'your ~/.aws/config profile name here',
});
[...]

PR #378, opened 2021-09-03
RFC 31: Integration testing for CDK apps

nija-at:
This is a request for comments about Integration testing for CDK apps.
See #31  for additional details. 

Readable version: https://github.com/aws/aws-cdk-rfcs/blob/nija-at/integ-tests/text/0031-integ-testing.md

APIs are signed off by @{BAR_RAISER}.

---

### DIFF
--- a/text/0031-integ-testing.md
+++ b/text/0031-integ-testing.md
@@ -9,7 +9,7 @@ tests that can be run locally or in their CI/CD pipeline.
 
 ## Working Backwards - README.md
 
-### contest
+### contest *[name not final]*
 
 🌩️ contest tests can verify that your CDK app can be deployed successfully on AWS with the desired resources.
 
@@ -20,6 +20,8 @@ tests that can be run locally or in their CI/CD pipeline.
 🧑‍💻 Write contest tests in any CDK supported language, and execute them as part of your CI pipeline.
 
 📸 Capture snapshots of your CDK apps to track how they are impacted by changes.
+
+🧹 contest will clean up any AWS resources that it had to provision, and keep your AWS costs to the minimum.
 
 ### Writing Tests
 
@@ -153,6 +155,8 @@ Detailed test results are available in a file, usually at `contest.out/results-x
 - In the pipeline
 - Snapshots
 - construct library / app upgrades.
+- clean up
+- authoring experience - interplay with 'cdk watch'
 
 -->
 
@@ -252,7 +256,7 @@ This needs to be incorporated later.
 
 ## Appendix A - Test Execution
 
-### Implementation
+### Design
 
 A project set up for contest will have the `contest` section in its `cdk.json`.
 In the case of a javascript CDK app:
@@ -294,7 +298,41 @@ We do not have a way to deploy CDK applications programmatically.
 ## Appendix B - Assertions
 
 All contest assertions are AWS Lambda functions at their core. Each assertion is associated to an
-AWS CloudFormation custom resource. They all use a single provider.
+AWS CloudFormation custom resource with a single provider.
 
-This provider simply invokes the lambda function specified by the assertion, collects its results
-and writes them into a pre-configured log stream in AWS Logs associated with this test run.
+This provider simply invokes the lambda function attached with the assertion, collect results
+and write them into a pre-configured log stream in AWS Logs associated with this test run.
+
+### Alternatives
+
+Alternatively, we can run these assertions from the test runner as part of test execution.
+In this approach, the developer machine or the CI server will interact with the deployed CDK app
+to execute the assertions.
+
+The proposed approach is better than this alternative for the following reasons.
+
+The proposed solution allows for users, typically enterprise users, to test "private" resources,
+such as a lambda function in a private vpc. With this alternative, the resource will need to be
+exposed to the developer machine and/or CI server.
+
+This type of testing being proposed here involves deploying and cleaning up AWS resources via
+CloudFormation stacks for each test. This is time intensive. A combined snapshot of the CDK app
+and assertions, taken once, provides a high level of confidence that the test does not need to be
+re-run. A fresh snapshot is taken every time the CDK app or any of the assertions in the test
+changes.
+Since the proposed solution models the assertions as AWS CDK constructs, the synthesized [cloud
+assembly] functions as a full and effective snapshot.
+This alternative, on the other hand, will require inventing an effective snapshotting solution.
+
+The proposed solution also provides a consistent execution environment for tests - the [AWS Lambda
+base images] and the ability to [modify the environment][Modifying runtime environment] via
+extensions and layers.
+As an example, if an assertion depends on the `sed` utility. OSX ships by default with BSD sed,
+while Linux ships with GNU sed, which have subtly different behaviours and option switches.
+In the proposed solution, the assertion will either use the `sed` utility in the Lambda base
+image or bring its own via Lambda Layers.
+If this is not accounted for, this alternative will cause the assertion to behave inconsistently.
+
+[cloud assembly]: https://docs.aws.amazon.com/cdk/api/latest/docs/cloud-assembly-schema-readme.html#cloud-assembly
+[AWS Lambda base images]: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-images.html
+[Modifying runtime environment]: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html
--- a/text/0031-integ-testing.md
+++ b/text/0031-integ-testing.md
@@ -1,6 +1,6 @@
 # Integration testing for CDK apps
 
-* **Original Author(s):**: [@nija-at](https://github.com/nija-at)
+* **Original Author(s)**: [@nija-at](https://github.com/nija-at)
 * **Tracking Issue**: [#31](https://github.com/aws/aws-cdk-rfcs/issues/31)
 * **API Bar Raiser**: @{BAR_RAISER_USER}
 
@@ -73,7 +73,7 @@ new AwsAssertion(repoTest, 'MessageReceived', {
 
 const testResult = repoTest.run();
 if (testResult.failures > 0) {
-  // test runner should fail the test
+  // notify the test runner to fail
 }
 ```
 
@@ -98,6 +98,19 @@ new AwsAssertion(test, 'GetObject', {
 new InvokeLambda(test, 'Invoke', {
   lambda: functionArn,
   throws: ...
+});
+```
+
+The `AwsAssertion` also allows multiple calls to be chained. The following calls Amazon S3 `GetObject` API
+based on the value returned from an Amazon DynamoDB `GetItem` call.
+
+```ts
+const response = new AwsAssertion(test, 'GetItem', {
+  request: AwsAssertionCall.DynamoDB.getItem({ ... }),
+});
+
+new AwsAssertion(test, 'GetObject', {
+  request: AwsAssertionCall.S3.getObject({ bucket: ..., key: response.Item.Payload }),
 });
 ```
 
@@ -129,7 +142,10 @@ To invoke this assertion, you can simply use the `invokeLambda()` API -
 
 ```ts
 const hdrAssertion = new StdHeaderAssertion(...);
-test.invokeLambda(hdrAssertion);
+
+new InvokeLambda(test, {
+  lambda: hdrAssertion,
+});
 ```
 
 > RFC Note: See [Appendix A](#appendix-a---assertions) for details on the underlying design.
@@ -301,8 +317,9 @@ of their app, such as, invoking an API endpoint or sending a request to a load b
 
 ### Why should I use this feature?
 
-This feature is beneficial to users who own a CDK app and want to run assertions on the deployed
-app in their pre-production stages as part of a CI/CD pipeline.
+This feature is beneficial to CDK app developers and want to run integration tests on their app
+in their pre-production stages as part of a CI/CD pipeline, as well as, to run smoke tests in
+their production stages.
 
 It is also beneficial to construct library owners who want to test the consumption of their
 construct library when it is deployed as part of an app.
@@ -319,7 +336,7 @@ With such changes, it is best to re-run all your tests in full (as is the defaul
 to ensure that all of the new changes do not have any unintended impact to your application.
 
 Rarely, the AWS CDK will change the synthesized template in backwards compatible ways, such as,
-changing the order of properties, or setting a property to its default. 
+changing the order of properties, or setting a property to its default.
 As stated previously, it is still safer to run all tests with invalidated snapshots.
 
 [assets]: https://docs.aws.amazon.com/cdk/latest/guide/assets.html
@@ -341,8 +358,11 @@ tests "in the cloud". This module aims to address this gap.
 
 ### Why should we _not_ do this?
 
-> Is there a way to address this use case with the current product? What are the
-> downsides of implementing this feature?
+There is no simple solution for users who wish to run integration tests on their CDK app
+or construct library without releasing these features.
+
+Users can roll their own solution here, but the problem and the effort would be significant.
+There is significant benefit in providing a testing system such as this.
 
 ### What is the technical solution (design) of this feature?
 
--- a/text/0031-integ-testing.md
+++ b/text/0031-integ-testing.md
@@ -15,11 +15,11 @@ tests that can be run locally or in their CI/CD pipeline.
 
 🤖 apptest can verify that your CDK app functions as expected, such as assertions against a public API endpoint.
 
+💨 apptest can be used to run smoke test against your production stack, and trigger a rollback when they fail.
+
 🧑‍💻 Write tests in any CDK supported language, and execute them as part of your CI pipeline.
 
 📸 Capture snapshots of your CDK apps to track how they are impacted by changes.
-
-🧹 apptest will clean up any AWS resources that it had to provision, and keep your AWS costs to the minimum.
 
 ### Writing Tests
 
@@ -71,7 +71,10 @@ new AwsAssertion(repoTest, 'MessageReceived', {
   returns: { Messages: [...] }
 });
 
-repoTest.run();
+const testResult = repoTest.run();
+if (testResult.failures > 0) {
+  // test runner should fail the test
+}
 ```
 
 ### Assertions
@@ -172,25 +175,33 @@ This setup will enable defining a separate script that runs the apptest suite - 
 Ant target, Maven phase, etc. - by using `*.apptest.*` as the file filter in your test runner.
 
 The test will be executed when the `run()` method is called on the `apptest.Test` class.
+This method will return a list of assertion failures (if any), that can then be used to notify
+the test runner of a test case failure.
 
 Test reports can be easily generated from the reporting mechanisms already available via these
 test runners.
 
 > RFC Note: See [Appendix B](#appendix-b---test-execution) for details on this design.
 
-### Developing & Debugging Tests
+### Testing AWS CDK Apps and construct libraries
 
-> TBD
+The industry standard practice is that app developers typically run pre-production instances
+of their applications or services as part of their deployment pipeline where integration tests
+are executed. This is no different for AWS CDK applications.
 
-<!--
+In such cases, the apptest suite should define *ONLY* the tests and point them at the existing
+pre-production app instance, such as via a well-known endpoint or CloudFormation stack outputs.
+The apptest suite, in this case, will deploy (and towards the end destroy) a "testing" stack
+where the tests would be executed.
 
-#### TBD
+The usage of `apptest` when testing construct libraries is different from that when it is used
+for apps.
 
-- In the pipeline
-- clean up
-- authoring experience - interplay with 'cdk watch'
-
--->
+The `apptest` suite for construct libraries can be used to verify the functionality of the construct
+library when it is used with various valid configurations.
+Each `apptest` case would define a new AWS CDK `Stack` or `App` that includes the construct
+library, followed by assertions. Upon execution, both the app and the assertions will be deployed
+(and finally destroyed) as part of the test execution.
 
 ## Working Backwards - README.md
 
@@ -216,8 +227,6 @@ The `deploy()` and `destroy()` APIs will block further program execution until t
 or destruction complete. If the action fails, the API will throw an error.
 Asynchronous support for `deploy()` and `destroy()` are not available yet.
 
-> TBD: configurable options of deploy and destroy
-
 By default, the APIs print their output and progress to stdout. This can be configured
 using options on the APIs.
 
@@ -228,7 +237,7 @@ app.deploy({
 });
 ```
 
-#### Credentials
+### Credentials
 
 This module uses the AWS SDK for Javascript under the hood to communicate with CloudFormation.
 We support most of the common ways in which [this SDK allows setting
@@ -348,8 +357,13 @@ modules, the project plan will be tracked separately and outside of this RFC.
 
 ### Are there any open issues that need to be addressed later?
 
-The RFC does not describe the experience of using apptest in CDK Pipelines.
-This needs to be incorporated later.
+The RFC does not describe:
+
+* the experience of using apptest in CDK Pipelines
+* the local development experience when writing tests and new assertions
+* clean up of resources that were deployed as part of the test
+
+These will be incorporated later.
 
 ## Appendix A - Assertions
 
--- a/text/0031-integ-testing.md
+++ b/text/0031-integ-testing.md
@@ -21,11 +21,13 @@ tests that can be run locally or in their CI/CD pipeline.
 
 📸 Capture snapshots of your CDK apps to track how they are impacted by changes.
 
-#### Getting Started
+### Getting Started
 
 > TBD: Installation instructions
 
-#### Writing Tests
+### Writing Tests
+
+contest tests can be written in the same programming language as your CDK project.
 
 Let's start with the below CDK construct that creates an artifact repository in an S3 bucket,
 and notifies of new artifacts to an SQS queue.
@@ -72,7 +74,7 @@ repoTest.assertAws.
      .returns({ Messages: [ ... ] });
 ```
 
-#### Assertions
+### Assertions
 
 `contest` comes packaged with a set of commonly used high level assertions.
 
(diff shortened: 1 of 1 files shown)


text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+* **API Bar Raiser**: @{BAR_RAISER_USER}
+
+CDK app developers and construct library authors can write integration
+tests that can be run locally or in their CI/CD pipeline.
+
+## Working Backwards - README.md
+
+### contest
+
+🌩️ contest tests can verify that your CDK app can be deployed successfully on AWS with the desired resources.
```
rix0rrr, line 14: Are we also taking care of cleanup? I think that'd be a selling point.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+* **Original Author(s):**: [@nija-at](https://github.com/nija-at)
+* **Tracking Issue**: [#31](https://github.com/aws/aws-cdk-rfcs/issues/31)
+* **API Bar Raiser**: @{BAR_RAISER_USER}
+
+CDK app developers and construct library authors can write integration
+tests that can be run locally or in their CI/CD pipeline.
+
+## Working Backwards - README.md
+
+### contest
```
rix0rrr, line 12: Haaaaaaaaaaaah!

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+     .returns({ Messages: [ ... ] });
+```
+
+### Assertions
+
+`contest` comes canned with a set of commonly used high level assertions.
+
+* `assertAws`: execute AWS service APIs and assert the response.
+* `invokeLambda`: invoke an AWS lambda function and assert the response.
+* `curlApi`: Execute the 'curl' command on an API endpoint. Used typically to test API Gateway resources.
```
rix0rrr, line 79: Too low level? Maybe just `httpRequest` ?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Working Backwards - README.md
+
+### contest
+
+🌩️ contest tests can verify that your CDK app can be deployed successfully on AWS with the desired resources.
+
+🤖 contest tests can verify that your CDK app functions as expected, such as assertions against a public API endpoint.
+
+🕰️ contest tests can verify that your CDK app can be updated from a previous version of the app.
```
rix0rrr, line 18: Not described in the doc.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+
+🧑‍💻 Write contest tests in any CDK supported language, and execute them as part of your CI pipeline.
+
+📸 Capture snapshots of your CDK apps to track how they are impacted by changes.
```
rix0rrr, line 22: Not described in the doc. I suppose we are leaving out the Custom Resource gunk from the snapshots?

rix0rrr (commented):
Interesting approach. I'm a little concerned that expressivity might suffer, and I'm also a little concerned about the runtime performance of these tests.

Ideally you'd want to keep test cases separated, but as doing so will require a full CloudFormation cycle for each tests (~multiple minutes per test) people will NOT do that and combine as many assertions as they can into a single test to avoid the performance overhead. Leading to worse tests, imo...

text/0031-integ-testing.md
nija-at: Yes I'm aware. The doc is incomplete.

text/0031-integ-testing.md
nija-at: Yes, the doc is not complete. I plan to add this in subsequent revisions.

text/0031-integ-testing.md
nija-at: Agreed. I'll add.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+repoTest.assertAws.
+  sqs.receiveMessage({ queue: stack.publishNotifs })
+     .returns({ Messages: [ ... ] });
+```
+
+### Assertions
+
+`contest` comes canned with a set of commonly used high level assertions.
+
+* `assertAws`: execute AWS service APIs and assert the response.
```
nija-at, line 77: Note to self: How to handle more complex handling of the SDK results?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+// repo.contest.ts
+import { ArtifactRepo } from './records';
+import { Test } from '@aws-cdk/contest';
+
+const repo = new ArtifactRepo(app, 'RequestRecord');
+
+const repoTest = Test.forConstruct(repo);
+
+repoTest.assertAws.
+  s3.putObject({ bucket: stack.artifacts, key: '...', body: '...' });
```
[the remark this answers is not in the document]
nija-at: > This feels like test setup, which makes the "assert" part of this statement read a bit awkwardly.

Hmm yes, you make a good point. Maybe just dropping the "assert" phrasing in the API will make this cleaner. I'll take a stab at this in the next rev.

> this may lead users to attempt to modify the return values of the SDK calls in the test file
> I think this will be the most confusing part of the paradigm you are describing here.

Good observation. I came to the similar realization after I published this. There's further work needed here.

nija-at:
@rix0rrr 

> I'm a little concerned that expressivity might suffer

Can you expand?

> also a little concerned about the runtime performance of these tests.

My approach was going to involve creating snapshots (app + assertions) for all tests by default and not "running" the test if the snapshots matched. Would this sufficiently address your concern?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+> TBD: Lifecycle of the log group, log stream
+
+### Alternatives
+
+Instead of adding a new CLI subcommand and updated init templates, we could look to reuse
+existing testing frameworks, such as, jest for Node.js, junit for Java, etc.
+This would imply a simpler and a more familiar experience of writing tests.
+
+However, to achieve this effectively, the test cases will need to deploy the app it defines.
+We do not have a way to deploy CDK applications programmatically.
```
skorfmann, line 292: Deploying programmatically seems to be a low hanging fruit and it's something people are asking for as well https://github.com/aws/aws-cdk-rfcs/issues/300 (there are similar requests in cdktf)

I actually built a jest test suite which deploys multiple AWS CDK test stack the other day - granted, it's all Typescript  but the experience is quite good.

skorfmann (commented):
I think this topic would be hugely beneficial beyond just AWS CDK. In CDK for Terraform were having similar thoughts and even beyond that, I think that whomever is provisioning AWS resources would benefit from having a strong testing support library which could be plugged into whatever testing framework people might be using.

It would be fantastic to have a polyglot testing support library, which abstracts away common use cases when interacting with AWS resources in test scenarios, like (off the top of my head, there's certainly more):

- triggering a Lambda function
- publishing / receiving messages (SQS, EventBridge)
- invoking Cloudfront Functions and fetching results
- Trigger StepFunction workflows / analyzing its state
- Dynamically override (stub) Lambda Function (e.g. in StepFunctions or API Gateway)
- AppSync resolvers
- Easily access logs and traces

I'd leave the actual assertions to the respective testing framework, let them run the tests locally by default. There could still be an glue layer on top, which integrates this neatly into AWS CDK, CDK for Terraform, etc...

Two questions:

- What is the motivation to run the tests as Lambda functions deployed via a Cloudformation Stack?
- Do you see away to extract the test helpers into an independent library to make these available beyond AWS CDK?

text/0031-integ-testing.md
nija-at: If we did get programmatic deploy-ability, this could work out.

As far as the AWS CDK goes, this will mean taking the 'deploy' logic in the AWS CDK CLI and converting it into an independent jsii module (so as to make this polyglot). I would not consider this low hanging fruit, but an rfc of its own.
However, this does not mean it's not worth pursuing.

@rix0rrr - given you know the most about the AWS CDK CLI, what do you think of this?

nija-at:
@skorfmann - thanks for your review.

> What is the motivation to run the tests as Lambda functions deployed via a Cloudformation Stack?

A few things -

It allows for users, typically enterprise users, to test "private" resources, such as a lambda function in a private vpc.

The biggest problem with this type of testing is the time of test execution which is going to be really long. The mitigation to this is snapshot testing (this section is due in the rfc and I hope to have it in the next rev). If the snapshot includes the app and the test, then we run the test only if the app and the test have not changed.

It also allows for predictable test execution environments, which allows for stable assertion execution. Let's say, an assertion depends on the `sed` command. If this is run on the Mac, it will run BSD `sed` and on unix, this will be GNU `sed`, which have subtly different behaviours.
With a lambda function, the execution environment and how to bring additional dependencies is well known.

Finally, a nice side effect is that we can use existing CLIs to deploy and run these tests, without any additional runtime logic.

This is a great question. I'll need to feed this back into the RFC.

> Do you see away to extract the test helpers into an independent library to make these available beyond AWS CDK?

What do you mean by test helpers?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+  }
+}
+```
+
+The following integ test invokes puts an object in the S3 bucket and verifies the message in the SQS queue.
+
+```ts
+// repo.contest.ts
+import { ArtifactRepo } from './records';
+import { Test } from '@aws-cdk/contest';
```
eladb, line 59:
```suggestion
import { Test } from '@aws-cdk/testing';
```

text/0031-integ-testing.md
eladb: Not a fan of the name... I'd just call this `cdk test`/`@aws-cdk/testing`

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+`cdk contest` will begin by discovering files and executing them one by one.
+As each test is completed, a one line summary of its pass/fail is printed to console.
+Detailed test results are available in a file, usually at `contest.out/results-xxx`.
+
+<!--
+
+#### TBD
+
+- In the pipeline
+- Snapshots
```
eladb, line 154: This is intrinsic to the whole model: Looking for the CLI workflow around updating snapshots (basically something similar to `cdk-integ` and/or how the workflow is implemented in construct hub.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,300 @@
...
+
+* `assertAws`: execute AWS service APIs and assert the response.
+* `invokeLambda`: invoke an AWS lambda function and assert the response.
+* `curlApi`: Execute the 'curl' command on an API endpoint. Used typically to test API Gateway resources.
+
+By default, these assertions will check that the underlying call succeeds. That behavior can be extended
+to also include verification of the response.
+
+```ts
+test.assertAws(
```
eladb, line 85: Ideally we want a programming model that will encourage an open ecosystem of assertions up to the point that even `assertAws` is not part of the AWS CDK but vended as a separate library. This likely implies that the APIs have separate entrypoints:

```ts
import { AwsAssertions } from 'cdk-aws-assertions';

const assert new AwsAssertions(test);
assert.s3.getObject({ ... }).returns(...);
```

Something like that.

nija-at:
> I think this topic would be hugely beneficial beyond just AWS CDK. In CDK for Terraform were having similar thoughts and even beyond that, I think that whomever is provisioning AWS resources would benefit from having a strong testing support library which could be plugged into whatever testing framework people might be using.

@skorfmann - agreed that this would be hugely beneficial.

I'm currently looking to find the right solution that works for one CDK, specifically the AWS CDK. Once we arrive at a happy solution, I intend to see what parts of the solution can be expanded to benefit other CDKs.

text/0031-integ-testing.md
rix0rrr: Definitely not an easy change to do this in a jsii-compatible way.

Of course, we could hack it and shell out to the CLI. But if we're going to do this, I'd be very interested in clearly defining and controlling the user experience. Having every test suite run wait for CFN deployments is not going to be great.

text/0031-integ-testing.md
```diff
@@ -58,39 +56,44 @@ The following integ test invokes puts an object in the S3 bucket and verifies th
...
 ```ts
-test.assertAws(
-  s3.getObject({ ... }).returns({ Data: ... });
-);
-
-test.invokeLambda(...).returns({ ... });
-test.invokeLambda(...).throws({ ... });
+new AwsAssertion(test, 'GetObject', {
+  request: AwsAssertionCall.S3.getObject({ ... }),
+  returns: { Data: ... },
```
BenChaimberg, line 90: to support the use case of allowing chained calls (call 1: `StartExecution`, call 2: `GetExecutionStatus(executionId = ...)`), can we use the `Capture` model?

nija-at: Let's not focus too deeply on this. We can get into a discussion about the exact API modelling when we get to implement (and implementation design of) this.
We can get as innovative on this as needed over time or even build service specific Assertion classes that do chaining, etc.

The idea here is to capture the overall direction in terms of API, to see there are no major gaps, or innovation blockers.

I would like to focus this RFC on areas around the designs of the test execution framework, assertion model, interaction between CLI and the test frameworks, etc.

BenChaimberg: Sure – this was I guess a poorly phrased way of saying: I would like to see first class support for call chaining in this design. I don't think it's explicitly mentioned but I hope it will be, since I think it's integral to the use of the framework.

nija-at: Ah, now I see what you mean.

This feels like the feature of the aws-assertions "module". If we look at the aws assertions as its own piece separate from the core testing system, we need to build some form of chaining (something like `Capture` perhaps as you previously alluded to). It's not necessarily something any other assertion "module" might require.

So it feels to me that this is part of the 'aws assertions' part of the system, and not something first class in the 'assertion' part of the system. But I'm also having a hard time judging that at this level - need to get my hands dirty to see how this will pan out.

Does that make sense?
Or do you see a piece that we need to add to the 'assertion' part of the system that will enable this for aws assertions?

BenChaimberg: I guess I don't yet see the dividing line between "core" and "aws" assertions – I'm sure this is in the works as the RFC takes shape! However, it seems to me that regardless of the [infrastructure manager (AWS/Azure/...)/call provider (AWS SDK/HTTP/...)], call chaining would be very useful. Not being able to track "state" in the assertions seems to really restrict what kind of tests the library can support. But, again, this is based just off my assumptions about core vs. aws. Excited to see this progress!

skorfmann:
> 
> > Do you see away to extract the test helpers into an independent library to make these available beyond AWS CDK?
> 
> What do you mean by test helpers?

Was aiming at what @eladb mentioned here https://github.com/aws/aws-cdk-rfcs/pull/378#discussion_r705166008

skorfmann:
I like where this is going 👍  What are your thoughts on traceability and logs for tests and their assertions? 

I've been playing around with `jest` and `xray` integration. My goal there was to get visibility into the resources and their potential logs, in the context of the tests they are running in. I still have to streamline this a bit, but then gonna publish a full code example. 

The rough idea goes like this:

```ts
// the jest test function wrapped into an xray starting segment 
// A setup block deployed a testing stack already, that's where the 'output' is coming from
  xray.test("render index file", async () => {
    const post = await fs.readFile(path.join(__dirname, 'fixtures', 'simple.md'), 'utf8');
    const index = await fs.readFile(path.join(__dirname, 'fixtures', 'index.md'), 'utf8');
    const bucketArn = output.renderOlap

   // the AWS SDK clients are instrumented with xray
    await putObject(bucketArn, `posts/simple.md`, post, 'text/markdown');
    await putObject(bucketArn, `index.md`, index, 'text/markdown');
    const result = await getObject(bucketArn, `index.md`);
    expect(result.toString()).toMatchSnapshot();
  }, 20_000);
```

Which allows to see the test execution for each single test as a full xray segment, including downstream calls of the tested infrastructure components.

Screenshot 2021-09-22 at 11 17 48

And more importantly, it's possible to collect all (cloudwatch) logs of the individual testing xray segments - here's an excerpt from my logging output (there's a delay of roughly 5 - 10 seconds for the logs to show up after the tests finished)

Screenshot 2021-09-22 at 11 18 37

I know that this is not that easy to pull off across all testing frameworks / languages, but perhaps it might be possible to keep this in mind and enable users to achieve something similar when they want to.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,483 @@
...
+language of your choice - jest for javascript or typescript, junit for Java, pytest for Python, etc.
+
+As shown in the above section, it is recommended to use the `apptest` suffix to your test files
+to differentiate them from unit and other kinds of tests - `xxx.apptest.ts` in a typescript
+project, `xxx.apptest.java` for Java project, etc.
+
+This setup will enable defining a separate script that runs the apptest suite - npm script,
+Ant target, Maven phase, etc. - by using `*.apptest.*` as the file filter in your test runner.
+
+The test will be executed when the `run()` method is called on the `apptest.Test` class.
```
nija-at, line 174: jsii (currently) does not support throwing error messages across languages.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,483 @@
...
+as part of their continuous deployment.
+
+These integration tests can be used to verify that their CDK app or construct library deploys
+successfully on AWS CloudFormation. Users can add additional assertions to verify the behaviour
+of their app, such as, invoking an API endpoint or sending a request to a load balancer.
+
+### Why should I use this feature?
+
+This feature is beneficial to users who own a CDK app and want to run assertions on the deployed
+app in their pre-production stages as part of a CI/CD pipeline.
```
polothy, line 271: So these tests are not deployed as part of the main application? If not, then they cannot trigger a rollback if any of the tests fail, which seems like a nice feature.

> pre-production stages

Why not allow tests against production? Just curious :)

We are currently using step function and/or lambdas to run integration tests.  Since they are part of the stack, they can roll it back if anything fails and we can run it against any deployment stage.  Maybe this RFC isn't meant to address this use case?

This RFC is very interesting! It'll be helpful no matter what.  Just trying to grasp how it is run and when (EG: how would this run in CodePipeline?).

text/0031-integ-testing.md
nija-at: I've addressed this in the latest change.

text/0031-integ-testing.md
nija-at: > Why not allow tests against production? Just curious :)

You are right. Take a look at the latest revision. I've included this.

nija-at:
@skorfmann -

> What are your thoughts on traceability and logs for tests and their assertions?

This should be possible with the apptest framework being proposed here. To get into the details on this, this is an entire topic we can go very deep on, so I've left it out.

If we build the right abstraction for an 'assertion', we can build assertions based on logs and, perhaps even, metrics.

text/0031-integ-testing.md
polothy: Think I caught your updates, looks like it is possible to include these in your main stack in order to trigger rollbacks, thanks! 👍 

For when it isn't part of your main stack, what's the expected workflow in a pipeline like CodePipeline?

- CloudFormation Action to deploy the main application that we want to test
- CodeBuild that runs CDK deploy/destroy or Another CloudFormation Action to deploy... and maybe a CloudFormation Action to destroy? (or maybe just keep it? No real cost if all serverless tech)

Think my concern about CodeBuild is if the deployment isn't in the same account, we have to do some role assumes.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+}
+```
+
+### Assertions
+
+`apptest` comes canned with a set of commonly used high level assertions.
+
+* `AwsAssertion`: execute AWS service APIs and assert the response.
+* `InvokeLambda`: invoke an AWS Lambda Function and assert its response.
+* `HttpRequest`: Execute an HTTP request against an API endpoint and assert the response.
```
eladb, line 86: The naming patterns are not consistent here (some use verbs, some use nouns). Also seems like all of them are "assertions", so either they all should have an `Assertion` suffix or none.

Here are some options:

- `AwsApiAssertion`, `LambdaAssertion`, `HttpAssertion`
- `AwsApiRequest`, `LambdaInvoke`, `HttpRequest`
- `AssertAwsApi`, `AssertLambda`, `AssertHttp`

etc...

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+To invoke this assertion, you can simply use the `invokeLambda()` API -
+
+```ts
+const hdrAssertion = new StdHeaderAssertion(...);
+
+new InvokeLambda(test, {
+  lambda: hdrAssertion,
+});
+```
+
```
eladb, line 150: I think this example should show how to vend an assertion class that looks and feels like the ones shipped with the CDK.

```ts
export class StdHeaderAssertion extends InvokeLambda {
  constructor(test, options) {
    super(test, {
      lambda: xxxx
    });
}
```

This implies that the programming model should give the derived class access to the construct tree, so requires a bit of thinking.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+> RFC Note: See [Appendix A](#appendix-a---assertions) for details on the underlying design.
+
+### Test snapshots
+
+On first run of a apptest test, a snapshot will be produced that *must be checked into source tree*.
+This is a snapshot of the [cloud assembly] containing both the CDK app (being tested) and the test
+assertions. The snapshot will be placed at a folder relative to the test file.
```
eladb, line 157: I am wondering if it makes sense to package the snapshot as a single [human readable text] file instead of a directory. WDYT?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+```
+
+On subsequent runs of this test, if the synthesized cloud assembly matches the
+snapshot, apptest will skip deployment and will not run any assertions. The test will be considered
+to have passed. This behaviour can be overridden either by setting the `ignoreSnapshot` option for
+individual tests or by setting the `CDK_APPTEST_IGNORE_SNAPSHOT` environment variable and will apply
+to all tests executed in that environment.
+
+However, when the synthesized cloud assembly does not match the snapshot, the test will be executed
+in full and the snapshot will be updated. As before, the changes to the snapshots must be checked into
```
eladb, line 176: Executing the test can only happen in an environment that has access to an AWS account, which is normally not the case for CI builds. So I think we should distinguish between re-evaluation and just snapshot assertion.

text/0031-integ-testing.md
eladb: Not sure what you mean by that? You can throw an exception in JavaScript and it will propagate to the host language...

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+the source tree.
+
+[cloud assembly]: https://docs.aws.amazon.com/cdk/api/latest/docs/cloud-assembly-schema-readme.html#cloud-assembly
+
+### Test Execution & Report
```
eladb, line 181: What are the pros/cons of this? Do we really see sufficient value from integrating into the existing unit test frameworks here? I feel like the complexity and cost and the need to implement a connector for each language might outweigh the value, but I am not sure.

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+
+The usage of `apptest` when testing construct libraries is different from that when it is used
+for apps.
+
+The `apptest` suite for construct libraries can be used to verify the functionality of the construct
+library when it is used with various valid configurations.
+Each `apptest` case would define a new AWS CDK `Stack` or `App` that includes the construct
+library, followed by assertions. Upon execution, both the app and the assertions will be deployed
+(and finally destroyed) as part of the test execution.
+
```
eladb, line 221: Can you provide some examples here? I am not sure I understand what this section is trying to say

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+To invoke an Amazon API Gateway endpoint as part of your smoke test, simply configure -
+
+```ts
+class MyStack extends Stack {
+  constructor(scope: Construct, id: string) {
+    // ...
+    const restApi = new apigateway.RestApi(this, 'MyRestApi');
+    // ...
+    new HttpInvoke(this, 'SmokeTest', {
```
eladb, line 238: This is cool, but also reinforces that these classes should have some common suffix like `Assertion`:

```suggestion
    new HttpAssertion(this, 'SmokeTest', {
```

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+
+This jsii module provides the ability to programmatically deploy and destroy AWS CDK apps.
+
+A CDK app can be deployed by calling -
+
+```ts
+import { App } from 'cdk-deploy';
+
+const app = new App('/path/to/cdk/app');
+app.deploy();
```
eladb, line 259: How do you plan to implement this synchronously?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+### `cdk-deploy`
+
+This jsii module provides the ability to programmatically deploy and destroy AWS CDK apps.
+
+A CDK app can be deployed by calling -
+
+```ts
+import { App } from 'cdk-deploy';
+
+const app = new App('/path/to/cdk/app');
```
eladb, line 258: Is this a path to a cloud assembly or to a directory with `cdk.json`? Kind of feels like it should accept `cdk.App` instance instead, no?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+```ts
+app.deploy({
+  stdout: false,
+  outfile: '/path/to/file' // path to file where the progress should be printed
+});
+```
+
+### Credentials
+
+This module uses the AWS SDK for Javascript under the hood to communicate with CloudFormation.
```
eladb, line 283:
```suggestion
This module uses the AWS SDK for Javascript under the hood to communicate with AWS services such as S3 and CloudFormation.
```

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+[Modifying runtime environment]: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html
+
+## Appendix B - Test Execution
+
+### Design
+
+The AWS CDK CLI (aka toolkit) is responsible for deploying and destroying AWS CDK apps.
+We will extract these two parts from the CLI into a separate jsii module.
+
+The `run()` method on the class `apptest.Test` will depend on this module to deploy and
```
eladb, line 492: Feels like this is a bit of a rabbit hole not directly needed to support integration tests. Why can't integration tests be simple CDK apps (like `cdk-integ` works) and then we can just use the CDK CLI (or a tool that wraps the CDK CLI) to deploy them?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+
+At the end of the run, the CLI will download the detailed test results from AWS Logs to the
+location `apptest.out/results-xxx`.
+
+The major downside to this solution is that it is re-inventing another test execution mechanism.
+
+Users will now need to learn how to organize apptest tests, learn how to execute it and read
+results. We are not leveraging existing "well known" test frameworks.
+
+Although this looks basic today, as the scope of apptest expands and more features are added,
```
eladb, line 537: I understand the desire to reuse execution&reporting from existing frameworks but I am not sure this makes total sense because users won't be able to use any assertion functions and the programming experience will likely be confusion.

As an exercise, can you provide a more complete example/prototype of a jest-based test suite that uses this? I am curious how this will look?

eladb (changes requested):

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+const repo = new ArtifactRepo(app, 'RequestRecord');
+
+const repoTest = Test.forConstruct(repo);
+
+new AwsAssertion(repoTest, 'RepoPutObject', {
+  request: AwsAssertionCall.S3.putObject({ bucket: stack.artifacts, key: '...', body: '...' })
+});
+
+new AwsAssertion(repoTest, 'MessageReceived', {
+  request: AwsAssertionCall.Sqs.receiveMessage({ queue: stack.publishNotifs }),
```
misterjoshua, line 70: When does this assertion issue the first request? How does it know when to stop long polling? (Does it send exactly one request?)

How might I handle a situation where the message takes 5 minutes to arrive?

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+  returns: { Messages: [...] }
```
misterjoshua, line 71: Are we supporting more than exact matches in `returns`? If the messages contain id numbers, I may not know those in advance, so I might want a language to match patterns.

Here's something like what we get in `expect` for jest:

```ts
new AwsAssertion(repoTest, 'MessageReceived', {
  request: AwsAssertionCall.Sqs.receiveMessage({ queue: stack.publishNotifs }),
  returns: AwsAssertion.objectContaining({
    Messages: AwsAssertion.arrayContaining([
      AwsAssertion.objectContaining({
        foo: 'bar'
      }),
    ]),
  }),
});
```

text/0031-integ-testing.md
```diff
@@ -0,0 +1,542 @@
...
+
+### Assertions
+
+`apptest` comes canned with a set of commonly used high level assertions.
+
+* `AwsAssertion`: execute AWS service APIs and assert the response.
+* `InvokeLambda`: invoke an AWS Lambda Function and assert its response.
+* `HttpRequest`: Execute an HTTP request against an API endpoint and assert the response.
+  Used typically to test API Gateway resources.
+
```
misterjoshua, line 88: What about these assertions:

`AwsAlarmStatusAssertion`: a CloudWatch alarm does not transition to 'ALARM' in a given timeframe.
`AwsStepFunctionsAssertion`: execute a Step Functions state machine - when it's done, check for a successful execution status.

text/0031-integ-testing.md
nija-at: > what's the expected workflow in a pipeline like CodePipeline?

You would run the tests as you would any other tests in your pipeline, which is probably CodeBuild.

A future addendum to this RFC will address this when [using CDK Pipelines](https://github.com/aws/aws-cdk/tree/master/packages/%40aws-cdk/pipelines#cdk-pipelines) and make this trivial. I've taken this out of scope to keep the RFC contained.

If you use your own CI/CD mechanism, you will need to set up cross-account assume role, yes.

PR #383, opened 2021-10-14
chore: add settings for the "cdk watch" command

skinny85:
This PR updates the "CDK Watch" RFC with details of how to customize the watching process by introducing new options to the `cdk.json` file:

* A `"build"` key that allows specifying what command(s) should be executed before the project is synthesized.
* A `"watch"` key with two sub-keys, `"include"` and `"exclude"`, that allows specifying what files should be observed for changes, and which ones should be ignored by the watcher.

---

### DIFF
--- a/text/0001-cdk-update.md
+++ b/text/0001-cdk-update.md
@@ -97,8 +97,8 @@ The "watch" functionality can be customized by setting a few new options in your
       "app": "mvn exec:java",
       "build": "mvn package",
       "watch": {
-        "include": ["src", "lambda/code/**/*"],
-        "exclude": ["cdk.out", "target"]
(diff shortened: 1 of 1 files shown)


marekq:
Thanks for adding this great feature! I've noticed that adding the following section to 'cdk.json' based on the example will pickup changes to my CDK stack, but also changes in other, non CDK files:

```
"watch": {
         "include": ["./**"],
         "exclude": ["target"]
       }
```

If I change something in my readme.md file, I see the following on my CLI:

`Detected change to 'README.md' (type: change). Triggering 'cdk deploy'
`       
Would it make sense to include only './lib', './bin' for example and to update this in the CDK docs?

PR #386, opened 2021-11-08
chore: update CDKv2 RFC guidance for v1 dist tags post-launch

njlynch:
The RFC currently specifies that post-launch of CDK v2, the v2 packages will get
an NPM distribution tag of `latest`, and the v1 packages will get a tag of
`v1`. However, tags that can be interpreted as valid semver ranges are not
allowed, and it is impossible to publish without a tag. Rename the tag to
`latest-1` to get around this. This is the same convention `npm` uses for its
releases.

---

### DIFF
--- a/text/0079-cdk-2.0.md
+++ b/text/0079-cdk-2.0.md
@@ -311,8 +311,8 @@ of CDKv1, i.e., `1.x.y`.
 
 When CDKv2 is declared GA, the `v2` distribution tag will be updated once to
 match the first GA release, and the `latest` distribution tag will start
-tracking the versions of CDKv2. A new distribution tag called `v1` will be
-created to track the latest release of CDK `1.x.y`.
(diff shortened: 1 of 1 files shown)


nija-at (commented):
As was discussed yesterday, do we even need a disttag here? Wouldn't `npx cdk@1` be sufficient?

njlynch:
> As was discussed yesterday, do we even need a disttag here? Wouldn't npx cdk@1 be sufficient?

That's sufficient from a *_consuming_* point of view, but not from a *_publishing_* point of view. `npm publish` automatically tags each new published artifact as `latest` unless a tag is explicitly provided; without setting an explicit tag for the v1 versions of the modules that have both v1 & v2 versions, we the `latest` tag will flip-flop between a v1 and v2 build.

eladb (approved):
Conforming to npm's standard sounds like a good idea to me.

nija-at (approved):

Issue #308, opened 2021-03-30
CLI notices


rix0rrr:
## Description

We don't have a good way of communicating urgent matters to our users. This becomes especially relevant when security issues or breakages are discovered--we want to push communication around this, but the AWS Blog is not appropriate, Personal Health Dashboard might not get looked at and email will get spammy.

The CLI is an obvious place to integrate.

## Working Backwards

The CDK CLI now automatically lets you know about issues that are important for the correctness, health and security of your CDK applications and toolchain.

Example messages we will push to you are:

* A security vulnerability has been detected in a (transitive) dependency of the CLI, and if you use Yarn you must take manual steps to protect yourself
* The Lambda runtime your CDK application is using is about to be deprecated and you should update it
* We fixed a serious issue in one of the CDK packages and you should update
* You are mixing CDKv1 package versions in your `package.json`, or you have carets on any of them (alternatively: you have multiple copies of the CDK in your dependency tree)

Every advisory is accompanied by a link information you about the details of the  advisory, and what you should do to protect yourself.

CDK will tell you about applicable advisories on every run. You can opt to ignore messages and never be notified of them again.

## Roles

* **Driver** (drives the proposal to completion): @user
* **Approver(s)**: (assigned by CDK team)

eladb:
Love this!

rix0rrr:
Another idea: have some rule that matches resources in templates, and throws up an advisory if we find a resource that matches some pattern or rule.

Chriscbr:
@otaviomacedo Can we mark this as done? 🎉

PR #389, opened 2021-11-29
RFC 308: CLI notices

otaviomacedo:
This is a request for comments about CLI notices. 
See #308 for additional details. 

APIs are signed off by @eladb.

[Readable version](https://github.com/aws/aws-cdk-rfcs/blob/otaviom/cli-advisories/text/0308-cli-advisories.md)

---

### DIFF
--- a/text/0308-cli-advisories.md
+++ b/text/0308-cli-advisories.md
@@ -2,7 +2,7 @@
 
 * **Original Author(s)**: [@otaviomacedo](https://github.com/otaviomacedo)
 * **Tracking Issue**: [#308](https://github.com/aws/aws-cdk-rfcs/issues/308)
-* **API Bar Raiser**: @{BAR_RAISER_USER}
+* **API Bar Raiser**: @eladb
 
 A new CLI feature to notify customers about urgent and important issues that
 require their attention.
@@ -115,11 +115,18 @@ In case of a high-impact issue, follow these steps:
 
 3. Create a PR with this change and wait for an approval. Only SDMs have
    permission to approve PRs in this repository.
-4. When the PR gets merged, the advisory will be visible to all CLI
+4. A PR check will validate the `advisories.json` file and verify that the
+   following conditions hold:
+   * The file is compliant with the schema.
+   * The issue exists.
+   * Title and overview are within the length constraints.
+   * The semantic version is valid.
+   * The component is valid.
+5. When the PR gets merged, the advisory will be visible to all CLI
    installations. The GitHub issue will also be automatically updated with the
    information contained in the file. All the necessary tags will also be added
    automatically. 
-5. You can keep updating the issue normally, as new information comes in, but
+6. You can keep updating the issue normally, as new information comes in, but
    you're not allowed to touch the sections auto-generated from the advisories
    file.
 
@@ -166,11 +173,12 @@ this feature.
 
 ### What is the technical solution (design) of this feature?
 
-Advisory information will be available as a static file, served from GitHub. 
-The CLI will consume this file from and apply a set of filters to narrow down 
-the list of advisories to the context in which it is being run. We will also 
-implement some GitHub actions to validate and copy the contents of the file over
-to the issue. For a more detailed explanation, see the Appendix.
+Advisory information will be available as a static file on a GitHub repository.
+This file will be automatically synced to an S3 bucket and distributed via
+CloudFront. The CLI will consume this file from and apply a set of filters to
+narrow down the list of advisories to the context in which it is being run. We
+will also implement some GitHub actions to validate and copy the contents of the
+file over to the issue. For a more detailed explanation, see the Appendix.
 
 ### Is this a breaking change?
 
@@ -182,18 +190,19 @@ On the publishing side:
 
 * Implementing a new internal REST service to manage the advisories. Overly
   complex for this use case.
-* Authoring the content directly on the GitHub issue. There is no good way to
-  implement an approval workflow that includes a human verification step.
+* Authoring the content directly on the GitHub issue. Hard to enfore
+  constraints on the content. 
 
 On the distribution side:
 
-* Assuming the advisories were stored in an S3 file (in case of the REST
-  service), we would use CloudFront to distribute them.
+* Serving the file directly form GitHub.
 * Using the GitHub API to query for special issues that are considered
   advisories (in case of using GitHub issues as the source of truth).  
 
 ### What is the high-level project plan?
 
+1. Implement and deploy the infastructure necessary for distribution (S3 +
+   CloudFront). 
 1. Implement the GitHub actions of validation and issue sync-up and issue
    protection.
 2. Add the construct library version to the cloud assembly metadata. 
@@ -245,7 +254,7 @@ list of advisories, each having the following fields:
 
 |     Field    |                           Description                          | Format                          | Mandatory? |
 |:------------:|:--------------------------------------------------------------:|---------------------------------|:----------:|
-| `title`      | The title of the incident                                      | Free form text                  | Yes        |
+| `title`      | The title of the incident (max length: 100)                    | Free form text                  | Yes        |
 | `issueUrl`   | A link to the GitHub issue where the incident is being tracked | URL                             | Yes        |
 | `overview`   | A paragraph with more information about the incident           | Free form text                  | Yes        |
 | `component`  | The CLI or the Framework                                       | Either `"cli"` or `"framework"` | Yes        |
@@ -261,11 +270,15 @@ We will also implement three GitHub actions on this repository:
 3. Issue protection. Every change to issues that are linked to some advisory
    will be checked by this action, to avoid corruption.
 
+For the distribution, we will implement a mechanism to sync the contents of the
+GitHub repository with the S3 bucket (e.g., a Event Bridge event). The S3
+content will be served by a CloudFront distribution.
+
 #### CLI logic
 
-The CLI will fetch the file from GitHub, parse the content and check whether the
-version range contained in the advisory matches the CLI version or the framework
-version (depending on the affected component).
+On every command, the CLI will fetch the file from the backend, parse the
+content and check whether the version range contained in the advisory matches
+the CLI version or the framework version (depending on the affected component).
 
 Since the CLI knows its own version, checking against the version range of the
 advisory is trivial. The version of the framework, however, is not readily
@@ -274,7 +287,8 @@ to the Cloud Assembly, in a place where the CLI can read it.
 
 Issues that pass this filter will be displayed on the standard output. If an 
 error or timeout occurs when retrieving the issues, the CLI will simply skip 
-the display of advisories and try again at the next command execution.
+the display of advisories and try again at the next command execution. Results
+will be cached for a period of one hour.
 
 The CLI will store the IDs of the acknowledged issues in the project specific 
 `./cdk.json` file.
--- a/text/0308-cli-advisories.md
+++ b/text/0308-cli-advisories.md
@@ -1,4 +1,4 @@
-# CLI advisories
+# CLI notices
 
 * **Original Author(s)**: [@otaviomacedo](https://github.com/otaviomacedo)
 * **Tracking Issue**: [#308](https://github.com/aws/aws-cdk-rfcs/issues/308)
@@ -20,7 +20,7 @@ $ cdk deploy
 
 ... # Normal output of the command
 
-ADVISORIES
+NOTICES
 
 16603   Toggling off auto_delete_objects for Bucket empties the bucket
 
@@ -44,15 +44,15 @@ ADVISORIES
 
         More information at: https://github.com/aws/aws-cdk/issues/17061 
 
-If you don’t want to see an advisory anymore, use "cdk acknowledge ID". For example, "cdk acknowledge 16603".
+If you don’t want to see an notice anymore, use "cdk acknowledge ID". For example, "cdk acknowledge 16603".
 ```
 
-By acknowledging a particular advisory, it won’t show anymore in subsequent calls:
+By acknowledging a particular notice, it won’t show anymore in subsequent calls:
 
 ```
 $ cdk acknowledge 16603
 
-ADVISORIES
+NOTICES
 
 17061   Error when building EKS cluster with monocdk import
 
@@ -64,44 +64,44 @@ ADVISORIES
 
         More information at: https://github.com/aws/aws-cdk/issues/17061 
 
-If you don’t want to see an advisory anymore, use "cdk acknowledge ID". For example, "cdk acknowledge 17061".
+If you don’t want to see an notice anymore, use "cdk acknowledge ID". For example, "cdk acknowledge 17061".
 ```
 
 You can suppress all warnings per individual execution:
 
 ```
-$ cdk deploy --no-advisories
+$ cdk deploy --no-notices
 ```
 
-And you can disable all advisories indefinitely by adding this entry to
+And you can disable all notices indefinitely by adding this entry to
 `~/.cdk.json`:
 
 ```
-"advisories": false
+"notices": false
 ```
 
-Regardless of the state of this flag and the advisories you have acknowledged,
-you can always show the currently active advisories:
+Regardless of the state of this flag and the notices you have acknowledged,
+you can always show the currently active notices:
 
 ```
-$ cdk advisories
+$ cdk notices
 ```
 
-This command returns zero if there is no advisory and non-zero otherwise. Users
+This command returns zero if there is no notice and non-zero otherwise. Users
 can then plug this into a pipeline approval workflow and expect manual review if
-there are any advisories.
+there are any notices.
 
 > Please note that the acknowledgements are made project by project. If you
-acknowledge an advisory in one CDK project, it will still appear on other 
+acknowledge an notice in one CDK project, it will still appear on other 
 projects when you run any CDK commands, unless you have suppressed or disabled 
-advisories.  
+notices.  
 
 ### Runbook section (internal to the CDK team)
 
 In case of a high-impact issue, follow these steps:
 
 1. Create or update an issue for this incident on the aws-cdk GitHub repository.
-2. Update the file `advisories.json` on repository `cdklabs/aws-cdk-advisories`, adding an entry for
+2. Update the file `notices.json` on repository `cdklabs/aws-cdk-notices`, adding an entry for
    the incident. Example:
 ```json
   {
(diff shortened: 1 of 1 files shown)


text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+
+```
+$ cdk deploy --no-advisories
+```
+
+And you can disable all advisories indefinitely by adding this entry to
+`~/.cdk.json`:
+
+```
+"supressAllAdvisories": true
```
eladb, line 78: This should just be `advisories: false` (which is exactly the same as `--no-advisories` in yargs).

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
+# CLI advisories
+
+* **Original Author(s)**: [@otaviomacedo](https://github.com/otaviomacedo)
+* **Tracking Issue**: [#308](https://github.com/aws/aws-cdk-rfcs/issues/308)
+* **API Bar Raiser**: @{BAR_RAISER_USER}
+
+A new CLI feature to notify customers about urgent and important issues that
+require their attention.
+
+## Working backwards
```
eladb, line 10: This is the "working backwards" from a user perspective (the README).

I would add another working backwards section which describes the publisher's experience (e.g. the content we will have in our runbook for how to publish an advisory).

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+
+### What is the technical solution (design) of this feature?
+
+To publish a new advisory, all that the CDK team will have to do is create a
+GitHub issue that fulfils some requirements. The CLI will consume these issues
+using the GitHub API and apply a set of filters to narrow down the list of
+advisories to the context in which it is being run. To improve the publishing
+experience, we will implement a GitHub action that will validate whether the
+issues that are candidates for advisories fulfil the requirements. For a more
+detailed explanation, see the Appendix.
```
eladb, line 142: See above - instead of appendix, describe the publisher experience in an additional "working backwards" section.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+
+* Assuming the advisories were stored in an S3 file (using one of the publishing
+  solutions considered above), we would use CloudFront to distribute them.
+* Use GitHub pages as our distribution mechanism, directly from the repository
+  were the data were stored.
+
+### What is the high-level project plan?
+
+1. Implement the issue validation logic as an NPM package.
+2. Implement the GitHub action, using the validation package.
```
eladb, line 169: What's the benefit of publishing this as a separate npm + GitHub Action? Feels like a simple GitHub workflow in the aws-cdk repo with some internal logic should be sufficient? (and less moving parts -- consider each public artifact like that an API we must maintain).

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+| Affected component     | The CLI or the construct library (or both)                                                                                                                 | A non-empty subset of {CLI, Framework} | Yes        |
+| Affected version range | Version range using the semver format                                                                                                                      | semver                                 | No         |
+| Suggested fix          | An action the user can take to mitigate the problem. If it requires a longer explanation, this field should be omitted in favor of the resource linked to. | Free-form text                         | No         |
+
+In order to keep the issue human readable as well as easily parseable, each of
+these fields should be indicated by a Markdown header (`### Summary`, `###
+Affected component` etc). The piece of text following the header will be
+interpreted as the content of the field.
+
```
eladb, line 228: Add an example

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+#### GitHub workflow
+
+To be considered an advisory, a GitHub issue must be pinned and have the
```
eladb, line 231: I am still wondering if from a workflow perspective it makes more sense to publish this by committing a JSON file in the aws-cdk's main branch. This will have the very important benefit of integrating with our PR workflow and will allow us to run validation checks and human approvals on these advisories before they are published.

If we want, we can easily create a GitHub workflow that will automatically create GitHub issues based on this JSON file so advisories will _also_ appear as pinned GitHub issues, but the source of truth will be the source repo and no the issues.

@rix0rrr @otaviomacedo what do you think?

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+The CLI will make unauthenticated requests to the GitHub Issues API and query
+for issues that match the tag system described in the **GitHub workflow**
+section. Then it will apply additional filters on the result:
+
+* Format filter: whether the body of the issue complies with the data format
+  specification.
+* Version filter: whether the version range contained in the issue matches the
+  CLI version or the framework version (depending on the affected component,
+  also present in the issue). To check which version of the framework is being
+  used, the CLI will read the cloud assembly (`tree.json` file, specifically).
```
eladb, line 251: tree? is that where the version appears? Can we rely on the fact that there is a cloud assembly? There are some CLI commands that won't synthesize a cloud assembly I believe.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,263 @@
...
+Issues that pass these filters will be displayed on the standard output.
+
+The GitHub API imposes a rate limit of 60 requests per hour for unauthenticated
+requests. To avoid hitting that limit, the CLI will cache the results for a set
+period of time (provisionally defined as 1 hour). The filtered advisories and
+the expiration time will be saved to a file in the `$HOME/.cdk/cache` folder.
+When the expiration time is reached, the cache is considered invalid and, at the
+next command execution, the CLI will make a new request to get fresh results.
+
+The CLI will also store the IDs of the acknowledged issues in a file in the
```
eladb, line 262: I think ACKs should be project-specific (i.e. stored in `./cdk.json`).

eladb (dismissed):

text/0308-cli-advisories.md
otaviomacedo: Done.

text/0308-cli-advisories.md
otaviomacedo: Done.

text/0308-cli-advisories.md
otaviomacedo: I just added a reference to the Working backwards section.

text/0308-cli-advisories.md
otaviomacedo: Agreed.

text/0308-cli-advisories.md
otaviomacedo: That's definitely an option. Let me think about the trade-offs involved.

kristianpaul:
This should be a report as well, advices could be forgotten after they keep showing on the screen

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,332 @@
...
+  used, the CLI will read the cloud assembly (`tree.json` file, specifically).
+
+Issues that pass these filters will be displayed on the standard output.
+
+The GitHub API imposes a rate limit of 60 requests per hour for unauthenticated
+requests. To avoid hitting that limit, the CLI will cache the results for a set
+period of time (provisionally defined as 1 hour). The filtered advisories and
+the expiration time will be saved to a file in the `$HOME/.cdk/cache` folder.
+When the expiration time is reached, the cache is considered invalid and, at the
+next command execution, the CLI will make a new request to get fresh results.
```
Chriscbr, line 329: Perhaps we could add a sentence or two describing what the expected behavior should be if the advisories can't be retrieved (either due to network reasons, or due to GitHub API rate limits).

text/0308-cli-advisories.md
otaviomacedo: @eladb the file wouldn't live only in `master`, but also in `v2-main`. Now that v2 is GA, both branches are reasonable candidates for the source of truth. Is it acceptable to just pick one?

eladb: I think a single file on `master` or even in a dedicated repository is preferable. I am not sure I see strong coupling to major version that justifies multiple sources of truth.

otaviomacedo: I'm still more inclined to authoring the issue directly. It already has a natural identifier and a URL for more information. And, in a way, the issue itself is also a source of truth for users, where they can see status updates, comments, other issues that linked to it etc. So we might as well embrace that and make it the single source of truth.

eladb: OK, but I think we need some kind of an approval and verification workflow for these advisories. 

We need to verify that: (a) that the issue is formatted exactly based on what the client expects and then we need some monitoring if the advisory issue is "broken"; and (b) that there's some approval process for publishing an advisory, the content is reviewed etc.

text/0308-cli-advisories.md
eladb: Sorry, I asked if `tree.json` is where we find the version? Isn't that in `manifest.json` somewhere?

otaviomacedo: Oh, my bad. The version in `tree.json` is of `constructs`. This information doesn't seem to be readily available anywhere, actually. We might need to add it to the metadata.

eladb: There's a version handshake between the CLI and the framework that is used to verify that the cloud assembly is compatible. Maybe we can utilize this. See [RFC 110](https://github.com/aws/aws-cdk-rfcs/blob/master/text/00110-cli-framework-compatibility-strategy.md) by @iliapolo.

otaviomacedo: Yes, but that is the version of the cloud assembly schema, which is too coarse-grained for this purpose.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+```
+$ cdk advisories
+```
+
+This command returns zero if there is no advisory and non-zero otherwise. Users
+can then plug this into a pipeline approval workflow and expect manual review if
+there are any advisories.
+
+> Please note that the acknowledgements are made project by project. If you
```
rix0rrr, line 94: To my mind, this makes sense only if the advisory is detected to apply to the current project.

In that case, it's a **risk that's deemed acceptable for this project**.

If we just blanket show anything that's available any time, I feel it makes more sense to store acknowledgements per user.

In those cases, it'sa  **push notification with a receipt confirmation**.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+    "issueUrl": "https://github.com/aws/aws-cdk/issues/16603",
+    "overview": "If a stack is deployed with an S3 bucket with auto_delete_objects=True, and then re-deployed with auto_delete_objects=False, all the objects in the bucket will be deleted.",
+    "component": ["framework"],
+    "version": "<1.126.0"
+  }
+```
+
+3. Create a PR with this change and wait for an approval. Only SDMs have
+   permission to approve PRs in this repository.
+4. When the PR gets merged, the advisory will be visible to all CLI
```
eladb, line 118: Add an item that says that a PR check will validate `advisories.json` for schema or logical issues, and maybe list a bunch of things we want to check. For example:

1. Schema of the file.
2. Check that the issue exists
3. Length limitation of the title & overview
4. Capitalization?
5. Valid semantic versions
6. Valid components

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+there are any advisories.
+
+> Please note that the acknowledgements are made project by project. If you
+acknowledge an advisory in one CDK project, it will still appear on other 
+projects when you run any CDK commands, unless you have suppressed or disabled 
+advisories.  
+
+### Runbook section (internal to the CDK team)
+
+In case of a high-impact issue, follow these steps:
```
RomainMuller, line 101: There should be a specific segment about SECURITY issues. Those have a specific, separate process that MUST be followed. The messaging in those cases has to be approved by Legal, SecOps, our GM, etc...

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+1. File validation on PR. It will block merging if the structure of the file is
+   not compliant with the specification above.
+2. Issue sync-up. When the PR is merged, this action will copy the content of
+   the file over to the GitHub issue it's linked to.
+3. Issue protection. Every change to issues that are linked to some advisory
+   will be checked by this action, to avoid corruption.
+
+#### CLI logic
+
+The CLI will fetch the file from GitHub, parse the content and check whether the
```
MrArnoldPalmer, line 266: What is the logic for fetching the file? Is it on every CLI command or is it cached for a period of time etc? What is the failure mode?

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+### What are we launching today?
+
+A new communication channel between AWS and users of the CDK. Starting on
+version x.y.z of the CDK CLI, customers will be notified, on every command,
+about security vulnerabilities, regressions and usage of unsupported versions.
+
+### Why should I use this feature?
+
+These advisories shown by the CLI contain very important and actionable
+information about problems that directly affect you. They give you an
```
eladb, line 139: They affect the CDK version you are using, but not necessarily your app.

Maybe for future development we can also publish advisories that match against the specific AWS CDK constructs you are using (we do have that information in the CDK analytics blob).

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+#### CLI logic
+
+The CLI will fetch the file from GitHub, parse the content and check whether the
+version range contained in the advisory matches the CLI version or the framework
+version (depending on the affected component).
+
+Since the CLI knows its own version, checking against the version range of the
+advisory is trivial. The version of the framework, however, is not readily
+available anywhere. To address this, we will start writing the framework version
+to the Cloud Assembly, in a place where the CLI can read it.
```
rix0rrr, line 273: Not strictly this, but a close proxy for it is already available in the form of constructs and their versions.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+3. Create a PR with this change and wait for an approval. Only SDMs have
+   permission to approve PRs in this repository.
+4. When the PR gets merged, the advisory will be visible to all CLI
+   installations. The GitHub issue will also be automatically updated with the
+   information contained in the file. All the necessary tags will also be added
+   automatically. 
+5. You can keep updating the issue normally, as new information comes in, but
+   you're not allowed to touch the sections auto-generated from the advisories
+   file.
```
eladb, line 124: Do we ever plan to remove advisories? We should have some kind of monitoring to make sure they don't linger for too long.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+RFC are not considered strong enough by the CDK team, we should not implement 
+this feature.
+
+### What is the technical solution (design) of this feature?
+
+Advisory information will be available as a static file, served from GitHub. 
+The CLI will consume this file from and apply a set of filters to narrow down 
+the list of advisories to the context in which it is being run. We will also 
+implement some GitHub actions to validate and copy the contents of the file over
+to the issue. For a more detailed explanation, see the Appendix.
```
RomainMuller, line 173: What happens in network-isolated/controlled environments? The documentation needs to be EXTREMELY clear that this is calling out to some GitHub-controlled URL. AppSec might also somewhat frown on that premise...

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+No.
+
+### What alternative solutions did you consider?
+
+On the publishing side:
+
+* Implementing a new internal REST service to manage the advisories. Overly
+  complex for this use case.
+* Authoring the content directly on the GitHub issue. There is good way to
```
eladb, line 185:
```suggestion
* Authoring the content directly on the GitHub issue. There is no good way to
```

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+### What is the expected lifetime of advisories?
+
+We should expect advisories to be available for months. This is usually the case
+when we want users to upgrade the construct library version.
+
+### Why do we need to give users the option to suppress advisories?
+
+This is useful when the CDK is running as part of larger script and users want
+to hide the output.
```
RomainMuller, line 210: ... or if users don't want GitHub to know when they run CDK CLI commands.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+We will create a new repository, `cdklabs/aws-cdk-advisories`, dedicated to host
+the advisories file. As usual, in order to be merged, any change to this file
+will have to be published as a PR and approved. Using a CODEOWNERS file, we will
+restrict the permission to approve PRs only to SDMs. The file will contain a
+list of advisories, each having the following fields:
+
+|     Field    |                           Description                          | Format                          | Mandatory? |
+|:------------:|:--------------------------------------------------------------:|---------------------------------|:----------:|
+| `title`      | The title of the incident                                      | Free form text                  | Yes        |
```
eladb, line 248: Length limits?

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+
+
+We will also implement three GitHub actions on this repository:
+
+1. File validation on PR. It will block merging if the structure of the file is
+   not compliant with the specification above.
+2. Issue sync-up. When the PR is merged, this action will copy the content of
+   the file over to the GitHub issue it's linked to.
+3. Issue protection. Every change to issues that are linked to some advisory
+   will be checked by this action, to avoid corruption.
```
eladb, line 262: What's the difference between 2 and 3? I'd assume it's a workflow that just iterates over all the advisories and syncs up their issues. No?

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+Regardless of the state of this flag and the advisories you have acknowledged,
+you can always show the currently active advisories:
+
+```
+$ cdk advisories
+```
+
+This command returns zero if there is no advisory and non-zero otherwise. Users
```
corymhall, line 90: I think that if we do return a non-zero code and advise users to use this in a pipeline then we need to be very purposeful about what we publish as advisories. If it is hit or miss then I think we should just always return 0 and not advertise this as a use case.

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,294 @@
+# CLI advisories
+
+* **Original Author(s)**: [@otaviomacedo](https://github.com/otaviomacedo)
+* **Tracking Issue**: [#308](https://github.com/aws/aws-cdk-rfcs/issues/308)
+* **API Bar Raiser**: @eladb
+
+A new CLI feature to notify customers about urgent and important issues that
```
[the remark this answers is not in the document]
otaviomacedo: Oh, that's right. I missed that one.

eladb (approved):

text/0308-cli-advisories.md
```diff
@@ -0,0 +1,294 @@
...
+particular, this feature should not be used for things like marketing campaigns, 
+blog post announcements and things like that. If the mechanisms proposed in this
+RFC are not considered strong enough by the CDK team, we should not implement 
+this feature.
+
+### What is the technical solution (design) of this feature?
+
+Advisory information will be available as a static file on a GitHub repository.
+This file will be automatically synced to an S3 bucket and distributed via
+CloudFront. The CLI will consume this file from and apply a set of filters to
```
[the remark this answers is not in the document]
otaviomacedo: Based on the internal guidance on the names, something like `notices.cdk.developer-tools.aws.dev` would be a good choice.

eladb:
Looking forward to seeing this in production.

Issue #52, opened 2019-12-16
Support resource import


NetaNir:
|PR|Champion|
|--|--------|
|#392 | @rix0rrr        |

## Description

CloudFormation released[ resource import](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-supported-resources.html
)
How should we support it in the CDK?
See #84 for a related feature/discussion.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
- [x] Released As Feature Preview
- [ ] Resolved (feature stable)

eschulma:
This looks like a duplicate of [this issue](https://github.com/aws/aws-cdk-rfcs/issues/84)

neg3ntropy:
I am confused about the Importing Existing External Resources from the [developer docs](https://docs.aws.amazon.com/cdk/latest/guide/resources.html).

It looks likes it is a different thing than this, right?
It's not meant as a way to import a resource and manage it, but just a way to access its attributes. Conflated terminology?

shearn89:
This would be of great help, especially when getting to grips with the CDK and learning about AWS services. For example I've set up some proof-of-concept stuff to show my team, but now want to manage it via a CDK template. I'm probably going to have to destroy and recreate, but if i was able to run the cdk with an `import` flag on the deploy then I'd be much happier!

marcindulak:
> I am confused about the Importing Existing External Resources from the [developer docs](https://docs.aws.amazon.com/cdk/latest/guide/resources.html).
> 
> It looks likes it is a different thing than this, right?
> It's not meant as a way to import a resource and manage it, but just a way to access its attributes. Conflated terminology?

Pulumi is more clear about it: they talk about `referencing` vs `adopting` existing resources https://www.pulumi.com/blog/adopting-existing-cloud-resources-into-pulumi/.

A clear distinction between `referencing` and `importing` resources will be needed in the cdk documentation, and I guess cdk will follow the naming used by cloudformation. I've just opened an issue https://github.com/aws/aws-cdk/issues/7370 which may require such clarification.

offbyone:
[...]

PR #392, opened 2021-12-20
RFC 52: Support for importing existing resources into a CDK stack

tomas-mazak:
This is a request for comments about "Support importing resources". See #52 for
additional details. 

[Rendered version](https://github.com/tomas-mazak/aws-cdk-rfcs/blob/0052-resource-importing-support/text/0052-resource-importing-support.md)

---

### DIFF
--- a/text/0052-resource-importing-support.md
+++ b/text/0052-resource-importing-support.md
@@ -248,18 +248,20 @@ CloudFormation API doesn't allow any non-import changes in the template for
 import operation. This is a challenge for CDK, as:
 
 * `CDKMetadata` resource changes on each synthesis, so the import operation
-  never succeeds with the unmodified template. To resolve this, the proposed
-  solution restores CDKMetadata to the currently deployed version before
-  passing the template to ChangeSet creation step.
+  never succeeds with the unmodified template. To resolve this, this solution
+  doesn't pass the newly synthesized template to the import operation - instead,
+  the currently deployed template is fetched and only the resources to be imported
+  are added to it (copied over from the new template), producing an interim
+  _import template_.
 * High-level constructs produce multiple underlying CloudFormation resources
   and only some of those might need to be imported - the rest might need to
   be created (e.g. `ec2.Vpc()` might require VPC being imported, but subnets
   being created). Furthermore, CDK can't automatically determine what
   resources to import and what to create. For this to behave correctly, the
   solution will prompt user for each resource - whether it should be imported
-  or not. If not, the resource is removed from the template before creating
-  the change set. The remaining (not imported) resources can be created by
-  a subsequent `cdk deploy` command.
+  or not. If not, the resource is omitted (not added to the import template).
+  The remaining (not imported) resources can be created by a subsequent
+  `cdk deploy` command.
 
 #### An explicit DeletionPolicy must be set on imported resources
 
@@ -267,7 +269,8 @@ CloudFormation import API requires all imported resources to have specified
 explicit DeletionPolicy. CDK satisfies this requirement for typical stateful
 resources (e.g. S3 buckets), but not _all_ resources (e.g. VPCs). The solution
 will inject a `DeletionPolicy` with a value of `Delete` (CloudFormation's
-default) in the template for all resources without `DeletionPolicy` set by CDK.
+default) in the import template for all resources without `DeletionPolicy` set
+by CDK.
 
 This manipulation will only happen during the import operation and therefore it
 is not persistent. On the next deploy, the property is dropped. However, as
@@ -324,12 +327,34 @@ feature and so it introduces the import feature's inherent issues in CDK, like:
 
 ### What is the high-level project plan?
 
-A simple [PoC Draft PR](https://github.com/aws/aws-cdk/pull/17666) was created
-in the main `aws-cdk` repository that is being iteratively expanded with the
-proposed features. Once the approach is agreed and the agreed solution is
-implemented, the PR will be moved from "Draft" to "Open" status and CDK team
-member code review will be requested. Once all approved, the PR will be ready
-to be merged.
+The RFC will be implemented in following stages:
+
+#### 1. Interactive import
+
+Basic wrapper around the CloudFormation Import API, `cdk import` CLI command
+with only basic options. Interactive prompts for resource identification. Use
+of the default credentials (same as `cdk deploy`) for all API calls. A simple
+[PoC Draft PR](https://github.com/aws/aws-cdk/pull/17666) was created in the
+main `aws-cdk` repository that implements this stage.
+
+#### 2. Non-interactive import
+
+Extend the CLI interface to allow passing "resource mapping" hints on command
+line, instead of the interactive prompts. Secondly, provide a helper to
+create the "resource mapping" file based on user input (interactive). This
+stage allows splitting the import workflow into prepare and execute phases,
+possibly using CI/CD
+
+#### 3. Using the lookup role
+
+Run read-only operations using the lookup role from CDK Toolkit stack instead
+of the default credentials.
+
+#### 4. Leverage Cloud Control API to improve user experience
+
+By using Cloud Control API, the CLI interface can be relatively easily extended
+to provide much better user experience, see
+[possible future expansions](#possible-future-expansions) section.
 
 ### Are there any open issues that need to be addressed later?
 
@@ -357,6 +382,12 @@ might still be needed.
 
 It might be considered to release such convenience features iteratively
 per resource type, beginning with the most frequently used types.
+
+This proposal also only concerns with a single CDK stack at a time - the PoC
+code fails if multilpe stacks are matched. For the app stacks refactoring
+use case it would be beneficial to support batch operations (same import flow
+on multiple stacks) and a _move operation_ - single command to remove a
+resource from one stack and import it into another).
 
 ### Links
 
--- a/text/0052-resource-importing-support.md
+++ b/text/0052-resource-importing-support.md
@@ -1,0 +1,268 @@
+# Support for importing existing resources into a CDK stack
+
+* **Original Author(s):**: @tomas-mazak
+* **Tracking Issue**: #52
+* **API Bar Raiser**: @rix0rrr
+
+Allow users to import resources, that were created manually or by different
+means, into CDK stacks.
+
+## Working Backwards - README
+
+```
+cdk import [STACK]
+
+Import existing resource(s) into the given STACK
+
+Options:
+[... standard CDK CLI and CDK CLI `deploy` options ...]
+  -o, --create-resource-mapping  If specified, CDK will generate a mapping of
+                                 existing physical resources to CDK constructs to
+                                 be imported as. The mapping will be written in
+                                 the given file path. No actual import operation
+                                 will be performed                      [string]
+  -m, --resource-mapping         If specified, CDK will use the given file to
+                                 map physical resources to CDK resources for
+                                 import, instead of interactively asking the
+                                 user. Can be run from scripts          [string]
+  -f, --force                    Allow importing resources with properties
+                                 different from the CDK construct configuration.
+                                 This will most likely lead to a stack drift.
+      --non-interactive          Suppress all interactive prompts. All
+                                 confirmation prompts assume yes, all resources
+                                 that can't be identified will be skipped from
+                                 import. (default: false if stdin is a TTY, true
+                                 otherwise)
+```
+
+Sometimes, it is useful to be able to import (enroll/adopt/...) AWS resources,
+that were created by different means (manually in AWS console, using different
+orchestrator, etc), into a CDK stack. Some resources can simply be deleted and
+recreated by CDK, but for others, this is not convenient: Typically stateful
+resources like S3 Buckets, DynamoDB tables etc, cannot be easily deleted without
+an impact on the service.
+
+To import an existing resource into a CDK stack:
+
+- Run a `cdk diff` to ensure there are no pending changes to the CDK stack you
+  want to import resources into - if there are, apply/discard them first.
+- Add corresponding constructs for the resources to be added in your stack - for
+  example, for an S3 bucket, add something like
+  `new s3.Bucket(this, 'ImportedS3Bucket', {});` - **no other changes must be
+  done to the stack before the import is completed**.
+- Run `cdk import` command - if there are multiple stacks in the CDK app, pass a
+  specific stack name as an argument.
+- If the resource definition contains all information needed for the import,
+  this happens automatically (e.g. an `s3.Bucket` construct has an explicit
+  `bucketName` set), otherwise, CDK will display an interactive prompt to
+  provide neccessary identification information (e.g. the bucket name).
+- After `cdk import` reports success, the resource is managed by CDK. Any
+  subsequent changes to the construct configuration will be reflected on the
+  resource.
+
+---
+
+Ticking the box below indicates that the public API of this RFC has been
+signed-off by the API bar raiser (the `api-approved` label was applied to the
+RFC pull request):
+
+```
+[ ] Signed-off by API Bar Raiser @rix0rrr
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+New CDK CLI sub-command `cdk import` that works just like `cdk deploy` but for
+newly added constructs in the stack, instead of creating new AWS resources, it
+attempts to import corresponding existing resources.
+
+### Why should I use this feature?
+
+When an already existing AWS resource needs to be brought under CDK management
+the most convenient way of achieving this is using this feature (`cdk import`).
+
+## Internal FAQ
+
+### Why are we doing this?
+
+This solution is exposing an existing [CloudFormation resource
+import](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html)
+feature in a CDK native way. CDK's goal is to provide convenient abstraction
+over sometimes complex CloudFormation aspects and this new feature is aligned
+with this goal. Additionally, other IaC orchestrators' also provide a
+corresponding feature, for example:
+
+* [terraform import](https://www.terraform.io/cli/import)
+* [pulumi import](https://www.pulumi.com/docs/guides/adopting/import/#pulumi-import-command)
+
+### Why should we _not_ do this?
+
+The feature might be considered independent enough to provide a separate tool
+for it (e.g. `cdk-import` node package).
+
+### What is the technical solution (design) of this feature?
+
+* [Proof-of-Concept PR](https://github.com/aws/aws-cdk/pull/17666)
+
+The solution leverages CloudFormation resource import API. The API is very
+similar to the regular deploy API, consisting of ChangeSet creation and
+ChangeSet execution actions.
+
+* ChangeSet of type `IMPORT` is created (for deploy it is `CREATE` or
+  `UPDATE`)
+* on ChangeSet creation, an additional structure has to be passed with the
+  action, an array of [ResourceToImport](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ResourceToImport.html)
+  structures to instruct CloudFormation what actual existing resources to
+  import as corresponding logical resources in the template
+* ChangeSet is executed as on regular deployment
+
+The solution is leveraging existing wrappers around ChangeSet creation and
+execution in `aws-cdk` codebase, extending them with the options required for
+importing.
+
+Following characteristics/limitations of the API pose a challenge to the
+solution:
+
+#### Inconsistent resource identifiers between resource types
+
+In the ResourceToImport structure, each type of resource is identified by a set
+of resource properties (typically of length 1 - by a single property). The
+property name is different between different resource types (e.g. `BucketName`
+for S3 buckets, or `VpcId` for VPCs).
+
+To be able to put the structure together, the import logic must be able to
+identify the required properties and then look up the values. For some
+resources, this can be automatically looked up from the CDK construct
+configuration (e.g. an S3 bucket having specific `bucketName` configured) but
+for some resources, the code has to rely on user input.
+
+To address this, the solution is divided into two steps:
+
+##### Physical resource identification
+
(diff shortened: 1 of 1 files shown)


text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+Sometimes, it is useful to be able to import (enroll/adopt/...) AWS resources,
+that were created by different means (manually in AWS console, using different
+orchestrator, etc), into a CDK stack. Some resources can simply be deleted and
+recreated by CDK, but for others, this is not convenient: Typically stateful
+resources like S3 Buckets, DynamoDB tables etc, cannot be easily deleted without
+an impact on the service.
+
+To import an existing resource into a CDK stack:
+
+- Run a `cdk diff` to ensure there are no pending changes to the CDK stack you
```
rix0rrr, line 47: Is this something the CLI can identify automatically? If so, worth calling out.

(We should assume users don't read docs and will get into all kinds of bad situations if we'll let them.)

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+  want to import resources into - if there are, apply/discard them first.
+- Add corresponding constructs for the resources to be added in your stack - for
+  example, for an S3 bucket, add something like
+  `new s3.Bucket(this, 'ImportedS3Bucket', {});` - **no other changes must be
+  done to the stack before the import is completed**.
+- Run `cdk import` command - if there are multiple stacks in the CDK app, pass a
+  specific stack name as an argument.
+- If the resource definition contains all information needed for the import,
+  this happens automatically (e.g. an `s3.Bucket` construct has an explicit
+  `bucketName` set), otherwise, CDK will display an interactive prompt to
```
rix0rrr, line 57: I think there's more to be said about the experience here (with the prompts and their meaning we already talked about on the PR)

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+### Why are we doing this?
+
+This solution is exposing an existing [CloudFormation resource
+import](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html)
+feature in a CDK native way. CDK's goal is to provide convenient abstraction
+over sometimes complex CloudFormation aspects and this new feature is aligned
+with this goal. Additionally, other IaC orchestrators' also provide a
+corresponding feature, for example:
+
+  * [terraform import](https://www.terraform.io/cli/import)
```
rix0rrr: Oh excellent research! That clinches the command's name then :)

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+  example, for an S3 bucket, add something like
+  `new s3.Bucket(this, 'ImportedS3Bucket', {});` - **no other changes must be
+  done to the stack before the import is completed**.
+- Run `cdk import` command - if there are multiple stacks in the CDK app, pass a
+  specific stack name as an argument.
+- If the resource definition contains all information needed for the import,
+  this happens automatically (e.g. an `s3.Bucket` construct has an explicit
+  `bucketName` set), otherwise, CDK will display an interactive prompt to
+  provide neccessary identification information (e.g. the bucket name).
+- After `cdk import` reports success, the resource is managed by CDK. Any
```
rix0rrr, line 59: What are typical ways in which this operation will fail? How are we displaying this to the user?

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+identify the required properties and then look up the values. For some
+resources, this can be automatically looked up from the CDK construct
+configuration (e.g. an S3 bucket having specific `bucketName` configured) but
+for some resources, the code has to rely on user input.
+
+To address this, the solution is divided into two steps:
+
+##### Physical resource identification
+
+First, identify what properties must be looked up/asked for for each resource to
```
rix0rrr, line 145: Sweet! Does this mean we don't need to hardcode support for each resource type anymore?

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+Import existing resource(s) into the given STACK
+
+Options:
+[... standard CDK CLI and CDK CLI `deploy` options ...]
+  -o, --create-resource-mapping  If specified, CDK will generate a mapping of
+                                 existing physical resources to CDK constructs to
+                                 be imported as. The mapping will be written in
+                                 the given file path. No actual import operation
+                                 will be performed                      [string]
+  -m, --resource-mapping         If specified, CDK will use the given file to
```
rix0rrr, line 24: Do you have a particular use case  in mind for these parameters? I see how it *might* be useful, but if this is with an eye towards a non-permissioned CI/CD workflow, just adding these flags is not going to be enough by itself.

Can you describe in the README a use case a user might find themselves in, in which they want this feature, and what the workflow looks like in that case?

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,272 @@
...
+Based on command line flags, the compiled "resource mapping" structure is either
+passed on to the next phase, or written to the specified file.
+
+##### Import execution
+
+Based on command line flags, this phase either reads "resource mapping" from the
+specified file, or it will be passed the resource mapping directly from the
+previous phase.
+
+First, the sanity checks are performed and then the change set is created. By
```
rix0rrr, line 165: Which sanity checks again?

text/0052-resource-importing-support.md
rix0rrr: Sounds like we're going to drop changes from the template before performing the import (from the text below). So that could/would include any unintended changes, and so this caveat doesn't really apply anymore?

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,268 @@
...
+  never succeeds with the unmodified template. To resolve this, the proposed
+  solution restores CDKMetadata to the currently deployed version before
+  passing the template to ChangeSet creation step.
+* High-level constructs produce multiple underlying CloudFormation resources
+  and only some of those might need to be imported - the rest might need to
+  be created (e.g. `ec2.Vpc()` might require VPC being imported, but subnets
+  being created). Furthermore, CDK can't automatically determine what
+  resources to import and what to create. For this to behave correctly, the
+  solution will prompt user for each resource - whether it should be imported
+  or not. If not, the resource is removed from the template before creating
```
rix0rrr, line 184: Just thinking: the other way around might be easier.

Why don't we download the *current template* from the Stack, and then add the resources we're trying to import? Seems safer than dropping resources from the target template.

So the workflow looks a bit like:

* Take the new template, identify candidate resources to import by looking at `cdk diff`
* Obtain the import map by a combination of heuristics and user input. Include the resource definitions of the resources in the import map

Then:

* Download the old template
* Add the resources from the previous step to the template
* (Potentially: check the integrity of the template)
* Do an IMPORT update with the import map

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,268 @@
...
+  the change set. The remaining (not imported) resources can be created by
+  a subsequent `cdk deploy` command.
+
+#### An explicit DeletionPolicy must be set on imported resources
+
+CloudFormation import API requires all imported resources to have specified
+explicit DeletionPolicy. CDK satisfies this requirement for typical stateful
+resources (e.g. S3 buckets), but not _all_ resources (e.g. VPCs). The solution
+will inject a `DeletionPolicy` with a value of `Delete` (CloudFormation's
+default) in the template for all resources without `DeletionPolicy` set by CDK.
```
rix0rrr, line 194: Fuck yeah 👍

You really thought through a lot of detail here, nice job!

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,268 @@
...
+#### CloudFormation doesn't check the physical resource's configuration
+
+If a physical resource identified by the provided identifiers does not exist,
+the import operation (change set creation) will fail with clear-enough message.
+However, if the resource does exist, but it's actual configuration doesn't match
+the CloudFormation configuration, import operation succeeds and *not update*
+the resource, causing the stack to be drifted. This can be left to the user
+to handle (manually checking the config prior to the import or running drift
+detection after the import), but for convenience, this solution uses
+CloudControl API to look up the physical resource and validate its configuration
```
rix0rrr, line 211: This is worth calling out in the above section as well.

I assume this happens before doing the actual import?

rix0rrr: I wonder if running a drift detection afterwards might be good enough for a start?

The CCAPI will require us to do a JSON diff, and I'm not sure it's complete enough. Have you experimented with this?

text/0052-resource-importing-support.md
rix0rrr: I wonder what we're going to do about dependencies on non-imported resources here... but I also don't have a good intuition for where and how those are going to crop up. I guess the best path forward is roll this feature out and seeing where it's going to creak 😅

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,268 @@
...
+  be up to date with the CDK configuration after the import operation, another
+  `deploy` is needed.
+
+### What is the high-level project plan?
+
+A simple [PoC Draft PR](https://github.com/aws/aws-cdk/pull/17666) was created
+in the main `aws-cdk` repository that is being iteratively expanded with the
+proposed features. Once the approach is agreed and the agreed solution is
+implemented, the PR will be moved from "Draft" to "Open" status and CDK team
+member code review will be requested. Once all approved, the PR will be ready
```
rix0rrr, line 255: Does that mean everything is going to be implemented in 1 big PR?

Is there a logical order in which we can deliver things incrementally?

What's the aftercare/future extensions story?

text/0052-resource-importing-support.md
rix0rrr: What's the structure of the resource mapping file? 

If this is going to be useful in an unattended scenario, it will have to be a 2-level map I think, right?

```
{ stack -> { logicalId -> physicalId } }
```

?

If so, you probably want to be able to combine the outputs of multiple `import` calls into the same output file.

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,268 @@
...
+configuration (e.g. an S3 bucket having specific `bucketName` configured) but
+for some resources, the code has to rely on user input.
+
+To address this, the solution is divided into two steps:
+
+##### Physical resource identification
+
+First, identify what properties must be looked up/asked for for each resource to
+be imported. This information can be obtained by executing CloudFormation API
+action `GetTemplateSummary` with the new template. The action returns the
```
rix0rrr, line 147: Does this require additional permissions for the `cdk-deploy` role in the bootstrap stack? Or are we going to use the `cdk-lookup` role for this?

rix0rrr (commented):
I love this! Thanks for your thoroughness here!

text/0052-resource-importing-support.md
tomas-mazak: So there are two kinds of possible changes: either an already existing resources has its configuration changed and this is easily detectable and it's already handled by the PR code: https://github.com/aws/aws-cdk/pull/17666/files#diff-f0f2c0ae1f83424770eef7a1108e226ea791642f88a7cc18b8f41443b62479b9R120-R128

But for newly defined constructs/resources, I can't see how we could determine if user intent is to create them or import them. We interactively prompt user anyway so they will be able to skip the import (and we delete it from the template in such case for the import operation).

So yes, this sentence can be dropped. I was just thinking that it's a good practice to see if there's any difference between declared and deployed configuration, before the changes are made for the import operation. On the other hand, this is probably a good practice before _any_ changes to the code, not just import. So it might not belong here.

text/0052-resource-importing-support.md
tomas-mazak: Sure, I will try to expand on this. On top of what we discussed on the PR (and I believe I've implemented accordingly), there's also an aspect of detecting if stdin is a TTY or not and behaving accordingly. If not, there's no point to use the prompts.

text/0052-resource-importing-support.md
tomas-mazak: Since I use the existing code for creating/executing the changeset, it will report the error exactly the same way as `cdk deploy`. Typically, the changeset creation fails, I haven't seen the actual execution to fail. If a changeset creation fails, I don't think any further action is needed as no changes were made to the stack yet. And from what I've seen, the CFN error messages are quite clear.

I can imagine a corner case though when a changeset creation passes, but then the physical resource is deleted, before the change set is executed. I am not sure what happens - whether the operation passes and the stack is drifted or it fails. Can try to simulate. Also, there might be different reasons for the changeset execution to fail that I am not aware. The question is if we need to handle those or simply report back to the user, same as with deploy.

text/0052-resource-importing-support.md
tomas-mazak: Exactly, this is a very neat feature from CFN :)

text/0052-resource-importing-support.md
tomas-mazak: Sanity checks means:
* check that no existing resource modifications are made in the new template
* check that the physical resources exist and are configured consistent with the template (using CloudControl API - this part is not implemented yet)

I didn't detail it here, because I mention it a bit later in the text. But I can do here as well if you want.

text/0052-resource-importing-support.md
tomas-mazak: Yes, I am happy to revert the logic here, although it should produce exactly the same template, as:
* we already download the current template and perform the diff
* for any modification (besides CDKMetadata that we copy over) we fail so non-addition difference will never get to the import action
* for any newly defined resource, we decide whether it is going to be imported (we keep it) or not (we remove it) ... which would be the same other way around: if it's going to be imported we add it, otherwise we skip it

But now thinking about it, the above applies only to the `Resources` section of the template. I will double-check what if `Parameters` or `Outputs` change - this can add some more complexity. Imagine a resource properties of a resource-to-be-imported reference a parameter that did not exist in the previous template.

text/0052-resource-importing-support.md
tomas-mazak: I have experimented with this and it works quite nicely. The greatest benefit is that CCAPI uses the same resource type nomenclature as CloudFormation so we can simply pass the type to the API call. There are two issues I am still looking at though: 

Firstly, I need to understand how the "import identifiers" of a resource relate to "primary" and "secondary" indentifiers that CCAPI uses. While for CFN import, explicit identifier _key_ has to be provided (e.g. `BucketName` for S3 Bucket, or `VpcId` for VPC), with CCAPI call you always provide `Identifier` key that has a value based on the type - it can be either primary or secondary identifier. My plan is to get all resource schemas from CFN resource catalog, extract primary/secondary identifiers info, then run get-template-summary for a template containing all resource types and do some analysis on that. It will be somewhat time consuming though, not sure when I can get to that.

Secondly, CCAPI returns _all_ resource properties, but some of them are not mandatory in CFN templates and if not specified, will be set to default values. So in order to do the comparison correctly, I would need to know the default values of all optional properties of all resources, didn't find a way to do that yet (in best-effort approach, we could only compare properties explicitly set in CFN).

I agree with your suggestion: If it is OK to get this merged without this extra validation, let's do it and either run drift detection automatically or display a post import message to tell user to run it. And leave the CCAPI extra validation for the next iteration.

text/0052-resource-importing-support.md
tomas-mazak: Not sure about the lifecycle of the code post-merge, I guess that is for you and your team to decide - I can't guarantee I will be able to support this afterwards, although I will certainly watch the space for some time as I will be interested how is my so-far-biggest contribution doing :) 

It might make sense to split the feature into multiple stages (and in fact, I believe the first stage might be pretty much ready now), I would take your guidance on how to split it.

text/0052-resource-importing-support.md
tomas-mazak: That's a very good point. Not sure but I will have a look

text/0052-resource-importing-support.md
tomas-mazak: cheers!

text/0052-resource-importing-support.md
tomas-mazak: thanks!

text/0052-resource-importing-support.md
tomas-mazak: I have updated this as requested.

text/0052-resource-importing-support.md
tomas-mazak: dropped

text/0052-resource-importing-support.md
tomas-mazak: Expanded accordingly

text/0052-resource-importing-support.md
tomas-mazak: As per different discussion, I would say the checks that require CloudControl API should be left for next phase - they are not strictly needed because some fail conditions will be evaluated on CFN change set creation and the drift can be detected by running drift detection. So for now, the sanity checks just check if there are no update/delete-type changes in the template (as that is not supported on import), I've updated the sentence to mention that explicitly.

text/0052-resource-importing-support.md
tomas-mazak: For now, I have updated the RFC to omit the CCAPI validations and I have added a "Possible future expansions" section with some thoughts on how to leverage CCAPI to provide a better user experience in future - perhaps beyond simple validations.

text/0052-resource-importing-support.md
tomas-mazak: So here I am actually using an existing `SdkProvider` that is initialized in the main file of CDK CLI (same as the `cdk deploy` command is using). If I understand the code right, it will only ever use the AWS CLI credentials - it doesn't seem to be assuming any role from the bootstrap stack or so:

https://github.com/aws/aws-cdk/blob/master/packages/aws-cdk/bin/cdk.ts#L212-L219

So I don't think the bootstrap stack needs to be updated, am I correct?

text/0052-resource-importing-support.md
rix0rrr: Probably fine.

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,365 @@
...
+  originally. As the project got more complex, the team opts for
+  infrastructure-as-code approach using CDK. To be able to organize the existing
+  resources in CDK stacks without recreation, the import feature is needed.
+- **CDK migration:** The team uses an infrastructure-as-code tool such as
+  Terraform, Pulumi or plain CloudFormation and chooses to migrate onto CDK. To
+  perform a smooth migration without resources recreation, it is important to:
+  1. make the original IaC orchestrator "abandon" the resource - remove it
+     from its orchestration without physically removing the resource
+  2. "import" the resource in a CDK stack
+- **App refactoring:** The team already uses CDK but a major refactor is needed
```
rix0rrr, line 154: This is good to identify as a use case. I like that you thought about it.

But in practice, we're going to have (in AWS) 20+ copies of every app deployed... so if you really want to refactor something across 20 copies, we're going to need more automation to take care of this.

In any case -- not for this RFC, we will deal with that when we get to it! Perhaps people will be clever and build something on top of the structured input file.

text/0052-resource-importing-support.md
rix0rrr: It probably will eventually, but for now I suppose the trivial solution will do.

Can you add a note in the RFC (execution plan probably) that we're going to initially support "current" credentials, and later on support using bootstrapped roles?

text/0052-resource-importing-support.md
rix0rrr: Exactly. It just feels to me that taking a known-good template and doing a very concrete set of operations to it that we thought about beforehand is going to be safer than taking an unknown, new template, and hoping we remember to take out all the unintended modifications.

On the other hand, some `Parameters` and `Mappings` might be necessary to support the new resource?

text/0052-resource-importing-support.md
rix0rrr: I'm thinking we slice the elephant here. We'll say something like:

- First we implement interactive import with current credentials
- Then noninteractive import
- Then support using the lookup role
- Then support CCAPI validations

I see you put some in the extensions piece, that's a good place too.

text/0052-resource-importing-support.md
```diff
@@ -0,0 +1,365 @@
...
+(Does the physical resource exist? Does it not belong to another CloudFormation
+stack? Is the resource configuration consistent with the actual state?).
+
+However, such features would be difficult to implement and might require
+resource type-specific logic. Using [Cloud Control API](https://aws.amazon.com/blogs/aws/announcing-aws-cloud-control-api/)
+might avoid using service-specific APIs, but some resource type specific code
+might still be needed.
+
+It might be considered to release such convenience features iteratively
+per resource type, beginning with the most frequently used types.
```
rix0rrr, line 359: I think mass refactoring support (across multiple stacks) would be a great extension we could write down here.

It's going to require some additional tooling I think.

rix0rrr (changes requested):
This thing is as good as done! I like it. Some small changes and then we can ship it. Thanks!

text/0052-resource-importing-support.md
tomas-mazak: rephrased to reflect this

text/0052-resource-importing-support.md
tomas-mazak: modified as suggested

text/0052-resource-importing-support.md
tomas-mazak: done

text/0052-resource-importing-support.md
tomas-mazak: added in the "possible future expansions" section

text/0052-resource-importing-support.md
tomas-mazak: done

rix0rrr (commented):
I like it! Going to merge this.

rix0rrr (approved):
But then I mis-clicked

PR #397, opened 2022-01-03
feat(RFC-204): propose go generics use in go bindings

RomainMuller:
This proposes making use of `go 1.18+` generics in generated go bindings in order to label optional values as such, and remove the pervasive use of pointers.

---

### DIFF
--- a/text/204-golang-bindings.md
+++ b/text/204-golang-bindings.md
@@ -1195,7 +1195,7 @@ interfaces with a non-exported implementing struct. A run-time check can be wove
 provide helpful error messages (this can be done regardless of how optionals are implemented, as the type model has the
 required information).
 
-While turning an required parameter to optional is not a usage-breaking change in this scenario, it is as an
+While turning a required parameter to optional is not a usage-breaking change in this scenario, it is as an
 override-breaking change, as the type signature of the function that must be implemented to satisfy an interface
 changes. This has consequences in case a type extends a go interface from a dependency.
 
--- a/text/204-golang-bindings.md
+++ b/text/204-golang-bindings.md
@@ -1179,6 +1179,29 @@ validity checks on structs (ensuring all required properties were set) without h
 of a custom validator for each struct type; instead relying solely on the type information that is available from
 the `reflect` standard library package.
 
+```go
+// Given this struct
+type WithRequiredProperties struct {
+    Required     InterfaceType
+    Optional     Option[InterfaceType]
+}
+
+// The following are all valid, compiler-wise
+a := WithRequiredProperties{}
+b := WithRequiredProperties{Optional: someValue}
+c := WithRequiredProperties{Required: someValue}
+d := WithRequiredProperties{Required: someValue, optional: someValue}
+
+// When that struct is passed over the wire to node, we can determine that a and b are invalid
+SomeStatic(a) // #=> Panic ("missing value for required property 'Required'")
+SomeStatic(b) // #=> Panic ("missing value for required property 'Required'")
+SomeStatic(c) // #=> OK
+SomeStatic(d) // #=> OK
+
+// Without `Option[T]`, this can only be achieved by generating a validator function, and calling that on EVERY usage
+// site of the struct.
+```
+
 Not using pointers everywhere also enables the runtime library to do away with a lot of pointer indirection work, as
 values may no longer be of pointer types and instead are always either a plain struct (in the case of jsii structs), or
 an interface type. This results in simplifications in some areas of the runtime library.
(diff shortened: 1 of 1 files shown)


text/204-golang-bindings.md
```diff
@@ -1068,12 +1068,160 @@ This allows construction like so:
...
+
+Not using pointers everywhere also enables the runtime library to do away with a lot of pointer indirection work, as
+values may no longer be of pointer types and instead are always either a plain struct (in the case of jsii structs), or
+an interface type. This results in simplifications in some areas of the runtime library.
+
+Non-optional jsii struct parameters are passed as go structs, which provides compile-time gurantee that the value is not
+nil (but does not gurantee all fields have been set to a value other than their zero-value). Thi is more idiomatic go
+than the option of passing pointers to structs.
+
+### Downsides
```
eladb, line 1190: Can you add code samples to demonstrate the downsides?

text/204-golang-bindings.md
```diff
@@ -1068,12 +1068,160 @@ This allows construction like so:
...
+
+In go, `nil` is a valid value at any location where an interface is expected. This means it is not possible to achieve
+compile-time safety with this approach in places where classes are passed as arguments, since those are represented as
+interfaces with a non-exported implementing struct. A run-time check can be woven into the generated bindings, to
+provide helpful error messages (this can be done regardless of how optionals are implemented, as the type model has the
+required information).
+
+While turning an required parameter to optional is not a usage-breaking change in this scenario, it is as an
```
eladb, line 1198:
```suggestion
While turning a required parameter to optional is not a usage-breaking change in this scenario, it is as an
```

text/204-golang-bindings.md
```diff
@@ -1068,12 +1068,160 @@ This allows construction like so:
...
+
+    fmt.printf("{}", has_string.OptionalString());  // => "I have a string"
+    fmt.printf("{}", not_so_much.OptionalString()); // => nil
+
+    fmt.printf("{}", example.MyClass_._SomeStatic(jsii.String("returns a string"))); // => "returns a string"
+    fmt.printf("{}", example.MyClass_._SomeStatic(nil));                             // => nil
+}
+```
+
+### Side Benefits
```
eladb, line 1175: Add code samples to demonstrate

eladb (dismissed):
What would be the migration story from the current generated bindings to the new style?

text/204-golang-bindings.md
```diff
@@ -1068,12 +1068,249 @@ This allows construction like so:
...
+}
+
+// No need to declare a separate variable up-front here.
+UncheckedCast[SpecificType](opaqueValue)
+```
+
+### Downsides
+
+In go, `nil` is a valid value at any location where an interface is expected. This means it is not possible to achieve
+compile-time safety with this approach in places where classes are passed as arguments, since those are represented as
```
MrArnoldPalmer, line 1253: is this not the current situation as well? Where a field in TS is type `SomeClass` (required) and in go is `SomeClass`, the interface is met by null, though the property is required?

text/204-golang-bindings.md
```diff
@@ -1159,8 +1394,6 @@ also true, an output value, IE a method's return value, cannot be changed from n
 With these downsides, it still seems like this is the most idiomatic route to take. Go users are familiar with using pointer types in this way from other
 popular modules. Additionally, it is the only strategy we can identify that allows us to maintain JSII's versioning compatibility strategy.
 
-### Explored Alternatives
-
 1. Don't use pointers.
```
MrArnoldPalmer, line 1397: I read this as now being a continuation of the "explored alternatives" list, therefore this should be 2 and "wrapper types" should be 3.

MrArnoldPalmer (commented):
@RomainMuller I wonder how you feel about creating a minimally functional implementation so we can allow users to compare these side by side and gather feedback? It seems like with some pretty minimal changes to codegen and `castAndSetToPtr` logic, we could get something working at least in the case of optional values, though maybe without a lot of the mentioned side benefits.

With the understanding that required/optional is still unable to be enforced at compile time, it doesn't feel like a clear improvement to the code needed to interact with JSII modules in Go. The level of indirection feels very similar, but more side by side examples might make a clearer case.

text/204-golang-bindings.md
RomainMuller: Correct. I just wanted to be very explicit about the guarantees the compiler will NOT provide.

RomainMuller:
This solution does not turn out to be viable... It prevents users from leveraging inheritance - i.e: it is not possible to pass an instance of `cdk.App` or `cdk.Stack` where `jsii.Option[constructs.Construct]` is expected, despite both `cdk.App` and `cdk.Stack` embed `constructs.Construct` - that is because they cannot implement `jsii.Option[constructs.Construct]`, as that would imply they carry two identically named methods with different return types.

Turns out, actually implementing this might require adding inordinate amounts of boilerplate to client code, which is obviously undesirable.

```go
package main

import (
	"github.com/RomainMuller/aws-cdk-go/awscdk/v2"
	"github.com/RomainMuller/aws-cdk-go/awscdk/v2/awskms"
	"github.com/RomainMuller/aws-cdk-go/awscdk/v2/awss3"
	"github.com/RomainMuller/jsii-runtime-go"
)

func main() {
	app := awscdk.NewApp(nil)
	// ❌ cannot use app (variable of type awscdk.App) as type jsii.Option[constructs.Construct] in argument to awscdk.NewStack:
	//   awscdk.App does not implement jsii.Option[constructs.Construct] (missing FromOption__ method)
	stack := awscdk.NewStack(app, jsii.String("ExampleStack"), nil)

	encryption_key := awskms.NewKey(stack, jsii.String("EncryptionKey"), nil)

	awss3.NewBucket(stack, jsii.String("Bucket"), awss3.BucketProps{
		Encryption: awss3.BucketEncryption_KMS_MANAGED,
		// ❌ cannot use encryption_key (variable of type awskms.Key) as type jsii.Option[awskms.IKey] in struct literal:
		//   awskms.Key does not implement jsii.Option[awskms.IKey] (missing FromOption__ method)
		EncryptionKey: encryption_key,
	})

	app.Synth(nil)
}
```

The `UnsafeCast` feature might still benefit from generics, as that provides better ergonomics from the fact it is not necessary to manually declare and type a variable for the cast result.

PR #398, opened 2022-01-06
1Pager: stateful resources

iliapolo:
This is a request for initial feedback about a possible RFC for modeling stateful resources in the CDK core framework. 

[Rendered Version](https://github.com/aws/aws-cdk-rfcs/blob/epolon/stateful-resources/text/stateful-resources-one-pager.md)

---


text/stateful-resources-one-pager.md
```diff
@@ -0,0 +1,258 @@
...
+
+By default, CloudFormation will **delete** resources that are removed from the stack,
+or when a property that requires replacement is changed.
+
+> - [UpdateReplacePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html)
+> - [DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html)
+
+To retain stateful resources, authors must remember to configure those policies
+with the `Retain` value. Having to remember this makes it also easy to forget.
+This means that stateful resources might be shipped with the incorrect policy.
```
rix0rrr, line 44: Most of this has been implemented in https://github.com/aws/aws-cdk/pull/12920

text/stateful-resources-one-pager.md
```diff
@@ -0,0 +1,258 @@
...
+to research how this would be done in CDK pipelines. My initial thoughts are to somehow incorporate
+`cdk diff` in the pipeline, along with a manual approval stage.
+
+### Would this have prevented [#16603](https://github.com/aws/aws-cdk/issues/16603)?
+
+Yes. If the API proposed here existed, we would have passed `stateful: true` in the
+implementation of the `autoDeleteObjects` property. Then, when the resource was
+removed due to the toggle, we would identify this as a removal of a stateful resource, and block it.
+
+### Curating a list of stateful resources is a maintenance burden - can we do it differently?
```
otaviomacedo, line 242: Modifying the file should be straightforward and infrequent. So I think it's only a burden to the extent that people can forget to update it when creating a new stateful resource. Is there any mechanism to remind developers to check this (like a lint rule)?

iliapolo:
@rix0rrr Has just informed my about [Stack Policies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html) so i'm going to do some research into this and see how it affects my proposal.

text/stateful-resources-one-pager.md
iliapolo: Taking a loog.

text/stateful-resources-one-pager.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+Using this file, we can generate the correct constructor, for example:
+
+```ts
+export class CfnBucket extends cdk.CfnResource implements cdk.IInspectable {
+
+  constructor(scope: cdk.Construct, id: string, props: CfnBucketProps = {}) {
+      super(scope, id, {
+        stateful: true,
```
rix0rrr, line 176: If we go the "stack policies" route, we need a way to turn off the protection, in a way that is user controllable (when they finally do want to get rid of a stateful resource).

Maybe it's a sort of `removalPolicy` though:

* `RemovalPolicy.DESTROY`
* `RemovalPolicy.RETAIN`
* `RemovalPolicy.PROTECTED`

When you want to get rid of it, you switch to `RemovalPolicy.DESTROY` first, then remove it.

rix0rrr (changes requested):

Issue #419, opened 2022-04-13
CDK environment setup for platform/system administrators


skinny85:
## Description

CDK is tricky to use in restricted environments where there is a split between the platform/system administrator teams, and the application development teams who use CDK.

This RFC proposes a set of features for dealing with that problem.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @skinny85
| Author(s)           | @skinny85
| API Bar Raiser      | @rix0rrr
| Stakeholders        | @cgarvis, @addihorowitz, @rix0rrr

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

skinny85:
@rix0rrr I've assigned you as the API bar raiser on this one, let me know if you disagree.

skinny85:
Pull Request with the RFC: https://github.com/aws/aws-cdk-rfcs/pull/421.

PR #421, opened 2022-04-13
RFC 419: CDK Launchpads

skinny85:
[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/cdk-launchpads/text/0419-cdk-launchpads.md)

This is a request for comments about "CDK environment setup for platform/system administrators" (a.k.a. CDK Launchpads).
See #419 for additional details. 

---

### DIFF
--- a/text/0419-cdk-launchpads.md
+++ b/text/0419-cdk-launchpads.md
@@ -1,0 +1,539 @@
+# CDK Launchpads
+
+## Working Backwards
+
+CDK Launchpads is a new functionality of the CDK project.
+It is geared primarily towards platform or system administrators in enterprise settings who are responsible for managing
+the AWS environments that are then used by application developers to deploy CDK applications into.
+
+Launchpads can be thought of as an extension of the
+[`cdk bootstrap` command](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html).
+The bootstrap command creates a set of resources -- S3 Bucket with a KMS Key, ECR Repository, several IAM Roles, etc. --
+that allow CDK applications to be deployed into a given AWS account and region.
+Launchpads are extensions of those bootstrap Stacks.
+They contain (by default) the same set of resources,
+but also add several features that are not available in the bootstrap command, such as:
+
+1. Ability to define compliance rules.
+  It is impossible to deploy a CDK application to the given environment if it breaks any compliance rule defined in the Launchpad.
+2. Easy customization of the bootstrap resources, like adding lifecycle policies to the ECR repository,
+  changing the names of the generated Roles, or any other customization needed.
+  With the existing bootstrap command,
+  the only way these customizations are supported is by providing your own CloudFormation template.
+3. Extending the predefined set of bootstrap resources with additional resources.
+  Common examples are VPC networks, IAM Roles or Permission Boundaries, or Security Groups.
+  These shared resources can then be easily referenced from CDK code by developers working in a given Launchpad
+  (see below for details).
+4. Launchpads integrate with AWS organizations,
+  allowing easy creation of new accounts that have the given Launchpad provisioned when the account is created,
+  alleviating the need for running any command, including `cdk bootstrap`, manually.
+5. Launchpads can be configured to use a given Managed Policy as a Permission Boundary for all CDK applications deployed into the given environment.
+
+Using Launchpads does not require any knowledge of the CDK --
+in fact, it does not require knowing a programming language at all!
+
+The functionality of CDK Launchpads can be accessed in three different ways:
+
+1. Through a dedicated CLI, `cdk-lpads`. This is a self-contained program
+  (does not have any dependencies -- for example, it does not depend on NodeJS)
+  that can be installed using the idiomatic package managers for a given platform
+  (`apt-get install cdk-lpads` on Ubuntu Linux, `yum install cdk-lpads` on CentOS Linux,
+  `brew install cdk-lpads` on MacOS, etc.).
+2. Through the existing CDK CLI, using the `cdk lpads` sub-command.
+3. Through the AWS Console, which now contains a new section dedicated to CDK,
+  which in turn contains options for managing your Launchpads.
+
+### Basic commands
+
+> **Note**: in this section, I show the commands only for the stand-alone CLI, `cdk-lpads ...`.
+> The same commands would work in the CDK CLI without the dash (`cdk lpads ...`),
+> but I'm only showing the standalone version for brevity.
+
+#### 1. `status`
+
+Shows the current status of any Launchpads defined in the given environment.
+
+This invocation will use the current AWS credentials in the console:
+
+```shell
+$ cdk-lpads status
+
+Environment - account: 123, region: us-east-1
+1 Launchpad found:
+1 | default CDK bootstrap Stack | version: 10   (UP-TO-DATE)
+```
+
+You can also check a specific account and region, using a profile defined in your AWS config files:
+
+```shell
+$ cdk-lpads status —profile acc-456 aws://456/eu-west-2
+
+Environment - account: 456, region: eu-west-2
+1 Launchpad found:
+1 | legacy (V1) CDK bootstrap Stack
+    Run `cdk-lpads update` to upgrade it to the newest version.
+```
+
+#### 2. `new`
+
+Creates a new Launchpad in the given account and region.
+By default, the command will include a wizard that will guide the user through the options:
+
+```shell
+$ cdk-lpads new
+
+Welcome! This wizard will guide you through creating a new Launchpad in the environment.
+To start, let's make sure we're operating in the correct AWS account and region:
+
+Environment - account: 123, region: eu-west-2
+
+Does this look correct? [Y/N] Y
+
+Looks like the environment aws://123/eu-west-2 doesn't have any existing Launchpads.
+
+Do you want to customize the qualifier used? [default: hnb659fds]:
+...
+```
+
+Alternatively, the `new` command can be customized directly, instead of going through the wizard;
+see "Customizing the Launchpad" below.
+
+#### 3. `update`
+
+This command attempts to update an existing Launchpad.
+In the case it's the default CDK bootstrap Launchpad, it will simply update it to the latest version.
+The `update` command can be customized directly, same as the `new` command;
+see "Customizing the Launchpad" below.
+
+#### 4. `diff`
+
+Allows you to see what differences would be applied if you updated your Launchpad.
+Useful to run before `update` to confirm the changes are what you expect.
+
+#### 5. `init json/yaml/cdk [ts/js/java/python/csharp/go]`
+
+Similar to the `cdk init` command,
+this command creates a skeleton that allows you get started with customizing your Launchpad.
+It supports both the JSON/YAML and CDK ways of customizing Launchpads (see below).
+
+### Ways of customizing the Launchpad
+
+The default Launchpad contains the same resources as the `cdk bootstrap` command (in version 2) creates.
+A CDK Launchpad can be customized with the following additional options:
+
+1. Compliance rules.
+2. Additional resources (deployed using CloudFormation, CDK, or in any other way, and simply referenced by the Launchpad).
+3. Modifications to the CDK bootstrap resources (using either CloudFormation templates or CDK).
+4. Setting IAM Permission Boundaries used by all applications deployed into a given environment.
+
+There are two ways of customizing a Launchpad: using JSON/YAML files, or using CDK code.
+
+#### Customizing the Launchpad with JSON/YAML files
+
+This option of customizing the Launchpad is geared primarily towards platform administrators
+who don't want to use a programming language (and thus, CDK code) to customize their Launchpads.
+Using a JSON or YAML file in a specific format allows customizing all aspects of the Launchpad,
+without the need to write a single line of code.
+
+The JSON file has the basic structure (the YAML file is very similar):
+
+```json
+{
+  "launchpad": {
+    ...
+  }
+}
+```
+
+We show examples of different parts of the JSON file in each of the sections below.
+
+You apply the changes using the JSON/YAML file by passing the file as an argument to the Launchpad commands,
+for example `cdk-lpads update my-launchpad.json/my-launchpad.yaml`.
+
+Deploying the Launchpads through a pipeline will also be supported with
+[CDK Pipelines](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines-readme.html).
+
+#### Customizing the Launchpad with CDK code
+
+It's also possible to customize the Launchpad using CDK code.
+In that scenario, the Launchpad is basically a CDK application, and you execute the Launchpads CLI commands,
+like `cdk-lpads update`, inside of it, similarly like you execute CDK CLI commands inside your CDK application.
+
+A CDK application for a Launchpad has a very similar structure to a "regular" CDK application,
+but the classes used are slightly different (the example is in TypeScript):
+
+```ts
+import * as lpads from '@aws-cdk/cdk-launchpads';
+
+const launchpad = new lpads.Launchpad();
+new lpads.StandardCdkBootstrap(launchpad, 'CdkBootstrap', {
+  // options to customize the CDK bootstrap can be passed here...
+});
+
+// additional things can be changed in the Launchpad here...
+
+launchpad.synth();
+```
+
+This CDK application will synthesize to the JSON configuration mentioned above,
+similarly like "regular" CDK applications synthesize to CloudFormation templates.
+
+We will show details of customizing the Launchpad using CDK code in each of the sections below.
+
+### Launchpad features
+
+#### Compliance
+
+Compliance rules allow administrators to restrict what kind of AWS resources can be deployed in a given environment.
+They can be used both for security purposes,
+but also for things like checking whether a given application conforms to the AWS Well-Architected principles.
+
+Launchpads allow defining compliance rules using the
+[CFN Guard 2.0 tool](https://github.com/aws-cloudformation/cloudformation-guard).
+Those compliance rules are then deployed as
+[CloudFormation Hooks](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/hooks.html),
+which means it's impossible to deploy a CDK application into a given environment that violates any compliance rule defined for it.
+
+Compliance rules can be defined in JSON/YAML directly:
+
+```json
+{
+  "launchpad": {
+    "compliance": {
+      "cfn-guard": {
+        "common": [
+          "let s3_buckets = Resources.[ Type == /S3::Bucket/ ]"
+        ],
+         "rules": {
+           "s3_bucket_name_encryption_check": {
+             "$when": "%s3_buckets !empty",
+           "%s3_buckets": {
+            "Properties": {
+              "BucketName": "== /^MyCompanyPrefix/",
+                "BucketEncryption.ServerSideEncryptionConfiguration[]": {
+                  "ServerSideEncryptionByDefault.SSEAlgorithm": "IN ['aws:KMS']"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Using CDK, you can either write them directly in the construct's body:
+
+```ts
+new lpads.CfnGuardComplianceRules(launchpad, 'Compliance', {
+  rules: lpads.CfnGuardRules.fromString(`
+    let s3_buckets = Resources.[ Type == /S3::Bucket/ ]
+
+    # Skip the checks if there are no S3 buckets present
+    rule s3_bucket_name_encryption_check when %s3_buckets !empty {
+      %s3_buckets {
+        Properties {
+          # common prefix
+          BucketName == /^MyCompanyPrefix/
+
+          # encryption MUST BE on
+          BucketEncryption.ServerSideEncryptionConfiguration[*] {
+            # only KMS
+            ServerSideEncryptionByDefault.SSEAlgorithm IN ["aws:KMS"]
+          }
+        }
+      }
+    }
+  `),
+});
+```
+
+Or use a special dialect of the CDK that allows creating CFN Guard Rules programmatically:
+
+```ts
+import * as cguard '@aws-cdk/launchpad-cfn-guard';
+
+const s3Buckets = new cguard.Variable(launchpad, 's3_buckets', cguard.Query.resources({
+  type: guard.Expr.matches(/S3::Bucket/),
+}));
+new lpads.CfnGuardComplianceRules(launchpad, 'Compliance', {
+  rules: lpads.CfnGuardRules.fromObjects(
+    new guard.Rule(launchpad, 's3_bucket_name_encryption_check', {
+      when: cguard.Expr.not(cguard.Expr.empty(s3Buckets)),
+      expr: guard.Query.properties({
+        'BucketName': cguard.Expr.matches(/^MyCompanyPrefix/),
+        'BucketEncryption.ServerSideEncryptionConfiguration[*]': cguard.Expr.in(
+          'ServerSideEncryptionByDefault.SSEAlgorith', ['aws:KMS']
+        ),
+      }),
+    }),
+  ),
+});
+```
+
+#### Customizing the base bootstrap resources
+
+The default set of resources that get created by the `cdk bootstrap` command can be customized using the Launchpad.
+
+It can be done both using JSON/YAML:
+
+```json
+{
+  "launchpad": {
+    "bootstrap": {
+      "patch": {
+        "description": "Turn off image scanning on push",
+        "operations": [
+          {
+            "op": "replace",
+            "path": "/Resources/ContainerAssetsRepository/Properties/ImageScanningConfiguration/ScanOnPush",
+            "value": false
+          }
+        ]
+      }
+    }
+  }
+}
+```
+
+And CDK code:
+
+```ts
+new lpads.StandardCdkBootstrap(launchpad, 'CdkBootstrap', {
+  imageRepository: lpdas.CdkBootstrapImageRepository.default({
+    scanOnPush: false, // default: true
+  }),
+});
+```
+
+The `cdk-lpads` command contains a validator that makes sure any changes you make to the bootstrap template are correct,
+and satisfy the minimal requirements needed by CDK in order to still deploy using those bootstrap resources.
+
+#### Adding new bootstrap resources
+
+You can also add new resources to the Launchpad that you want to be shared between all applications deployed into this environment.
+Application developers can then easily import these resources into their CDK applications.
+
+You can add the new resources using JSON/YAML files:
+
+```json
+{
+  "launchpad": {
+    "resources": {
+      "cfn": {
+        "Resources": {
+          "PermissionBoundaryPolicy": {
+            "Type": "AWS::IAM::ManagedPolicy",
+            "Properties": {
+              "PolicyDocument": {
+                "Statement": [
+                  {
+                    "Action": "iam",
+                    "Effect": "Deny",
+                    "Resource": "*"
+                  }
+                ],
+                "Version": "2012-10-17"
+              },
+              "Description": "The managed policy used as the permission boundary"
+            }
+          }
+        }
+      },
+      "export": {
+        "permissionBoundaryPolicy": {
+          "Ref": "PermissionBoundaryPolicy"
+        }
+      }
+    }
+  }
+}
+```
+
+And also with CDK, using the same CDK libraries that are used for applications:
+
+```ts
+import * as iam from 'aws-cdk-lib/aws-iam';
+
+const managedPolicy = new iam.ManagedPolicy(launchpad, 'Boundary', {
+  statements: [
+    new iam.PolicyStatement({
+      effect: iam.Effect.DENY,
+      actions: ['iam:'],
+      resources: [''],
+    }),
+  ],
+});
+// surface this resource to application developers using this Launchpad
+// (seee below)
+launchpad.export(managedPolicy);
+```
+
+#### IAM Permission Boundaries
+
+Launchpads allow setting a given Managed Policy as the Permissions Boundary to be used by every application deployed into a given environment.
+
+You can do it through JSON/YAML:
+
+```json
+{
+  "launchpad": {
+    "iam": {
+      "permissionBoundary": {
+        "Ref": "PermissionBoundaryPolicy"
+      }
+    }
+  }
+}
+```
+
+Or in CDK code:
+
+```ts
+launchpad.setPermissionBoundaryPolicy(managedPolicy);
+```
+
+### Importing the Launchpad in CDK code
+
+Once a Launchpad has been deployed by the administrators,
+application developers can import it using the
+[`cdk-import` tool](https://github.com/cdklabs/cdk-import)
+(whose functionality will be incorporated into the main CDK CLI in the future).
+
+Executing the following command:
+
+```shell
+$ cdk-import launchpad -l typescript
+```
+
+Will generate code in your project that you then use in your CDK app definition:
+
+```ts
+import * as cdk from 'aws-cdk-lib';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import Launchpad from '../build';
+
+const app = new cdk.App({
+    launchpad: Launchpad, // adds the appropriate Permission Boundary automatically
+});
+const stack = new cdk.Stack(app, 'MyStack');
+const fn = new lambda.Function(stack, 'MyLambda', {
+  code: new lambda.Code.fromAsset('lambda-code'),
+  handler: 'index.handler',
+  runtime: lambda.Runtime.PYTHON_3_9,
+  // refer to a resource defined in the Launchpad
+  role: Launchpad.Resources.functionRole,
+});
+```
+
+Additionally, importing the Launchpad in that way will produce a warning during `cdk deploy`
+if the version of the Launchpad in your environment is older than the current latest version of the Launchpad,
+and will ask you to update.
+
+### Integrating Launchpads with AWS Organizations
+
+Launchpads will integrate with AWS Organizations.
+When creating a new account using the AWS Console for Organizations,
+you'll be able to specify which Launchpad that account should use.
+This means the account will be provisioned ready for CDK deployments from the start -
+there will be no need for any manual actions, like running `cdk bootstrap`.
+
+## Public FAQ
+
+### What are we launching today?
+
+We are launching CDK Launchpads,
+a product that allows platform administrators in enterprise environments to easily configure AWS environments for deploying CDK applications.
+
+### Why should I use this feature?
+
+You should use this feature if you're a platform administrator who is responsible in your company/organization for the management of AWS environments,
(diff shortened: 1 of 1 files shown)


text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,539 @@
...
+They can be used both for security purposes,
+but also for things like checking whether a given application conforms to the AWS Well-Architected principles.
+
+Launchpads allow defining compliance rules using the
+[CFN Guard 2.0 tool](https://github.com/aws-cloudformation/cloudformation-guard).
+Those compliance rules are then deployed as
+[CloudFormation Hooks](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/hooks.html),
+which means it's impossible to deploy a CDK application into a given environment that violates any compliance rule defined for it.
+
+Compliance rules can be defined in JSON/YAML directly:
```
alexpulver, line 197: What about being able to provide a path to CFN Guard rules file? It will then allow to develop the rules separately, including capabilities such as [testing](https://docs.aws.amazon.com/cfn-guard/latest/ug/testing-rules.html).

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,539 @@
...
+there will be no need for any manual actions, like running `cdk bootstrap`.
+
+## Public FAQ
+
+### What are we launching today?
+
+We are launching CDK Launchpads,
+a product that allows platform administrators in enterprise environments to easily configure AWS environments for deploying CDK applications.
+
+### Why should I use this feature?
```
alexpulver, line 449: Can platform administrators use Launchpad for applications deployed using other provisioning engines? Why does the RFC mention CDK applications only? Does it mean the platform administrators will need to invest a separate effort for other provisioning engines?

RichiCoder1:
Apologies for possibly very early questions but:

* Is Launchpads something that's provisioned via CloudFormation as a first class resource? Both the cmd and the integration into AWS Organizations make it kind of vague how Launchpads are actually implemented (thought it looks like thats an open question mentioned in the RFC

* I would echo the concern here, but from the consumer side in there's a worry that Launchpad may lag behind the CDK project itself (bootstrap) or slow CDK's velocity

> Integrating with AWS Organizations, and having a presence in the AWS Console, will require owning server-side components. Maintaining them will put additional strain on the CDK team.

* For this part: https://github.com/aws/aws-cdk-rfcs/blob/cdk-launchpads/text/0419-cdk-launchpads.md#importing-the-launchpad-in-cdk-code. Out of curiosity, why is getting the environmental launchpad an import and not a lookup?


Wanted to say I'm extremely bullish about this! Centrally bootstrapping as well as allowing multiple team/persona specific "launchpads" would/will be amazing for easing team dev and setting up gaurdrails

text/0419-cdk-launchpads.md
skinny85: Yes, that's a great point! Will add.

text/0419-cdk-launchpads.md
skinny85: The part that is CDK-specific is the bootstrap resources that are used are CDK-specific.

Other provisioning engines could be supported by customizing the Launchpad shared resources, I guess. I'll try to add it to the doc.

skinny85:
> Apologies for possibly very early questions but:

No worries - this is exactly the time to ask those questions!

> * Is Launchpads something that's provisioned via CloudFormation as a first class resource? Both the cmd and the integration into AWS Organizations make it kind of vague how Launchpads are actually implemented (thought it looks like thats an open question mentioned in the RFC)

No. A Launchpad would be represented as a CloudFormation Stack (and possibly some extra data, like the Compliance rules, stored in something like S3, most likely using simple JSON/YAML files).

> * I would echo the concern here, but from the consumer side in there's a worry that Launchpad may lag behind the CDK project itself (bootstrap) or slow CDK's velocity

I wouldn't be concerned about lagging behind bootstrap (we would share the code, so that it wouldn't be duplicated), but the second concern is definitely valid.

> > Integrating with AWS Organizations, and having a presence in the AWS Console, will require owning server-side components. Maintaining them will put additional strain on the CDK team.
> 
> * For this part: https://github.com/aws/aws-cdk-rfcs/blob/cdk-launchpads/text/0419-cdk-launchpads.md#importing-the-launchpad-in-cdk-code. Out of curiosity, why is getting the environmental launchpad an import and not a lookup?

Because we want to generate code based on the Launchpad, not force the CDK application users of the Launchpad to import things directly (you could do that of course, it would just be a worse experience).

> Wanted to say I'm extremely bullish about this! Centrally bootstrapping as well as allowing multiple team/persona specific "launchpads" would/will be amazing for easing team dev and setting up gaurdrails

Glad you like it, that's what we're hoping for 🙂.

RichiCoder1:
Thank you for the great answers!

> Because we want to generate code based on the Launchpad, not force the CDK application users of the Launchpad to import things directly (you could do that of course, it would just be a worse experience).

For this, is the intent that each application would generate it's own binding code? Or is something that could be generated and shared via, say, an NPM library? I just worry about adding extra steps or something that can easily get out of sync for consuming devs.

text/0419-cdk-launchpads.md
alexpulver: @skinny85 thanks! Thinking about it, what is CDK-specific about bootstrap resources? The way I think about is that these resources support a certain design approach, and CDK makes use of them to implement it. Other tools might use these resources similarly, or extend them through customization (as you mentioned).

skinny85:
You definitely could generate it once, and then publish as an NPM library. The complicated part is that you most likely want to make it company-private, as it will be very specific to the way your environment operates. It won't be useful to the entire open-source community. That complicates things a little bit, but there are ways of handling that (like using npmjs.com's private registries, or something like CodeArtifact).

text/0419-cdk-launchpads.md
skinny85: I mean, it's really hard to judge in advance whether other tools could use these resources in the same way as CDK does, without doing some sort of dive-deep into a specific tool 🙂.

Taking a step back, the bootstrap resources CDK needs are:
1. An S3 Bucket (along with a BucketPolicy) and ECR repository (for Asset support).
2. Several IAM Roles (for Asset publishing, for CDK context lookups, and also for CloudFormation deployments, including cross-account deployments).
3. Optional KMS Keys for encrypting the Bucket and the repo, and Aliases to make these Keys easier to manage in the Console.
4. An SSM parameters that stores the current version of the bootstrap Stack.
5. All of the above resources (besides the KMS Keys) need to have a set physical name according to a specific convention, otherwise CDK will not be able to use them.
6. We need a few CloudFormation Outputs (again, with specific physical names), used the CDK CLI.

That's a pretty specific list! Of course, I'm not saying none of these can be used by other tools - but looking at how specific the CDK requirements are, it's very likely other tools will have similarly detailed requirements of their own, and re-using the same resources for both will be tricky.

text/0419-cdk-launchpads.md
skinny85: Done in the latest revision.

text/0419-cdk-launchpads.md
```diff
@@ -194,65 +194,88 @@ Those compliance rules are then deployed as
...
+}
+```
+
+Or you can reference a file on disk that contains the Guard rules:
+
+```json
+{
+  "launchpad": {
+    "compliance": {
+      "cfn-guard": ["file:my-guard-rules.guard"]
```
alexpulver, line 239: Perhaps worth mentioning or adding an example of validating input data against multiple data files? CDK creates cloud assembly with potentially multiple templates. Perhaps the default behavior should use cloud assembly directory for CFN Guard's `--data`. `--rules` can be a file or a directory.

RichiCoder1:
> You definitely could generate it once, and then publish as an NPM library. The complicated part is that you most likely want to make it company-private, as it will be very specific to the way your environment operates. It won't be useful to the entire open-source community. That complicates things a little bit, but there are ways of handling that (like using npmjs.com's private registries, or something like CodeArtifact).

Ah, to be clear we do this today to share opinionated constructs and company specific helpers today via a private NPM registry. So not to worried about that, just that it'd be feasible.

text/0419-cdk-launchpads.md
skinny85: I feel like that's a little obvious. Yes, the Guard rules will be applied automatically to each template CDK generates 🙂.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+To achieve this goal, we will allow administrators to define a StackSet, using CDK constructs
+(similarly to the [`ProductStack` class](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-servicecatalog.ProductStack.html)
+in the ServiceCatalog Construct Library).
+We will also create construct(s) that represent the CDK bootstrap resources,
+and allow customizing them.
+
+In order to automatically bootstrap new accounts that get added to a given Organization,
+we will use the [CDK Pipelines library](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines-readme.html).
+The StackSet will live in a separate Stack, deployed to the master account.
+Any time a new account is added, a CloudWatch Event will trigger the pipeline,
```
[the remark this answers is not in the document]
skinny85: Unfortunately, the problem is that we don't know what region we need to deploy the StackSet instance to, using only this integration 😕. StackSet are not really very flexible in that regard.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+```
+
+Additionally, importing the Launchpad in that way will produce a warning during `cdk deploy`
+if the version of the Launchpad in your environment is older than the current latest version of the Launchpad,
+and will ask you to update.
+
+### Integrating Launchpads with AWS Organizations
+
+Launchpads will integrate with AWS Organizations.
+When creating a new account using the AWS Console for Organizations,
```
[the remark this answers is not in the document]
skinny85: Unfortunately, the problem is that `cdk bootstrap` is a regional operation, while the StackSet OU integration is only in a single region.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+
+Launchpads can be thought of as an extension of the
+[`cdk bootstrap` command](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html).
+The bootstrap command creates a set of resources -- S3 Bucket with a KMS Key, ECR Repository, several IAM Roles, etc. --
+that allow CDK applications to be deployed into a given AWS account and region.
+Launchpads are extensions of those bootstrap Stacks.
+They contain (by default) the same set of resources,
+but also add several features that are not available in the bootstrap command, such as:
+
+1. Ability to define compliance rules.
```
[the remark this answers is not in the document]
skinny85: In theory, we could add more rules type there in the future - I didn't specify it, but it's definitely possible (but most likely at a later stage in the project).

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+}
+```
+
+Or in CDK code:
+
+```ts
+launchpad.setPermissionBoundaryPolicy(managedPolicy);
+```
+
+### Importing the Launchpad in CDK code
```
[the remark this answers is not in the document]
skinny85: I think the problem is that we need additional information to correctly do this importing. That's why, in the RFC, we have the `"export"` element in JSON, or the `export()` element in the CDK code of the Launchpad. Without this extra information, I don't think we'll have enough data to correctly generate CDK code from these resources.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+  allowing easy creation of new accounts that have the given Launchpad provisioned when the account is created,
+  alleviating the need for running any command, including `cdk bootstrap`, manually.
+5. Launchpads can be configured to use a given Managed Policy as a Permission Boundary for all CDK applications deployed into the given environment.
+
+Using Launchpads does not require any knowledge of the CDK --
+in fact, it does not require knowing a programming language at all!
+
+The functionality of CDK Launchpads can be accessed in three different ways:
+
+1. Through a dedicated CLI, `cdk-lpads`. This is a self-contained program
```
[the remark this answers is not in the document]
skinny85: > Is the only reason for the CLI that it can be installed without NodeJS? Maybe we should just rewrite the CDK CLI to not depend on NodeJS.

The idea is also to simplify (the CDK CLI has a _ton_ of stuff in it that's unrelated to Launchpads - removing that complexity seems valuable).

> Also, I'm wondering if this functionality needs to exist separate from `cdk bootstrap` at all?

It's possible we should make this functionality part of the `cdk bootstrap` command - I think that's a valid idea. The question is: is it a more ergonomic design, than a separate command? Historically, we were very hesitant to add more options to the `bootstrap` command, and that's why I went with a separate command in this proposal. But we can certainly discuss this - we can always change our mind on making more changes to `cdk bootstrap`.

> It almost feels like launchpads should exist as a group of construct libraries (or CFN public resources) where we don't necessarily provide an opinionated way to deploy them. What if you are an organization that already has a process for provisioning new accounts and deploying a baseline set of resources (maybe using something like [this](https://aws.amazon.com/solutions/implementations/customizations-for-aws-control-tower/))? You would want a way to provide a customizable CFN template that deploys CDK specific resources.

I think public CFN resources are out of the picture for now (too much development effort before we validate this idea), and CDK resources for this are planned (the RFC contains some examples), but we also need the ability to define Launchpads without the requirement of writing CDK code, to cater to a segment of customers who don't want to use a programming language.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,629 @@
...
+A CDK Launchpad can be customized with the following additional options:
+
+1. Compliance rules.
+2. Additional resources (deployed using CloudFormation, CDK, or in any other way, and simply referenced by the Launchpad).
+3. Modifications to the CDK bootstrap resources (using either CloudFormation templates or CDK).
+4. Setting IAM Permission Boundaries used by all applications deployed into a given environment.
+
+There are two ways of customizing a Launchpad: using JSON/YAML files, or using CDK code.
+
+#### Customizing the Launchpad with JSON/YAML files
```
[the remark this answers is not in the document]
skinny85: Yes, a section of the JSON/YAML file will be translated to a CloudFormation template. We need our own JSON/YAML structure for all of those things that are not part of CloudFormation, like Compliance rules, and shared exported resources (I mentioned them in https://github.com/aws/aws-cdk-rfcs/pull/421#discussion_r861169550).

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
RichiCoder1: I personally like the idea of a simple command like bootstrap (or literally an extension of bootstrap) as the low-friction onboard, with a set of opinionated CDK constructs to allow for customization or a more comfortable IaC approach.

If the behavior of bootstrap is significantly different than launchpads, or would add a significant amount of overhead to bootstrap, I would endorse a seperate cmd. But it's largely a super set of bootstrap (which is sounds like it is), I'd love to have it be an opt in behavior of bootstrap.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
RichiCoder1: To echo this concern, I'd say I'd still maybe separating those out as otherwise you end up with a bizzare `JSON/YAML` that ends up having 2-3 sub-DSLs because of all the knob it turns. On it's face it makes it simple (on file) but I think you're aggregating complexity.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
RichiCoder1: Is there physical information/known properties that launchpad information can be encoded too and looked up from like how bootstrap works today? We do this internally and using when using [CDK's ECS Service Extensions](https://www.npmjs.com/package/@aws-cdk-containers/ecs-service-extensions/v/2.0.0-alpha.17) (which has the notion of a shared "Environment").

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
corymhall: When you create the stackset you specify the OU _and_ the regions you want to deploy to right?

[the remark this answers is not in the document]
skinny85: Hmm, the page you linked to (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacksets-orgs-manage-auto-deployment.html) doesn't mention choosing a region when enabling the automated deployments?

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
skinny85: That's true. Maybe having multiple separate files, and gluing them with CLI options, is the better design here.

text/0419-cdk-launchpads.md
alexpulver: Even though it makes a total sense, it wasn't obvious to me from the name of the configuration key :). Perhaps rename it to `cfn-guard-rules`, and give an example for accepting either a file or a directory, like CFN Guard does?

```suggestion
      "cfn-guard-rules": ["file:my-guard-rules.guard"] // or "dir:my-guard-rules"
```

text/0419-cdk-launchpads.md
alexpulver: It makes sense and seems like a 2-way door :)

text/0419-cdk-launchpads.md
skinny85: That's fair! Changed the example (I've kept the `"cfn-guard"` name of the key though - I feel like the "rules" part belongs more on the actual files than on this key, let me know if you strongly disagree 🙂).

alexpulver: Thanks! I do feel quite strongly about the key name, because the file or directory names are provided by the user, and can't be enforced :). Neither tye extension is mandatory from what I remember in my last use of CFN Guard.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+}
+```
+
+Or you can reference a file on disk that contains the Guard rules:
+
+```json
+{
+  "launchpad": {
+    "compliance": {
+      "cfn-guard": [
```
alexpulver, line 239: @skinny85 are you sure that CFN Guard accepts multiple values? Or are you planning to call CFN Guard multiple times and merge the results if so?

alexpulver: \+ referencing discussion we started which became outdated after the last commit:
https://github.com/aws/aws-cdk-rfcs/pull/421#discussion_r861462006

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+  statements: [
+    new iam.PolicyStatement({
+      effect: iam.Effect.DENY,
+      actions: ['iam:'],
+      resources: [''],
+    }),
+  ],
+});
+// surface this resource to application developers using this Launchpad
+// (seee below)
```
eladb, line 395:
```suggestion
// (see below)
```

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+```shell
+$ cdk-import launchpad -l typescript
+```
+
+Will generate code in your project that you then use in your CDK app definition:
+
+```ts
+import * as cdk from 'aws-cdk-lib';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import Launchpad from '../build';
```
eladb, line 441: Does this export a type of an object? Seems like it's an object, so use lowercase.

```suggestion
import launchpad from '../build';
```

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+#### Customizing the Launchpad with JSON/YAML files
+
+This option of customizing the Launchpad is geared primarily towards platform administrators
+who don't want to use a programming language (and thus, CDK code) to customize their Launchpads.
+Using a JSON or YAML file in a specific format allows customizing all aspects of the Launchpad,
+without the need to write a single line of code.
+
+The JSON file has the basic structure (the YAML file is very similar):
+
+```json
```
eladb, line 140: Is there a spec somewhere for this file?

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+    }
+  }
+}
+```
+
+Using CDK, you can either write them directly in the construct's body,
+or again reference a file on disk:
+
+```ts
+new lpads.CfnGuardComplianceRules(launchpad, 'Compliance', {
```
eladb, line 252: The "Cfn" prefix implies this is an L1... Is that true? Feel this should be `@aws-cdk/cfn-guard.ComplianceRules`.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+    `),
+    lpads.CfnGuardRules.fromFile('my-guard-rules.guard'),
+  ],
+});
+```
+
+Or use a special dialect of the CDK that allows creating CFN Guard Rules programmatically:
+
+```ts
+import * as cguard from '@aws-cdk/launchpad-cfn-guard';
```
eladb, line 281: Curious why is this tightly coupled to launchpads? Is there a use case for using the guard constructs in other contexts, and then this would be `@aws-cdk/cfn-guard`.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+
+Will generate code in your project that you then use in your CDK app definition:
+
+```ts
+import * as cdk from 'aws-cdk-lib';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import Launchpad from '../build';
+
+const app = new cdk.App({
+    launchpad: Launchpad, // adds the appropriate Permission Boundary automatically
```
eladb, line 444: What happens if users forget to do this? Does the permission boundary also apply? Feels like this shouldn't be required.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+});
+const stack = new cdk.Stack(app, 'MyStack');
+const fn = new lambda.Function(stack, 'MyLambda', {
+  code: new lambda.Code.fromAsset('lambda-code'),
+  handler: 'index.handler',
+  runtime: lambda.Runtime.PYTHON_3_9,
+  // refer to a resource defined in the Launchpad
+  role: Launchpad.Resources.functionRole,
```
eladb, line 452: Why is `Resources` pascal case?

eladb (commented):
Looks pretty good. I think it will be helpful to formally define the launchpad spec, as it will give readers an overview of the capabilities.

text/0419-cdk-launchpads.md
skinny85: I feel like that's fine - then name of the file should already include the fact that it's a Guard rules file.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
skinny85: I think the problem is that you can't really cache these lookups, because the Launchpads can change with time. And so, you would need to re-do the lookup then, and we currently assume lookups can be cached forever (there's no easy way to invalidate a lookup, other than fiddling with the context yourself).

Code generation make this "invalidation" step easy (just re-run the generation part!).

text/0419-cdk-launchpads.md
skinny85: Yes, I think we can easily run the command multiple times, and merge the results.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
RichiCoder1: That's fair (and one of my issues with lookups in general). I guess my wish is just that it be more integrated than another step in the pipeline (and hopefully a cheap one), but maybe that can be a post-MVP thing?

text/0419-cdk-launchpads.md
alexpulver: The "should" part is not enforceable by any means, but I will stop nudging you on this :)

alexpulver: Sounds like a plan! 😄

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+```shell
+$ cdk-lpads status —profile acc-456 aws://456/eu-west-2
+
+Environment - account: 456, region: eu-west-2
+1 Launchpad found:
+1 | legacy (V1) CDK bootstrap Stack
+    Run `cdk-lpads update` to upgrade it to the newest version.
+```
+
+#### 2. `new`
```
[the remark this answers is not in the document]
skinny85: Changed.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
skinny85: I feel like that would be a ton of work on the CDK team's side to own that resource, so I'd rather avoid it, at least in the first iteration (we can transition to that model once we prove the concept out more).

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,632 @@
...
+in fact, it does not require knowing a programming language at all!
+
+The functionality of CDK Launchpads can be accessed in three different ways:
+
+1. Through a dedicated CLI, `cdk-lpads`. This is a self-contained program
+  (does not have any dependencies -- for example, it does not depend on NodeJS)
+  that can be installed using the idiomatic package managers for a given platform
+  (`apt-get install cdk-lpads` on Ubuntu Linux, `yum install cdk-lpads` on CentOS Linux,
+  `brew install cdk-lpads` on MacOS, etc.).
+2. Through the existing CDK CLI, using the `cdk lpads` sub-command.
```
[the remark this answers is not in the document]
skinny85: Ok, fair enough. I was on the fence about it myself.

Changed `lpads` to `launchpad` everywhere.

text/0419-cdk-launchpads.md
skinny85: Good eye! Fixed.

text/0419-cdk-launchpads.md
skinny85: I thought lowercase was traditionally reserved for modules, but I don't have a strong opinion on this. Changed.

text/0419-cdk-launchpads.md
skinny85: There is not. The spec is kind of gradually introduced throughout the RFC. I don't want to focus too much on the exact details of this spec, because I don't think it's that important - we can iron out those details when doing the implementation (kind of what I was alluding to during the discussion with @alexpulver in https://github.com/aws/aws-cdk-rfcs/pull/421#discussion_r862308094).

What I'm more interested in is an opinion on whether we should create a new format at all, or rather have separate files, and piggy back 100% on existing formats (what @RichiCoder1 is saying here: https://github.com/aws/aws-cdk-rfcs/pull/421#discussion_r861233400).

text/0419-cdk-launchpads.md
skinny85: That's a great point. It's very likely these would be useful outside just Launchpads.

Changed to the more generic `@aws-cdk/cfn-guard`.

text/0419-cdk-launchpads.md
skinny85: I wanted this to be available from a Launchpad without the requirement of using the `@aws-cdk/cfn-guard` DSL (if you just wanted to import your existing CFN Guard rules that you already had).

text/0419-cdk-launchpads.md
skinny85: If a user doesn't set this, then so, the Permission Boundary wouldn't be applied in that case.

How do you see this working without this being provided? How would the CDK app know it needed to apply a Permission Boundary, and what Policy should it use for that?

text/0419-cdk-launchpads.md
skinny85: No great reason. Just wanted to underscore that `Resources` is a collection. But I don't have a strong preference either way, so if you feel like it should be `launchpad.resources.functonRole` instead, I don't have a problem with that.

skinny85:
@eladb thanks a lot for the review! I've included most of your comments, let me know if this makes sense!

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
corymhall: I think it's because it's a feature of _stack sets_ and not specifically automated deployments. You create a stack set (and specify OU/regions) and _then_ you enable/disable automated deployments.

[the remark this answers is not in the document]
skinny85: I guess that could work, the major downside being that customers would have to decide up front which regions they will bootstrap in (and this decision has to be made both for developer accounts, and for shared environment accounts - they have to be in the same regions).

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,647 @@
...
+This CDK application will synthesize to the JSON configuration mentioned above,
+similarly like "regular" CDK applications synthesize to CloudFormation templates.
+
+We will show details of customizing the Launchpad using CDK code in each of the sections below.
+
+### Launchpad features
+
+#### Compliance
+
+Compliance rules allow administrators to restrict what kind of AWS resources can be deployed in a given environment.
```
blimmer, line 187: Are compliance rules similar to AWS Control Tower [Guardrails](https://docs.aws.amazon.com/controltower/latest/userguide/guardrails.html)?

Guardrails use AWS Config and these seem to use CfnGuard. However, as a Control Tower user, it would be really nice to have a single view into how accounts are governed and whether or not they're in compliance.

text/0419-cdk-launchpads.md
```diff
@@ -0,0 +1,647 @@
...
+```
+
+Additionally, importing the Launchpad in that way will produce a warning during `cdk deploy`
+if the version of the Launchpad in your environment is older than the current latest version of the Launchpad,
+and will ask you to update.
+
+### Integrating Launchpads with AWS Organizations
+
+Launchpads will integrate with AWS Organizations.
+When creating a new account using the AWS Console for Organizations,
```
blimmer, line 463: Similar to my other comment, would it be possible to use this with AWS Control Tower? Most of my clients now create new accounts via the AWS Control Tower "Landing Zone". "Landing Zone" and "Launchpad" even feels like a similar term 😄 

Since I'm assuming Control Tower is managed by another team(s), it's possible that it isn't feasible to integrate this feature with Control Tower. However, as an end-user, would be really great to integrate with an existing tool for simplicity.

text/0419-cdk-launchpads.md
skinny85: The main reason for using CFN Guard over Config is that, at least as far as I know, Config rules cannot fail deployments - they are post-deployment validation. While what we want here is to prevent deployments of things that violate rules completely, which CloudFormation Hooks give us.

text/0419-cdk-launchpads.md
skinny85: Yes, that's a good point. I'm not sure what exactly will that entail though, but this is an option we should explore internally.

text/0419-cdk-launchpads.md
eladb: Can we support both options? Feels like there are valid uses for both technologies and I am wondering if the launchpad spec should be open to supporting both.

skinny85: Sure. I think it's viable to support multiple compliance tools.

I would start with CFN Gard + CloudFormation Hooks though, as the first supported one.

text/0419-cdk-launchpads.md
rix0rrr: There is an Aspect that applies a PB to all roles in a Stack. We can apply it transparently.

skinny85: > There is an Aspect that applies a PB to all roles in a Stack. We can apply it transparently.

But we need some input from the customer for this to be applied, right? Which I think was the original question.

text/0419-cdk-launchpads.md
blimmer: This AWS Solution might be interesting to consider as part of this: https://aws.amazon.com/blogs/architecture/fast-and-secure-account-governance-with-customizations-for-aws-control-tower/

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
rix0rrr: Stack Sets deploy to the same region(s) in all accounts. Not typically something we recommend for users, we recommend they use a different account for every region, so that makes it hard(er).

image

I'm in favor of using stacksets, but we'll need same additional work to allocate regions and accounts to stacksets.

text/0419-cdk-launchpads.md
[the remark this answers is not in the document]
rix0rrr: Here's a thought:

* If bootstrap resources were all in a construct...
* ...and therefore bootstrapping is equivalent to deploying a CDK app...
* ...then `cdk bootstrap` is the same as `cdk deploy`

And we don't need a special command at all anymore.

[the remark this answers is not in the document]
skinny85: Doesn't that lead to a "chicken and egg" problem though?

`cdk deploy` requires bootstrapping, but you can't bootstrap, because you can't `cdk deploy`?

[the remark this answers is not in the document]
hoegertn: Are you sure that `cdk deploy` always needs bootstrapping? I thought without assets it works un-bootstrapped.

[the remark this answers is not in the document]
skinny85: Pretty sure that in V2, bootstrapping is required, even without Assets (there is a Rule that will make it fail if the version parameter stored in the bootstrap Stack is either not found, or not greater or equal than the expected version).

henrysachs:
Hey There, just as a extra perspective I would like to emphasize on the custom bootstrapping points. We as a team kinda built this ourselves with tranforming the bootstrapping template to cdk and having a permission boundary aspect. We would love to have this natively in cdk!

text/0419-cdk-launchpads.md
bitshop: Lots of great ideas in this RFC, great reading -- As a customer I'd prefer to see Control Tower have an option to deploy the CDK bootstrap (no reason first person using CDK in the account should need to do that) as well as the LaunchPad config for the account.  It should happen both when onboarding existing accounts or when creating new accts (with Account Factory).  Pushing centrally managed rules out would be a very nice feature to have.

text/0419-cdk-launchpads.md
bitshop: Config supports preventative controls to block - https://aws.amazon.com/blogs/aws/new-aws-config-rules-now-support-proactive-compliance/

text/0419-cdk-launchpads.md
bitshop: Having these files in s3 or codecommit would be nice for when this is centrally managed w/ control tower (hopefully).  codecommit://my-central-repo/permission-name/ could be a directory for example, obviously roles for access complicate this but central storage pointing to the base set of OU wide rules would be nice.  Perhaps how I'm thinking quickly evolves into a custom rule in config that prevents CF updates that don't meet the cfn guard rules? As I'm typing I think this is going beyond the scope of your goal (?).

Issue #39, opened 2019-12-10
Release public artifacts (lambda layers for custom resources, docker images)


eladb:
|PR|Champion|
|--|--------|
|# |        |

## Description

When the AWS construct library includes an asset such as custom resource providers, lambda bundles, docker images or static files, we currently bundle these into the library and then build and publish it to user accounts. Since the CDK is public and open-source, it should be possible to publish these assets to S3/ECR in all regions and consume them as public assets. This will reduce the size of our libraries and the time it takes for these constructs to be deployed.

## Progress

- [x] Tracking Issue Created
- [x] RFC PR Created 
- [x] Core Team Member Assigned 
- [x] Initial Approval / Final Comment Period
- [x] Ready For Implementation
    
    - [ ] implementation issue 1
- [x] Resolved

eladb:
Let's focus on S3 for Lambda bundles because that is a pertinent use case we have to reduce the side of the mono-module. Specifically, the aws-s3-deployment module currently needs to bundle the entire AWS CLI because it leverages `s3 sync`. Another use case we already have is the aws-eks which currently uses an open source AWS Lambda Layer published as a SAR app which contains `kubectl`, `helm` and the AWS CLI.

The only value of using SAR here would be that they take care of replication but that's not a problem with S3 since S3 now supports [cross-region replication](https://docs.aws.amazon.com/AmazonS3/latest/dev/replication-how-setup.html).

So the design approach would be to create a public S3 bucket in each region we support with a well-known name and cross-region replication. We will store .zip files for AWS Lambda layers there which will be mounted into custom resources of AWS constructs such as s3-deployment and EKS.

To achieve this, we will add another type of artifact to our build system (e.g. add `dist/s3` or something like that) and during publishing we will upload all the files from this directory to the replicated S3 bucket (with a key prefix that denotes the CDK version (e.g. `s3://aws-cdk-files-eu-west-2/v1.22.0/awscli-layer.zip`). We can also decide to just upload it to all the regional buckets. Shouldn't be too problematic and might be better because it will give us a synchronous confirmation that we are done.

Note that this is susceptible to an eventual consistency issue with the publishing of the actual modules that reference the artifact, so perhaps it has to be the first publisher (i.e. a separate pipeline stage or run-order).

The publisher will no-op if there is already a file in that location, which is aligned with how we publish (publishing only really happens when we actually bump the version and published artifacts are immutable).

pahud:
[...]

PR #430, opened 2022-04-26
RFC 39: Reduce aws-cdk-lib Package Size

madeline-k:
[Rendered Version](https://github.com/aws/aws-cdk-rfcs/blob/madeline-k/reduce-module-size/text/0039-reduce-module-size.md)

This is a request for comments about Reduce aws-cdk-lib Package Size. See #39 for
additional details. 

APIs are signed off by @otaviomacedo. 

---


text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,518 @@
...
+
+* **Original Author(s):**: @madeline-k
+* **Tracking Issue**: #39
+* **API Bar Raiser**: tbd
+
+The AWS CDK v2 framework library, `aws-cdk-lib` contains all packages from AWS
+CDK v1 in one monolithic package. While this is a great value for customers in
+making dependency management much easier, it has also resulted in increasing the
+unpacked npm package size to 238 MB, at the time of writing. This feature
+reduces the size of the package to 100 MB.
```
kaizencc, line 11: Do we have a section that breaks down, with all the proposed changes, what we will get the size down to?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,518 @@
...
+These files currently makeup 22% of the total size of `aws-cdk-lib`. Source map
+files map from the transformed source to the original source. In the case of
+`aws-cdk-lib`, these are not actually useful today, since we do not ship the
+original .ts files. We will remove the source maps from the released package.
+
+#### Lambda Layer Zip Files
+
+There are three zip files included in the `aws-cdk-lib` package. They are listed in the table below.
+
+| Filename                                    | Includes                                                                             | Usage                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | Size (MB) | Percentage of aws-cdk-lib |
```
kaizencc, line 207: minor nit: these calculations (along with the .jsii files one) suggests that the full module size is calculated as ~241 MB, not ~238 MB. I think we should replace the places that say 238 with 241.

text/0039-reduce-module-size.md
madeline-k: Not precisely. It is mentioned in the execution plan section at the end. I don't want to make this super concrete, because we haven't done enough prototyping to be super confident in these estimates. So far, we can be reasonably confident that the proposed solutions can reduce the size by more than 100 MB. But it's hard to make a reliable estimate of future size, especially since we will keep adding more code and features to the library before the solutions proposed here are implemented.

text/0039-reduce-module-size.md
madeline-k: The current total size according to npm is 238 MB, so I'd like to leave the total at that. So, there is some other discrepancy in my numbers, I'm not sure if it's worth it to fix that. 

These calculations are approximate. The goal of having numbers here at all is to give the reader an idea for which solutions are highest priority, and whether they will actually succeed in reducing the package size significantly. The precision of the numbers is not as important. Maybe we should include a disclaimer here that all of the numbers are a little bit rough.

kaizencc: For me, first thing I did when I read 

> Note: The percentages do not add up to 100. Some files, e.g. various .json
> files, are excluded and contribute very little to the size.

was add up all the numbers to see exactly what percentage of the total size was excluded. Maybe that's just me, but I figure I'd add in a "total" section in that table in case other people have the same thought as me.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,518 @@
...
+
+
+
+### Why should we _not_ do this?
+
+The main reason to not do this is that it will inevitably break up the
+monolothic `aws-cdk-lib` package in some manner, and we will lose some of the
+wonderful simplicity of releasing and using a single package. But, we can't keep
+everything in one place, and reduce the size enough to have a significant
+positive impact.
```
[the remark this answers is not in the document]
madeline-k: Yeah, these are good points. The current biggest contributor of size is not the Lambda Layers, they are just the biggest potential contributor. I called this out in the execution plan, but would be good to add it here. Here we should also call out the approximate size of monolithic aws-cdk-lib with all of the extra Lambda Layers. With .jsii compression and removing source maps it would still be nearly 200 MB.

text/0039-reduce-module-size.md
madeline-k: Another reason for discrepancies in numbers is I calculated most of them a few releases ago, and the package size changes with every release.

text/0039-reduce-module-size.md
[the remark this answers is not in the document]
kaizencc: It feels like the argument of this RFC hinges on this future size that includes additional lambda layers. I think it's a bit confusing to have both current and future estimates here, and instead we should benchmark _only_ on the future estimate. Then all the sections will have the same reference point and we don't have to switch context between comapring the impact of .jsii file compression to the current size and lambda layers to the future size.

I'm open to the response "too much effort / not needed" but that's just my $0.02.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,517 @@
...
+to include these dependencies, and they will not be required to access a public
+endpoint besides npm during runtime of their CDK apps. Practically, this
+solution breaks down into the following steps.
+
+1. Publish v2 compatible versions of `@aws-cdk/lambda-layer-aws-cli`,
+   `@aws-cdk/lambda-layer-aws-cli`, `@aws-cdk/lambda-layer-node-proxy-agent` as
+   their own npm packages, separate from `aws-cdk-lib`.
+2. Modify the `lambda-layer-X` submodules in `aws-cdk-lib` to not bundle these large
+   dependencies themselves, but instead reference the appropriate `Layer` class
+   from the packages in step 1. We will need to do something clever here to make
```
kaizencc, line 258: Do we have a high level idea for this "something clever"? If so, I think it's worth adding in an appendix section so as not to clutter this section but also remove the ambiguity.

text/0039-reduce-module-size.md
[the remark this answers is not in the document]
madeline-k: Hmm, I hadn't thought of it that way. Good idea. Let me switch to framing it that way, and see how it looks. I want to avoid including within the scope of work actually adding the future lambda layers. This project should just set up the mechanism to add them.

text/0039-reduce-module-size.md
madeline-k: That is something we still need to figure out with a POC :) Unless @rix0rrr has any idea? The seed of this idea to use npm packages that we automatically bring in for the customer came from him.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,518 @@
...
+2. Additional versions of [awscli](https://pypi.org/project/awscli/). See
+   [issue](https://github.com/aws/aws-cdk/issues/13993).
+3. [Boto3](https://pypi.org/project/boto3/), and
+   [aws-sdk](https://www.npmjs.com/package/aws-sdk). All custom resources in the
+   AWS CDK framework currently rely on whatever version of the AWS SDK that is
+   available in the Lambda runtime by default. To make the custom resources more
+   robust, and allow us to choose the exact version of the AWS SDK used by them,
+   we should bundle these dependencies into each custom resource’s Lambda code.
+
+If we add all of these with the current design, that would increase the size of `aws-cdk-lib` by ~91 MB*.
```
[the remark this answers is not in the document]
madeline-k: Changed my mind, this is going in the main doc, based on the conversation we had in the other comment about focusing more on future size.

text/0039-reduce-module-size.md
madeline-k: Added an Appendix that hopefully make this all WAY more clear. Please add more comments if not.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,534 @@
...
+Use separate npm packages for each `lambda-layer-X` module, which the CDK CLI
+will automatically install if not already present in the `node_modules`
+directory. With this solution, customers will not have to figure out when they
+do or do not need to include these dependencies, and they will not be required
+to access a public endpoint besides npm during runtime of their CDK apps.
+Practically, this solution breaks down into the following steps.
+
+1. Publish v2 compatible versions of `@aws-cdk/lambda-layer-aws-cli`,
+   `@aws-cdk/lambda-layer-aws-cli`, `@aws-cdk/lambda-layer-node-proxy-agent` as
+   their own npm packages, separate from `aws-cdk-lib`.
```
kaizencc, line 232: will these versions still live in the `aws-cdk` repo? will a new version get released with every `aws-cdk` release or will they be decoupled?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,534 @@
...
+package in ADC regions, the right dependencies will already be present in the
+`node_modules` directory.
+
+#### .jsii Files
+
+There are two files, .jsii.tabl.json and .jsii, bundled in `aws-cdk-lib` for the
+JSII runtime to work for any non-NodeJS languages. And, to enable code and
+documentation generation from the published npm package. Today, they makeup 42 %
+of the package size. And in the future, after adding more Lambda Layers would
+still compose 30%.
```
kaizencc:
```suggestion
documentation generation from the published npm package. Today, they make up 42%
of the package size. And in the future, after adding more Lambda Layers, .jsii files would
still compose 30% of the total size.
```

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+package in ADC regions, the right dependencies will already be present in the
+`node_modules` directory.
+
+#### .jsii Files
+
+There are two files, .jsii.tabl.json and .jsii, bundled in `aws-cdk-lib` for the
+JSII runtime to work for any non-NodeJS languages. And, to enable code and
+documentation generation from the published npm package. Today, they makeup 42%
+of the package size. And in the future, after adding more Lambda Layers would
+compose 30%.
```
kaizencc, line 325:
```suggestion
of the package size. And in the future, after adding more Lambda Layers, the .jsii files would
compose 30% of the package size.
```

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+CDK and publishing them into the ADC regions, e.g. `constructs`. We will add the
+lambda-layer packages to this process, and when customers use the aws-cdk npm
+package in ADC regions, the right dependencies will already be present in the
+`node_modules` directory.
+
+#### .jsii Files
+
+There are two files, .jsii.tabl.json and .jsii, bundled in `aws-cdk-lib` for the
+JSII runtime to work for any non-NodeJS languages. And, to enable code and
+documentation generation from the published npm package. Today, they makeup 42%
```
kaizencc, line 323:
```suggestion
documentation generation from the published npm package. Today, they make up 42%
```

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+   app in a new GitHub repository. The CDK CLI will download and cache these zip
+   files at synth time. The existing `lambda-layer-X` modules will be modified
+   to reference the large dependencies from a local cache. The CDK CLI will be
+   modified to update that cache at synth time, if needed. Then, the rest of
+   their usage will remain the same. The URLs for each dependency would follow a
+   predictable format, including each dependency’s version. Some automated
+   mechanism would be required to add new artifacts to the website, as new
+   versions of each dependency are released. And, to add new versions of each
+   dependency to aws-cdk-lib. The problem with this solution is that it requires
+   access to a public endpoint during synth of CDK applications. Many customers
```
skinny85, line 408: > The problem with this solution is that it requires
   access to a public endpoint during synth of CDK applications.

Doesn't the proposed chosen solution suffer from the some problem?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+
+This question raises an important drawback of this solution. It will result in a
+breaking change for customers who are using these Lambda Layers in
+network-restricted environments. If customers are able to use the CDK CLI in
+their network-restricted environments, then it is reasonable to assume they were
+able to acquire the `aws-cdk` package from npm somehow. This implies customers
+will be able to also acquire the lambda layer packages from npm. They can either
+make sure the necessary packages are already installed and available in the
+`node_modules` directory where their CDK applications are executed, or include
+them in a private npm registry.
```
skinny85, line 302: I think they can just depend on these packages directly in their CDK app, right? That will solve this problem?

I guess it doesn't solve it for non-NodeJS languages...

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+to access a public endpoint besides npm during runtime of their CDK apps.
+Practically, this solution breaks down into the following steps.
+
+1. Publish v2 compatible versions of `@aws-cdk/lambda-layer-aws-cli`,
+   `@aws-cdk/lambda-layer-aws-cli`, `@aws-cdk/lambda-layer-node-proxy-agent` as
+   their own npm packages, separate from `aws-cdk-lib`.
+2. Modify the `lambda-layer-X` submodules in `aws-cdk-lib` to not bundle these
+   large dependencies themselves, but instead reference the appropriate `Layer`
+   class from the packages in step 1. We will need to do something clever here
+   to make sure customers do not have IDE or compilation errors before the CDK
```
skinny85, line 236: What about the development experience (working on the CDK codebase) of these modules? Probably worth it to call that out as well. Perhaps just a `devDependency` is enough, but I'm not 100% sure.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+   CLI has an opportunity to download and install the lambda layer packages.
+3. Modify the CDK CLI to verify that the correct packages are available in the
+   `node_modules` directory during synthesis. If they are not available,
+   download them from npm.
```
skinny85, line 240: How does this work for non-NodeJS languages, that might not have a `node_modules` directory at all?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+* aws_eks.KubernetesObjectValue
+* aws_eks.KubernetesPatch
+* lambda_layer_awscli.AwsCliLayer
+* aws_s3_deployment.BucketDeploymentß
+* aws_stepfunctions_tasks.EmrContainersStartJobRun
+* lambda_layer_node_proxy_agent.NodeProxyAgentLayer
+* aws_eks.FargateProfile
+* aws_eks.Cluster
+
+Please see [#1234](https://github.com/aws/aws-cdk/issues/1234) for more details,
```
skinny85, line 35: I guess this is meant as a placeholder, but this actually links to an unrelated issue right now 😛.

Maybe use #1234567 instead?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+| KubectlLayer                                                                              | @aws-cdk/lambda-layer-kubectl                                                                           |
+| KubectlProvider, HelmChart, KubernetesManifest, KubernetesObjectValue, or KubernetesPatch | @aws-cdk/lambda-layer-kubectl and @aws-cdk/lambda-layer-awscli                                          |
+| AwsCliLayer, BucketDeployment, or EmrContainersStartJobRun                                | @aws-cdk/lambda-layer-awscli                                                                            |
+| NodeProxyAgentLayer, or FargateProfile                                                    | @aws-cdk/lambda-layer-node-proxy-agent                                                                  |
+| aws_eks.Cluster                                                                           | @aws-cdk/lambda-layer-kubectl, @aws-cdk/lambda-layer-awscli, and @aws-cdk/lambda-layer-node-proxy-agent |
```
skinny85, line 59: Can we use the module prefix everywhere in the table? I think it will make it easier to see on first glance. And maybe the full prefix? (`aws-cdk-lib/aws_eks`).

Maybe put each construct on its own line?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+caused performance problems for others. It also prevented the CDK team from
+releasing certain new features. Now, the size of the package is less than 100
+MB. This smaller version also has a mechanism for adding new lambda-layer
+dependencies. Because of this launch, the AWS CDK framework will be able to add
+support for more versions of kubectl, and awscliv2.
+
+### Why should I use this feature?
+
+You should upgrade your dependencies on `aws-cdk-lib` to get the benefits of a
+smaller package size. If you use `aws-cdk-lib` in a lambda function, now you can
```
skinny85, line 88: Capitalize "lambda" to "Lambda"?

(This comment applies to other places in the doc as well)

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+
+1. Publish v2 compatible versions of `@aws-cdk/lambda-layer-aws-cli`,
+   `@aws-cdk/lambda-layer-aws-cli`, `@aws-cdk/lambda-layer-node-proxy-agent` as
+   their own npm packages, separate from `aws-cdk-lib`.
+2. Modify the `lambda-layer-X` submodules in `aws-cdk-lib` to not bundle these
+   large dependencies themselves, but instead reference the appropriate `Layer`
+   class from the packages in step 1. We will need to do something clever here
+   to make sure customers do not have IDE or compilation errors before the CDK
+   CLI has an opportunity to download and install the lambda layer packages.
+3. Modify the CDK CLI to verify that the correct packages are available in the
```
skinny85, line 238: Do we have any idea how will we implement that? Can we use [optional dependencies](https://betterprogramming.pub/what-are-npms-optional-dependencies-and-when-should-we-use-them-796a6a964e73)?

skinny85 (commented):
Awesome docs!

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+We can deliver a notice to customers based on what specific constructs they are
+using. We will define separate notices for each different scenario.
+
+Title: (aws-cdk-lib): upcoming change in packaging structure might require
+action to upgrade
+
+Body: We’ve identified that you are using at least one of the [List of
+Constructs] Constructs. If you are running your aws-cdk commands in an
+environment that does not have access to npm, then you will need to make a
+change to make version 2.x of [npm packages] available in your environment. See
```
otaviomacedo, line 49: What change is this? I feel this summary could already give some hint on what the user would need to do. Otherwise this sounds super scary before you actually read the issue.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+234 MB. This large size prevented many customers from being able to use it, and
+caused performance problems for others. It also prevented the CDK team from
+releasing certain new features. Now, the size of the package is less than 100
+MB. This smaller version also has a mechanism for adding new Lambda Layer
+dependencies. Because of this launch, the AWS CDK framework will be able to add
+support for more versions of kubectl, and awscliv2.
+
+### Why should I use this feature?
+
+You should upgrade your dependencies on `aws-cdk-lib` to get the benefits of a
```
otaviomacedo, line 87: Well framed!

otaviomacedo, line 87: Well framed!

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+
+Note: The percentages do not add up to 100. Some files, e.g. various .json
+files, are excluded and contribute very little to the size.
+
+| Category                       | Percentage of aws-cdk-lib size |
+| ------------------------------ | ------------------------------ |
+| .jsii files                    | 41.64%                         |
+| Source Maps (.js.map)          | 21.93%                         |
+| Lambda Layer zip files         | 15.48%                         |
+| Javascript code (.js)          | 9.17%                          |
```
otaviomacedo, line 156: Does this include `.warnings.jsii.js` files? There is some room for improvement there: if there is no warning to print for a given class, it will still generate a corresponding function, taking up unnecessary space. We can change the warnings generation in JSII to not generate these functions and not call them from the compiled .js files. But, of course, it depends on how much this particular issue contributes to the whole. Might be worth it getting some data around this.

otaviomacedo, line 156: Does this include `.warnings.jsii.js` files? There is some room for improvement there: if there is no warning to print for a given class, it will still generate a corresponding function, taking up unnecessary space. We can change the warnings generation in JSII to not generate these functions and not call them from the compiled .js files. But, of course, it depends on how much this particular issue contributes to the whole. Might be worth it getting some data around this.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+4. We need to keep the dependencies up-to-date to get security updates and new
+   features. Each Lambda Layer will need to be treated slightly differently.
+    1. `@aws-cdk/lambda-layer-kubectl` - This one has two dependencies to keep
+       updated. All updates described below will result in a minor version bump
+       to this package.
+        1. [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
+            1. Kubernetes users will need to be able to choose the minor version
+               of kubectl that they want to use. And the aws-eks module will
+               need to as well.
+            2. The public API for this package will have a `KubectlV1_20Layer`
```
otaviomacedo, line 250: Are we planning to extend this mechanism to arbitrary lambda layers? Suppose I'm a member of the community and I want to write a new custom resource that depends on a different lambda layer. Am I allowed to do this? If so, what are the steps I should take to get the lambda layer available for consumption?

otaviomacedo, line 250: Are we planning to extend this mechanism to arbitrary lambda layers? Suppose I'm a member of the community and I want to write a new custom resource that depends on a different lambda layer. Am I allowed to do this? If so, what are the steps I should take to get the lambda layer available for consumption?

text/0039-reduce-module-size.md
madeline-k: I still need to work on a POC to see how this will actually be implemented. Thanks for the optional dependencies suggestion, that looks like it would work. I will dive deeper and then update this section!

text/0039-reduce-module-size.md
madeline-k: True. I need a better way to phrase this. The chosen solution is slightly better because the public endpoint is one that we already know customers will have access to.

text/0039-reduce-module-size.md
madeline-k: Good idea, that's another easier solution for them. For non-NodeJS languages,customers still have to download the CLI from npm... I'm still not 100% clear on how they will workaround this.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,533 @@
...
+These files currently makeup 22% of the total size of `aws-cdk-lib`. Source map
+files map from the transformed source to the original source. In the case of
+`aws-cdk-lib`, these are not actually useful today, since we do not ship the
+original `.ts` files. We will remove the source maps from the released package.
+
+#### Javascript Files (.js)
+
+The Javascript files make up about 10% of the overall package size. Today, they
+are already minified in the released package. There is not much else that can be
+done to make them smaller.
```
Chriscbr, line 345: It might be possible that we could make the JavaScript files smaller by reducing the amount of code in them. Specifically, if we change the ECMAScript version targeted by JSII/TypeScript, we could reduce the amount of polyfills that are added. (If I run `grep -r '"target": "ES' .` in the root of the the aws cdk repo, it looks as though we're mostly targeting ES2018 -- but the minimum version of Node we support is 14, which I believe supports all of the features in ES2019).

All that said, changing this might not make much of a dent. 🤷‍♂️

text/0039-reduce-module-size.md
madeline-k: Looks like they are about 1% of the size. I don't think it's worth it to optimize those for size at this time.

text/0039-reduce-module-size.md
madeline-k: I added this to the section, but it is still a bit hard to define until we actually implement this. Open to ideas on how to make those steps more concrete at this point.

text/0039-reduce-module-size.md
madeline-k: Just saw this comment - It should still be part of `aws-cdk` repo. Added some more detail on this.

otaviomacedo (dismissed):

text/0039-reduce-module-size.md
Chriscbr: I think the change would be updating all of the `tsconfig.json` files in the repo. Unfortunately I'm not sure which polyfills are currently added.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,534 @@
...
+
+* **Original Author(s):**: @madeline-k
+* **Tracking Issue**: #39
+* **API Bar Raiser**: tbd
+
+The AWS CDK v2 framework library, `aws-cdk-lib` contains all packages from AWS
+CDK v1 in one monolithic package. While this is a great value for customers in
+making dependency management much easier, it has also resulted in increasing the
+unpacked npm package size to 234 MB, at the time of writing. This feature
+reduces the size of the package to under 100 MB.
```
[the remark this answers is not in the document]
madeline-k: I was thinking this is a somewhat conservative estimate based on the size reduction we can expect from each section below. However, it is really hard to predict what the size will actually be after we've implemented it. We could either add an Appendix to explicitly justify why it will end up under 100 MB, or remove it entirely. What do you think?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+| README, etc (.md)              | 0.76%                          |
+| bundled npm dependencies       | 0.69%                          |
+| .ts-fixture                    | 0.37%                          |
+| .jsiirc                        | 0.36%                          |
+
+### Why is the package going to be larger in the future?
+
+The CDK team is currently blocked on adding more Lambda Layer zip files to the
+AWS CDK framework, because adding them would dramatically increase the size of
+the package. Each of these zips are in a module whose only purpose is to bundle
```
dbartholomae, line 167: I'm not sure I fully understand the plan here: Is the plan to reduce the package size so that it can be increased again with these layers? If so, then the arguments here are contradictory: We won't achieve both adding the layers and the benefits mentioned above of smaller package size at the same time.
Also, then the problem will come up again once there is a new layer which reaches the limit again.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+2. A large download size might be problematic for customers with weak internet
+   connections or limited bandwidth.
+3. Customers who download and install `aws-cdk-lib` on every CI/CD pipeline run
+   may be frustrated to spend resources frequently downloading a large package.
+4. AWS Lambda limits the size of your function code to 250 MB.
+5. The AWS Lambda limit for all uploaded assets for a Function is 75 GB. If a
+   customer is using Lambda Versions, and wants to maintain a Version for every
+   version they have published, they might end up hitting this limit because of
+   the large size of `aws-cdk-lib`.
+6. The developer experience of using `aws-cdk-lib` in an IDE might be slower in
```
rix0rrr, line 120: Pretty sure this one is not going to be true. If this is the case, it will be because of the large swathes of tiny `.d.ts` files, not a couple of large `.zip` files.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+If we add all of these with the current design, that would increase the size of
+`aws-cdk-lib` by approximately 130 MiB*, and make the Lambda Layers the biggest
+contributor of size at 39%.
+
+*See Appendix A to see where this estimate came from.
+
+| Category                       | Percentage of aws-cdk-lib size |
+| ------------------------------ | ------------------------------ |
+| Lambda Layer zip files         | 38.83%                         |
```
rix0rrr, line 196: If this table is intended to show the new distribution AFTER hypothetical future changes, I'm not sure that's necessary. The text is sufficient, and this table confused me initially because the numbers were so much different from the previous one. 

If you do want to show this data, combine it into 1 table I suppose.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+The main reason to not do this is that it will inevitably break up the
+monolothic `aws-cdk-lib` package in some manner, and we will lose some of the
+wonderful simplicity of releasing and using a single package. But, we can't keep
+everything in one place, and reduce the size enough to have a significant
+positive impact.
+
+### What is the technical solution (design) of this feature?
+
+There are different categories of files contributing to the size of aws-cdk-lib,
+and each category will have it’s own solution. The following sections cover the
```
rix0rrr, line 217:
```suggestion
and each category will have its own solution. The following sections cover the
```

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+top five categories in size contribution in order from largest to smallest
+potential size.
+
+#### Lambda Layer Zip Files
+
+Use separate npm packages for each `lambda-layer-X` module, which the CDK CLI
+will automatically install if not already present in the `node_modules`
+directory. With this solution, customers will not have to figure out when they
```
rix0rrr, line 225: Or globally installed.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+   their own npm packages, separate from `aws-cdk-lib`. This publishing process
+   will be an extension of the existing AWS CDK V2 release machinery. Any
+   submodules with the prefix `lambda-layer` will be published the same way.
+2. Modify the `lambda-layer-X` submodules in `aws-cdk-lib` to not bundle these
+   large dependencies themselves, but instead reference the appropriate `Layer`
+   class from the packages in step 1. Since `aws-cdk-lib`'s source code is
+   auto-generated from the v1 submodules, this could be an autogenerated
+   wrapper. We will need to do something clever here to make sure customers do
+   not have IDE or compilation errors before the CDK CLI has an opportunity to
+   download and install the packages.
```
rix0rrr, line 241: If the external NPM package is just the binary (with some wrapping JS to interface with), this would be trivial.

text/0039-reduce-module-size.md
rix0rrr: `~/.cache/cdk/large-packages` ?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+3. Modify the CDK CLI to verify that the correct packages are available in the
+   `node_modules` directory during synthesis. If they are not available,
+   download them from npm.
+4. We need to keep the dependencies up-to-date to get security updates and new
+   features. Each Lambda Layer will need to be treated slightly differently.
+    1. `@aws-cdk/lambda-layer-kubectl` - This one has two dependencies to keep
```
rix0rrr, line 247: Why don't we have `@aws-cdk/lambda-layer-kubectl-120` and `@aws-cdk/lambda-layer-kubectl-122` etc?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+               available version of Helm which is compatible with that version
+               of Kubernetes.
+            3. A weekly automated task will update the minor and patch versions
+               of helm within the supported ranges.
+            4. The automated task that checks for new supported minor versions
+               of kubectl, will also attempt to figure out what is the latest
+               version of helm compatible with that version of kubectl. This
+               will be verified by a human in PR review to make sure it is
+               correct.
+    2. `@aws-cdk/lambda-layer-aws-cli` - Major versions 1 and 2 will both be
```
rix0rrr, line 280: Also here why not 2 packages?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+#### .jsii Files
+
+There are two files, .jsii.tabl.json and .jsii, bundled in `aws-cdk-lib` for the
+JSII runtime to work for any non-NodeJS languages. And, to enable code and
+documentation generation from the published npm package. Today, they make up 42%
+of the package size. And in the future, after adding more Lambda Layers, the .jsii files would
+compose 30% of the package size.
+
+We will compress the .jsii.tabl.json and .jsii files. Compressing with gzip
```
rix0rrr, line 344: Is this valuable? I would argue it doesn't matter, since the download size will always be the already-compressed size. 

If we're going to do anything, we should probably try to find a more efficient encoding of the assembly format (with less redundancy), potentially stop pretty-printing it.

text/0039-reduce-module-size.md
rix0rrr: @Chriscbr do you want to take a stab at figuring out the impact of a change like this?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+#### Javascript Files (.js)
+
+The Javascript files make up about 10% of the overall package size. Today, they
+are already minified in the released package. There is not much else that can be
+done to make them smaller.
+
+#### Type Declaration Files (.d.ts)
+
+We can minify these files, which could result in a space saving of up to 20 MB.
```
rix0rrr, line 366: This is a tempting number, but people command-clicking to CDK definitions will end up in a file with unreadable garbage. Not sure this would be worth it.

I wonder if there are gains to be had if we combine the `.d.ts` files for submodules together (into 1 file per submodule). I'm guessing this would lead to better compression, though the difference might not be very significant. If nothing else,  would have performance gains as well for downstream `tsc` invocations, since the compiler wouldn't have to read so many files.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+Each category of files has different alternative solutions to reduce the size.
+
+#### .jsii Files
+
+1. The .jsii.tabl.json  and .jsii files are both “pretty-printed.” We can reduce
+   their size to 42 MB (-23%) and 30 MB (-33%) by removing all the whitespace
+   and newlines. We will not take this solution, because the space savings is
+   not as significant as compression. And, then we can maintain the benefit of
+   using pretty-printed JSON when uncompressed.
```
rix0rrr, line 391: I would argue the benefit of pretty-printed JSON isn't that large, and depending on the JSON decoded/parser loading the files would be faster if we strip whitespace. Might not be a big impact, but probably won't hurt either.

I would be in favor of this I think.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+they are the biggest potential contributor to size. And, there is customer need
+to add more Lambda Layers, which we are not able to do until the large
+dependencies are not required to be bundled in aws-cdk-lib.
+
+The next category will be the .jsii files. After both of these solutions are
+implemented, the the aws-cdk-lib unpacked npm package size will be about 100 MB.
+
+The lowest priority category will be removing the Source Map files and minifying
+the Type Declaration files. The final size of the package after this work is
+still to be determined.
```
rix0rrr, line 521: It's low priority but let's do it early on anyway, it's an easy win for about 5 minutes of work!

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+Work will be tracked on [this project
+board](https://github.com/aws/aws-cdk/projects/15).
+
+## Appendix A - Notes on size calculations
+
+1. All size calculations in this document were done based on
+   `aws-cdk-lib@2.19.0`.
+2. `du -k` was used to get sizes in Kb (Kebibytes), for calculating the
```
rix0rrr, line 530:
```suggestion
2. `du -k` was used to get sizes in Kb (Kibibytes), for calculating the
```

🤓

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,550 @@
...
+
+There are different categories of files contributing to the size of aws-cdk-lib,
+and each category will have it’s own solution. The following sections cover the
+top five categories in size contribution in order from largest to smallest
+potential size.
+
+#### Lambda Layer Zip Files
+
+Use separate npm packages for each `lambda-layer-X` module, which the CDK CLI
+will automatically install if not already present in the `node_modules`
```
rix0rrr, line 224: Needs a bit of elaboration on how the CLI will know what it needs to install, and how the app will know what has been installed and where.

text/0039-reduce-module-size.md
madeline-k: I need to clarify this better in the doc, since you are asking, it must not be clear! But the goal is to remove large dependencies for these layers, and also create a mechanism for adding more layers in the same manner. So that we can add more Lambda Layer dependencies without ballooning the package size after this project is implemented.

text/0039-reduce-module-size.md
madeline-k: This will be purely an implementation detail of the `Layer` constructs, and actually should not show errors for end users prior to the CDK CLI dynamically loading the dependencies from npm.

text/0039-reduce-module-size.md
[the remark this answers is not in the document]
madeline-k: Removed, let me know if you disagree.

text/0039-reduce-module-size.md
madeline-k: With the current proposed solution, these packages will remain in the aws-cdk git repo. So we will just keep the current `devDependency` on the `@aws-cdk/lambda-layer-x` packages that exists today.

text/0039-reduce-module-size.md
madeline-k: They still need to have a global `node_modules` directory to have the CDK CLI installed. I think we can attempt to install into the current application's `node_modules`, if there isn't one, then install into the global one. Although, will elevated permissions be required for that?

text/0039-reduce-module-size.md
madeline-k: Good idea, done :)

text/0039-reduce-module-size.md
madeline-k: Hmm okay, I will just remove this then. I think we have enough other data for why we need to do this project.

text/0039-reduce-module-size.md
madeline-k: I think this table is necessary to show why it is important to address the lambda layer zips. If not for the future things we want to add, then we could just address the .jsii files and remove the source maps and call it a day.

madeline-k: I am not sure how to logically combine this into one table and still have it makes sense in the flow of the doc. I change the title of the 2nd column of this one. Does that help make it less confusing? How would you like to see it represented?

text/0039-reduce-module-size.md
madeline-k: Oof, that's embarrassing 😂

text/0039-reduce-module-size.md
madeline-k: Yeah you are right. This package will be the `Layer` construct including the binary, but it will still be an implementation detail of the constructs exposed in `aws-cdk-lib`. Updated this section.

text/0039-reduce-module-size.md
madeline-k: I didn't think about this. I was thinking it makes logical sense to group all versions of one dependency into the same package. But it might actually be simpler to manage if they are separate. What do you think will be the major difference between the two options?

text/0039-reduce-module-size.md
madeline-k: Yes, definitely agree with you. This section only describes priority, they can all be done in parallel. So I don't think this plan prevents us from doing it early on. Do I need to specify that?

text/0039-reduce-module-size.md
madeline-k: The "unpacked" npm size is also important. I think this is a concrete step we should take right now to improve that. We can still pursue a more efficient assembly format as well. I am not sure what the options are there, and haven't done any investigation into it. Do you have any ideas?

text/0039-reduce-module-size.md
madeline-k: Oh, I thought we only needed to worry about API reference in IDEs. But yes this will be an issue as well.

text/0039-reduce-module-size.md
madeline-k: So humans never look at these files? In that case, I agree, we should stop pretty-printing. Currently thinking in addition to the compression.

text/0039-reduce-module-size.md
skinny85: > I think we can attempt to install into the current application's `node_modules`, if there isn't one, then install into the global one. Although, will elevated permissions be required for that?

Also, won't that cause problems if the customer has multiple CDK apps with different versions on their machine?

comcalvi: Why not create an application level `node_modules` if it doesn't exist?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,555 @@
...
+network-restricted environments. If customers are able to use the CDK CLI in
+their network-restricted environments, then it is reasonable to assume they were
+able to acquire the `aws-cdk` package from npm somehow. This implies customers
+will be able to acquire the `@aws-cdk/lambda-layer-X` packages from npm. There
+are a few options for these customers:
+1. If using TypeScript or JavaScript, add each necessary package the
+   dependencies of their CDK app.
+2. Make sure the necessary packages are already installed and available in the
+   `node_modules` directory where their CDK CLI is installed.
+3. Include them in a private npm registry. The
```
comcalvi, line 325: leftover 'The'?

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,555 @@
...
+
+This question raises an important drawback of this solution. It will result in a
+breaking change for customers who are using these Lambda Layers in
+network-restricted environments. If customers are able to use the CDK CLI in
+their network-restricted environments, then it is reasonable to assume they were
+able to acquire the `aws-cdk` package from npm somehow. This implies customers
+will be able to acquire the `@aws-cdk/lambda-layer-X` packages from npm. There
+are a few options for these customers:
+1. If using TypeScript or JavaScript, add each necessary package the
+   dependencies of their CDK app.
```
comcalvi, line 322: [Above](https://github.com/aws/aws-cdk-rfcs/pull/430/files#diff-94270ba514d2edadf1d22d85d22900af665e6aa9cbc8ecbea733105b83d39d35R219-R221) it says that we would automatically install the lambda layer dependencies. Why are we explaining how customers can go install these manually here?

text/0039-reduce-module-size.md
rix0rrr: Even though download size is the more important metric, reducing unpacked size will have some benefits for disk usage of projects (and the aforementioned 250MB Lambda footprint limit--even though it's possible to work around that limit, let's try to not force people to have to do that). Compressing is an easy way to get that size down, so why not.

One thing I missed originally is that the JSII manifest embedded in NPM packages will not (typically) be used at runtime. It will only be loaded by jsii build tooling--and this is especially true for tablets: so gzipping them will not usually have a big performance impact.

The only place we need to pay attention is jsii client languages. The Python bundle currently looks like this (containing the full NPM package, this pattern is repeated for all jsii languages):

```
┌───────────────────────────────────────────────────┐                    
│                  aws_cdk_lib.whl                  │                    
├───────────────────────────────────────────────────┤                    
│                                                   │                    
│                                                   │                    
│                                                   │                    
│               ┌───────────────────────────────┐   │                    
│               │     aws-cdk-lib.jsii.tgz      │   │                    
│               ├───────────────────────────────┤   │                    
│               │                               │   │                    
│   ┌──────┐    │  ┌──────┐   ┌─────────────┐   │   │       jsii kernel  
│   │ *.py │    │  │ *.js │   │    .jsii    │◀──┼───┼─────   needs this  
│   └──────┘    │  └──────┘   └─────────────┘   │   │           file     
│               │                               │   │                    
│               │   ┌───────────────────────┐   │   │                    
│               │   │    .jsii.tabl.json    │◀──┼───┼─────  nobody needs 
│               │   └───────────────────────┘   │   │        this file.  
│               │                               │   │         kill it.   
│               └───────────────────────────────┘   │                    
│                                                   │                    
└───────────────────────────────────────────────────┘                    
```

Two observations:

- The `.jsii.tabl.json` file is in there, which nobody needs. We can totally remove it to save a bunch of space.
- The jsii kernel is going to need to load the `.jsii` file, which turns into a `.jsii.gz` file if we go through with the compression strategy, paying a small performance penalty to load and gunzip the library on startup (we need to quantify how much this costs!). Either that or we put the uncompressed version in the jsii client archive, but then our disk space savings only materialize for TypeScript users, not any other language users.

text/0039-reduce-module-size.md
madeline-k: @skinny85 , you were right! They can depend on the packages directly from their CDK app, and it will work both for NodeJS and non-NodeJS languages. Non-Node users will just add direct dependencies in the same manner that they  usually do (Pip, Maven, etc). See this repo for the POC: https://github.com/madeline-k/dynamically-load-npm-dependencies

text/0039-reduce-module-size.md
madeline-k: The proposal has changed a bit since this conversation. The dependencies will be downloaded at runtime from the framework, not the CLI. Check out the latest version, and let me know if there is any further concerns about this topic!

text/0039-reduce-module-size.md
madeline-k: I've added some more detail in this section to explain. Thanks for your comment @dbartholomae!

text/0039-reduce-module-size.md
madeline-k: Resolving this comment, please re-open if you have a suggestion here.

text/0039-reduce-module-size.md
madeline-k: I've given this some more thought, and I think it will be best to have separate libraries for each minor version like you suggested. This way, the runtime only needs to download the kubectl version(s) that will actually be used. Updated the doc to reflect this.

text/0039-reduce-module-size.md
madeline-k: Good point here as well! Updated the doc :)

text/0039-reduce-module-size.md
madeline-k: I am no longer planning to use the CLI for installing/downloading.

text/0039-reduce-module-size.md
madeline-k: Thanks for catching!

text/0039-reduce-module-size.md
madeline-k: This is for customers using the CDK in environments where we *can't* automatically download these for them. So we need to provide a manual workaround for them. Let me know if that's not clear from the doc!

text/0039-reduce-module-size.md
RomainMuller: The tablet file is used by ConstructHub to not time out when generating documentation...

madeline-k: Thanks for confirming that we need to keep the table file, @RomainMuller .

madeline-k:
This RFC is now in the [final comments period](https://github.com/aws/aws-cdk-rfcs#7-final-comments-period). It will be merged by next week.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,583 @@
...
+   not impact the current developer workflow of working with these `Layer`
+   constructs. Developers will work with each `@aws-cdk` module in its
+   corresponding source directory. The publishing pipeline will publish the
+   `@aws-cdk/lambda-layer-X` packages, and the `aws-cdk-lib` code generation
+   process will create the wrappers that reference them.
+    1. I have implemented a POC that shows the proposed dynamic loading
+       mechanism will work with node languages and jsii target languages. See
+       [this
+       repo](https://github.com/madeline-k/dynamically-load-npm-dependencies)
+       for more details.
```
skinny85, line 252: It's so awesome that you managed to implement this, nice work! 😊

text/0039-reduce-module-size.md
skinny85: Love the new direction!

text/0039-reduce-module-size.md
iliapolo: How will we know which version of the `lambda-layer-X` package to install at synth time? We have to lock it somehow otherwise users get a non deterministic build. I have some ideas but wondering if you considered this.

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,583 @@
...
+   own. This solution is the same as the one described in this
+   [comment](https://github.com/aws/aws-cdk-rfcs/issues/39#issuecomment-593092612).
+4. Separate packages that `aws-cdk-lib` peer depends on. With this solution,
+   customers do not need to include the large `lambda-layer-X` packages in their
+   dependencies unless they are actually using one of the Constructs that
+   requires them. This would be confusing for customers to know when they need
+   these peer dependencies, and when they do not. It would also be difficult for
+   customers to manage versions and to know which versions of each package are
+   compatible with each other.
+5. Upstream the custom resources where the zip files are used. Since all of
```
iliapolo, line 481: A similar approach that suffers from them same problem would have been to setup an internal pipeline that publishes AWS provided layers, such as:

Screen Shot 2022-07-17 at 3 57 08 PM

Thought I'd put it on here anyway for posterity. Also - it might be worth while investigating how exactly this later is published.

text/0039-reduce-module-size.md
skinny85: We can probably use the same version as the package the depends on `lambda-layer-X` has. But I agree it's non-trivial to get out that information (we'll probably have to do some `package.json` parsing in the "dynamic require" code.

text/0039-reduce-module-size.md
madeline-k: Good callout, I did look into this approach. It basically has the same drawbacks as an internal pipeline publishing to s3. I heard from someone, but did not verify myself, that with Lambda Layers, you publish under your own account id, and so it is not possible to use convention-based naming.

text/0039-reduce-module-size.md
madeline-k: I've tested this with the [POC jsii library](https://github.com/madeline-k/dynamically-load-npm-dependencies), and if we add a package as an optional dependency, then the following happens:
1. The optional dependency is not included in the jsii packaged bundle for npm or PyPI
2. When you run `npm install` in a node project that lists the jsii library as a dependency, then the optional dependency is automatically installed as well.
3. You can run `npm install --no-optional` to avoid installing the optional dependencies
4. This has no impact on the experience for target languages (python). 

So, this would only impact TS/JS users. The argument for using optional dependencies is that this could make it easier for TS/JS users, because the optional dependencies would be installed by default. The argument against is that TS/JS users will need to change their current workflow to avoid installing things they don't need. At first, I don't think we should include this. Let's wait to see if there are real customer issues that this would alleviate.

text/0039-reduce-module-size.md
madeline-k: Yes, turns out package.json reflection is the way to go for this one. I've started implementing it for one of the layers in this PR: https://github.com/aws/aws-cdk/pull/21799

After you both raised this question, I realized this also affects how we release the `lamdba-layer-X` packages! I summarized the options in this issue https://github.com/aws/aws-cdk/issues/21605, and also updated the RFC to reflect the decision.

text/0039-reduce-module-size.md
madeline-k: I added this solution to the list of alternatives.

madeline-k:
PR Lint is failing because of a different RFC document

text/0039-reduce-module-size.md
```diff
@@ -0,0 +1,601 @@
...
+               correct.
+    2. AWS CLI - We will release two separate packages to make major versions 1
+       and 2 both available, and automatically update to the latest minor or
+       patch version.
+        1. Two packages: `@aws-cdk/lambda-layer-aws-cli-v1`, and
+           `@aws-cdk/lambda-layer-aws-cli-v2`.
+        2. Automatically check for new minor and patch versions from PyPi -
+           [awscli](https://pypi.org/project/awscli/),
+           [awscliv2](https://pypi.org/project/awscliv2/). Each update will
+           result in a minor version bump of the corresponding package.
```
kaizencc, line 307: should i change this to reflect the actual awscliv2 implementation?

madeline-k: Yes! It would be great if you could add that!

kaizencc (dismissed):

kaizencc (approved):

Issue #431, opened 2022-04-28
SageMaker Model Hosting L2 Constructs


petermeansrock:
## Description

Amazon SageMaker provides every developer and data scientist with the ability to build, train, and deploy machine learning models quickly. Amazon SageMaker is a fully-managed service that covers the entire machine learning workflow to label and prepare your data, choose an algorithm, train the model, tune and optimize it for deployment, make predictions, and take action. Your models get to production faster with much less effort and lower cost.

This RFC will specifically target the delivery of an L2 construct for a SageMaker `Endpoint` resource which represents a hosted endpoint for real-time inference. As an endpoint itself dependents on other resources, this RFC will extend to the following SageMaker resources as well:
* `EndpointConfig`
* `Model`

**Note:** The additional authors listed below contributed to the design of CDK implementation PR [6107](https://github.com/aws/aws-cdk/pull/) (which has been newly re-created in PR [20113](https://github.com/aws/aws-cdk/pull/20113)), but are not currently engaged on the drafting of this RFC.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @petermeansrock
| Author(s)           | @petermeansrock, @mattmcclean, @jetterdj, @foxpro24, @rangoju, *TBD*
| API Bar Raiser      | *TBD*
| Stakeholders        | *TBD*

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [ ] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

petermeansrock:
Per @kaizencc from https://github.com/aws/aws-cdk/pull/22886:

[...]

PR #433, opened 2022-05-04
RFC 431: SageMaker Model Hosting L2 Constructs

petermeansrock:
Co-authored-by: Matt McClean <mmcclean@amazon.com>
Co-authored-by: Long Yao <yl1984108@gmail.com>
Co-authored-by: Drew Jetter <60628154+jetterdj@users.noreply.github.com>
Co-authored-by: Murali Ganesh <59461079+foxpro24@users.noreply.github.com>
Co-authored-by: Abilash Rangoju <988529+rangoju@users.noreply.github.com>

This is a request for comments about SageMaker model hosting L2 constructs. See #431 for
additional details. 

APIs are signed off by @kaizencc.

---

[Rendered Version](https://github.com/petermeansrock/aws-cdk-rfcs/blob/sagemaker-l2-endpoint/text/0431-sagemaker-l2-endpoint.md)

---

### DIFF
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -264,7 +264,7 @@ import * as sagemaker from '@aws-cdk/aws-sagemaker';
 declare const endpointConfig: sagemaker.EndpointConfig;
 
 const endpoint = new sagemaker.Endpoint(this, 'Endpoint', { endpointConfig });
-const productionVariant = endpoint.findProductionVariant('variantName');
+const productionVariant = endpoint.findInstanceProductionVariant('variantName');
 const instanceCount = productionVariant.autoScaleInstanceCount({
   maxCapacity: 3
 });
@@ -287,7 +287,7 @@ import * as sagemaker from '@aws-cdk/aws-sagemaker';
 declare const endpointConfig: sagemaker.EndpointConfig;
 
 const endpoint = new sagemaker.Endpoint(this, 'Endpoint', { endpointConfig });
-const productionVariant = endpoint.findProductionVariant('variantName');
+const productionVariant = endpoint.findInstanceProductionVariant('variantName');
 productionVariant.metricModelLatency().createAlarm(this, 'ModelLatencyAlarm', {
   threshold: 100000,
   evaluationPeriods: 3,
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -1194,7 +1194,15 @@ proposed APIs are appropriately extensible in order to support these use-cases.
        been prefixed with the string `[Ii]nstance` to designate that they are only associated with
        instance-based hosting. When later adding serverless support to the SageMaker module,
        `[Ss]erverless`-prefixed analogs can be created with attributes appropriate for the use-case
-       with appropriate plumbing to the L1 constructs.
+       with appropriate plumbing to the L1 constructs. Note, there are a [number of features which
+       do not yet work with serverless variants][serverless-exclusions], so it may be necessary to
+       incorporate a number of new synthesis-time checks or compile-time contracts to guard against
+       mixing incompatible features. For example, as [discussed with the bar
+       raiser][design-conversation], alongside the proposed `EndpointConfigProps` attribute
+       `instanceProductionVariants?: InstanceProductionVariantProps[]`, a new mutually exclusive
+       attribute `serverlessProductionVariant?: ServerlessProductionVariantProps` (as only a single
+       variant is supported with serverless inference) could be added with a synthesis-time check
+       confirming that the customer hasn't configured both instance-based and serverless production variants.
     1. [Asynchronous Inference][async-inference]: By default, a deployed endpoint is synchronous:
        a customer issues an InvokeEndpoint operation to SageMaker with an attached input payload and
        the resulting response contains the output payload from the endpoint. To instead support
@@ -1203,7 +1211,7 @@ proposed APIs are appropriately extensible in order to support these use-cases.
        an InvokeEndpointAsync operation to SageMaker with an attached input location in S3;
        SageMaker asynchronously reads the input from S3, invokes the endpoint, and writes the output
        to an S3 location specified within the `AsyncInferenceClientConfig` attribute. As [discussed
-       with the RFC bar raiser here][async-conversation], there are a few ways to tackle the
+       with the RFC bar raiser here][design-conversation], there are a few ways to tackle the
        addition of this functionlity. One option is to add attribute(s) to the L2 endpoint config
        construct to support asynchronous inference along with synthesis-time error handling to
        catch configuration conflicts (e.g., asynchronous endpoints are only capable of supporting
@@ -1326,7 +1334,8 @@ proposed APIs are appropriately extensible in order to support these use-cases.
 [network-isolation]: https://aws.amazon.com/blogs/security/secure-deployment-of-amazon-sagemaker-resources/
 [marketplace-models]: https://aws.amazon.com/blogs/awsmarketplace/using-amazon-augmented-ai-with-aws-marketplace-machine-learning-models/
 
-[async-conversation]: https://github.com/aws/aws-cdk-rfcs/pull/433#discussion_r952949608
+[serverless-exclusions]: https://docs.aws.amazon.com/sagemaker/latest/dg/serverless-endpoints.html#serverless-endpoints-how-it-works-exclusions
+[design-conversation]: https://github.com/aws/aws-cdk-rfcs/pull/433#discussion_r952949608
 [deployment-guardrails-exclusions]: https://docs.aws.amazon.com/sagemaker/latest/dg/deployment-guardrails-exclusions.html
 [container-image-from-registry]: https://github.com/aws/aws-cdk/blob/v1-main/packages/%40aws-cdk/aws-ecs/lib/container-image.ts#L14-L19
 
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -2,7 +2,7 @@ SageMaker Model Hosting L2 Constructs
 
 * **Original Author(s):** @petermeansrock, @mattmcclean, @l2yao, @jetterdj, @foxpro24, @rangoju
 * **Tracking Issue**: #431
-* **API Bar Raiser**: *TBD*
+* **API Bar Raiser**: @conroyka
 
 This feature supports the creation of Amazon SageMaker real-time inference hosted endpoints using a
 new set of L2 constructs for the `Endpoint`, `EndpointConfig`, and `Model` CloudFormation resources.
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -209,11 +209,11 @@ By using the `EndpointConfig` construct, you can define a set of endpoint config
 used to provision one or more endpoints. In this configuration, you identify one or more models to
 deploy and the resources that you want Amazon SageMaker to provision. You define one or more
 production variants, each of which identifies a model. Each production variant also describes the
-resources that you want Amazon SageMaker to provision. This includes the number and type of ML
-compute instances to deploy. If you are hosting multiple models, you also assign a variant weight to
-specify how much traffic you want to allocate to each model. For example, suppose that you want to
-host two models, A and B, and you assign traffic weight 2 for model A and 1 for model B. Amazon
-SageMaker distributes two-thirds of the traffic to Model A, and one-third to model B:
+resources that you want Amazon SageMaker to provision. If you are hosting multiple models, you also
+assign a variant weight to specify how much traffic you want to allocate to each model. For example,
+suppose that you want to host two models, A and B, and you assign traffic weight 2 for model A and 1
+for model B. Amazon SageMaker distributes two-thirds of the traffic to Model A, and one-third to
+model B:
 
 ```typescript
 import * as sagemaker from '@aws-cdk/aws-sagemaker';
@@ -222,7 +222,7 @@ declare const modelA: sagemaker.Model;
 declare const modelB: sagemaker.Model;
 
 const endpointConfig = new sagemaker.EndpointConfig(this, 'EndpointConfig', {
-  productionVariants: [
+  instanceProductionVariants: [
     {
       model: modelA,
       variantName: 'modelA',
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -1,6 +1,6 @@
 SageMaker Model Hosting L2 Constructs
 
-* **Original Author(s):**: @pvanlund
+* **Original Author(s):** @pvanlund, @mattmcclean, @l2yao, @jetterdj, @foxpro24, @rangoju
 * **Tracking Issue**: #431
 * **API Bar Raiser**: *TBD*
 
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -62,32 +62,22 @@ import * as sagemaker from '@aws-cdk/aws-sagemaker';
 
 ## Model
 
-By creating a model, you tell Amazon SageMaker where it can find the model components. This includes
-the S3 path where the model artifacts are stored and the Docker registry path for the image that
-contains the inference code. The `ContainerDefinition` interface encapsulates both the specification
-of model inference code as a `ContainerImage` and an optional set of artifacts as `ModelData`.
+In machine learning, a model is used to make predictions, or inferences. A deployable model in
+SageMaker consists of inference code and model artifacts. Model artifacts are the results of model
+training by using a machine learning algorithm. The inference code must be packaged in a Docker
+container and stored in Amazon ECR. You can either package the model artifacts in the same container
+as the inference code, or store them in Amazon S3. As model artifacts may change each time a new
+model is trained (while the inference code may remain static), artifact separation in S3 may act as
+a natural decoupling point for your application.
 
-### Container Images
+When instantiating the `Model` construct, you tell Amazon SageMaker where it can find these model
+components. The `ContainerDefinition` interface encapsulates both the specification of model
+inference code as a `ContainerImage` and an optional set of separate artifacts as `ModelData`.
 
-Inference code can be stored in the Amazon EC2 Container Registry (Amazon ECR), which is specified
-via `ContainerDefinition`'s `image` property which accepts a class that extends the `ContainerImage`
-abstract base class.
+### Single Container Model
 
-#### `EcrImage`
-
-Reference an image available within ECR:
-
-```typescript
-import * as ecr from '@aws-cdk/aws-ecr';
-import * as sagemaker from '@aws-cdk/aws-sagemaker';
-
-const repository = ecr.Repository.fromRepositoryName(this, 'Repository', 'repo');
-const image = sagemaker.ContainerImage.fromEcrRepository(repository, 'tag');
-```
-
-#### `AssetImage`
-
-Reference a local directory containing a Dockerfile:
+In the event that a single container is sufficient for your inference use-case, you can define a
+single-container model:
 
 ```typescript
 import * as sagemaker from '@aws-cdk/aws-sagemaker';
@@ -96,49 +86,8 @@ import * as path from 'path';
 const image = sagemaker.ContainerImage.fromAsset(this, 'Image', {
   directory: path.join('path', 'to', 'Dockerfile', 'directory')
 });
-```
-
-### Model Artifacts
-
-Models are often associated with model artifacts, which are specified via the `modelData` property
-which accepts a class that extends the `ModelData` abstract base class. The default is to have no
-model artifacts associated with a model.
-
-#### `S3ModelData`
-
-Reference an S3 bucket and object key as the artifacts for a model:
-
-```typescript
-import * as s3 from '@aws-cdk/aws-s3';
-import * as sagemaker from '@aws-cdk/aws-sagemaker';
-
-const bucket = new s3.Bucket(this, 'MyBucket');
-const modelData = sagemaker.ModelData.fromBucket(bucket, 'path/to/artifact/file.tar.gz');
-```
-
-#### `AssetModelData`
-
-Reference local model data:
-
-```typescript
-import * as sagemaker from '@aws-cdk/aws-sagemaker';
-import * as path from 'path';
-
 const modelData = sagemaker.ModelData.fromAsset(this, 'ModelData',
   path.join('path', 'to', 'artifact', 'file.tar.gz'));
-```
-
-### `Model`
-
-The `Model` construct associates container images with their optional model data.
-
-#### Single Container Model
-
-In the event that a single container is sufficient for your inference use-case, you can define a
-single-container model:
-
-```typescript fixture=with-assets
-import * as sagemaker from '@aws-cdk/aws-sagemaker';
 
 const model = new sagemaker.Model(this, 'PrimaryContainerModel', {
   container: {
@@ -148,7 +97,7 @@ const model = new sagemaker.Model(this, 'PrimaryContainerModel', {
 });
 ```
 
-#### Inference Pipeline Model
+### Inference Pipeline Model
 
 An inference pipeline is an Amazon SageMaker model that is composed of a linear sequence of two to
 five containers that process requests for inferences on data. You use an inference pipeline to
@@ -170,6 +119,67 @@ const model = new sagemaker.Model(this, 'InferencePipelineModel', {
     { image: image3, modelData: modelData3 }
   ],
 });
+```
+
+### Container Images
+
+Inference code can be stored in the Amazon EC2 Container Registry (Amazon ECR), which is specified
+via `ContainerDefinition`'s `image` property which accepts a class that extends the `ContainerImage`
+abstract base class.
+
+#### Asset Image
+
+Reference a local directory containing a Dockerfile:
+
+```typescript
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+import * as path from 'path';
+
+const image = sagemaker.ContainerImage.fromAsset(this, 'Image', {
+  directory: path.join('path', 'to', 'Dockerfile', 'directory')
+});
+```
+
+#### ECR Image
+
+Reference an image available within ECR:
+
+```typescript
+import * as ecr from '@aws-cdk/aws-ecr';
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const repository = ecr.Repository.fromRepositoryName(this, 'Repository', 'repo');
+const image = sagemaker.ContainerImage.fromEcrRepository(repository, 'tag');
+```
+
+### Model Artifacts
+
+If you choose to decouple your model artifacts from your inference code, the artifacts can be
+specified via the `modelData` property which accepts a class that extends the `ModelData` abstract
+base class. The default is to have no model artifacts associated with a model.
+
+#### Asset Model Data
+
+Reference local model data:
+
+```typescript
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+import * as path from 'path';
+
+const modelData = sagemaker.ModelData.fromAsset(this, 'ModelData',
+  path.join('path', 'to', 'artifact', 'file.tar.gz'));
+```
+
+#### S3 Model Data
+
+Reference an S3 bucket and object key as the artifacts for a model:
+
+```typescript
+import * as s3 from '@aws-cdk/aws-s3';
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const bucket = new s3.Bucket(this, 'MyBucket');
+const modelData = sagemaker.ModelData.fromBucket(bucket, 'path/to/artifact/file.tar.gz');
 ```
 
 ## Model Hosting
--- a/text/0431-sagemaker-l2-endpoint.md
+++ b/text/0431-sagemaker-l2-endpoint.md
@@ -1107,39 +1107,15 @@ No.
 
 ### What are the drawbacks of this solution?
 
-This RFC and its [original associated implementation PR][original-pr] were based on a Q3 2019
-feature set of SageMaker real-time inference endpoints. Since that point in time, SageMaker has
-launched the following features which would require further additions to the L2 API contracts:
-
-* [Multi-model endpoints][multi-model]
-* [Model Monitor][model-monitor]
-* [Asynchronous inference][async-inference]
-* [Deployment guardrails][deployment-guardrails]
-* [Serverless inference][serverless-inference]
-
-Although some of these changes would be small and additive (e.g., `DataCaptureConfig` for Model
-Monitor), features like asynchronous and serverless inference represent more significant shifts in
-functionality. For example, SageMaker hosts real-time inference endpoints on EC2 instances, meaning
-that CloudWatch alarms and Application Auto Scaling rules operate on instance-based metrics. In
-contrast, serverless inference does not expose any instance-based metrics nor does it yet support
-auto-scaling. Since both features are specified via the CloudFormation resource
-`AWS::SageMaker::EndpointConfig`, the current recommendation of this RFC would be to support the
-specification of both use-cases through a single L2 `EndointConfig` construct. However, this
-presents a challenge when modeling helper APIs like `metricCPUUtilization` or
-`autoScaleInstanceCount` on a related construct as those methods would not universally apply to
-*all* endpoint types.
-
-Given that (1) RFC reviewers may have idiomatic recommendations to solve such modeling challenges
-and (2) the current proposed constructs are still viable for creating a SageMaker endpoint, the
-first draft of this RFC is being published without further revisions accommodating the above newer
-SageMaker features.
-
-[original-pr]: https://github.com/aws/aws-cdk/pull/6107
-[multi-model]: https://aws.amazon.com/blogs/machine-learning/save-on-inference-costs-by-using-amazon-sagemaker-multi-model-endpoints/
-[model-monitor]: https://aws.amazon.com/about-aws/whats-new/2019/12/introducing-amazon-sagemaker-model-monitor/
-[async-inference]: https://aws.amazon.com/about-aws/whats-new/2021/08/amazon-sagemaker-asynchronous-new-inference-option/
-[deployment-guardrails]: https://aws.amazon.com/about-aws/whats-new/2021/11/new-deployment-guardrails-amazon-sagemaker-inference-endpoints/
-[serverless-inference]: https://aws.amazon.com/about-aws/whats-new/2021/12/amazon-sagemaker-serverless-inference/
+Since production variants are configured via the `EndpointConfig` construct while the monitoring and
+auto-scaling of a deployed production variant is only possible once the `EndpointConfig` has
+been associated to an `Endpoint` (i.e., the dimension for most SageMaker model hosting metrics
+consists of endpoint name and production variant name), this RFC proposes the implementation of the
+function `Endpoint.findProductionVariant(string)`, the [return value for
+which](#endpoint-production-variants) contains `metric*` and `autoScaleInstanceCount` helper methods
+as demonstrated in the [README](#metrics). Although not necessarily a drawback, this separation of
+configuration-time and deploy-time APIs appears to be a novel pattern for the CDK, and thus, has the
+potential to be confusing to customers.
 
 ### What is the high-level project plan?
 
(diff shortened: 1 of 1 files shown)


text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+Import it into your code:
+
+```typescript
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+```
+
+## Model
+
+By creating a model, you tell Amazon SageMaker where it can find the model components. This includes
```
rix0rrr, line 65: I don't know yet what a model is, and did you meant to say "by creating a `Model`" ? As in, refer to a class here?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+```typescript
+import * as ecr from '@aws-cdk/aws-ecr';
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const repository = ecr.Repository.fromRepositoryName(this, 'Repository', 'repo');
+const image = sagemaker.ContainerImage.fromEcrRepository(repository, 'tag');
+```
+
+#### `AssetImage`
```
rix0rrr, line 88: People don't actually see the class `AssetImage` themselves, so the section title might be confusing.

We generally recommend people use assets. Can we move this paragraph up?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+single-container model:
+
+```typescript fixture=with-assets
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const model = new sagemaker.Model(this, 'PrimaryContainerModel', {
+  container: {
+    image: image,
+    modelData: modelData,
+  }
```
rix0rrr, line 147: We prefer a [flat namespace whenever possible](https://github.com/aws/aws-cdk/blob/master/docs/DESIGN_GUIDELINES.md#flat). Does the additional nesting of `container` here provide a lot of value?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const model = new sagemaker.Model(this, 'InferencePipelineModel', {
+  container: {
+    image: image1, modelData: modelData1
+  },
+  extraContainers: [
+    { image: image2, modelData: modelData2 },
+    { image: image3, modelData: modelData3 }
+  ],
```
rix0rrr, line 171: Oh I see.

Why the departure in syntax here, between `container` and `extraContainers` ? Why don't we have a single list of `containers`, which can have 1..* elements?

text/0431-sagemaker-l2-endpoint.md
rix0rrr: Would it make sense for `Container` to be a class?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+  productionVariant: {
+    model: modelA,
+    variantName: 'modelA',
+    initialVariantWeight: 2.0,
+  },
+  extraProductionVariants: [{
+    model: modelB,
+    variantName: 'variantB',
+    initialVariantWeight: 1.0,
+  }]
```
rix0rrr, line 204: Same question here about the non-regularity

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+    image: image,
+    modelData: modelData,
+  }
+});
+```
+
+#### Inference Pipeline Model
+
+An inference pipeline is an Amazon SageMaker model that is composed of a linear sequence of two to
+five containers that process requests for inferences on data. You use an inference pipeline to
```
comcalvi, line 154: Why two to five containers? The docs indicate a maximum of 15 containers: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sagemaker-model.html#cfn-sagemaker-model-containers

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+instances and deploys the model or models as specified in the configuration. To get inferences from
+the model, client applications send requests to the Amazon SageMaker Runtime HTTPS endpoint. For
+more information about the API, see the
+[InvokeEndpoint](https://docs.aws.amazon.com/sagemaker/latest/dg/API_runtime_InvokeEndpoint.html)
+API. Defining an endpoint requires at minimum the associated endpoint configuration:
+
+```typescript fixture=with-endpoint-config
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const endpoint = new sagemaker.Endpoint(this, 'Endpoint', { endpointConfig });
```
rix0rrr, line 220: Is it valuable to have `EndpointConfig` as its own separate visible resource in the first place? Why don't I just directly declare an `Endpoint` which has 1..* models?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+```
+
+### AutoScaling
+
+The `autoScaleInstanceCount` method on the `IEndpointProductionVariant` interface can be used to
```
rix0rrr, line 225: Too much detail on the interface here I think :).

comcalvi (commented):
First quick pass, will take another look later.

rix0rrr (commented):
Haven't read everything yet, comments so far

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+   L2 construct for endpoint configuration was incorporated into this RFC.
+
+[earliest-pr]: https://github.com/aws/aws-cdk/pull/2888
+
+### What are the drawbacks of this solution?
+
+This RFC and its [original associated implementation PR][original-pr] were based on a Q3 2019
+feature set of SageMaker real-time inference endpoints. Since that point in time, SageMaker has
+launched the following features which would require further additions to the L2 API contracts:
+
```
TheRealAmazonKendra, line 1089: If these are part of SageMaker's core feature set, I think they need to be taken into account in this RFP. We may be creating one way doors with contracts set here that will make the user experience more complicated if these are added later.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+As the proposed design has been fully implemented in
+[CDK PR #20113](https://github.com/aws/aws-cdk/pull/20113), the delivery timeline of the
+implementation of this RFC will be contingent upon the scope of changes requested by reviewers. For
+baking, the L2 constructs for this module would be marked as experimental, leaving room for further
+adjustments prior to marking the APIs as stable.
+
+### Are there any open issues that need to be addressed later?
+
+1. Please see the [drawbacks section above](#what-are-the-drawbacks-of-this-solution) for potential
+   follow-on work (assuming it need not be incorporated into this RFC).
```
TheRealAmazonKendra, line 1131: I think we at least need an assessment of how leaving out that work could potentially create one way doors or impact our ability to add the feature set not included in this RFP.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+1. As observed with [Lambda][lambda-eni-issue] and [EKS][eks-eni-issue], the Elastic Network
+   Interfaces (ENIs) associated with a SageMaker model's VPC are not always cleaned up in a timely
+   manner after downstream compute resources are deleted. As a result, attempts to delete a
+   SageMaker endpoint along with its networking resources (e.g., subnets, security groups) from a
+   CloudFormation stack may cause the stack operation to fail as the ENIs are still in use. From a
+   CDK integration test perspective, specifying `--no-clean` will allow the generation of a snapshot
+   regardless of whether stack deletion will succeed or fail but may hinder snapshot re-generation
+   by subsequent CDK contributors. For this reason, it may be helpful to exclude VPC specification
+   from the endpoint integration test at this time.
```
TheRealAmazonKendra, line 1140: Rather than excluding testing, we should add retries to the tests that take into account these delays.

petermeansrock:
Thanks for all the feedback so far! I'll be OOTO for the remainder of the week, so I'll begin fielding comments early next week.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+  ```ts
+  export interface ProductionVariantProps {
+    /**
+     * The size of the Elastic Inference (EI) instance to use for the production variant. EI instances
+     * provide on-demand GPU computing for inference.
+     *
+     * @default none
+     */
+    readonly acceleratorType?: AcceleratorType;
```
rix0rrr, line 759: How does this relate to instanceType?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+     * name.
+     */
+    readonly endpointName?: string;
+
+    /**
+     * The endpoint configuration to use for this endpoint.
+     *
+     * [disable-awslint:ref-via-interface]
+     */
+    readonly endpointConfig: EndpointConfig;
```
rix0rrr, line 894: Why is this not an interface?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+     *
+     * @default average over 5 minutes
+     */
+    metricMemoryUtilization(props?: cloudwatch.MetricOptions): cloudwatch.Metric;
+    /**
+     * Metric for GPU utilization
+     *
+     * @default average over 5 minutes
+     */
+    metricGPUUtilization(props?: cloudwatch.MetricOptions): cloudwatch.Metric;
```
rix0rrr, line 1040: No uppercase abbreviations. `metricGpuUtilization`, `metricCpuUtilization`, etc.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+   for both use-cases to `bind` using an `IGrantable` (adjusting ECS's `TaskDefinition`
+   accordingly). However, it's unclear within which module such a unified API should reside as
+   support for private repositories makes it a bad fit for `@aws-cdk/ecr` and `@aws-cdk/ecr-assets`,
+   and it would be unintuitive for `@aws-cdk/sagemaker` to declare a dependency on `@aws-cdk/ecs`.
+
+   Package concerns aside, historically, there was a period during which SageMaker only supported
+   ECR as an image source while ECS was capable of sourcing images from either ECR or a
+   customer-owned private repository. Given the fact that these two products' supported images
+   sources may yet again diverge in the future, maybe it would be best to keep their
+   `ContainerImage` APIs separate within their respective modules.
```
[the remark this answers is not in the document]
rix0rrr: I agree that reuse is better, but in this particular case reusing from `ecs` doesn't sound great. We might at some point support a more generic `ContainerImage`--for now this is okay.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+1. In the [earliest PR][earliest-pr] attempting to add SageMaker L2 constructs to the CDK, the
+   author did not create an `EndpointConfig` construct, instead hiding the resource's creation
+   behind `Endpoint` (to which production variants could be added). Although a simplifier, this
+   prevents customers from reusing configuration across endpoints. For this reason, an explicit
+   L2 construct for endpoint configuration was incorporated into this RFC.
```
rix0rrr, line 1080: Thanks for getting into this. What would be a typical use case to use the same config for multiple endpoints?

text/0431-sagemaker-l2-endpoint.md
rix0rrr: At least I'd like a good story for how the API would evolve.

text/0431-sagemaker-l2-endpoint.md
rix0rrr: I agree. This is out of scope for the current RFC, but it may give @corymhall some inspiration 😜

comcalvi (dismissed):

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+multiple models, you also assign a variant weight to specify how much traffic you want to allocate
+to each model. For example, suppose that you want to host two models, A and B, and you assign
+traffic weight 2 for model A and 1 for model B. Amazon SageMaker distributes two-thirds of the
+traffic to Model A, and one-third to model B:
+
+```typescript fixture=with-assets
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const endpointConfig = new sagemaker.EndpointConfig(this, 'EndpointConfig', {
+  productionVariant: {
```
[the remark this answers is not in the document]
petermeansrock: In the earliest PR for these constructs, [@skinny85 commented](https://github.com/aws/aws-cdk/pull/2888#discussion_r311259586):

> So here's an interesting pattern that you can use here.
> 
> Since at least one ProductionVariant is required, have in props:
> 
> ```ts
> productionVariant: ProductionVariant; // required
> extraProductionVariants?: ProductionVariant[]; // optional
> ```
> 
> This way, you communicate to the clients of this class, at compile time, that at least one variant has to be provided.

Given that this new feedback contradicts Adam's original guidance, in light of this context, I just wanted to double-check that I should indeed recombine these attributes back into a single array.

---

CC-ing @rix0rrr as he gave the same feedback in another comment

text/0431-sagemaker-l2-endpoint.md
petermeansrock: Most of the current verbiage in the README was taken from SageMaker's public documentation ([link relevant to this section](https://docs.aws.amazon.com/sagemaker/latest/dg/realtime-endpoints-deployment.html#realtime-endpoints-deployment-create-model)), which uses "*model*" to describe both the SageMaker resource and its associated abstract ML concept.

Which of the following are you looking for?

1. Use/paraphrase a one-liner from the [AWS glossary](https://docs.aws.amazon.com/general/latest/gr/glos-chap.html#ml-model), as in the following
    > In machine learning (ML), a mathematical model that generates predictions by finding patterns in data.
1. A more substantial write-up where I describe ML concepts to non-ML customers
1. Simply use `Model` in place of the first use of "*model*"?

text/0431-sagemaker-l2-endpoint.md
petermeansrock: Good catch. As described in the [drawback section](https://github.com/petermeansrock/aws-cdk-rfcs/blob/75dbf2360c7e6bb0e046e89954703a44723da203/text/0431-sagemaker-l2-endpoint.md#what-are-the-drawbacks-of-this-solution), the initial draft of this RFC was based on a Q3 2019 snapshot of the SageMaker feature set; SageMaker has since increased the number of supported containers. As part of authoring the second revision of this RFC, I'll audit the SageMaker CloudFormation resources to bring the RFC in line with the current supported attributes/features.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
kaizencc: Want to hear from others as well but I feel like a synth-time check for at least one production variant would suffice.

petermeansrock (commented):
Responding to remaining comments on the first draft.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+     */
+    readonly variantName: string;
+  }
+  ```
+
+- `ProductionVariant` -- represents a production variant that has been associated with an
+  `EndpointConfig`
+
+  ```ts
+  export interface ProductionVariant {
```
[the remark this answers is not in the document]
petermeansrock: Please see [the above related comment thread](https://github.com/aws/aws-cdk-rfcs/pull/433/files#r870700436).

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+#### `S3ModelData`
+
+Reference an S3 bucket and object key as the artifacts for a model:
+
+```typescript
+import * as s3 from '@aws-cdk/aws-s3';
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const bucket = new s3.Bucket(this, 'MyBucket');
+const modelData = sagemaker.ModelData.fromBucket(bucket, 'path/to/artifact/file.tar.gz');
```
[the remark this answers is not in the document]
kaizencc: I think this is okay. If @rix0rrr thinks otherwise we can discuss further.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+to each model. For example, suppose that you want to host two models, A and B, and you assign
+traffic weight 2 for model A and 1 for model B. Amazon SageMaker distributes two-thirds of the
+traffic to Model A, and one-third to model B:
+
+```typescript fixture=with-assets
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+
+const endpointConfig = new sagemaker.EndpointConfig(this, 'EndpointConfig', {
+  productionVariant: {
+    model: modelA,
```
kaizencc, line 196: Is `modelA` of type `sagemaker.Model`? It's hard to tell. In general, we've moved away from providing specific rosetta fixtures beyond the default, for this exact reason. It's hard to tell the type of the things you've put in the fixture, and it won't show up on our docs page. We mitigate this issue now by using `declare const modelA: sagemaker.Model;` now, so at least the type of a property is clear in the example.

Can you remove any other examples with fixture-with-assets and instead utilize the `declare` pattern?

text/0431-sagemaker-l2-endpoint.md
kaizencc: Reopening to say that the related comment convinces me that having an `EndpointConfig` L2 makes sense. However, we should upstream that reasoning here so that our users will understand the use case as well.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+```
+[ ] Signed-off by API Bar Raiser @xxxxx
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+We are launching the first set of L2 constructs for an existing module (`@aws-cdk/aws-sagemaker`),
```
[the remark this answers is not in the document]
kaizencc: The above is all correct, and the reword is good. We currently do some magic to turn the `import * as blah from '@aws-cdk/aws-blah'` into `import * as blah from 'aws-cdk-lib/aws-blah'`, but that only happens in the examples, I believe.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
kaizencc: Combining `ContainerImage` is out of scope for this RFC. You're good to go ahead with SageMaker specific `ContainerImage` API. Creating a generic `ContainerImage` is our problem to solve, you don't really need to worry about it :). 

(Unless if you have ideas in which case feel free to remember to open a github issue after this L2 is merged!)

text/0431-sagemaker-l2-endpoint.md
kaizencc: Responded on a different comment but this makes sense, just please upstream it to the EndpointConfig readme section (just a simple "By using the `EndpointConfig` construct, you can define one endpoint and reuse it on multiple `Endpoint` constructs")

text/0431-sagemaker-l2-endpoint.md
kaizencc: Just to be clear. We won't actually be creating any one-way doors because sagemaker will start out as an alpha module. I think we should be thinking about how these constructs can be extended to support other SageMaker resources in the future, but it's not necessary to include them in the current RFC.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
kaizencc: I think its okay to go forward with the consensus that we need just one property `productionVariants` and one property `containers`. These will be required properties so it stands to reason that at least one variant and one container must be provided. If we have to go back on this decision I will eat crow :)

text/0431-sagemaker-l2-endpoint.md
kaizencc: Simply "To enable autoscaling on the production variant, use the `autoScaleInstanceCount` method:"

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+The proposed design has been fully implemented in
+[CDK PR #20113](https://github.com/aws/aws-cdk/pull/20113). Each of the following sections lays out
+the proposed interfaces needed for each L2 construct along with any supporting classes.
+
+#### Model
+
+- `IModel` -- interface for defined and imported models
+
+  ```ts
+  export interface IModel extends cdk.IResource, iam.IGrantable, ec2.IConnectable {
```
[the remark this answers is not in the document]
kaizencc: @rix0rrr

kaizencc (changes requested):
Hi @petermeansrock! Just doing some housekeeping and responding to some of your inquiries. Will be waiting for the next revision before I take another look. In general, let's focus mostly on the README, and far less on the "technical solution" FAQ. That question is meant to be more high-level anyway, you're in a unique position where you have actually implemented most of what we're designing.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+    public static fromEcrRepository(repository: ecr.IRepository, tag: string = 'latest'): ContainerImage { ... }
+
+    /**
+     * Reference an image that's constructed directly from sources on disk
+     *
+     * @param scope The scope within which to create the image asset
+     * @param id The id to assign to the image asset
+     * @param props The properties of a Docker image asset
+     */
+    public static fromAsset(scope: Construct, id: string, props: assets.DockerImageAssetProps): ContainerImage { ... }
```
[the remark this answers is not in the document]
petermeansrock: @rix0rrr, as I was implementing this change, I noticed that the [ECS module's `AssetImage` hardcodes the construct ID as `'AssetImage'`](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-ecs/lib/images/asset-image.ts#L26). Bringing this updated API as-is into the SageMaker module would present a rough edge to customers as it would prevent customers from using the API to define inference pipelines (represented as multiple containers), the reason being that the second `bind` invocation (for the second container) would attempt to reuse a hardcoded construct ID resulting in a an error like:
```
Error: There is already a Construct with name 'AssetImage' in Model [InferencePipelineModel]
```

I see that [Elad introduced into the ECS module the API `ContainerImage.fromDockerImageAsset`](https://github.com/aws/aws-cdk/commit/b94577a0acb0aa76e257fb15267faf6c0954700a) which operates on an `DockerImageAsset` (the construct ID for which the customer has full control).

To sidestep the rough edge for customers, I'm considering replacing `ContainerImage.fromAsset` with `ContainerImage.fromDockerImageAsset`. Does that seem reasonable or would you prefer I keep both APIs?

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+     *
+     * @default none
+     */
+    readonly acceleratorType?: AcceleratorType;
+    /**
+     * Number of instances to launch initially.
+     *
+     * @default 1
+     */
+    readonly initialInstanceCount?: number;
```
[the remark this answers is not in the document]
petermeansrock: I confirmed the following with a handful of CloudFormation endpoint deployments:
1. The default behavior for an `AWS::SageMaker::Endpoint` resource is that, when updating it to use a new `AWS::SageMaker::EndpointConfig` resource (as would happen when changing the `InitialInstanceCount` property for a production variant), the desired instance count on the endpoint is updated to the `InitialInstanceCount` from the newly updated EndpointConfig.
2. This default behavior can be overridden by setting [`RetainAllVariantProperties`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sagemaker-endpointconfig-productionvariant.html#cfn-sagemaker-endpoint-retainallvariantproperties) to `true` on the `AWS::SageMaker::Endpoint` resource. The values that will be retained from the current endpoint include the desired instance count, desired weight, and data capture config for the endpoint.
3. The retainment of variant properties can be further customized to disable retainment for individual properties by specifying them under the [`ExcludeRetainedVariantProperties`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sagemaker-endpointconfig-productionvariant.html#cfn-sagemaker-endpoint-excluderetainedvariantproperties) setting on the `AWS::SageMaker::Endpoint` resource.

As I'm bring this RFC up-to-date with the current state of the world, I'll look to incorporate these retainment features into the `Endpoint` construct.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: @kaizencc I've familiarized myself with asynchronous inference and serverless inference (the two feature launches with the biggest impact on the proposed APIs in this RFC) by deploying a few sample endpoints via CloudFormation and found a number of feature disparities that I was hoping to get some design guidance on questions related to compile-time, synthesis-time, and deploy-time detection of issues. Particularly, strong compile-time seem like a good idea in practice, but in reality, I would hope that some of these feature gaps disappear. To avoid that future evolution from preventing convergence of the API contracts, using synthesis-time validation seems ideal, but then again, I have no guarantee that these products will actually converge. Anyway, to go into more detail:

1. **Lack of feature parity across inference products**: The [serverless inference documentation](https://docs.aws.amazon.com/sagemaker/latest/dg/serverless-endpoints.html) states that the following features are absent from SageMaker's serverless feature:
    > Some of the features currently available for SageMaker Real-time Inference are not supported for Serverless Inference, including GPUs, AWS marketplace model packages, private Docker registries, Multi-Model Endpoints, VPC configuration, network isolation, data capture, multiple production variants, Model Monitor, and inference pipelines.

    Although I haven't found corresponding documentation, asynchronous inference has similar limitations: you cannot use multiple production variants nor can the single supported variant be serverless (it must used instance-based hosting). This led me down two separate paths:
   1. **Asynchronous Endpoints:** Whether or not an endpoint should be treated as sync vs async is controlled cross-variant at the level of the `EndpointConfig` resource. When an `EndpointConfig` resource is configured as async, it is currently only capable of supporting a single production variant. Should I use a compile-time solution (e.g. two different compile-time `*Props` interfaces to differentiate between the supported features of sync vs async) or instead simply use synthesis-time validation to identify a misconfigured resource (e.g., an async endpoint config with multiple variants)?
    1. **Serverless Production Variants:** Similarly, even though the instance-vs-serverless specification is per variant, the presence of one serverless production variant prevents customers from using additional variants (i.e., an aspect of one variant influences the behavior of the entire EndpointConfig resource). Would a synthesis-time check be sufficient here to error out on the incompatible variants or would a compile-time contract guarding this scenario be preferred?
1. **Feature-specific metrics:** Asynchronous endpoints have unique metrics like `ApproximateBacklogSize` on the endpoint name dimension whereas serverless variants have unique metrics like `ModelSetupTime` on the endpoint name + production variant dimension. When adding `metric*` helper APIs for these scenarios, should they be implemented in such a way to guarantee that the metric is available for the specific use-case from a compile-time perspective (i.e., each disparate use-case has an appropriate interface with only its suitable methods) or is it sufficient to use synthesis-time validation in this case?

kaizencc: Hi @petermeansrock! These are all great questions, and thank you for doing the research on this.

`Asynchronous Endpoints`: I don't have a great answer. Sounds like `AsyncEndpointConfig` would be nice, but in the future, if `Async` supported multiple variants, suddenly`AsyncEndpointConfig` would be no different from `SyncEndpointConfig` and thus be redundant. Given this, I think it's safer to stay with one `EndpointConfig` construct with synth-time checks to make sure we catch misconfigured resources prior to deploy. They are not so different that having a generic `EndpointConfig` makes it awkward, so I think we're okay with keeping the two together.

`Serverless Production Variants`: Let me make sure I'm understanding right -- for the `productionVariant` property, there can be one or more serverless variants OR one or more instance variants. Just cannot have a mix of serverless and instance variants. In that case, I like the idea of having 2 properties, `serverlessProductionVariants` and `instanceProductionVariants` that are mutually exclusive (documented as such and validated at synth time).

`Feature-specific metrics`: I think the RFC should not worry about specific `XxxMetric` APIs and rather just expose the generic `metric` API and allow the user to specify exactly what metrics they want to use. This item can be a fast-follow item but isn't required for an alpha release, and there's already a lot of ideas to implement in this RFC. To answer your question though, I think synth-time validation makes sense / is the best we can do.

Feel free to override any of these decisions if you have a compelling reason -- just let me know why so I can be convinced :).

petermeansrock: @kaizencc 

> Given this, I think it's safer to stay with one `EndpointConfig` construct with synth-time checks to make sure we catch misconfigured resources prior to deploy. They are not so different that having a generic `EndpointConfig` makes it awkward, so I think we're okay with keeping the two together.

Sounds good, that makes perfect sense to me.

> `Serverless Production Variants`: Let me make sure I'm understanding right -- for the `productionVariant` property, there can be one or more serverless variants OR one or more instance variants. Just cannot have a mix of serverless and instance variants.

No, it's unfortunately even more limited:

1. If your `EndpointConfig` is asynchronous, you can have a single instance-based production variant.
2. For a synchronous `EndpointConfig`, you can either:
    1. Have a _single_ serverless production variant **OR**
    2. Have one or more instance-based production variants

In other words, multiple production variants are _only_ supported today for a synchronous `EndpointConfig` using instance-based hosting.

> `Feature-specific metrics`: I think the RFC should not worry about specific `XxxMetric` APIs and rather just expose the generic `metric` API and allow the user to specify exactly what metrics they want to use. This item can be a fast-follow item but isn't required for an alpha release, and there's already a lot of ideas to implement in this RFC. **To answer your question though, I think synth-time validation makes sense / is the best we can do.**

Sounds good!

kaizencc: Thanks for clarifying. This does seem confusing to navigate, so its a great call out. The CDK can and should do the best it can to make things simpler for its users. The clarification has given me a change of heart, and now I'm interested in the following idea: two constructs, one generic `EndpointConfig`, and one "L2.5" `AsyncEndpointConfig` that provides further restrictions on the `productionVariant` property. My thinking is that users can drop down to use `EndpointConfig` for both sync and async endpoints if they want to, but we still have provided a higher level abstraction for async endpoints as well.

What do you think?

petermeansrock: @kaizencc 

> The clarification has given me a change of heart, and now I'm interested in the following idea: two constructs, one generic `EndpointConfig`, and one "L2.5" `AsyncEndpointConfig` that provides further restrictions on the productionVariant property.

## L2.5 `AsyncEndpointConfig` construct

For should such an `AsyncEndpointConfig`, which of the following would apply?
1. **`AsyncEndpointConfig extends Resource` and instantiates an `EndpointConfig` construct**: This is the classic L2 strategy, but would (I believe) introduce an intermediate async construct in the hierarchy, meaning customers couldn't replace an `AsyncEndpointConfig` with an `EndpointConfig` without changing their CloudFormation logical IDs.
1. **`AsyncEndpointConfig extends EndpointConfig`**: I assume this would retain the identical construct hierarchy but would violate the following [documented CDK design guideline](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#construct-class):
    > As a rule of thumb, most constructs should directly extend the Construct or Resource instead of another construct. Prefer representing polymorphic behavior through interfaces and not through inheritance.
1. I'm assuming there could be another scenario where `AsyncEndpointConfig` instantiates `EndpointConfig` with a customer-provided scope and ID without extending `Construct`/`Resource` to avoid impacting the hierarchy, but I'm not familiar with this strategy being employed in the CDK codebase.

## Serverless Production Variants

The other limitation called out above, for synchronous endpoints, is that the presence of a single serverless variant prevents other variants from being configured. Would a synthesis-time approach be appropriate here (i.e., taking `ProductionVariant[]` on the `EndpointConfigProps` but erroring out if multiple are present alongside a serverless variant) or would you prefer a model more similar to your serverless-vs-instance specific properties?

kaizencc: Thanks for reading the design guidelines! You are one of the few :). 

We don't have to implement `AsyncEndpointConfig` day 1 of the library. Lets add it to an "extensions" section and then turn those extensions into github issues later. 

I think this may be a special case where we are okay with extending `EndpointConfig`. We generally don't want to support that many L2.5s in the library, which is why we document that we want to extend the resource, but this could be a good candidate. I could be wrong, and be forced to change my mind later, but I'm looking at `NodeJsFunction` and other L2.5s in the main repo and they extend another construct. At any rate, this doesn't need to be implemented along with `EndpointConfig` so we don't need to finalize how to instantiate it just yet.

---

How about mutually exclusive properties `serverlessProductionVariant: ServerlessProductionVariant` and `productionVariants: ProductionVariant[]`? That seems like a good mix of property differentiation and synth time validation (that only one of the two props are set).

petermeansrock: @kaizencc 

> We don't have to implement `AsyncEndpointConfig` day 1 of the library. Lets add it to an "extensions" section and then turn those extensions into github issues later.

Sounds good! I assume the serverless pieces can be added to the same section.

> How about mutually exclusive properties `serverlessProductionVariant: ServerlessProductionVariant` and `productionVariants: ProductionVariant[]`? That seems like a good mix of property differentiation and synth time validation (that only one of the two props are set).

Two quick follow-up questions:

1. As both types of variants share a variant name & weight (and ideally could be used as siblings if the SageMaker folks release support for multi-heterogenous variants per endpoint config), to adjust your proposal slightly, would it make sense to define a shared interface like `ProductionVariant` and then have two interfaces extending it: `ServerlessProductionVariant` and `InstanceProductionVariant`? We could simply not export `ProductionVariant` yet as we don't have a need for polymorphic representation (e.g., `ProductionVariant[]`) and that way, both exported interfaces are specifically targeting their use-case (e.g., serverless vs instance-based).
2. Challenging one statement I made above: if the SageMaker folks later do support heterogeneous variants, would it make sense to collapse both variant specifications into a single `ProductionVariant[]` on the `EndpointConfigProps` (i.e., is that polymorphic API `jsii`-compatible)? If so (and assuming we can foresee SageMaker's product evolution), would it be unusual to present two separate variant specifications now if we _know_ they will eventually be "_doomed_" APIs (i.e., ideal for removal during a major version bump)?

***

Thanks so much for the help, Kaizen; this conversation has been immensely helpful! I'll start incorporating your recommended changes as my final pair of questions are mostly focused on minor remaining details that can be adjusted rather easily.

kaizencc: Answer for 1 is yes. 

Answer for 2 is hmmm. One the one hand, predicting what SageMaker is going to do in the future is future-proofing and we want to avoid doing that. We have no guarantee that that will ever happen. If it happens when SageMaker is in alpha, then we can make that breaking change. If it happens after SageMaker is stabilized, then we can deprecate `serverlessProductionVariant` and `instanceProductionVariants` and introduce a new `productionVariant` property. Or, we can keep both APIs and just remove/modify the synth-time validation -- they're not necessarily _doomed_ and it might be fine to specify different production variants in different properties. I think if that happens it's okay, we're not stuck, and any changes we have to make won't really hurt the API too much.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+  loaded from an S3 location
+
+  ```ts
+  export abstract class ModelData {
+    /**
+     * Constructs model data which is already available within S3.
+     * @param bucket The S3 bucket within which the model artifacts are stored
+     * @param objectKey The S3 object key at which the model artifacts are stored
+     */
+    public static fromBucket(bucket: s3.IBucket, objectKey: string): ModelData { ... }
```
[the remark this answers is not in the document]
petermeansrock: As far as I can tell (from reading documentation & experimenting with SageMaker operations), SageMaker does _not_ support versioned objects as a source of model data. I've also [posted a question to repost.aws](https://repost.aws/questions/QUmyE-SweqTvazF0y_5OJtpA/does-sage-makers-create-model-api-support-model-data-from-versioned-s-3-objects) to confirm, but I haven't gotten any responses.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: FYI: As I had already expanded the wording a bit more (to provide a bit more ML related context), the version in the next revision will be slightly longer.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+```typescript
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+```
+
+## Model
+
+By creating a model, you tell Amazon SageMaker where it can find the model components. This includes
+the S3 path where the model artifacts are stored and the Docker registry path for the image that
+contains the inference code. The `ContainerDefinition` interface encapsulates both the specification
```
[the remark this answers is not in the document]
petermeansrock: In rev2, I've moved up the `Model` examples prior to diving into container image and model data assets.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+
+Reference a local directory containing a Dockerfile:
+
+```typescript
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+import * as path from 'path';
+
+const image = sagemaker.ContainerImage.fromAsset(this, 'Image', {
+  directory: path.join('path', 'to', 'Dockerfile', 'directory')
+});
```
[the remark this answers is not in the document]
petermeansrock: In rev2, I've pivoted to having the new proposed APIs take an `Asset` from '@aws-cdk/aws-s3-assets' (for `ModelData`) and `DockerImageAsset` from '@aws-cdk/aws-ecr-assets' (for `ContainerImage`) to avoid adding any assets into the hierarchy myself.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: In rev2, I've mentioned that an `EndpointConfig` can be used to provision one or more `Endpoint` resources.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: In rev2, I've adjusted both references to `IEndpointProductionVariant` in the README accordingly.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
petermeansrock: In rev2, I've decided to pivot to having the new proposed APIs take an `Asset` from '@aws-cdk/aws-s3-assets' (for `ModelData`) and `DockerImageAsset` from '@aws-cdk/aws-ecr-assets' (for `ContainerImage`) to avoid adding any assets into the hierarchy myself.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1150 @@
...
+     */
+    public static fromBucket(bucket: s3.IBucket, objectKey: string): ModelData { ... }
+
+    /**
+     * Constructs model data that will be uploaded to S3 as part of the CDK app deployment.
+     * @param scope The scope within which to create a new asset
+     * @param id The id to associate with the new asset
+     * @param path The local path to a model artifact file as a gzipped tar file
+     */
+    public static fromAsset(scope: Construct, id: string, path: string): ModelData { ... }
```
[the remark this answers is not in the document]
petermeansrock: Noted above but copying here as well:

> In rev2, I've pivoted to having the new proposed APIs take an `Asset` from '@aws-cdk/aws-s3-assets' (for `ModelData`) and `DockerImageAsset` from '@aws-cdk/aws-ecr-assets' (for `ContainerImage`) to avoid adding any assets into the hierarchy myself.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
petermeansrock: In rev2, I've called these features out explicitly in the "_Are there any open issues that need to be addressed later?_" section.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
petermeansrock: In rev2, I've reworded this bit and moved it into the "_Are there any open issues that need to be addressed later?_" section.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: In rev2, I now mention reuse of `EndpointConfig` across `Endpoint` resources in the README.

petermeansrock (commented):
Responding to a few comments about the direction that was taken in authoring revision 2.

text/0431-sagemaker-l2-endpoint.md
petermeansrock: > `Feature-specific metrics`: I think the RFC should not worry about specific `XxxMetric` APIs and rather just expose the generic metric API and allow the user to specify exactly what metrics they want to use. This item can be a fast-follow item but isn't required for an alpha release, and there's already a lot of ideas to implement in this RFC. To answer your question though, I think synth-time validation makes sense / is the best we can do.

@kaizencc 

As I just published rev2, I wanted to highlight one bit: I kept the previously proposed `metric*` APIs on the newly renamed `IEndpointInstanceProductionVariant` as they had already been implemented and would be easy enough to remove if you feel strongly that they shouldn't be included.

text/0431-sagemaker-l2-endpoint.md
```diff
@@ -0,0 +1,1371 @@
...
+```typescript
+import * as ecr_assets from '@aws-cdk/aws-ecr-assets';
+import * as s3_assets from '@aws-cdk/aws-s3-assets';
+import * as sagemaker from '@aws-cdk/aws-sagemaker';
+import * as path from 'path';
+
+const imageAsset = new ecr_assets.DockerImageAsset(this, 'Image', {
+  directory: path.join('path', 'to', 'Dockerfile', 'directory')
+});
+const image = sagemaker.ContainerImage.fromAsset(imageAsset);
```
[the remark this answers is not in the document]
kaizencc: same goes for the below `sagemaker.ModelData.fromAsset`

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
kaizencc: Commented elsewhere but I think the correct API here is to create the asset for the user (and let them supply only a path to a directory, and optional options.

kaizencc (commented):
Hi @petermeansrock, I'm pretty happy with this RFC modulo some small comments. And we may get further involvement from the sagemaker team as well. 

I also didn't really look at the code in this RFC. i'll review that on the actual PR when the time comes, but I'm much less worried about the actual implementation. i focused primarily on the readme and the other parts of the FAQ.

text/0431-sagemaker-l2-endpoint.md
kaizencc: I think this should take an `IEndpointConfig` as is the traditional CDK pattern. 

the implementation for `public get instanceProductionVariants(): IEndpointInstanceProductionVariant[] {` should take this into consideration and throw an error if the EndpointConfig is imported. 

This seems reasonable to me - you can create an `Endpoint` from an imported `EndpointConfig`, but you can't actually reference anything from within the config. Otherwise, there is no point in differentiating an `Endpoint` from an `EndpointConfig`, because users will be importing configs from elsewhere to reuse the config construct.

text/0431-sagemaker-l2-endpoint.md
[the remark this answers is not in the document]
TheRealAmazonKendra: Yeah, I was still new and too dumb to know this.

kaizencc (approved):
Yay @petermeansrock! We're calling this approved :).

PR #449, opened 2022-07-27
RFC309: weak cross-stack references

rix0rrr:
This is a request for comments about Cross stack References. See #309 for
additional details. 

APIs are signed off by @Naumel.

---

### DIFF
--- a/text/0309-weak-cross-stack-references.md
+++ b/text/0309-weak-cross-stack-references.md
@@ -2,7 +2,7 @@
 
 * **Original Author(s):**: @rix0rrr
 * **Tracking Issue**: #309
-* **API Bar Raiser**: @otaviomacedo
+* **API Bar Raiser**: @Naumel
 
 A mechanism to make sure that CDK doesn't prevent resource replacement if there is no impact on downstream stacks.
 
@@ -18,7 +18,7 @@ We introduce a mechanism for turning strong references into weak references, and
 
 ### Cross-stack references
 
-When CDK detects that you are using resources define in one Stack in another Stack, it will automatically add a dependency between the two stacks, and generate the declarations necessary in both stacks to make sure the value gets transported between them (this feature only works for Stacks in the same account and region--if stacks are in either a different account or different region, the reference will lead to a synthesis error).
+When CDK detects that you are using resources defined in one Stack in another Stack, it will automatically add a dependency between the two stacks, and generate the declarations necessary in both stacks to make sure the value gets transported between them (this feature only works for Stacks in the same account and region--if stacks are in either a different account or different region, the reference will lead to a synthesis error).
 
 By default all references are **strong references**, which introduce some limitations on the resources being referenced:
 
@@ -54,7 +54,9 @@ const fn = new lambda.Function(this, 'MyFunction', {
 });
 ```
 
-`Reference.weakStringReference()` only has an effect if the reference is to a construct defined in a different stack: it will use SSM Parameter Store to transport the value between stacks. If the reference is to a construct in the same stack, nothing happens (do be aware that in the example above, if `Bucket` gets renamed, there is a certain time window in which the Lambda Function may continue to be executed with the old bucket name!).
+`Reference.weakStringReference()` only applies if the referenced construct is defined in a different stack: it will use SSM Parameter Store to transport the value between stacks. If the reference is to a construct in the same stack, nothing happens.
+
+> N.B: in the example above, if the reference is weak and `Bucket` gets renamed there is a certain time window in which the Lambda Function may continue to be executed with the old bucket name!
 
 If the value passed to `Reference.weakStringReference()` contains multiple references (either because it's a string with multiple references or it's an array or complex data structure), all of them are made weak.
 
--- a/text/0309-weak-cross-stack-references.md
+++ b/text/0309-weak-cross-stack-references.md
@@ -1,0 +1,194 @@
+# Weak cross-stack references
+
+* **Original Author(s):**: @rix0rrr
+* **Tracking Issue**: #309
+* **API Bar Raiser**: @otaviomacedo
+
+A mechanism to make sure that CDK doesn't prevent resource replacement if there is no impact on downstream stacks.
+
+## Motivation
+
+One of the most magical (and hated) features of the CDK is to transparently transport values from one CloudFormation stack to the other. It makes it possible to pass constructs from one stack and pass them into another, and use the construct as if it was defined locally. CDK will detect that a value originating from one stack is used in a different one, and insert the right CloudFormation incantations to make sure the value is properly referenced.
+
+However, CDK has chosen a mechanism to do this (called **CloudFormation Exports**) which have a specific side effect: *values which are currently being consumed by a different Stack are not allowed to be changed or removed*. This RFC introduces "weak cross-stack references", a way to make certain cross-stack references selectively get moved via SSM Parameters instead of by CloudFormation Exports.
+
+We introduce a mechanism for turning strong references into weak references, and turn this on by default for CloudWatch Dashboards.
+
+## README
+
+### Cross-stack references
+
+When CDK detects that you are using resources define in one Stack in another Stack, it will automatically add a dependency between the two stacks, and generate the declarations necessary in both stacks to make sure the value gets transported between them (this feature only works for Stacks in the same account and region--if stacks are in either a different account or different region, the reference will lead to a synthesis error).
+
+By default all references are **strong references**, which introduce some limitations on the resources being referenced:
+
+- The resource that is being referenced cannot be deleted.
+- The resource that is being referenced cannot have its location in the construct tree change.
+- The resource that is being referenced cannot have any properties changed on it that requires *replacement* of the resource.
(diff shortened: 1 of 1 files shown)


text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
+# Weak cross-stack references
+
+* **Original Author(s):**: @rix0rrr
+* **Tracking Issue**: #309
+* **API Bar Raiser**: @otaviomacedo
```
Naumel, line 5: I think he's out today :-)

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+
+A mechanism to make sure that CDK doesn't prevent resource replacement if there is no impact on downstream stacks.
+
+## Motivation
+
+One of the most magical (and hated) features of the CDK is to transparently transport values from one CloudFormation stack to the other. It makes it possible to pass constructs from one stack and pass them into another, and use the construct as if it was defined locally. CDK will detect that a value originating from one stack is used in a different one, and insert the right CloudFormation incantations to make sure the value is properly referenced.
+
+However, CDK has chosen a mechanism to do this (called **CloudFormation Exports**) which have a specific side effect: *values which are currently being consumed by a different Stack are not allowed to be changed or removed*. This RFC introduces "weak cross-stack references", a way to make certain cross-stack references selectively get moved via SSM Parameters instead of by CloudFormation Exports.
+
+We introduce a mechanism for turning strong references into weak references, and turn this on by default for CloudWatch Dashboards.
```
Naumel, line 15: >turn this on by default for CloudWatch Dashboards.

Default only for it, am I reading this correctly?

Naumel: Yes, explained as an use-case in a subsequent section.

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+const fn = new lambda.Function(this, 'MyFunction', {
+  // ...
+
+  environment: {
+    BUCKET_NAME: Reference.weakStringReference(bucket.bucketName),
+  },
+});
+```
+
+`Reference.weakStringReference()` only has an effect if the reference is to a construct defined in a different stack: it will use SSM Parameter Store to transport the value between stacks. If the reference is to a construct in the same stack, nothing happens (do be aware that in the example above, if `Bucket` gets renamed, there is a certain time window in which the Lambda Function may continue to be executed with the old bucket name!).
```
Naumel, line 57: I'm having a rough time reading 

> only has an effect if the reference is to a construct defined in a different stack

mind rewording it maybe like

> the weak reference works only if its construct is defined in a different stack

?

Naumel: Should the time-window of doom warning get its own section with a couple of examples or do you think it's better to repeat it throughout the doc?

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+After you have deployed your stack with `strongToWeak` enabled, you can remove the flag again and do another deployment:
+
+```ts
+// Second deployment, stop generating strong export
+Reference.weakStringReference(bucket.bucketName)
+```
+
+### Defaults for stacks
+
+You can choose to configure the default reference style for an entire stack passing the `defaultReferenceStyle` property to a stack constructor:
```
Naumel, line 83: What happens if a team is silly and one change passes a stack-style and the other change passes the style per reference, what's the precedence order -- the more granular, per-reference one?

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+});
+```
+
+It is recommended to select reference style on a reference-by-reference basis, and not use this method, but you can if you want to.
+
+### Defaults for dashboards
+
+Strong references cause a lot of usability problems for CloudWatch Dashboards. That's why starting in new applications, all CloudWatch Dashboards will use weak references by default.
+
+This is controlled by the feature flag `@aws-cdk/aws-cloudwatch:dashboardWeakReferences`. It can take on the values `true` (for weak references) or `"strongToWeak"`, for the first phase of the two-phase deployment.
```
Naumel, line 100: What case can be made that **not** switching to weak references in CloudWatch Dashboards is the right thing to do?

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+
+```ts
+class Bucket {
+  public export(options?: ExportOptions): IBucket;
+}
+```
+
+And use either `Exports` or SSM parameters to reference itself.
+
+The explicitness puts users in full control, able to generate exports in either or both styles, even without a consumer.
```
Naumel, line 184: This makes it more generic when compared to the feature flag, right?

text/0309-weak-cross-stack-references.md
rix0rrr: Yes.

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,194 @@
...
+First, pass `strongToWeak: true` for the reference you are converting from strong to weak. This will generate both the strong and weak exports in the producing stack, while only consuming the weak reference in the consuming stack:
+
+```ts
+// First deployment, generate both exports
+Reference.weakStringReference(bucket.bucketName, {
+  strongToWeak: true,
+})
+```
+
+After you have deployed your stack with `strongToWeak` enabled, you can remove the flag again and do another deployment:
```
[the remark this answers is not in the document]
rix0rrr: Just duplication, not even cost.

Just not as clean.

text/0309-weak-cross-stack-references.md
rix0rrr: Yes.

text/0309-weak-cross-stack-references.md
rix0rrr: I don't think there is any. Strong references only make sense if the consumer would suffer availability impact if the producer disappears. This is definitely not the case for dashboards: the line will just have a gap in it.

Note: explicitly *not* saying this for *Alarms*: they may have impact if the producer disappears, depending on whether the `MissingDataPoints` property is set to `BREACHING` or `IGNORE`.

text/0309-weak-cross-stack-references.md
rix0rrr: I don't know what generic means in this context. But it sure makes it more *explicit*, at the cost of:

* having to type more
* having to be aware when you are passing constructs across stacks (might not be obvious when writing higher-level code) 
* losing the ability to make decisions on a per-consumption basis (person passing into the consumer stack makes the choice for the entire stack)
* losing the ability to change the producing construct from the consumer stack

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,196 @@
...
+
+* **Original Author(s):**: @rix0rrr
+* **Tracking Issue**: #309
+* **API Bar Raiser**: @Naumel
+
+A mechanism to make sure that CDK doesn't prevent resource replacement if there is no impact on downstream stacks.
+
+## Motivation
+
+One of the most magical (and hated) features of the CDK is to transparently transport values from one CloudFormation stack to the other. It makes it possible to pass constructs from one stack and pass them into another, and use the construct as if it was defined locally. CDK will detect that a value originating from one stack is used in a different one, and insert the right CloudFormation incantations to make sure the value is properly referenced.
```
comcalvi, line 11: minor: typo

```suggestion
One of the most magical (and hated) features of the CDK is to transparently transport values from one CloudFormation stack to the other. It makes it possible to pass constructs from one stack into another, and use the construct as if it was defined locally. CDK will detect that a value originating from one stack is used in a different one, and insert the right CloudFormation incantations to make sure the value is properly referenced.
```

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,196 @@
...
+After you have deployed your stack with `strongToWeak` enabled, you can remove the flag again and do another deployment:
+
+```ts
+// Second deployment, stop generating strong export
+Reference.weakStringReference(bucket.bucketName)
+```
+
+### Defaults for stacks
+
+You can choose to configure the default reference style for an entire stack passing the `defaultReferenceStyle` property to a stack constructor:
```
comcalvi, line 85:
```suggestion
You can choose to configure the default reference style for an entire stack passing the `defaultReferenceStyle` (defaults to `ReferenceStyle.STRONG`) property to a stack constructor:
```

text/0309-weak-cross-stack-references.md
```diff
@@ -0,0 +1,196 @@
...
+### Why are we doing this?
+
+It comes up frequently that users are stuck trying to remove or refactor a resource, only to find out that they can't because the resource is referenced in another stack. This takes a lot of work to disentangle. This problem is so common that it has been named the "deadly embrace" problem (mostly occurs in Pipelines, where users are unable to play with the deployment order of stacks).
+
+A better mechanism to set up references, combined with sensible defaults for resources that are not impacted by such changes like
+Dashboards, should make this problem less frequent.
+
+### Why should we _not_ do this?
+
+It might not be worth the additional complexity in the token and reference system (which is already pretty complex).
```
comcalvi, line 143: Could we elaborate on this?

Issue #201, opened 2020-07-16
Construct scope relocation


eladb:
|PR|Champion|
|--|--------|
|# |    @RomainMuller     |

## Description

To allow preserving paths and generated identities when refactoring construct trees, we propose to introduce `construct.node.relocate(path)` which will allow specifying the root path for a scope and all its children.

For example, if I have a set of stacks in production which I want to bundle into a new `Stage` type so that I can deploy them through CDK Pipelines and deploy them without replacement of resources, I would use `relocate()` to ensure that all paths and unique IDs are preserved as I refactor my code.

## Progress

- [x] Tracking Issue Created
- [ ] RFC PR Created 
- [ ] Core Team Member Assigned 
- [ ] Initial Approval / Final Comment Period
- [ ] Ready For Implementation
    
    - [ ] implementation issue 1
- [ ] Resolved

mrgrain:
Marking this RFCs as `stale` since there has been little recent activity and it is not currently close to getting accepted as-is. We appreciate the effort that has gone into this proposal. Marking an RFCs as stale is not a one-way door. If you have made substantial changes to the proposal, please open a new issue/RFC. You might also consider raising a PR to [aws/aws-cdk](https://github.com/aws/aws-cdk/) directly or self-publishing to Construct Hub.

PR #455, opened 2022-09-05
RFC 201: Refactoring support

rix0rrr:
This is a request for comments about refactoring support. See #201 for
additional details. 

APIs are signed off by @RomainMuller.

[Rendered Link](https://github.com/aws/aws-cdk-rfcs/blob/huijbers/0201-refactoring-support/text/0201-refactoring-support.md)

---

### DIFF
--- a/text/0201-refactoring-support.md
+++ b/text/0201-refactoring-support.md
@@ -1,0 +1,193 @@
+# Refactoring Support
+
+* **Original Author(s):**: @rix0rrr
+* **Tracking Issue**: #201
+* **API Bar Raiser**: @RomainMailler
+
+You can now refactor the shape of your construct tree without incurring resource replacement in CloudFormation. To use refactoring support, you must be on `constructs >= 10.x.y`.
+
+## README
+
+When the CDK needs to allocate unique identifiers for your constructs (such as allocating *logical id*s in CloudFormation templates), it will use the full construct path of your construct to base the identifier on. The result of this is that if you refactor any part of your construct tree (for example, if you change a construct identifier, or extract a resource to new construct, thereby introducing a new scope), the generated identifiers will change. This will lead CloudFormation to think a new resource has been introduced and the old one has been removed.
+
+To prevent this from happening, whenever you refactor part of your construct tree you should record the refactoring by calling specific methods indicating what the change was you performed. Let's run through an example.
+
+### Example: extracting a new construct
+
+```ts
+class MyStack extends Stack {
+  constructor(scope: Scope, id: string, props: MyStackProps) {
+    super(scope, id, props);
+
+    new Bucket(this, 'MyBucket');
+  }
+}
+```
+
+Let's say you introduce a new construct called `MyConstruct` to hold the S3 Bucket (and perhaps a couple of other resources). This will make the construct path to the bucket change from `Stack/MyBucket` to `Stack/MyConstruct/MyBucket`, and hence make it allocate a different ID which will lead to CloudFormation trying to delete the Bucket.
+
+When you perform the refactoring, add a call to `this.node.refactor()`:
+
+```ts
+// New extracted construct
+class MyConstruct extends Construct {
+  constructor(scope: Scope, id: string) {
+    super(scope, id);
+
+    new Bucket(this, 'MyBucket');
+  }
+}
+
+// New version of stack construct
+class MyStack extends Stack {
+  constructor(scope: Scope, id: string, props: MyStackProps) {
+    super(scope, id, props);
+
+    new MyConstruct(this, 'MyConstruct');
+    this.node.refactor(this, 'MyBucket', 'MyConstruct/MyBucket');
+  }
+}
+```
+
+Using this information, CDK will keep on generating the same unique identifiers as from before the refactoring.
+
+Put the refactoring calls in the closest where you removed or renamed the resource.
+
+### Supported refactorings
+
+The following refactorings are supported:
+
+* Renaming a construct.
+* Moving a construct from the current scope into a deeper scope, or vice versa.
+* Moving a construct between two deeper scopes.
+
+It is not possible to move a construct into a bigger scope, or reference a construct from a bigger scope.
+
+It is also not possible to refactor across Stack boundaries. Refactoring only affects the unique identifiers generated for constructs, it does not actually move the construct to a new parent.
+
+To encode a literal `/` into a construct id, double it (`//`).
+
+> Pre-emptive FAQ:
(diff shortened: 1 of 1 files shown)


text/0201-refactoring-support.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Supported refactorings
+
+The following refactorings are supported:
+
+* Renaming a construct.
+* Moving a construct from the current scope into a deeper scope, or vice versa.
+* Moving a construct between two deeper scopes.
+
+It is not possible to move a construct into a bigger scope, or reference a construct from a bigger scope.
```
straygar, line 64: Any particular reason this is the case?

text/0201-refactoring-support.md
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+// New version of stack construct
+class MyStack extends Stack {
+  constructor(scope: Scope, id: string, props: MyStackProps) {
+    super(scope, id, props);
+
+    new MyConstruct(this, 'MyConstruct');
+    this.node.refactor(this, 'MyBucket', 'MyConstruct/MyBucket');
```
eladb, line 47: API thoughts:

* The term "refactor" doesn't necessarily mean "rename". Perhaps be a bit more explicit about the name of this method. 
* It might be useful to *also* support directly referencing the target object (less potential for mistakes based on string matching)
* Why do we need `this` as the first argument if we are already in the `node`?

```ts
this.node.moveToPath("MyBucket", "MyConstruct/MyBucket");
this.node.moveTo("MyBucket", myConstruct.myBucket);
```

text/0201-refactoring-support.md
```diff
@@ -0,0 +1,193 @@
...
+
+- The "real" construct path
+- The "identifier" construct path
+
+All methods that calculate identifiers will be rewritten to use the second root path.
+
+We will add misuse protection. It is important that when a construct is "virtually" moved into the location of another, that that location is not reused by a meaningful construct, as their identifiers would start to collide. In the example given above:
+
+```ts
+this.node.refactor(this, 'MyBucket', 'MyConstruct/MyBucket');
```
Chriscbr, line 162: Since the API is a method call and not "new"-ing something, it reads to me like an operation, not a declaration, so as a user I'd expect I could call this multiple times. Would something like this be possible?

```ts
this.node.refactor(this, 'MyBucket', 'MyConstructV1/MyBucket');
this.node.refactor(this, 'MyConstructV1/MyBucket', 'MyConstructV2/MyBucket');
```

Chriscbr (commented):
I feel like refactorings in the other direction could also be useful -- for example, if I have an app structured like this, and I wanted to move ChildA to be part of the root (while keeping its old ID at synthesis-time).

```
- root
  - Parent1
    - ChildA
    - ChildB
```

---

side note: Reading the RFC, I wondered if we would gain any expressive power in constructs by taking a page from the world of filesystems and introducing a symlink-like concept. This would be a special construct that redirects to some other path in the construct tree, and would be handled specially during construct tree traversal. Something like:

```ts
new Link(this, "MyBucket", "MyConstruct/MyBucket");
new Link(this, "MyBucket", myConstruct.myBucket);
new Portal(this, "MyBucket", myConstruct.myBucket);
```

But I'm not sure if this API as much sense 🤷

NetaNir (commented):
First, thank you so much for thinking about this problem. As someone who owns several large scale application managed by the CDK, I can attest to how painful hitting this can be, especially how risky it is, as sometimes, the only way out (beside reverting the code) is to delete resources from the console 😱.

The CDK default behavior for allocation logical ids is unsafe, its a good getting starting experience, but it comes at the expense of long term maintenance pain. For that reason I think the solution should be on the getting started path, and not after the user hits this sharp edge (thinking of an emergency deployment, not so fun realizing you cant deploy your app). Having said that, I realize this might not be possible. 

This RFC propose a solution in the `construct` level, but the problem mostly impacts the AWS CDK users, due to how CloudFormation handles logical ids, so maybe it will be easier to solve it on the Stack level.

The "easiest" solution from the framework perspective would be to simply use the user provided id, and let them deal with the consequence of collisions. In case of a collisions, a detailed error message will be thrown, and the user will have to fix it. I agree that this might not be a great getting started experience, but this has the advantage of letting the user make a decision when they build a production app. Working backward from a (imaginary) CDK best practices section:

> ## Things to consider when building a production CDK application
> By default the CDK will allocate logical ids based on the construct path in the construct tree. This means that if you move your construct around the code, it might result in a change to its logical id, that will trigger a resource replacement, if you use "named resources", e.g. `BucketName`, the deployment will fail when trying to create a new resource. If this is indeed an issue for you, you can opt out of this CDK default behavior by setting the context key:
```
construct-id:no-default-generation
```
> This will override the default behavior and use the user provided id (the second arg when initializing a construct). Note that this means that you will have to manage the Ids uniqueness yourself. 

While not perfect it will allow the user to make a concise decision before deploying a full application. 

Also worth noting that there is support for refactoring today, by using the `Default` id on the added layer, from [constructs](https://github.com/aws/constructs/blob/c37850f79500b748727ec828a122230e28f3bc0b/src/private/uniqueid.ts#L3-L13)



```ts
/**
 * Resources with this ID are complete hidden from the logical ID calculation.
 */
const HIDDEN_ID = 'Default';

/**
 * Calculates the construct uid based on path components.
 *
 * Components named `Default` (case sensitive) are excluded from uid calculation
 * to allow tree refactorings.
 *
```

royby-cyberark:
I was just wondering about this and tried to ask in discussions (https://github.com/aws/aws-cdk/discussions/22016)
and I suggested a way of side stepping the entire logical ids issues when refactoring.

My suggestion was to **avoid using construct classes (avoid inheriting Construct) and instead arrange your code with regular classes which takes the scope of the stack/parent construct and use this as the scope for all child resources.**

Pros:
* No logical id changes when refactoring
* get all the value of refactoring and code encapsulation

Cons: 
* You lose the CloudFormation hierarchy which comes out of the box
  * you no longer get the ids of the parent concatenated
  * you can't see the nice hierarchy in CloudFormatoin new Tree view which is really nice (also other tools that might do the same things for visualization)

What do you think about this approach? do you think it goes against the "spirit" of cdk/cloudformation, or does it seem like a legitimate approach? 

Thank you.

humanzz:
How different is this from renaming logical ids? Is the difference that relocating can happen at some intermediate construct level therefore might save multiple logical Id renames?

jonodrew:
Was this ever merged?

Issue #477, opened 2023-02-08
Add policy validation to CDK


otaviomacedo:
## Description

Add a new feature to the AWS CDK, that allows users to validate their CDK applications against externally defined policies. These policies are constraints on the properties and shape of the resources that can be created in an AWS account. For instance, a policy may specify that all S3 buckets should be encrypted using a customer managed KMS key, and that the KMS key, in turn, must have certain policies attached to it.

There are many tools in the market that allow customers to define these policies using a domain specific language (DSL) or a library written in a general purpose programming language. Taking a set of CloudFormation templates and a set of policies, these tools check whether any template violates any policy and report the violations to the user accordingly. This launch allows users to integrate one or more of these tools in the CDK synthesis flow, so that, if their application produces a non-compliant CloudFormation template, they get the feedback immediately.

This feature should be extensible, to allow any user to integrate with their own tools, without depending on changes to the CDK CLI or framework.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @otaviomacedo
| Author(s)           | @otaviomacedo, @corymhall 
| API Bar Raiser      | @iliapolo 
| Stakeholders        | @evgenyka, @udaypant 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [x] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #478, opened 2023-02-08
RFC 477: Policy Validation

otaviomacedo:
This is a request for comments about validation of policies by the CDK at synth time. See https://github.com/aws/aws-cdk-rfcs/issues/477 for additional details. 

APIs are signed off by @iliapolo.

[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/otaviom/rfc-0477/text/0477-policy-validation.md)

---


text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+
+Instead of integrating with each policy tool individually, we have considered the option of developing a single
+mechanism, that would integrate with CloudFormation hooks. The idea was to invoke the Lambda function underlying a
+CloudFormation hook and interpret its results in a similar way CloudFormation, that is, by failing synthesis if the
+function reported a validation failure.
+
+However, the Lambda functions invoked by CloudFormation hooks are not deployed to the customer’s account, but to an
+account owned by AWS. This makes this solution not feasible.
+
+### What are the drawbacks of this solution?
```
iliapolo, line 314: Another drawback I'd like to point out:

- If an organization wants validations to apply to all languages in the company, they have to setup a jsii publishing pipeline. But if the validation is invoked from the CLI, than all that is needed is one node package.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+
+No.
+
+## Appendix
+
+### Appendix A - High level design
+
+![Class diagram for policy validation](https://github.com/aws/aws-cdk-rfcs/blob/otaviom/rfc-0477/images/policy-validation/validation.png?raw=true)
+
+The main abstraction of the model is that of a validation plugin. A plugin encapsulates a piece of logic that includes
```
iliapolo, line 341: Who installs the plugin itself? The framework or the user?

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+
+This behavior can be disabled by providing the `validationMetadata: false` property.
+
+```ts
+new App({
+  validationMetadata: false,
+});
+```
+
+**Example of how CFN Guard rules handle suppression:**
```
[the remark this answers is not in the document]
otaviomacedo: Removed.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+}
+```
+
+The CLI would read this file, load the appropriate plugins either from the filesystem or directly from NPM, and call
+them to get a validation report.
+
+We eventually discarded this option for the following reasons:
+
+* All plugins would have to be written in TypeScript/JavaScript for them to be consumable by the CLI.
+* The CLI lacks contextual information that would be useful for tracing the problem back to where it was defined.
```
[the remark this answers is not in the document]
otaviomacedo: Done.

text/0477-policy-validation.md
iliapolo: Yeap

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+
+* Developer writes a CDK application without the correct validations config.
+* Developer deploys non-compliant stacks.
+* Deployment guardrails catch these violations, and instruct the developer to add a validations property to their
+  application.
+* Developer adds the validations property, and avoids these violations going forward.
+
+Policy as code tools are integrated with CDK through a plugin mechanism so in order to add policy enforcement for a
+specific tool you need to first specify the plugin to use. This can be done by importing the plugin and applying it to a
+scope.
```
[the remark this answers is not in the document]
corymhall: yeah it should be "stage"

text/0477-policy-validation.md
otaviomacedo: That should be the responsibility of the plugin.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+    const bucket = new Bucket(this, 'MyBucket');
+    Validations.of(bucket).addSuppression({
+      rule: 'S3BucketEncryption',
+      reason: 'This bucket does not require encryption because xyz',
+    });
+  }
+}
+```
+
+Each plugin will determine how the suppressions are handled for the specific tool used, but the CDK will also synthesize
```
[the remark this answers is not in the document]
corymhall: I think the plugins have to access suppressions via the template. The rules that are run locally should be the same rules that are eventually run as part of CFN hooks so it needs to be the rules themselves that have the logic for identifying suppressions and not the plugin.

I think the suppressions need to be written to each individual resource. Maybe instead of the metadata we should use tags instead since those will be available in the template as well as the resource itself (and available to Config). 

Since the plugins run after the template is synthesized I don't think we can have the plugin report back anything since it is too late to modify the template at this point.

> And again going back to my previous question, how will plugins know that a specific suppression relates to them and not to some other plugin?

I don't think this matters. If there are multiple plugins that both have the same rule I think a single suppression would be fine.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+const prodStage = new Stage(app, 'ProdStage', {
+  validationPlugins: [...],
+});
+```
+
+#### Alternative
+
+```ts
+const app = new App();
+Validations.of(app).add(new CfnGuardValidator());
```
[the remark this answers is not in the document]
corymhall: Done.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,553 @@
...
+```json
+{
+  "Resources": {
+    "MyBucket": {
+      "Type": "AWS::S3::Bucket",
+      "Properties": {
+        ...
+      },
+      "Metadata": {
+        "guard": {
```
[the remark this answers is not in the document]
corymhall: This is changed so that now we will not write any metadata ourselves. Only the plugin will.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,548 @@
...
+
+The CLI would read this file, load the appropriate plugins either from the filesystem or directly from NPM, and call
+them to get a validation report.
+
+We eventually discarded this option for the following reasons:
+
+* All plugins would have to be written in TypeScript/JavaScript for them to be consumable by the CLI.
+* The framework has access to the construct tree in memory, with all the contextual information that is needed to
+  provide a better error message, such as stack traces. This is useful to point out, for example, in which line 
+  an offending construct was defined. The CLI, on the other hand, only has access to the synthesized template.
```
iliapolo, line 273: Not accurate, the CLI also has access to the `manifest.json` file, which can contain metadata produced by the framework. Its just easier to access this information from the framework, but possible via the CLI as well.

iliapolo (commented):
@otaviomacedo @corymhall I think the major decisions in the RFC are solid. That is:

1. The decision not to use CloudFormation hooks.
2. The decision to invoke validations in the framework.

Can we move on to API discussions?

text/0477-policy-validation.md
```diff
@@ -0,0 +1,544 @@
...
+const prodStage = new Stage(app, 'ProdStage', {
+  validationPlugins: [...],
+});
+```
+
+The details of how to configure the plugin will be specific to each plugin. For example, the `CfnGuardValidator` shown
+above has a `rules` property which is an array of `Rules` objects. The `Rules` class is a helper class that can be used
+to specify where the rules are located.
+
+CDK Policy Validation should never be used as a deployment gate or as the enforcement layer. It should instead be
```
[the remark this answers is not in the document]
corymhall: Yeah we should add some more explanation here.  We're trying to say that you should not rely on this feature in order to prevent the deployment of misconfigured infrastructure. It's fine if it runs as part of a deployment pipeline and fails the pipeline if it is not compliant, but it should be viewed as more of "fail fast(er)" tool rather than a preventative measure. I guess both could be viewed as a type of "gate" though.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,548 @@
...
+    CfnGuardValidator.addSuppression(bucket, {
+      rule: 'S3BucketEncryption',
+      reason: 'This bucket does not require encryption because xyz',
+    });
+  }
+}
+```
+
+Each plugin will determine how the suppressions are handled for the specific tool used, for example a plugin might
+add to the resource metadata.
```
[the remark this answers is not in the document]
corymhall: yeah it would then read it during validation. That way you could have a single guard rule, for example

```
let s3_bucket_public_read_prohibited = Resources.*[ Type == 'AWS::S3::Bucket'
  Metadata.guard.SuppressedRules not exists or
  Metadata.guard.SuppressedRules.* != "S3_BUCKET_PUBLIC_READ_PROHIBITED"
]

rule S3_BUCKET_PUBLIC_READ_PROHIBITED when %s3_bucket_public_read_prohibited !empty {
  %s3_bucket_public_read_prohibited.Properties.PublicAccessBlockConfiguration exists
  %s3_bucket_public_read_prohibited.Properties.PublicAccessBlockConfiguration.BlockPublicAcls == true
  %s3_bucket_public_read_prohibited.Properties.PublicAccessBlockConfiguration.BlockPublicPolicy == true
  %s3_bucket_public_read_prohibited.Properties.PublicAccessBlockConfiguration.IgnorePublicAcls == true
  %s3_bucket_public_read_prohibited.Properties.PublicAccessBlockConfiguration.RestrictPublicBuckets == true
  <<
    Violation: S3 Bucket Public Write Access controls need to be restricted.
    Fix: Set S3 Bucket PublicAccessBlockConfiguration properties for BlockPublicAcls, BlockPublicPolicy, IgnorePublicAcls, RestrictPublicBuckets parameters to true.
  >>
}
```

This rule could be evaluated by running guard as part of a guard plugin _or_ it could be run in a CFN Hook when the cloudformation template is actually deployed. The suppression needs to be persisted in a way that allows it to be evaluated during deployment as well, so rather than create a separate suppression mechanism that only works for CDK plugins, we're trying to create one that works for both.

text/0477-policy-validation.md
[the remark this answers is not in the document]
corymhall: I'm changing this to `an enforcement layer` I don't think it should be thought of as an enforcement layer at all.

text/0477-policy-validation.md
otaviomacedo: Fair enough. Item removed.

text/0477-policy-validation.md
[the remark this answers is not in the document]
corymhall: I've added a note.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+developers and development teams. It's up to them to opt in to this feature, and to ensure that the policy 
+validation is applied. The goal of CDK Policy Validation is to minimize the amount of set up needed during development, and make it as easy
+as possible. For example, a typical workflow could be:
+
+> A note on the term "deployment gate" above. It is perfectly acceptable to have
+> policy at synth run as part of you CI or CD pipeline and prevent deployments.
+> What we want to ensure is that users do not think of that "deployment gate" as
+> a way to _ensure_ that their infrastructure deployment is compliant. It should
+> instead be thought of as a way to "fail fast" and report issues before
+> deployment.
```
rix0rrr, line 72: I know what you're trying to say, but I would say it differently. How about something a long the lines of:

*The validation performed by the CDK at synth time can be bypassed by developers, and can therefore not be relied on as the sole mechanism of validation in large organizations. Some other mechanism to validate the same rule set more authoritatively should be set up independently, like CloudFormation Hooks or AWS Config. Nevertheless, CDK's ability to evaluate the rule set during development is still useful as it will improve detection speed and developer productivity.*

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+
+* Developer writes a CDK application without the correct validations config.
+* Developer deploys non-compliant stacks.
+* Deployment guardrails catch these violations, and instruct the developer to add a validations property to their
+  application.
+* Developer adds the validations property, and avoids these violations going forward.
```
rix0rrr, line 78: I think this is covered? Whaddayatrynasay?

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+             │ Library Version: 2.50.0
+             │ Location: new Bucket (/home/hallcor/tmp/cdk-tmp-app/node_modules/aws-cdk-lib/aws-s3/lib/bucket.js:1:12889)
+             └──  MyCustomL3Construct (MyStack/MyCustomL3Construct)
+                  │ Library: N/A - (Local Construct)
+                  │ Library Version: N/A
+                  │ Location: new MyCustomL3Construct (/home/hallcor/tmp/cdk-tmp-app/src/custom-construct.ts:9:20)
+                  └──  MyStack (MyStack)
+                       │ Library: aws-cdk-lib.Stack
+                       │ Library Version: 2.50.0
+                       │ Location: new MyStack (/home/hallcor/tmp/cdk-tmp-app/src/main.ts:9:20)
```
rix0rrr, line 123: This tree is the wrong way round :) 

We can iterate on this a little. I love the idea but not the specifics of this particular rendering.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+    - Resource Name: my-bucket
+    - Locations:
+      > BucketEncryption/ServerSideEncryptionConfiguration/0/ServerSideEncryptionByDefault/SSEAlgorithm
+
+  Recommendation: Missing value for key `SSEAlgorithm` - must specify `aws:kms`
+  How to fix:
+    > Using override `app.findChild('my-bucket').addPropertyOverride('SSEAlgorithm', 'aws:kms');`
+    > Add to construct properties for `cdk-app/MyStack/Bucket`
+      `encryption: BucketEncryption.KMS`
```
rix0rrr, line 132: I like this, but this needs be propped further up front that there will be hints attached for fixing violations, and that's one of the value adds of this feature!

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+                       │ Library: aws-cdk-lib.Stack
+                       │ Library Version: 2.50.0
+                       │ Location: new MyStack (/home/hallcor/tmp/cdk-tmp-app/src/main.ts:9:20)
+    - Resource Name: my-bucket
+    - Locations:
+      > BucketEncryption/ServerSideEncryptionConfiguration/0/ServerSideEncryptionByDefault/SSEAlgorithm
+
+  Recommendation: Missing value for key `SSEAlgorithm` - must specify `aws:kms`
+  How to fix:
+    > Using override `app.findChild('my-bucket').addPropertyOverride('SSEAlgorithm', 'aws:kms');`
```
rix0rrr, line 130: Oh my God can we generate the correct code here? That'd be neat!

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+  
+Validation failed. See above reports for details
+```
+
+### Adding Exemptions/Suppressions
+
+> Prior art: [cdk-nag](https://github.com/cdklabs/cdk-nag).
+
+There may be cases where you would like to suppress a certain rule. To do so you must specify the rule to suppress and
+the reason for the suppression. Each plugin is responsible for handling it's own
```
rix0rrr, line 143:
```suggestion
the reason for the suppression. Each plugin is responsible for handling its own
```

text/0477-policy-validation.md
```diff
@@ -0,0 +1,561 @@
...
+  }
+}
+```
+
+Each plugin will determine how the suppressions are handled for the specific tool used, for example a plugin might
+add to the resource metadata.
+
+```ts
+class CfnGuardValidator implements IValidatorPlugin {
+  public static addSuppression(scope: Construct, props: SuppressionProps) {
```
rix0rrr, line 181: * `suppress()` ?
* `ignore()` ?

text/0477-policy-validation.md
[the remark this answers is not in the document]
rix0rrr: What you are describing is worse than "the plugin takes care of it" though. This is "the user-written ruleset must take suppressions into account".

I wonder if these aren't in actuality two different features:

* One to actually suppress findings (I think this might be good to have anyway, like how people would like suppressions for our warnings, which we currently don't offer)
* One to smuggle out-of-band data from the CDK app to the rule set, so that the rule set can use it in its rules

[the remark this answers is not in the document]
corymhall: > What you are describing is worse than "the plugin takes care of it" though. This is "the user-written ruleset must take suppressions into account".

I agree it's not great, but what are the alternatives? Using the CFN guard example above, the _only_ option for suppressions is to have the suppression somewhere in the template. Guard doesn't have the ability to read data from somewhere else. So if we wrote some suppression logic into the plugin it would allow the developer to get passed local development, but then it would fail at deployment because the deployment system doesn't know about the suppression you added.

The only other alternative I can think of is to also build a feature into CFN hooks around suppressions. CFN hooks would read suppression info from somewhere (not sure it can be given access to anything other than the template though).

text/0477-policy-validation.md
corymhall: I actually did it this direction on purpose since it shows you the violation resource first, but I'm not tied to it. The `Location` info is also a little weird, but I'm not sure what the best option is for it. For example

```
new MyCustomL3Construct (/home/hallcor/tmp/cdk-tmp-app/src/custom-construct.ts:9:20)
```

This file location is actually the location of `new s3.Bucket`

text/0477-policy-validation.md
corymhall: ooh I like this better

text/0477-policy-validation.md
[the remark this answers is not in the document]
corymhall: After talking to Rico I split this section into two sections, 1 for each of suppressions and exemptions. @rix0rrr let me know what you think.

text/0477-policy-validation.md
corymhall: I've been convinced that the order should go top down. Updated.

text/0477-policy-validation.md
corymhall: went with `suppress`

text/0477-policy-validation.md
```diff
@@ -0,0 +1,621 @@
...
+
+### Plugins
+
+The CDK core framework is responsible for registering and invoking plugins and
+then displaying the formatted validation report. The responsibility of the
+plugin is to act as the translation layer between the CDK framework and the
+policy validation tool. Responsibilities of the plugin may include things like:
+
+- Bundling or installing the policy tool (cfn-guard, opa, etc)
+- Invoking the policy tool, parsing the output, and returning a
```
iliapolo, line 127:
```suggestion
- Bundling or fetching the rules to evaluate. 
- Invoking the policy tool, parsing the output, and returning a
```

text/0477-policy-validation.md
corymhall: I was convinced by the argument that we should go from most familiar to the user to least familiar. At the top they see something they are familiar with and can easily trace down until they see things they are not familiar with. Bottom up they start with something they might not even know exists and then have to dig down the trace until they hit something they recognize.

If we take this example, the user probably did something like

```ts
class MyStack extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);
      new MyCustomL3Construct(this, 'MyCustomL3Construct');
  }
}
```

They would recognize the stack and the L3 construct that they created. They might not even know that the L3 construct creates an S3 bucket. If we show the Bucket trace first they will have to dig down the trace until the reach the L3 construct which they recognize.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,620 @@
...
+  application.
+
+#### Alternative
+
+```ts
+const app = new App();
+Policy.of(app).add(new CfnGuardValidator());
+```
+
+### Validation Report
```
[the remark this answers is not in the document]
otaviomacedo: Yes. Added.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,739 @@
...
+Validation failed. See above reports for details
+```
+
+By default, the report will be printed in a human readable format. If you want a
+report in JSON format, enable it using the `@aws-cdk/core:validationReportJson`
+via the CLI or passing it directly to the application:
+
+```ts
+const app = new App({ 
+  context: { '@aws-cdk/core:validationReportJson': true },
```
iliapolo, line 125: This is a bit strange. Normally JSON output is needed in order for a machine to parse the output, but running `cdk synth` also produces additional stuff that will make the output non parsable. That is, it will contains some lines, then a JSON report, and some more lines? 

I think its going to be hard to make sure the CLI doesn't print anything when this feature is turned on, so I would suggest allowing to pass a file where the JSON report will be written to. Is this possible via context?

text/0477-policy-validation.md
```diff
@@ -0,0 +1,739 @@
...
+  /**
+   * The absolute path of all templates to be processed
+   */
+  readonly templatePaths: string[];
+}
+
+```
+
+The `validate` method returns an instance of `ValidationPluginReport`, which
+tells the CDK whether the template is compliant, which violations were found
```
iliapolo, line 189:
```suggestion
tells the CDK whether the templates are compliant, which violations were found
```

text/0477-policy-validation.md
corymhall: If the report returns anything that means it fails and synth will not write anything to `stdout`. The only thing written to `stdout` will be the report.

text/0477-policy-validation.md
```diff
@@ -0,0 +1,739 @@
...
+(if any), and any metadata about the report. These are the report related 
+interfaces:
```
iliapolo, line 191: nit: this should be in the appendix or design sections, not the working backwards.

text/0477-policy-validation.md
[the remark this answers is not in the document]
corymhall: Just trying to stay consistent and match the feature flags https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.cx_api-readme.html.

iliapolo (approved):

PR #484, opened 2023-02-28
RFC 485: AWS Batch Constructs

comcalvi:
This is a request for comments on a set of Batch L2 Constructs. See #485 for
additional details. 

---

### DIFF
--- a/batch-l2.md
+++ b/batch-l2.md
@@ -783,11 +783,34 @@ interface IManagedEc2ComputeEnvironment extends IManagedComputeEnvironment {
   readonly spotBidPercentage?: number;
   readonly spot?: boolean;
   readonly desiredvCpus?: number;
-  readonly instanceTypes?: ec2.InstanceType[];
+  readonly instanceTypes?: InstanceType[];
   readonly instanceRole?: iam.IRole;
   readonly launchTemplate?: ec2.ILaunchTemplate;
   readonly minvCpus?: number;
   readonly placementGroup?: ec2.IPlacementGroup;
+}
+
+interface IBatchMachineImage {
+  image: ec2.IMachineImage;
+  imageType: BatchMachineImageType;
+  imageKubernetesVersion?: string;
+}
+
+enum BatchMachineImageType {
+  ECS_AL2 = 'ECS_AL2',
+  ECS_AL2_NVIDIA = 'ECS_AL2_NVIDIA',
+  EKS_AL2 = 'EKS_AL2',
+  EKS_AL2_NVIDIA = 'EKS_AL2_NVIDIA',
+}
+
+enum InstanceType extends ec2.InstanceType {
+  OPTIMAL = 'optimal',
+}
+
+enum AllocationStrategy {
+  BEST_FIT = 'BEST_FIT',
+  BEST_FIT_PROGRESSIVE = 'BEST_FIT_PROGRESSIVE',
+  SPOT_CAPACITY_OPTIMIZED = 'SPOT_CAPACITY_OPTIMIZED',
 }
 
 interface ManagedEc2ComputeEnvironmentProps extends ComputeEnvironmentProps {
@@ -796,7 +819,8 @@ interface ManagedEc2ComputeEnvironmentProps extends ComputeEnvironmentProps {
   readonly spotBidPercentage?: number;
   readonly spot?: boolean;
   readonly desiredvCpus?: number;
-  readonly instanceTypes?: ec2.InstanceType[];
+  readonly instanceTypes?: InstanceType[];
+  readonly instanceClasses?: ec2.InstanceClass[];
   readonly instanceRole?: iam.IRole;
   readonly launchTemplate?: ec2.ILaunchTemplate;
   readonly minvCpus?: number;
@@ -806,18 +830,7 @@ interface ManagedEc2ComputeEnvironmentProps extends ComputeEnvironmentProps {
 class ManagedEc2ComputeEnvironment extends ManagedComputeEnvironmentBase implements IManagedEc2ComputeEnvironment {
   constructor(readonly props: ManagedEc2ComputeEnvironmentProps = {}) {}
   public addInstanceType() {} 
-}
-```
-
-* `IFargateComputeEnvironment` -- interface to define and deploy Batch Managed Fargate Compute Environments
-
-```ts
-interface IFargateComputeEnvironment extends IManagedComputeEnvironment {}
-
-interface FargateComputeEnvironmentProps extends ComputeEnvironmentProps {}
-
-class FargateComputeEnvironment extends ManagedComputeEnvironmentBase implements IFargateComputeEnvironment {
-  constructor(readonly props: FargateComputeEnvironmentProps = {}) {}
+  public addInstanceClass() {} 
 }
 ```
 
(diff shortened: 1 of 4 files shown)


batch-l2.md
```diff
@@ -0,0 +1,1494 @@
...
+});
+jobDefn.addHostPathVolume({
+  name: 'hostPath',
+  hostPath: '/sys', 
+  mountPath: '/Volumes/hostPath',
+});
+jobDefn.addSecretVolume({
+  name: 'secret', 
+  secret: new Secret(this, 'mySecret'),
+  mountPath: '/Volumes/secret',
```
rupard-aws, line 358: https://kubernetes.io/docs/concepts/storage/volumes/#secret
https://docs.aws.amazon.com/batch/latest/APIReference/API_EksSecret.html

Batch supports secret volumes in kube pods.  That means the actual secret value is stored in Kube, life-cycled and authorized by the customer within Kube. The Batch job definition just points to that secret location; setting or specifying secret values is not support in Batch.

batch-l2.md
```diff
@@ -0,0 +1,1494 @@
...
+  medium?: MediumType;
+  sizeLimit?: number;
+}
+
+class HostPathVolume extends EksVolume {
+  path: string;
+}
+
+class SecretPathVolume extends EksVolume {
+  secret: ssm.Secret;
```
rupard-aws, line 1053: Kube (eks) secret is just a volume location pointing a a kube secret resource.  SSM is not directly involved in this mapping.

customers have the ability to populate kube secrets from SSM, but those are still exposed to pods as a generic volume location and secretName string.

batch-l2.md
```diff
@@ -0,0 +1,1494 @@
...
+  spot: true, // optional, defaults to false
+  enabled: true, // optional, defaults to true
+  updateTimeout: cdk.Duration.minutes(30), // optional, defaults to 30 (CFN default)
+  terimnateOnUpdate: false, // optional, defaults to false (CFN default)
+  allocationStrategy: batch.AllocationStrategy.SPOT_CAPACITY_OPTIMIZED, // optional (not spot exclusive)
+  spotBidPercentage: 20, // optional (Spot exclusive)
+  desiredvCpus: 3, // optional: default???
+  instanceTypes: [ec2.InstanceType.of(ec2.InstanceClass.M5AD, ec2.InstanceSize.LARGE)], // optional, defaults to ['Instance.OPTIMAL']
+  instanceRole: myRole, // optional, creates it if not supplied
+  serviceRole: myServiceRole, // optional, creates it if not supplied
```
rupard-aws, line 461: This comment is for all occurrences of `serviceRole` in this RFC.

For ALL managed CEs (ECS, EKS, Fargate) in Batch, if serviceRole is null at CreateCE API call time, Batch will create a Service-Linked-Role (SLR) on the fly for the customer or it will re use an existing one. So CDK should totally get out of the business of creating role/policy for serviceRole with Managed CEs and let the service use default SLR. CDK can keep that create role logic for Unmanaged type.

SLR is the more secure and less error prone method of giving Batch access to orchestrate jobs and scaling in customer's account.  It should be used by default for all managed CEs.

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+computeEnv.addInstanceType(batch.InstanceType.OPTIMAL);
+
+Note: this is equivalent to specifying
+
+computeEnv.addInstanceClass(ec2.InstanceClass.C4);
+computeEnv.addInstanceClass(ec2.InstanceClass.M4);
+computeEnv.addInstanceClass(ec2.InstanceClass.R4);
+```
+
+If your workflow does not tolerate interruptions and you want to minimize your costs, use `AllocationStrategy.BEST_FIT`.
```
rix0rrr, line 117: Introduce the tradeoff here as well. Right now this is phrased as if it doesn't have downsides.

```suggestion
If your workflow does not tolerate interruptions but it *does* tolerate delay and you want to minimize your costs, use `AllocationStrategy.BEST_FIT`.
```

rix0rrr: I wonder if all this prose about allocation strategies could be summarized more scannably in a table:

| Strategy | Optimized for | Downsides | Notes |
|-------|----------|----------|---------|
| BEST_FIT | Cost | May lower throughput | |
| BEST_FIT_PROGRESSIVE | Throughput | May increase cost | |
| SPOT_CAPACITY_OPTIMIZED | Least interruption | (None?) | Spot only |

etc.

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+This proposes a new set of L2 constructs for the aws-batch module. 
+
+Existing (experimental) L2 API docs: https://docs.aws.amazon.com/cdk/api/v1/docs/aws-batch-readme.html
+CloudFormation resource reference: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_Batch.html
+
+## Batch Overview
+
+Batch has four resources:
+
+* `ComputeEnvironment`
```
rix0rrr, line 16: Please also explain Instance classes and types (note on naming: in EC2, "instance TYPE" is a combination of "instance CLASS+instance SIZE" ... if here TYPE in Batch means SIZE only, that's a bit confusing)

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+### Controlling vCPU allocation
+
+With a managed `ComputeEnvironment`, you can specify the maximum and minimum vCPUs it can have at any given time.
+You cannot do this with an unmanaged `ComputeEnvironment`, because you must provision and manage the instances yourself;
+that is, Batch will not scale them up and down as needed. This example shows how to configure these properties:
+
+```ts
+new batch.ManagedEC2ComputeEnvironment(this, 'myEc2ComputeEnv', {
+  instanceClasses: [ec2.InstanceClass.A1],
+  minvCPUs: 10,
```
rix0rrr, line 145:
```suggestion
  minvCpus: 10,
```

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+
+const highPriorityQueue = new batch.JobQueue(this, 'JobQueue', {
+   priority: 10,
+});
+
+lowPriorityQueue.addComputeEnvironment(sharedComputeEnv, 1);
+highPriorityQueue.addComputeEnvironment(sharedComputeEnv, 1);
+```
+
+### Fairshare Scheduling
```
rix0rrr, line 177: This whole thing is still clear as muck to me, but it's not clear to me that it can be cleared up anyway.

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+Jobs can specify retry policies to respond to different failures with different actions.
+There are three different ways information about the way a Job exited can be conveyed;
+
+* `exitCode`: the exit code returned from the process executed by the container. Will only match non-zero exit codes.
+* `reason`: any middleware errors, like your Docker registry being down.
+* `statusReason`: infrastructure errors, most commonly your spot instance being reclaimed. 
+
+You can specify a glob string to match each of these and react to different failures accordingly.
+Up to five different retry strategies can be configured for each Job,
+and each strategy can match against some or all of `exitCode`, `reason`, and `statusReason`.
```
rix0rrr, line 305: From the description above, I'm not sure whether it makes sense to match on multiple of these together. Does it? Can BOTH my instance be reclaimed AS WELL AS my process exits with an exitCode?

I'd rather see an enum-like class that is more reasonable and hides the distinction.

```
retryStrategies [
{
  action: Action.RETRY,
  reason: Reason.SPOT_INSTANCE_RECLAIMED,
},
{
  action: Action.EXIT,
  reason: Reason.CANNOT_PULL_CONTAINER,
},
{
  action: Action.RETRY,
  reason: Reason.nonZeroExitCode(),
},
{
  action: Action.Exit,
  reason: Reason.custom({
    onExitCode: '40*', // This is so weird
    onStatusReason: 'bananas',
  }),
},
]
```

rix0rrr: And I also don't particularly care for the list of `{ action, reason }` pairs, but that's necessary if order is important.

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+
+jobDefn.addRetryStrategy({
+  action: Action.RETRY,
+  onStatusReason: StatusReason.custom('Some other reason*'),
+});
+```
+
+### Running single-container ECS workflows
+
+Batch supports ECS workflows natively. This examples creates a `JobDefinition` that runs a single container with ECS:
```
rix0rrr, line 344: > Batch supports ECS workflows natively.

I'm not sure why this is relevant. You're not using types from the ECS library anyway. Is this to say, "Batch can run jobs using either ECS or EKS" ?

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+
+For workflows that need persistent storage, batch supports mounting `Volume`s to the container.
+You can both provision the volume and mount it to the container in a single operation:
+
+```ts
+jobDefn.addVolume({
+  name: 'myVolume',
+  efsVolumeConfiguration: {
+    fileSystem: myFileSystem
+  },
```
rix0rrr, line 364:
```suggestion
  volume: Volume.efs(myFileSystem),
```

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+You can mount `Volume`s to these containers in a single operation:
+
+```ts
+jobDefn.addEmptyDirVolume({
+  name: 'emptyDir',
+  mountPath: '/Volumes/emptyDir',
+});
+jobDefn.addHostPathVolume({
+  name: 'hostPath',
+  hostPath: '/sys',
```
rix0rrr, line 406:
```suggestion
  volume: Volume.hostPath('/sys'),
```

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+jobDefn.addSecretVolume('secret', new Secret(this, 'mySecret'));
+```
+
+You can mount these later:
+
+```ts
+// Mounts them so they are accessible to the containers
+jobDefn.addVolumeMount('emptyDir', '/Volumes/emptyDir');
+jobDefn.addVolumeMount('hostPath', '/Volumes/hostPath');
+jobDefn.addVolumeMount('secret', '/Volumes/secret');
```
rix0rrr, line 431: Don't do this no more, I think.

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+```
+
+### Running Distributed Workflows
+
+Some workflows benefit from parallellization and are most powerful when run in a distributed environment,
```
rix0rrr, line 436: Is this ultimately to say: one job in the queue can be run on multiple instances in parallel?

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+Message Passing Interface (MPI) is often used with these workflows.
+You must configure your containers to use MPI properly,
+but Batch allows different nodes running different containers to communicate easily with one another.
+You must configure your containers to use certain environment variables that Batch will provide them,
+which lets them know which one is the main node, among other information.
+In particular, the environment variable that tells the containers which one is the main node can be configured on your `MultiNodeJobDefinition` as follows:
+
+```ts
+const multiNodeJob = new batch.MultiNodeJobDefinition(this, 'JobDefinition', {
+  mainNode: 0,
```
rix0rrr, line 447: Please make 0 implicit though, right?

text/batch-l2.md
```diff
@@ -0,0 +1,1540 @@
...
+```
+
+### Running Distributed Workflows
+
+Some workflows benefit from parallellization and are most powerful when run in a distributed environment,
+such as certain numerical calculations or simulations. Batch offers `MultiNodeJobDefinition`s for this purpose.
+Message Passing Interface (MPI) is often used with these workflows.
+You must configure your containers to use MPI properly,
+but Batch allows different nodes running different containers to communicate easily with one another.
+You must configure your containers to use certain environment variables that Batch will provide them,
```
rix0rrr, line 441: Link to the details?

text/batch-l2.md
rix0rrr: Will the total amount of instances automatically be `max(endNode) - min(startNode) + 1` ? Or will we always start counting from 0, even if it doesn't run any code?

rix0rrr (dismissed):
I approve of most of this, couple of details to work out but otherwise good!

batch-l2.md
comcalvi: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-eksvolume.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-secret.html

I used the Secret class here because the EKS Volume Secret requires you to specify a Secrets Manager or SSM Parameter ARN as well as the secret name. The CDK Secret construct exposes both it's ARN and name, so it seems natural to use it here. 

How is the Secret ARN used by the volume? I thought that the Volume would pull the relevant information from the Secret to set it up, since you have to specify the ARN.

text/batch-l2.md
comcalvi: Batch uses the exact same conventions. However, batch allows you to specify either the class or the type (class + size). If you specify only the class, it will choose the size for you based off what it thinks the job will need.

text/batch-l2.md
comcalvi: Yes, exactly that

text/batch-l2.md
comcalvi: we always start from 0, but the total amount of instances is always `endNode + 1`. So if we have 4 nodes, we must define node 0, node 1, node 2, and node 3.

text/batch-l2.md
comcalvi: I've opted to add the table for conciseness, but I still think the explanations of what they actually mean are worth while.

text/batch-l2.md
comcalvi: There's a new section, `Choosing your Instance Types`, that should address this.

text/batch-l2.md
comcalvi: I've completely reworked this section based off new information

text/batch-l2.md
comcalvi: Not sure I understand this one, it would become:

```
jobDefn.addHostPathVolume({
  name: 'hostPath',
  volume: Volume.hostPath('/sys'),
  mountPath: '/Volumes/hostPath',
});
```
I could see:

```
Volume.hostPath({
  name: 'hostPath',
  hostPath: '/sys',
  ...
});

File: text/batch-l2.md (hunk 8 above)
Comment (comcalvi) [reply text/batch-l2.md:305]: Order is important here, batch will perform the first `action` whose `reason` matches

Comment (comcalvi) [reply text/batch-l2.md:305]: I'm not sure why this is possible, but it's most common to use just one at a time, so this is the better UX

File: text/batch-l2.md (hunk 10 above)
Comment (comcalvi) [reply text/batch-l2.md:364]: I'll address this tomorrow, I think there's some room to improve on this API (I basically lifted it from aws-ecs)

File: batch-l2.md (hunk 1 above)
Comment (comcalvi) [reply batch-l2.md]: Discussed offline, the CFN types are wrong here

File: text/0485-aws-batch.md [hunk 16]
```diff
@@ -0,0 +1,1586 @@
...
+Instead, the scheduler will use their `weightFactor`s to decide which jobs to schedule;
+the only difference is that instead of competing for 100% of the max capacity, jobs compete for 57.8125% of the max capacity.
+
+This example specifies a `computeReservation` of 75% that will behave as explained in the example above:
+
+```ts
+const fairsharePolicy = new FairshareSchedulingPolicy(this, 'myFairsharePolicy', {
+  computeReservation: 75,
+  shares: [
+    { id: 'A },
```
rix0rrr, line 332:
```suggestion
    { id: 'A' },
```

rix0rrr (approved):

Issue #474, opened 2023-01-08
EventBridge Scheduler L2 Construct


filletofish:
## Description

EventBridge Scheduler was released on 2022/11. It is basically a powerful version of the existing EventBridge scheduled rules, with more scalability, timezone support, time window schedules, etc. [Introducing Amazon EventBridge Scheduler](https://aws.amazon.com/blogs/compute/introducing-amazon-eventbridge-scheduler/)

This RFC will propose a L2 construct for  Amazon EventBridge Scheduler, which will be a more user-friendly version of existing L1 (CloudFormation) Construct (see [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html)).

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         |  @ffilipp, @Jacco
| Author(s)           | @ffilipp, @Jacco
| API Bar Raiser      | @kaizencc 
| Stakeholders        | 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [x] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

roshni06:
Hi what is the status of this feature in CDK and when will this be released?

jakegrabowski:
Also waiting for it :)

mrgrain:
`scheduler-alpha` has been released for a while 🎉 

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-scheduler-alpha-readme.html

deeTEEcee:
@mrgrain any idea when is it gonna get out of alpha state?

mrgrain:
[...]

PR #486, opened 2023-03-05
RFC 474: Event Bridge Scheduler L2

filletofish:
This is a request for comments about EventBridge Scheduler L2 Construct. See https://github.com/aws/aws-cdk-rfcs/issues/474 for
additional details. 

APIs are **NOT** signed off by @{BAR_RAISER}.


text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+>
+> * **PRESS RELEASE**: If this is a major feature (~6 months of work), write the
+>   press release which announces this feature. The press release is a single
+>   page that includes 7 paragraphs: (1) summary, (2) problem, (3) solution, (4)
+>   leader quote, (5) user experience, (6) customer testimonial and (7) one
+>   sentence call to action.
+
+**CHANGELOG**:
+
+`feat(events-scheduler): Event Bridge Scheduler L2 constructs`
```
kaizencc, line 42: If possible, lets break the proposed changelog into atomic pieces. Can we introduce schedule/targets/schedule-groups separately?

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+3. **Schedule Group**: Scheduling groups help you organize your schedules. Scheduling groups support tags that you can use for cost allocation, access control, and resource organization. When creating a new schedule, you can add it to a scheduling group.
+
+This module is part of the [AWS Cloud Development Kit](https://github.com/aws/aws-cdk) project. It allows you to define Event Bridge Schedules.
+
+
+## Defining a schedule 
+
+```ts
+
+const target = new targets.LambdaInvoke(props.func, {
```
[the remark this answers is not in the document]
kaizencc: and if so, `events` already has a concept of a schedule. Can you explain what this resource achieves that the schedule class in `events` does not?

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+});
+```
+
+A one-time schedule is a schedule that invokes a target only once. You configure a one-time schedule when by specifying the time, date, and time zone in which EventBridge Scheduler evaluates the schedule.
+
+```ts
+const oneTimeSchedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.at(new Date(2022, 10, 20, 19, 20, 23)),
+    target,
+    scheduleTimeZone: 'America/New_York'
```
kaizencc, line 101: TimeZone enum is added to core: https://github.com/aws/aws-cdk/pull/24127. We should use it.

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+});
+```
+
+### Grouping Schedules
+
+When creating a new schedule, you can add it to a scheduling group:
+
+```ts
+const group = new Group(this, "Group", {
+    groupName: "MyGroup"
```
kaizencc, line 112: Are there going to be more props to a group? Or is it simply a name.

Also, should we add APIs like `group.addSchedules(schedules: Schedule[])`?

kaizencc (commented):
Hi @filletofish, @jacco, starting off with a few comments on the proposed APIs. Nothing too major. I'll be back to look at the rest soon.

text/0474-event-bridge-scheduler-l2.md
filletofish: Could do so, but could you please explain to me what is the value of this? We will likely commit the whole implantation in a single PR to CDK repo?

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
filletofish: > this is coming from the existing events-targets module right?

No it's a separate module `targets` for schedule targets. I do not think that we should attempt to reuse the same module for EB Rules Targets and EB Schedule Targets. I think as in CFN those targets are different CFN declarations ([Rule Target CFN](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-scheduler-schedule-target.html), [EB Target CFN](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-events-rule-target.html)), attempting to somewhat reuse these classes in L2 constructs will create more problems for edge-cases when behaviour between targets is different. 

Curious to know what do you think should be the reasons for reusing module targets? I would guess, that maybe in customer code it would be easier to use the same targets for EB rules and EB schedules, right?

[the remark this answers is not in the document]
filletofish: > Can you explain what this resource achieves that the schedule class in events does not?

In EB Rules a `Schedule` is just a schedule expression – a string that defines on what schedule the rule should be invoked. 

In EB Scheduler a `Schedule` is a separate CFN resource. It has much more properties than just a schedule expression (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html). For example, `Schedule` supports timezone, groups, encryption, it can be disabled. `Schedule` in Scheduler plays similar role to  `Rule`: it is a central concept of the package that connects event input and event target.

Also, in the prototype implementation of L2 Scheduler constructs, we have a class `ScheduleExpression` that inherits `eb.Schedule`. We have added a method `at(..)` that allows to create a one-time schedule from a date. See https://github.com/filletofish/cdk-eb-scheduler/blob/main/lib/aws-scheduler/lib/expression.ts#L10

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+```ts
+
+const target = new targets.LambdaInvoke(props.func, {
+    input: ScheduleTargetInput.fromObject({
+    "payload": "useful"
+    })
+});
+    
+const schedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.rate(Duration.minutes(10)),
```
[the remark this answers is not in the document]
filletofish: Yes, see previous comment (https://github.com/aws/aws-cdk-rfcs/pull/486#discussion_r1141427352) - we are inheriting EB `Schedule` class to add implementation of `at(...)` method to create a one-time schedules, which are only supported in Scheduler (AFAIU).

text/0474-event-bridge-scheduler-l2.md
filletofish: Amazing, thanks!

text/0474-event-bridge-scheduler-l2.md
filletofish: > Are there going to be more props to a group? Or is it simply a name.

By them you mean the owners of AWS EB Scheduler service? I do not know, how can we check that? 

> Also, should we add APIs like group.addSchedules(schedules: Schedule[])?

Sounds like a good idea! 

A question for implementation - currently, we create `CfnSchedule` in a constructor of `Schedule` and pass `groupName` parameter to `CfnSchedule`. How can we delay creation of `CfnSchedule` so that we can change the `groupName` parameter dynamically? Is there a good example of that behaviour somewhere in CDK lib already?

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,481 @@
...
+    group: group,
+});
+```
+
+If group is not provided schedule will be created in default group, which you can access in CDK with: 
+```ts
+const defaultGroup = Group.fromDefaultGroup(this, "DefaultGroup");
+```
+
+## Scheduler Targets
```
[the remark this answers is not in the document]
filletofish: See discussion above - https://github.com/aws/aws-cdk-rfcs/pull/486#discussion_r1141424181

text/0474-event-bridge-scheduler-l2.md
kaizencc: Yes. I think you're asking for Lazy :). We use it all over the place. [here](https://github.com/aws/aws-cdk/blob/cabff71139bf6a3b8ecd4cdf4a8dea093e67b8e6/packages/%40aws-cdk/aws-events/lib/rule.ts#L112) is an example.

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
kaizencc: We can still do that, if the `events.Schedule` class was in core. I guess I understand that there is a `Schedule` prop in `events` (and other places) and then a `Schedule` resource that has a `ScheduleExpression` in this RFC, but I think they are the same thing represented in different ways.

A big win for the CDK is that we can take these two different _ways_ of expressing a concept like "schedule" and expose only one CDK "schedule". Then behind the scenes we do the translation into what the different resources want. That is sort of what you are doing in https://github.com/filletofish/cdk-eb-scheduler/blob/main/lib/aws-scheduler/lib/expression.ts#L10. Looks like you aren't creating a brand new `ScheduleExpression` class but extending the existing `Schedule`. That is fine for now, and https://github.com/aws/aws-cdk/issues/8979 can continue to be a ticket asking for `events.Schedule` to be moved to `core`.

text/0474-event-bridge-scheduler-l2.md
filletofish: > The value would be that its easier to review.

What exactly easier to review, the change log?

I thought this is a proposal for how the future CDK change-log would look like. See for example L2 constructs release in https://github.com/aws/aws-cdk/releases/tag/v2.65.0. In this case, isn't it easier for the reader to have a single entry for Scheduler L2 constructs rather than multiple entries for each L2 Construct Class separately? As those construct classes on their own do not provide much value, I wouldn't break them up. We could still list the new classes in the description though.

kaizencc: I mean easier for me to review the code if the PRs are separated out into chunks. The smaller the PR, the less likely I am to miss things. If these constructs are not able to be separated then I understand, although I don't feel like that's necessarily the case.

For example, `ScheduleGroup` can and should be a separate PR imo. Just helps declutter the initial PR.

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
kaizencc: No, what you are doing is enough I think.

```ts
class SpecialSchedule extends core.Schedule {
  at(...) {}
}

File: text/0474-event-bridge-scheduler-l2.md [hunk 7]
```diff
@@ -0,0 +1,545 @@
...
+
+Target can be invoked with a custom input. Class `ScheduleTargetInput` supports free form text input and JSON-formatted object input:
+
+```ts
+const input = ScheduleTargetInput.fromObject({
+    'QueueName': 'MyQueue'
+});
+```
+
+You can include context attributes in your target payload. EventBridge Scheduler will replace each keyword with its respective value and deliver it to the target. Supported context attributes:
```
kaizencc, line 188: Can we link to any relevant EB docs here?

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+const role = new iam.Role(this, 'Role', {
+  assumedBy: new iam.ServicePrincipal('scheduler.amazonaws.com'),
+});
+
+const target = new targets.LambdaInvoke({
+    input: ScheduleTargetInput.fromObject({ 
+        "payload": "useful"
+    }),
+    role: role
+}, props.func);
```
kaizencc, line 222: replace `props.func` with `declare const fn: lambda.Function;`

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+> Describe your plan on how to deliver this feature from prototyping to GA.
+> Especially think about how to "bake" it in the open and get constant feedback
+> from users before you stabilize the APIs.
+>
+> If you have a project board with your implementation plan, this is a good
+> place to link to it.
+
+Most part of the L2 constructs is already implemented. We have a set of relatively (1-2 day of work) todos that can be implemented in parallel: 
+
+1. Decide what removal policy for Group should be
```
kaizencc, line 523: Do we need to discuss this any further?

kaizencc (changes requested):
Hi! @filletofish @jacco. I am back, and have just a few more comments on this RFC. Once they are addressed I will swiftly approve and we can get to the implementation. Thanks for all your work thus far!

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+    schedule: ScheduleExpression.at(new Date(2022, 10, 20, 19, 20, 23)),
+    target,
+    scheduleTimeZone: TimeZone.AMERICA_NEW_YORK,
+    description: 'This is a one-time schedule in New York timezone',
+});
+```
+
+### Grouping Schedules
+
+When creating a new schedule, you can add it to a scheduling group:
```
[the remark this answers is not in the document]
Jacco: Reasons for grouping:
- access control
- list all schedules in a group
- delete group -> delete all schedules in it

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+});
+```
+
+A one-time schedule is a schedule that invokes a target only once. You configure a one-time schedule when by specifying the time, date, and time zone in which EventBridge Scheduler evaluates the schedule.
+
+```ts
+const oneTimeSchedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.at(new Date(2022, 10, 20, 19, 20, 23)),
+    target,
+    scheduleTimeZone: TimeZone.AMERICA_NEW_YORK,
```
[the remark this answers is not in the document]
Jacco: I need to check. Because I am reusing a class that already has `cron`, I only added `at`.

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+});
+    
+const schedule1 = new Schedule(this, 'Schedule1', {
+    schedule: ScheduleExpression.rate(Duration.minutes(10)),
+    target
+});
+
+const schedule2 = new Schedule(this, 'Schedule2', {
+    schedule: ScheduleExpression.rate(Duration.minutes(5)),
+    target
```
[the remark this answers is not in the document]
Jacco: Yeah we are still playing around in a separate repo. Once the APIs are solid we will move to aws-cdk fork

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,545 @@
...
+});
+
+group.addSchedules(schedule1, schedule2);
+```
+
+If group is not provided schedule will be created in default group, which you can access in CDK with: 
+```ts
+const defaultGroup = Group.fromDefaultGroup(this, "DefaultGroup");
```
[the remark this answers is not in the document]
Jacco: Only for being able to do grantRead / grantWrite on the default group :-)

text/0474-event-bridge-scheduler-l2.md
Jacco: We need to do some tests. Main point to figure out is if a group you create will actually delete all schedules in it when the group is deleted. That would make it a resource with state and its state would deserve protection.

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
filletofish: >  seems a bit odd to have schedule and scheduleTimeZone since scheduleTimeZone seems to be operating on whatever information is in schedule

Yeah, I agree it's odd. However, I don't think we have any other option if we want to reuse class `Schedule` from `events` module, as class `Schedule` already has a method `cron` that does not accept a timezone. Alternatively, we could create our own class `ScheduleExpression` that does not inherit from `Schedule`. What do you think?

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
filletofish: > We'll have to worry about dependencies because we will use the defaultGroup resource as if it already exists and not as a to-be-created resource.

Yeah, as Jacco mentioned above the default group already exists in every AWS account. 

See https://docs.aws.amazon.com/scheduler/latest/UserGuide/managing-schedule-group.html

> Only for being able to do grantRead / grantWrite on the default group :-)

And also for getting its metrics.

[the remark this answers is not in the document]
kaizencc: Sounds good. If it already exists, shouldn't be a dependency problem then.

text/0474-event-bridge-scheduler-l2.md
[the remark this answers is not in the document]
kaizencc: Lets just go with as is for now. I'm fine with `scheduleTimeZone`.

comcalvi (changes requested):

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,546 @@
...
+const schedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.rate(Duration.minutes(10)),
+    target,
+    description: 'This is a test schedule that invokes lambda function every 10 minutes.',
+});
+```
+
+### Schedule Expressions
+
+Both rate-based and cron-based schedules are recurring schedules. You configure each recurring schedule type using a schedule expression for the type of schedule you want to configure, and specifying a time zone in which EventBridge Scheduler evaluates the expression.
```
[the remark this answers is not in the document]
filletofish: No, just those two. Will update the doc to make it clear

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,546 @@
...
+});
+
+group.addSchedules(schedule1, schedule2);
+```
+
+Groups can be used to organize the schedules logically, access the schedule metrics and manage permissions at group granularity (see below).
+
+## Scheduler Targets
+
+The `@aws-cdk/aws-schedule-targets` (ADD LINK) module includes classes that implement the `IScheduleTarget` interface for various AWS services.
```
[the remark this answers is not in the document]
filletofish: Wanted to add a link once the module is realised. Better to remove for now. Thanks for highlighting

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,546 @@
...
+Most part of the L2 constructs is already implemented. We have a set of relatively (1-2 day of work) todos that can be implemented in parallel: 
+
+1. Decide what removal policy for Group should be 
+2. Support and Test Encrypted Schedules
+3. Decide or clean up how IAM Roles are used in Schedules and Targets
+4. Check why name generation fails if env is not passed to the Stack
+5. Unit tests for class schedule
+6. Unit tests for class targets 
+7. Unit tests for classes: groups, schedule expressions, input 
+8. Integration tests #15
```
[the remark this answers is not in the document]
filletofish: Accidental typo during copy paste (copied these items from a list of GH issues).

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,546 @@
...
+
+```ts
+const dlq = new Queue(this, "DLQ", {
+      queueName: 'MyDLQ',
+});
+
+const target = new targets.LambdaInvoke(props.func, {
+    deadLetterQueue: dlq,
+    maximumEventAge: Duration.minutes(1),
+    maximumRetryAttempts: 3
```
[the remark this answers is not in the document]
filletofish: In CFN only maximum event age and maximum retry attempts are part of retry policy. Would you recommend to organise them similarly in a `RetryPolicy` structure as well? 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-scheduler-schedule-retrypolicy.html

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,546 @@
...
+
+const target = new targets.LambdaInvoke({
+    input: ScheduleTargetInput.fromObject({ 
+        "payload": "useful"
+    }),
+    role: role
+}, fn);
+```
+
+For universal targets you must grant the required IAM permissions yourself.
```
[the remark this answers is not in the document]
filletofish: Updated the RFC

filletofish:
Thank you for your feedback @comcalvi! I have made some updates per your comments.

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,569 @@
...
+const rateBasedSchedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.rate(Duration.minutes(10)),
+    target,
+    description: 'This is a test rate-based schedule',
+});
+
+const cronBasedSchedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.cron({ minute: '0', hour: '23', day: '20', month: '11' }),
+    target,
+    scheduleTimeZone: TimeZone.AMERICA_NEW_YORK,
```
tmokmss, line 90: How about renaming the prop `scheduleTimezone` to `timezone` to make it short? I feel it is already clear enough that this timezone property is for the schedule.

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,569 @@
...
+    description: 'This is a test cron-based schedule that will run at 11:00 PM, on day 20 of the month, only in November in New York timezone',
+});
+```
+
+A one-time schedule is a schedule that invokes a target only once. You configure a one-time schedule when by specifying the time, date, and time zone in which EventBridge Scheduler evaluates the schedule.
+
+```ts
+const oneTimeSchedule = new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.at(new Date(2022, 10, 20, 19, 20, 23)),
```
tmokmss, line 99: I guess this `Date` class will be defined somewhere in CDK, as we cannot use Node's built-in Date class here due to jsii limitation.

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -0,0 +1,569 @@
...
+    Group <|-- Schedule : group
+    Input <|-- ScheduleTargetBase : input
+    Targets <.. ScheduleTargetBase : implements
+
+    class Schedule {
+        + String name
+        + Duration flexibleTimeWindow 
+        + Date startDate
+        + Date endDate
+        + String scheduleTimeZone
```
tmokmss, line 475: `String` should be replaced to `TimeZone` :)

tmokmss (commented):
A few minor comments from one of the users' perspective. Looking forward to using this feature! 🚀

text/0474-event-bridge-scheduler-l2.md
Jacco: I think I only implemented it this way when I saw another instance where Date was used in the CDK libraries. Will have a look :-)

text/0474-event-bridge-scheduler-l2.md
filletofish: I am concerned that the customer might think that timezone is also applied to other properties. For example, `Schedule` also has properties `startDate` and `endDate`, which should be specified in UTC timezone and do not depend on timezone.

So we have these options: 

1 - Do nothing

Keep interface as-is with a separate `scheduleTimeZone` parameter. It's simple and matches CFN structure https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html#cfn-scheduler-schedule-scheduleexpressiontimezone

2 - Rename `scheduleTimeZone` to `timeZone` 

Not obvious that the property `timeZone` does not apply to `startDate` and `endDate`

3 -  Add timezone parameter to `ScheduleExpression` methods

Example: 
```ts
const cronBasedSchedule = new Schedule(this, 'Schedule', {
    schedule: ScheduleExpression.cron({ minute: '0', hour: '23', day: '20', month: '11', timezone: TimeZone.AMERICA_NEW_YORK}),
    target,
    description: 'This is a test cron-based schedule that will run at 11:00 PM, on day 20 of the month, only in November in New York timezone',
});
```

In implementation we will create a class `ScheduleExpression` that will inherit from [`events.Schedule`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-events.Schedule.html) and will have methods `cron(..)`, `at(..)` and `rate(..)` with timezone as an additional parameter. 

Not a bad option. Maybe `timezone` will not be that discoverable as in other options. 


4 - Create a separate interface  with two parameters `expression` and `timezone`

Example: 

```ts
const cronBasedSchedule = new Schedule(this, 'Schedule', {
    scheduleExpression: {
          expression: ScheduleExpression.cron({ minute: '0', hour: '23', day: '20', month: '11' }),
          timeZone: TimeZone.AMERICA_NEW_YORK
     },
    target,
    description: 'This is a test cron-based schedule that will run at 11:00 PM, on day 20 of the month, only in November in New York timezone',
});
```

However, I am not sure what name to choose for the new interface as names `Schedule` and `ScheduleExpression` are already taken. Any recommendations? 

---

All in all, I am probably still leaning to the first as it's the simplest option. Curious if you guys also have any opinion on this question @Jacco, @kaizencc  and @comcalvi?

text/0474-event-bridge-scheduler-l2.md
filletofish: Thanks, will fix in the next commit

text/0474-event-bridge-scheduler-l2.md
tmokmss: That makes sense, thanks! To me, option 3 or 4 look nice and more cohesive, but at the same time option 1 sounds reasonable. I'm fine with whatever decision is made :)

kaizencc: I think I like option 4, with `scheduleExpression: { schedule: Schedule, timeZone: TimeZone }`

Jacco: Few things to consider: 

How would the original `events.Schedule` be changed if timezone support is added for `aws-events`. I think it would best added `CronOptions` in CDK since it has no meaning for a `rate` expressions `timeZone` has no meaning. It would be kept separate though because in cloudFormation is would probably not be part of the expression but a separate property like here.

So maybe we should still inherit but add the timeZone parameter from `cron` and `at`.

Unflat interfaces give less good code completion in some languages (like python).

Another thing to consider: what if `aws-events` at some point adds another type of schedule that is unsupported by `aws-scheduler`. We could add an error raising override, but maybe we'd better not inherited?

Maybe implement like this:

```ts
import { Duration, TimeZone } from 'aws-cdk-lib';
import * as events from 'aws-cdk-lib/aws-events';

/**
 * Schedule for EventBridge Schedule
 *
 * Note that rates cannot be defined in fractions of minutes.
 *
 * @see https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html
 */
export abstract class ScheduleExpression {
  /**
   * Construct a one-time schedule from a date.
   *
   * @param date The date and time to use. The millisecond part will be ignored.
   */
  public static at(date: Date, timeZone?: TimeZone): ScheduleExpression {
    try {
      const literal = date.toISOString().split('.')[0];
      return new LiteralScheduleExpression(`at(${literal})`, timeZone)
    } catch(e) {
      if (e instanceof RangeError){
        throw new Error('Invalid date');
      }
      throw e;
    }
  }

  public static cron(options: events.CronOptions, timeZone?: TimeZone): ScheduleExpression {
    return new LiteralScheduleExpression(events.Schedule.cron(options).expressionString, timeZone);
  }

  public static rate(duration: Duration): ScheduleExpression {
    return new LiteralScheduleExpression(events.Schedule.rate(duration).expressionString);
  }

  public static expression(expression: string, timeZone?: TimeZone) {
    return new LiteralScheduleExpression(expression, timeZone);
  }

  /**
   * Retrieve the expression for this schedule
   */
   public abstract readonly expressionString: string;

  /**
   * Retrieve the timeZone for this schedule
   */
   public abstract readonly expressionTimeZone?: TimeZone;
}

class LiteralScheduleExpression extends ScheduleExpression {
  constructor(public readonly expressionString: string, public readonly expressionTimeZone?: TimeZone) {
    super();
  }

  public _bind() {}
}
```

text/0474-event-bridge-scheduler-l2.md
Jacco: There are many examples of Date being used. Here is an example:

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-applicationautoscaling/lib/scalable-target.ts

tmokmss: TIL about that! Reading [this doc](https://aws.github.io/jsii/specification/2-type-system/), it seems Date is also  supported (a bit misleading though). I'll archive this 👍

text/0474-event-bridge-scheduler-l2.md
filletofish: I like that suggestion @Jacco. It's similar to option 3. I was also thinking about the way to reuse functionality of `events.Schedule` with composition, but couldn't come up with a good enough option like yours. 


I will add it to the RFC. Thanks!

kaizencc: Agree on not inheriting for now. The ideal scenario is still to put a generic Schedule into core that can be extended by the 4 resources that have the concept of schedule today. But that's out of scope of this RFC. 

(autoscaling, app-autoscaling, scheduler, events are the 4 i can think of)

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -221,15 +231,34 @@ const target = new targets.LambdaInvoke({
     input: ScheduleTargetInput.fromObject({ 
         "payload": "useful"
     }),
-    role: role
+    role,
 }, fn);
 ```
 
-For universal targets you must grant the required IAM permissions yourself.
+Class for universal target also automatically creates an IAM role if you do no specify your own IAM role. However, in comparison with templated targets, for universal targets you must grant the required IAM permissions yourself.
```
comcalvi, line 238:
```suggestion
Universal target automatically create an IAM role if you do not specify your own IAM role. However, in comparison with templated targets, for universal targets you must grant the required IAM permissions yourself.
```

comcalvi (changes requested):
This is great work, I just have a few minor comments

text/0474-event-bridge-scheduler-l2.md
```diff
@@ -65,45 +65,51 @@ const target = new targets.LambdaInvoke(props.func, {
...
     target,
     description: 'This is a test rate-based schedule',
 });
 
 const cronBasedSchedule = new Schedule(this, 'Schedule', {
-    schedule: ScheduleExpression.cron({ day: '20', month: '11' }),
+    scheduleExpression: ScheduleExpression.cron(
+        { minute: '0', hour: '23', day: '20', month: '11' },
+        TimeZone.AMERICA_NEW_YORK
+    ),
```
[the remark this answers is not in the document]
filletofish: nice one, will add

comcalvi (dismissed):

kaizencc (dismissed):

kaizencc (approved):

PR #498, opened 2023-04-19
RFC for Glue L2 Construct

natalie-white-aws:
This is a request for comments about the Glue L2 Construct. See https://github.com/aws/aws-cdk-rfcs/issues/497for
additional details. 

APIs are signed off by @TheRealAmazonKendra.

---


text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+  minWorkers: 20,
+  numberOfWorkers: 50
+});
+```
+
+### Job Triggers
+
+We will add convenience functions for adding triggers to jobs. Standalone
+triggers are an anti-pattern, so we will only create triggers from within a
+workflow.
```
TheRealAmazonKendra, line 274: As mentioned above, I'm not seeing a section on `Workflows` so I think I'm missing a little context here, however, instead of creating functions for each type of trigger, I'd suggest making `trigger`\`triggers` an optional prop on jobs that takes in an ITrigger and then do something more along the lines of:
```
trigger: Trigger.onDemand()
```
or 
```
trigger: Trigger.scheduled(...)
```
or 
```
trigger: Trigger.conditional(...)
```
In addition, you could also have a convenience function like:
```
myGlueJob.addTrigger(Trigger.notifyEvent(...));
```

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+#### **4. Conditional Trigger**
+
+Conditional trigger has predicate and action associated with it. Based on
+predicate, trigger action will be executed.
+
+```
+// Triggers on Job and Crawler status
+myGlueJob.addConditionalTrigger(
+  jobs: [
+    {jobArn: "job1-arn", status: glue.JobStatus.SUCCEEDED},
```
TheRealAmazonKendra, line 351: Any time you want to take in an ARN, you should instead use IJob or ICrawler

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+    {jobArn: "job2-arn", status: glue.JobStatus.FAILED},
+  ], crawlers: [
+    {crawlerArn: "crawler1-arn", status: glue.CrawlerStatus.SUCCEEDED},
+    {crawlerArn: "crawler2-arn", status: glue.CrawlerStatus.TIMEOUT},
+]);
+
+```
+
+### Connection Properties
```
TheRealAmazonKendra, line 360: Could you provide examples of what this contract would look like?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+
+* **Networking - CDK determines the best fit subnet for Glue Connection
+configuration
+    **The current glue-alpha-module requires the developer to specify the
+    subnet of the Connection when it’s defined. This L2 RFC will make the
+    best choice selection for subnet by default by using the data source
+    provided during Job provisioning, traverse the source’s existing networking
+    configuration, and determine the best subnet to provide to the Glue Job
+    parameters to allow the Job to access the data source. The developer can
+    override this subnet parameter, but no longer has to provide it directly.
```
TheRealAmazonKendra, line 378: Can you provide a little more technical detail on how you plan to do this? If you intend to make an API call to accomplish this, I would strongly suggest storing the value in `context` so that it only has to be done the first time.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+Validating Glue version and feature use per AWS region at synth time is also
+not in scope. AWS’ intention is for all features to eventually be propagated to
+all Global regions, so the complexity involved in creating and updating region-
+specific configuration to match shifting feature sets does not out-weigh the
+likelihood that a developer will use this construct to deploy resources to a
+region without a particular new feature to a region that doesn’t yet support
+it without researching or manually attempting to use that feature before
+developing it via IaC. The developer will, of course, still get feedback from
+the underlying Glue APIs as CloudFormation deploys the resources similar to the
+current CDK L1 Glue experience.
```
TheRealAmazonKendra, line 418: While these limitations will not necessarily need to be validated in code, they will need to be documented in docstring.

TheRealAmazonKendra (changes requested):
This is a great start! Please see my comments inline.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+   numberOfWorkers: 20
+});
+```
+
+1. **Streaming Jobs**
+
+A Streaming job is similar to an ETL job, except that it performs ETL on data
+streams. It uses the Apache Spark Structured Streaming framework. Some Spark
+job features are not available to streaming ETL jobs. These jobs will default
+to use Python 3.6.
```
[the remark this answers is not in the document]
natalie-white-aws: This is an error - python version isn't required for streaming spark jobs. For all of the defaults listed here, we're going off what Glue Studio currently does so customers can have the same experience. Agreed that we will default to 3.9 where it's required.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+recommendations when creating a Glue Job such use of Secrets Management for
+Connection JDBC strings, Glue Job Autoscaling, least privileges in terms of
+IAM permissions and also sane defaults for Glue job specification (more details
+are mentioned in the below table).
+
+This RFC will introduce breaking changes to the existing glue-alpha-module to
+streamline the developer experience and introduce new constants and validations.
+The L2 construct will determine the job type by the job type and language
+provided by the developer, rather than having separate methods in every
+permutation that Glue jobs allow.
```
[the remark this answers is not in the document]
natalie-white-aws: This makes a lot of sense, especially after reviewing the AppSync example. We'll update the RFC to reflect this contract.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
natalie-white-aws: We will add a note that we will log a warning if developers attempt to override the best practice default with deprecated / EOL versions.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+description:
+
+```
+glue.Job(this, 'ScalaSparkEtlJob', {
+   jobType: glue.JobType.ETL,
+   jobLanguage: glue.JobLanguage.SCALA_SPARK,
+   glueVersion: glue.GlueVersion.V3_0,
+   scriptS3Url: 's3://bucket-name/path-to-scala-jar',
+   className: 'com.example.HelloWorld',
+   extraJarsS3Url: ['s3://bucket-name/path-to-extra-scala-jar',],
```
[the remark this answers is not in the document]
natalie-white-aws: You only have one script per job - multiple scrips per job is not supported.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+   description: 'an example Scala Spark ETL job',
+   numberOfWorkers: 20
+});
+
+glue.Job(this, 'pySparkEtlJob', {
+   jobType: glue.JobType.ETL,
+   jobLanguage: glue.JobLanguage.PYSPARK,
+   glueVersion: glue.GlueVersion.V3_0,
+   pythonVersion: glue.PythonVersion.3_6,
```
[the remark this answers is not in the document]
natalie-white-aws: Agreed.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+
+myGlueJob.createHourlyTrigger(this, 'MyHourlyTrigger');
+
+myGlueJob.createWeeklyTrigger(this, 'MyWeeklyTrigger');
+
+myGlueJob.createScheduledTrigger(this, 'MyScheduledTrigger', {
+  description: 'Scheduled run for ' + crawler.name,
+  schedule: '`cron(15 12 * * ? *)'`` //``every day at 12:15 UTC`
+});
+```
```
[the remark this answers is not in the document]
natalie-white-aws: 1. No, triggers must be associated with a specific job, and we wouldn't want to modify an existing resource. 
2.  Everything is UTC so the daily trigger would run at midnight UTC, the hourly would run at the top of the hour, weekly would run Sunday at midnight UTC. We can add this to the spec.
3. We'll look at whether this supports all of the cron expression sections. Thanks for the suggestion, gotta leverage everything we can!

text/0497-glue-l2-construct.md
natalie-white-aws: Janny will work on codifying this for clarity.

text/0497-glue-l2-construct.md
natalie-white-aws: Janny will add more context around this, via VPC subnet selection

text/0497-glue-l2-construct.md
natalie-white-aws: Noted.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+
+glue.Job(this, 'pySparkEtlJob', {
+    jobType: glue.JobType.ETL,
+    jobLanguage: glue.JobLanguage.PYSPARK,
+    scriptS3Url: 's3://bucket-name/path-to-python-script',
+});
+```
+
+Optionally, developers can override the glueVersion and add extra jars and a
+description:
```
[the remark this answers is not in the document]
natalie-white-aws: Will do.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+default enable several features for ETL jobs these are
+`—enable-metrics, —enable-spark-ui, —enable-continuous-cloudwatch-log.` We
+recommend to use these feature for ETL jobs. You can find more details about
+version, worker type and other feature on public documentation.
+
+```
+glue.Job(this, 'ScalaSparkEtlJob', {
+    jobType: glue.JobType.ETL,
+    jobLanguage: glue.JobLanguage.SCALA_SPARK,
+    scriptS3Url: 's3://bucket-name/path-to-scala-jar',
```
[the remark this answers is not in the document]
humanzz: Isn't that already handled by `Code.fromBucker` and `Code.fromAsset` in https://github.com/aws/aws-cdk/blob/4038274b812a68993bf6b6cb5b0a3fe971a059ad/packages/%40aws-cdk/aws-glue-alpha/lib/code.ts?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+    className: 'com.example.HelloWorld',
```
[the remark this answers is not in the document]
humanzz: As users of Glue Jobs in my team, we do have a single jar that's re-used across our jobs, and we specify the class that's gonna be used by each job

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
humanzz: Same comment about the extraJars, this is already handled by `Code`. I'd find the experience with having to provide the full s3 url as a string, a degradation from the current experience (which was originally inspired by `lambda.Code`

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
natalie-white-aws: @vaughngit will research the current implementation and we'll update the RFC scope accordingly.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
natalie-white-aws: Discussed internally, the L2 will not support creating new resources using deprecated Python versions. Will update the RFC accordlingly.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
natalie-white-aws: @mjanardhan will take this update.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+
+myGlueJob.createNotifyEventNonBatchingTrigger(this, 'MyNotifyTrigger',
+    workFlowName: workflow.name
+);
+
+myGlueCrawler.createNotifyEventNonBatchingTrigger(this, 'MyNotifyTrigger',
+    workFlowName: workflow.name
+);
+
+```
```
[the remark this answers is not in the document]
natalie-white-aws: @parag-shah-aws will research this

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Create a Glue Job
+
+The glue-alpha-module already supports three of the four common types of Glue
+Jobs: Spark (ETL and Streaming), Python Shell, Ray. This RFC will add the
+more recent Flex Job. The construct also implements AWS practice
+recommendations when creating a Glue Job such use of Secrets Management for
+Connection JDBC strings, Glue Job Autoscaling, least privileges in terms of
+IAM permissions and also sane defaults for Glue job specification (more details
+are mentioned in the below table).
```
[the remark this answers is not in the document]
natalie-white-aws: We removed the table from the main body of the RFC, will update this reference

text/0497-glue-l2-construct.md
mjanardhan: See https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ec2/SubnetSelection.html

text/0497-glue-l2-construct.md
parag-shah-aws: Adding function at job level will be ambiguous, we'll have to add additional parameter to distinguish between action/predicate list remove vs add for removing existing jobs from trigger or add new jobs to list. Rather I think we should add trigger functions in workflow

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
parag-shah-aws: Agree, based on batch size parameter batching or non-batching trigger can be created.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
TheRealAmazonKendra: This line looks like it still needs to be removed.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+
+This RFC will introduce breaking changes to the existing glue-alpha-module to
+streamline the developer experience and introduce new constants and validations.
+The L2 construct will determine the job type by the job type and language
+provided by the developer, rather than having separate methods in every
+permutation that Glue jobs allow. As an opinionated construct, it will enforce
```
TheRealAmazonKendra, line 46: Based off my last set of feedback, I think this is no longer true?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+   glueVersion: glue.GlueVersion.V3_0,
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-scala-jar'),
+   className: 'com.example.HelloWorld',
+   extraJarsS3Url: [glue.Code.fromBucket('bucket-name', 'path-to-extra-scala-jar'),],
+   description: 'an example Scala Spark ETL job',
+   numberOfWorkers: 20
+});
+
+glue.pySparkEtlJob(this, 'pySparkEtlJob', {
+   jobType: glue.JobType.ETL,
```
TheRealAmazonKendra, line 87: I don't think this prop makes sense here. It's implied in the name. Is it serving a purpose I might be missing?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+permutation that Glue jobs allow. As an opinionated construct, it will enforce
+best practices and not allow developers to create resources that use deprecated
+libraries and tool sets (e.g. deprecated versions of Python).
+
+### Spark Jobs
+
+1. **ETL Jobs**
+
+ETL jobs supports python and Scala language. ETL job type supports G1, G2, G4
+and G8 worker type default as G2, which customer can override. It wil default to
```
TheRealAmazonKendra, line 55: Just curious why G2 is chosen as the default here?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+
+```
+
+Optionally, developers can override the glueVersion and add extraJars and a
+description:
+
+```
+new glue.PythonSparkStreamingJob(this, 'PythonSparkStreamingJob', {
+   glueVersion: glue.GlueVersion.V3_0,
+   pythonVersion: glue.PythonVersion.3_6,
```
TheRealAmazonKendra, line 128: Python 3.6 is deprecated so we would not support it.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-python-script'),
+   description: 'an example Python Streaming job',
+   numberOfWorkers: 20,
+});
+
+
+new glue.ScalaSparkStreamingJob(this, 'ScalaSparkStreamingJob', {
+   glueVersion: glue.GlueVersion.V3_0,
+   pythonVersion: glue.PythonVersion.3_6,
```
TheRealAmazonKendra, line 137: Same comment as above.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+glue.pySparkFlexEtlJob(this, 'pySparkFlexEtlJob', {
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-python-script'),
+});
+```
+
+Optionally, developers can override the glue version, python version,
+provide extra jars, and a description
+
+```
+glue.ScalaSparkFlexEtlJob(this, 'ScalaSparkFlexEtlJob', {
```
TheRealAmazonKendra, line 169: Looks like pythonVersion is missing from this one. Are these the complete set of props proposed?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-scala-jar'),
+   className: 'com.example.HelloWorld',
+   extraJarsS3Url: [glue.Code.fromBucket('bucket-name', 'path-to-extra-python-scripts')],
+   description: 'an example pySpark ETL job',
+   numberOfWorkers: 20,
+});
+
+new glue.pySparkFlexEtlJob(this, 'pySparkFlexEtlJob', {
+   glueVersion: glue.GlueVersion.V3_0,
+   pythonVersion: glue.PythonVersion.3_6,
```
TheRealAmazonKendra, line 180: Same comment about deprecation.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+```
+
+### Python Shell Jobs
+
+A Python shell job runs Python scripts as a shell and supports a Python
+version that depends on the AWS Glue version you are using. This can be used
+to schedule and run tasks that don't require an Apache Spark environment.
+
+We’ll default to `PythonVersion.3_9`. Python shell jobs don't support different
+worker typesbut do have they have a MaxDPU feature. Developers can choose
```
TheRealAmazonKendra, line 194: Can you clarify what you mean here by `Python shell jobs don't support different
worker types`?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-python-script'),
+});
+```
+
+Optional overrides:
+
+```
+new glue.PythonShellJob(this, 'PythonShellJob', {
+    glueVersion: glue.GlueVersion.V1_0,
+    pythonVersion: glue.PythonVersion.3_6,
```
TheRealAmazonKendra, line 210: Same deprecation comment

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+```
+new glue.PythonShellJob(this, 'PythonShellJob', {
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-python-script'),
+});
+```
+
+Optional overrides:
+
+```
+new glue.PythonShellJob(this, 'PythonShellJob', {
```
TheRealAmazonKendra, line 208: This is missing `MaxDPU` from the props.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+});
+```
+
+Developers can override min workers and other Glue job fields
+
+```
+new glue.GlueRayJob(this, 'GlueRayJob', {
+  runtime: glue.Runtime.RAY_2_2,
+  script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-python-script'),
+  minWorkers: 20,
```
TheRealAmazonKendra, line 234: I don't see this set in any examples above. Is this prop unique to this one?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+
+// Create Custom schedule, e.g. Monthly on the 7th day at 15:30 UTC
+myWorkflow.createCustomScheduleJobTrigger((this, 'TriggerCrawlerOnCustomSchedule', {
+    description: 'Scheduled run for ' + myGlueJob.name,
+    actions: [glueJob1, glueJob2, glueJob3, glueCrawler]
+    schedule: events.Schedule.cron(day: '7', hour: '15', minute: '30')
+});
+```
+
+#### **3. Notify  Event Triggers**
```
TheRealAmazonKendra, line 319: This seems to be missing some props in the examples based off the examples.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,410 @@
...
+
+This type of trigger is only supported with Glue workflows. There are two types
+of notify event triggers, batching and non-batching trigger. For batching triggers,
+developers must specify `BatchSize` but for non-batching `BatchSize` will be set
+to 1. For both triggers, `BatchWindow` will be default to 900 seconds.
+
+```
+myWorkflow.createNotifyEventTrigger(this, 'MyNotifyTriggerBatching', {
+    batchSize: batchSize,
```
TheRealAmazonKendra, line 328: I would suggest making this an optional prop that's something like `batching: Batch.set(size: number, window?: Duration)` with `Batch` being a class and `set` being a static function.

TheRealAmazonKendra (dismissed):
This is still missing the full set of proposed props in some places. Please fill those in so I can get an overall view of the user experience for each construct. This is looking very close to where I can provide sign-off but I'm going to also ask for a second set of reviews from a team mate.

text/0497-glue-l2-construct.md
parag-shah-aws: It can support majority of the workload and it was last known released worker type. We recently released G4 and G8 but since its released very recently wanted to keep G2 as default

text/0497-glue-l2-construct.md
parag-shah-aws: Scala jobs doesn't require python version which is why didn't specified

text/0497-glue-l2-construct.md
parag-shah-aws: pySpark is known name for starting spark job written in python.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
parag-shah-aws: We are providing a way to create python shell, ETL (scala and python) or RAY jobs with default values so alpha module style validation won't be required.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
TheRealAmazonKendra: Edit still needed here.

text/0497-glue-l2-construct.md
TheRealAmazonKendra: Our general practice is to make the most up-to-date version the default so unless you have major objections, I'd prefer that.

text/0497-glue-l2-construct.md
TheRealAmazonKendra: But the fact that it's an ETL job is in the name so users shouldn't need to input the job type.

text/0497-glue-l2-construct.md
TheRealAmazonKendra: Got it. Thanks for the clarification!

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+ETL jobs supports python and Scala language. ETL job type supports G1, G2, G4
+and G8 worker type default as G2, which customer can override. It wil default to
+the best practice version of ETL 4.0, but allow developers to override to 3.0.
+We will also default to best practice enablement the following ETL features:
+`—enable-metrics, —enable-spark-ui, —enable-continuous-cloudwatch-log.`
+You can find more details about version, worker type and other features in
+[Glue's public documentation](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-job.html).
+
+```
+glue.ScalaSparkEtlJob(this, 'ScalaSparkEtlJob', {
```
otaviomacedo, line 61: I don't understand the benefit of having one construct per combination of (language, type, subtype). From what I've seen in the `aws-glue-alpha` readme, `JobExecutable` already takes care of these combinations.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+myWorkflow.createCustomScheduleJobTrigger(this, 'TriggerCrawlerOnCustomSchedule', {
+    description: 'Scheduled run for ' + myGlueJob.name,
+    actions: [glueJob1, glueJob2, glueJob3, glueCrawler]
+    schedule: events.Schedule.cron(day: '7', hour: '15', minute: '30')
+});
+```
+
+#### **3. Notify  Event Triggers**
+
+This type of trigger is only supported with Glue workflows. There are two types
```
otaviomacedo, line 346: > This type of trigger is only supported with Glue workflows.

As opposed to what? All examples here show triggers being used with Glue workflows.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+|ExecutionClass|No|STANDARD / FLEX|No|
+|Command PythonVersion|No|3 for ETL/Streaming, 3.9 for PythonShell|
+|LogUri|No|None|This option is not widely used|
+|MaxCapacity (Python Shell Jobs only)|No|0.0625|Yes|
+
+### Uploading scripts from the same repo to S3
+
+Similar to other L2 constructs, the Glue L2 will automate uploading / updating
+scripts to S3 via an optional fromAsset parameter pointing to a script
+in the local file structure. Developers will provide an existing S3 bucket and
```
otaviomacedo, line 268: > Developers will provide an existing S3 bucket and the path to which they'd like the script to be uploaded.

Why do we need this? The whole point of using `fromAsset()` is to hide this implementation detail from the user.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+
+The glue-alpha-module already supports three of the four common types of Glue
+Jobs: Spark (ETL and Streaming), Python Shell, Ray. This RFC will add the
+more recent Flex Job. The construct also implements AWS practice
+recommendations when creating a Glue Job such use of Secrets Management for
+Connection JDBC strings, Glue Job Autoscaling, least privileges in terms of
+IAM permissions and also sane defaults for Glue job specification (more details
+are mentioned in the below table).
+
+This RFC will introduce breaking changes to the existing glue-alpha-module to
```
otaviomacedo, line 42: Can you call out these breaking changes when you introduce them in the text?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+    provide the Secrets Manager Key name as a property to the Job connection
+    property.
+
+* **Networking - CDK determines the best fit subnet for Glue Connection
+configuration
+    **The current glue-alpha-module requires the developer to specify the
+    subnet of the Connection when it’s defined. The developer can still specify the
+    specific subnet they want to use, but no longer have to. This Glue L2 RFC will
+    allow developers to provide only a VPC and either a public or private subnet
+    selection. The L2 will then leverage the existing [EC2 Subnet Selection](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ec2/SubnetSelection.html)
```
otaviomacedo, line 395: Nice! It should have been like this from the beginning.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+convenience features and abstractions for the existing
+[L1 (CloudFormation) Constructs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_Glue.html) building on the
+functionality already supported in the [@aws-cdk/aws-glue-alpha module](https://github.com/aws/aws-cdk/blob/v2.51.1/packages/%40aws-cdk/aws-glue/README.md).
+
+## Create a Glue Job
+
+The glue-alpha-module already supports three of the four common types of Glue
+Jobs: Spark (ETL and Streaming), Python Shell, Ray. This RFC will add the
+more recent Flex Job. The construct also implements AWS practice
+recommendations when creating a Glue Job such use of Secrets Management for
```
kaizencc, line 37:
```suggestion
recommendations when creating a Glue Job such as use of Secrets Management for
```

but honestly this sentence probably could use a bit of rewording. maybe have the best practice recommendations as part of a list?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+
+ETL jobs supports python and Scala language. ETL job type supports G1, G2, G4
+and G8 worker type default as G2, which customer can override. It wil default to
+the best practice version of ETL 4.0, but allow developers to override to 3.0.
+We will also default to best practice enablement the following ETL features:
+`—enable-metrics, —enable-spark-ui, —enable-continuous-cloudwatch-log.`
+You can find more details about version, worker type and other features in
+[Glue's public documentation](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-job.html).
+
+```
```
kaizencc, line 60: can you mark all these examples as ```ts so that they show up with ts syntax highlighting on the rich diff?

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+list using the job or crawler objects using conditional types.
+
+```
+myWorkflow = new glue.Workflow(this, "GlueWorkflow", {
+    name: "MyWorkflow";
+    description: "New Workflow";
+    properties: {'key', 'value'};
+});
+
+myWorkflow.createOnDemandTrigger(this, 'TriggerJobOnDemand', {
```
kaizencc, line 306: Others can chime in if I'm being overly opinionated here, but I feel like CDK has a subtle standardization on avoiding action words like create/add/etc when we can. I think the api here should read `myWorkflow.onDemandTrigger`. 

And my opinion continues for the other APIs that utilize the word `create`

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
+# RFC - Glue CDK L2 Construct
+
+## L2 Construct for AWS Glue Connections, Jobs, and Workflows
+
+* **Original Author(s):** @natalie-white-aws, @mjanardhan @parag-shah-aws
+* **Tracking Issue:**
+* **API Bar Raiser:** @TheRealAmazonKendra
+[Link to RFC Issue](https://github.com/aws/aws-cdk-rfcs/issues/497)
+
+## Working Backwards - README
```
kaizencc, line 10: My comment on this first section of working backwards is that we're not really working backwards from a README. A lot of the material here is (necessary) context, but would not be information found in a README.

kaizencc (changes requested):
Aggregating some important comments that when addressed will make this RFC very strong:

- move context out of README section and treat the README as an actual artifact for what you would like the README to look like when the feature is released. It should follow similar flow/story to our other READMEs, which could mean some research to get the right tone.
- because we are building on an existing Glue L2, we are in a unique situation where BREAKING CHANGES need to be called out and scrutinized in their own section.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,434 @@
...
+Conditional triggers have a predicate and actions associated with them.
+When the predicateCondition is true, the trigger actions will be executed.
+
+```
+// Triggers on Job and Crawler status
+myWorkflow.createConditionalTrigger(this, 'conditionalTrigger', {
+    description: 'Conditional trigger for ' + myGlueJob.name,
+    actions: [glueJob1, glueJob2, glueJob3, glueCrawler]
+    predicateCondition: glue.TriggerPredicateCondition.AND,
+    jobPredicates: [{'job': glueJobPred, 'state': glue.JobRunState.FAILED},
```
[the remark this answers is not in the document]
natalie-white-aws: Would it help if we list the object types rather than variable name placeholders?

text/0497-glue-l2-construct.md
natalie-white-aws: We initially were going to use a factory pattern with separate sub-classes to decide based on the parameters provided, but that puts the onus on the developer to remember what parameters are required and optional for each job type. Kendra preferred having separate methods to instantiate each permutation of valid job type and its associated parameters, which enforces the contract proactively and simplifies the developer experience. The actual objects created will all be JobExecutables, not sub-classes or sub-constructs.

text/0497-glue-l2-construct.md
natalie-white-aws: Technically you can create a Trigger outside of a Glue workflow in the console, CLi, or using L1s, but the L2 will not allow you to do so because it's an anti-pattern. We can remove this or add this construct if it causes confusion.

text/0497-glue-l2-construct.md
natalie-white-aws: Originally there was a question of whether we would support uploading scripts from a local file to S3 or only support existing files in S3. We decided to support both, and all of the other examples show the latter. This section specifically calls out that we'll support the former as well. However, it's not something we can hide from the developer because they have to specify where the file is coming from using fromAsset (local) or fromBucket (already in S3).

text/0497-glue-l2-construct.md
natalie-white-aws: The breaking changes will be related to renaming functions to be more accurately linked to the job types and languages supported, function parameters, and new restrictions on Python and Glue versions (not supporting deprecated versions).  We could add a sentence for each job type saying this maps to [existing alpha function] if that would help clarify the specific breaking changes, but Kendra said the expectation was that since it's an alpha, there will potentially be breaking changes until it's stabilized, and didn't ask for a comprehensive migration guide.

text/0497-glue-l2-construct.md
natalie-white-aws: Agreed!  :)

natalie-white-aws:
> Aggregating some important comments that when addressed will make this RFC very strong:
> 
> * move context out of README section and treat the README as an actual artifact for what you would like the README to look like when the feature is released. It should follow similar flow/story to our other READMEs, which could mean some research to get the right tone.
> * because we are building on an existing Glue L2, we are in a unique situation where BREAKING CHANGES need to be called out and scrutinized in their own section.

I'm not following your guidance on the README section - the format of prior RFCs doesn't seem to have a trend or consistency. Would it be better if we removed the README label altogether and just had "Working Backwards" or "Glue Overview"?

text/0497-glue-l2-construct.md
natalie-white-aws: We'll reword and create a list. We also need to revisit the IAM role comment since I think we decided the developer would provide it since they will have to know what the script is going to do and add the corresponding permissions to it anyway.

text/0497-glue-l2-construct.md
natalie-white-aws: Done.

text/0497-glue-l2-construct.md
natalie-white-aws: Will update.

text/0497-glue-l2-construct.md
natalie-white-aws: We'll rename the section.

text/0497-glue-l2-construct.md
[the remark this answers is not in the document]
otaviomacedo: Yes, I think it would.

text/0497-glue-l2-construct.md
otaviomacedo: Sure, but these methods should have different signatures:

```ts
fromAsset('/path/to/file') // <-- user specifies the local file, but the bucket is decided by the CDK
fromBucket(bucketName, objectKey)
```

text/0497-glue-l2-construct.md
otaviomacedo: I think you should rephrase this to say something along the lines "workflows are mandatory".

natalie-white-aws: They are for the L2 but not for the Glue service itself. I'll add some language to clarify.

text/0497-glue-l2-construct.md
natalie-white-aws: Having two separate methods won't hide the implementation detail from the user, they still have to specify the path or bucket attributes, and they'll have to remember to use a different method to create the job as well. We're following the pattern of Lambda.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,1207 @@
...
+     * @default 3.0 for ETL
+     * */
+    glueVersion?: glue.GlueVersion;
+}
+```
+
+pySpark ETL Job Property Interface:
+
+```ts
+pySparkEtlJobProps{
```
humanzz, line 212: Please note that the pySpark Jobs should also support the `extraJars` property. Eventhough spark can be interfaced with using Python, libraries from jars can still be loaded/leveraged

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,1207 @@
...
+streamline the developer experience and introduce new constants and validations.
+As an opinionated construct, the Glue L2 construct will enforce
+best practices and not allow developers to create resources that use deprecated
+libraries and tool sets (e.g. deprecated versions of Python).
+
+Optional and required parameters for each job will be enforced via interface
+rather than validation; see [Glue's public documentation](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api.html)
+for more granular details.
+
+### Spark Jobs
```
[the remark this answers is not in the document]
mjanardhan: role is something customer is required to provide to be able to run the Glue job with required permissions. We cannot make it optional or use a default role.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,1301 @@
...
+     * */
+    numberOrWorkers?: int;
+
+    /**
+     * Worker Type (optional)
+     * Type of Worker for Glue to use during Job execution
+     * Enum options: Standard, G.1X, G.2X, G.025X. G.4X, G.8X, Z.2X
+     * @default G.2X
+     * */
+    workerType?: glue.WorkerType;
```
TheRealAmazonKendra, line 165: I'm just leaving the comment in this one place, but it stands for each instance here. Since these two are dependent on each other, it should be one field that takes in an enum like class. Something like `Worker.G_1X(200)` with the input being the number of workers.

TheRealAmazonKendra (changes requested):

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,1207 @@
...
+
+Optionally, developers can override the glueVersion and add extra jars and a
+description:
+
+```ts
+glue.ScalaSparkEtlJob(this, 'ScalaSparkEtlJob', {
+   glueVersion: glue.GlueVersion.V3_0,
+   script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-scala-jar'),
+   className: 'com.example.HelloWorld',
+   extraJarsS3Url: [glue.Code.fromBucket('bucket-name', 'path-to-extra-scala-jar'),],
```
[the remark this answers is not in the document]
natalie-white-aws: Addressed this in latest push.

text/0497-glue-l2-construct.md
natalie-white-aws: Addressed this in latest push.

text/0497-glue-l2-construct.md
```diff
@@ -0,0 +1,1207 @@
...
+and G8 worker type default as G2, which customer can override. It wil default to
+the best practice version of ETL 4.0, but allow developers to override to 3.0.
+We will also default to best practice enablement the following ETL features:
+`—enable-metrics, —enable-spark-ui, —enable-continuous-cloudwatch-log.`
+You can find more details about version, worker type and other features in
+[Glue's public documentation](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-job.html).
+
+```ts
+glue.ScalaSparkEtlJob(this, 'ScalaSparkEtlJob', {
+    script: glue.Code.fromBucket('bucket-name', 's3prefix/path-to-scala-jar'),
```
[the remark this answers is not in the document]
natalie-white-aws: Addressed this in latest push.

text/0497-glue-l2-construct.md
parag-shah-aws: Worker type and number of workers are independent of each other which is why kept two different field. Customer can set number of workers and keep worker type as default (G.2X)

text/0497-glue-l2-construct.md
```diff
@@ -7,57 +7,77 @@
...
-* defaults for Glue job specification
-
-This RFC will introduce breaking changes to the existing glue-alpha-module to
-streamline the developer experience and introduce new constants and validations.
-As an opinionated construct, the Glue L2 construct will enforce
-best practices and not allow developers to create resources that use deprecated
-libraries and tool sets (e.g. deprecated versions of Python).
-
-Optional and required parameters for each job will be enforced via interface
+A Job encapsulates a script that connects to data sources, processes
```
kaizencc, line 46: Nice! When you go and write the actual readme in aws-cdk-lib, the first paragraph is the perfect intro + an example. The second paragraph (starting with "This iteraton of the L2 construct") is less readme-ey. We'll call out the breaking changes in the PR description so they show up in the changelog, and customers better know that L2 === opinionated well before reading the readme.

text/0497-glue-l2-construct.md
```diff
@@ -7,57 +7,77 @@
...
 1. **ETL Jobs**
 
-ETL jobs supports python and Scala language. ETL job type supports G1, G2, G4
-and G8 worker type default as G2, which customer can override. It wil default to
-the best practice version of ETL 4.0, but allow developers to override to 3.0.
-We will also default to best practice enablement the following ETL features:
+ETL jobs support pySpark and Scala languages, for which there are separate but
+similar constructors. ETL jobs default to the G2 worker type, but you can
+override this default with other supported worker type values (G1, G2, G4
+and G8). ETL jobs defaults to Glue version 4.0, which you can override to 3.0.
```
kaizencc, line 79: Prefer this sentence to be shown in code actually. Consider removing "which you can override to 3.0" and instead add this to your example:

```ts
glue.ScalaSparkEtlJob(this, 'SSEJ', {
  script: glue.Code.fromBucket('bucket', 'path'),
  className: 'com.example.Hi',
  role: iam.IRole,
  workerType: WorkerType.3_0 // default is WorkerType.4_0
```

Rationale: our examples travel farther than our words. As part of our release process, we search for all constructs to see if they have an inline code example, and if not, we try to source them from the readme. So this example would be a valid option to be the code example for `glue.ScalaSparkEtlJob` and could show up in the cdk docs for the construct. We don't do any magic for our prose, so that will definitely only be in the readme.

This is just one example, but if you see opportunity to improve examples elsewhere, I encourage you to take it with the knowledge that examples end up having more influence.

kaizencc (commented):
Thanks for the changes, @natalie-white-aws! I have 2 non-blocking comments so consider this my approval. Not actually clicking approve because I was more focused on the readme section and @TheRealAmazonKendra may have more comments on implementation.

Edit: sorry I see Kendra added `api-approved` so maybe no more comments on implementation :)

TheRealAmazonKendra (approved):

Issue #507, opened 2023-05-26
Full control over VPC and subnet configuration


otaviomacedo:
## Description

New L2 constructs to allow for greater control over subnet configuration. There are a number of use cases that are not supported by the `Vpc` construct.

See https://github.com/aws/aws-cdk/issues/5927 for more details.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @otaviomacedo
| Author(s)           | @otaviomacedo 
| API Bar Raiser      | @corymhall 
| Stakeholders        | @udaypant, @evgenyka

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #508, opened 2023-05-26
RFC 507: Full control over VPC and subnet configuration

otaviomacedo:
This is a request for comments about a new API in the `aws-ec2` module to give users full control over VPC and subnet configuration. See #507 for additional details. 

APIs are signed off by @corymhall 

[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/otaviom/subnets/text/0507-subnets.md)

---


text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+one or more secondary IP blocks, by providing them in the CIDR format:
+
+```ts
+const vpc = new Vpc(this, 'vpc', {
+  // Primary address block
+  ipAddresses: IpAddresses.ipv4('10.1.0.0/16'),
+});
+
+// The secondary address block must be in the same RFC 1918 range
+vpc.addSecondaryAddressBlock(IpAddresses.ipv4('10.2.0.0/16'));
```
rix0rrr, line 25: I imagine there will also be a `secondaryAddressBlocks?: IpAddresses[]` property?

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+vpc.selectSubnets({subnetType: ec2.SubnetType.PUBLIC}).includes(subnet);
+```
+
+To create a route table that sends DynamoDB and S3 traffic through VPC
+endpoints:
+
+```ts
+const routeTable = new RouteTable(this, 'routeTable', {
+  routes: [
+    Route.DYNAMODB_ENDPOINT,
```
rix0rrr, line 137: Does this implicitly create the endpoint? In all AZs?

Ultimately our enum will have ~200 values if we go this way. Maybe we should go `Route.endpoint(Endpoint.DYNAMODB)` ? That will be more amenable to codegen.

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Working Backwards
+
+### IP addressing
+
+When you create a VPC, in addition to the primary IP block, you can provide
+one or more secondary IP blocks, by providing them in the CIDR format:
+
+```ts
+const vpc = new Vpc(this, 'vpc', {
```
rix0rrr, line 19: Honestly I think we can also go `VpcV2`.

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+const routeTable = vpc.addRouteTable('routeTable', {
+  routes: [
+    Route.to({
+      destination: '0.0.0.0/0',
+
+      // targets must implement the IRouter interface
+      target: natGateway,
+    }),
+  ],
+});
```
rix0rrr, line 166: Do we expect people to do this 3 times? What does that look like, and are we still happy with that?

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+You can also create the same kind of routing pattern, but with a NAT instance
+instead:
+
+```ts
+const natInstance = publicSubnet.addNatInstance('natinst', {
+  instanceType: ec2.InstanceType.of(ec2.InstanceClass.BURSTABLE2, ec2.InstanceSize.MICRO),
+  machineImage: ec2.MachineImage.latestAmazonLinux({
```
rix0rrr, line 181: Wait don't you need a special AMI to serve as NAT instance?

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+    }),
+  ],
+});
+
+const subnet = vpc.addSubnet('publicSubnet', {
+  cidrBlock: '10.2.0.0/20',
+  availabilityZone: 'us-west-2a',
+  routeTable: publicRouteTable,
+  mapPublicIpOnLaunch: false, // default: true for public subnets
+});
```
rix0rrr, line 122: Subnet can create its own route table by default, and then we can also do:

```ts
subnet.routeTable.addRoute(...)
```

Right? Let's support both and leave this pattern for people who want to have exactly one route table the same for all subnets?

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+// 3. Using an Amazon-provided IPv6 CIDR block:
+vpc.addSecondaryAddressBlock(IpAddresses.amazonProvidedIpv6());
+```
+
+### Defining your own subnets
+
+By default, `Vpc` will create subnets for you (by default, a public and a
+private subnet per availability zone). If you want to define your own
+subnets, use the `addSubnet()` method, which will turn off this default
+behavior:
```
rix0rrr, line 71: Is the plan to lazily create the default subnets when `selectSubnets` is called, if `addSubnet` hasn't been called at that point yet? (And then disallow `addSubnet` afterwards?)

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+  routes: [
+    Route.local('10.0.0.0/16'),
+
+    // Static route to a VPN Gateway. Takes priority over propagated ones
+    // Causes VPN gateway to be enabled in the VPC
+    Route.toVpnGateway('172.31.0.0/24'),
+  ],
+
+  // To make VPN Gateway routes propagate to this route table
+  enableVpnGatewayRoutePropagation: true, // default: false
```
rix0rrr, line 235: Is either of these fine or do we need both? The text says either but the code comments don't make it clear.

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+For IPv6 traffic, you create this pattern by using an egress-only internet
+gateway:
+
+```ts
+const routeTable = vpc.addRouteTable('routeTable', {
+  routes: [
+    Route.local('10.0.0.0/16'),
+    Route.local('2001:db8:1234:1a00:/64'),
```
rix0rrr, line 213: Is this worth autoderiving or is that a problem?

text/0507-subnets.md
```diff
@@ -0,0 +1,402 @@
...
+instance where you want to place a component, you can provide it directly.
+For example, to create an `ApplicationLoadBalancer`:
+
+```ts
+const subnet1 = vpc.addSubnet(/*...*/);
+const subnet2 = vpc.addSubnet(/*...*/);
+
+const lb = new elbv2.ApplicationLoadBalancer(this, 'LB', {
+  vpc,
+  subnets: [subnet1, subnet2],
```
rix0rrr, line 270: `subnetSelection` already has these properties, we don't need to invent new ones.

rix0rrr (commented):
I like where this is going!

text/0507-subnets.md
otaviomacedo: Yes, we can add that, too.

text/0507-subnets.md
otaviomacedo: I was thinking of doing it the other way around: let the default subnets be created eagerly (as it is today). But the first `addSubnet` call will override them. Are you concerned about `selectSubnets` returning different values at different points in time?

text/0507-subnets.md
otaviomacedo: Yeah, the text is wrong. You can have both, with the static one taking precedence.

text/0507-subnets.md
otaviomacedo: The parameter will be optional, with the default value being the entire VPC CIDR. But we do need to have a parameter in case you want to route part of the traffic to the local VPC and part to a peer VPC, for example.

text/0507-subnets.md
otaviomacedo: Yes, you're right. I picked a random example, and guess what: it doesn't make sense 🤦

text/0507-subnets.md
otaviomacedo: Agreed.

text/0507-subnets.md
otaviomacedo: What's an alternative that allows users to define 3 identical route tables without repeating the properties? A `clone()` method, perhaps?

```ts
const routeTable = vpc.addRouteTable(...);

vpc.addSubnet('sn1', {
  ...,
  routeTable,
});

vpc.addSubnet('sn2', {
  ...,
  routeTable.clone(),
});

vpc.addSubnet('sn3', {
  ...,
  routeTable.clone(),
});
```

But also, if you are defining your own route tables, why would you do this, instead of reusing the same one?

text/0507-subnets.md
otaviomacedo: What about:

```ts
const dockerEndpoint  = vpc.addInterfaceEndpoint('EcrDockerEndpoint', {
  service: ec2.InterfaceVpcEndpointAwsService.ECR_DOCKER, 
});

new RouteTable(this, 'routeTable', {
  routes: [
    Route.interfaceEndpoint(dockerEndpoint),
  ],
});
```

and

```ts
const dynamoDbEndpoint = vpc.addGatewayEndpoint('DynamoDbEndpoint', {
  service: ec2.GatewayVpcEndpointAwsService.DYNAMODB,
});

new RouteTable(this, 'routeTable', {
  routes: [
    Route.gatewayEndpoint(dynamoDbEndpoint),
  ],
});
```

otaviomacedo: Alright, after doing some experiments, I've changed my mind about this. I think this should create an endpoint. For the gateway endpoints (at most one per service type). For the interface endpoints, the framework also creates them, but the user has to specify the subnets. Changed the text to reflect this.

text/0507-subnets.md
otaviomacedo: I've changed my mind about this, as well. I think it's safer to keep them separate. Doc updated.

tmokmss (commented):
A few minor comments and questions from a frequent user of Vpc. Looking forward to using this feature soon! 🚀

mrpackethead:
Its not clear to me, but it seems SubnetGroups have dissapeared somewhere.  

Subnet groups are an exceptionally powerful things.;    I like having an abstracted 'routing table' associated with a subnetgroup.   ( but underneath it still has individual routing tables for each subnet )..    

I don't think that its desirable to have people needing to add a routing table...   

If you want to do this kind of thing, you can ( using the L1 Constructs which is painful ).

ec2.Vpc, is a L3 Construct,  and this ec2.Vpc2 will be to.     I think the way forward here, is not to create another L3 Construct in the library.   ( But by all means, create them for your own use cases ..    What would be much more valuable is to create a solid set of L2 Constructs. 


For me,  I' have been able to extend ec2.Vpc to do some things that are not easy otherwise.     But really, if there had been some Good L2 Constructs, i'd probalby have started there. 



Underneath, i need subnets to use endpoints ( for cloudwan, transit gateways, firewalls  etc ) that are in their specific AZ,  but i don't want to have to consider that when i'm building my vpc.. Its a deployment detail, that the construct can deal with. 

I also need to be able to route to specific subnets, within my vpc from other subnets, so that i can force traffic via firewalls. this is a feature that was added a while ago, so you can route to smaller routes than the vpc's cidr.

I've essentially created my own EVpc which wraps around eVpc to provide some extra methods to the exisiting vpc.      This is to meet MY specific use-cases.    They are opinionated and solve my problems.     


My fear is that the Vpc2 construct seems more opinionated than the Vpc construct, and while it will meet the needs of some people, for their quite specific use cases, it wont' actually help most people. 

Could we have some examples of where the current ec2.Vpc does not 
  



```
// Define subnetGroups, for use in the enterprise VPC

    const linknet = new network.SubnetGroup(this, 'linknet', {name: 'linknet', subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS, cidrMask: 26});
    const outside = new network.SubnetGroup(this, 'outside', {name: 'outside', subnetType: ec2.SubnetType.PUBLIC, cidrMask: 28});
    const endpoints = new network.SubnetGroup(this, 'endpoints', {name: 'endpoints', subnetType: ec2.SubnetType.PRIVATE_ISOLATED, cidrMask: 24});


sharedServiceVpc.router([
      {
        subnetGroup: linknet,
        routes: [
          { cidr: '10.0.0.0/8', destination: network.Destination.CLOUDWAN, description: 'linknetAllPrivateNetworks' }
        ]
      },
      {
        subnetGroup: endpoints,
        routes: [
          { cidr: '10.0.0.0/8', destination: network.Destination.CLOUDWAN, description: 'endpointsAllPrivateNetworks' }
        ]
      },
      {
        subnetGroup: outside,
        routes: [
          { cidr: '10.0.0.0/8', destination: network.Destination.CLOUDWAN, description: 'outsideAllPrivateNetworks' }
        ]
      },
    ]);
   ```

**Comment by corymhall** [comment]
> ec2.Vpc, is a L3 Construct, and this ec2.Vpc2 will be to. I think the way forward here, is not to create another L3 Construct in the library. ( But by all means, create them for your own use cases .. What would be much more valuable is to create a solid set of L2 Constructs.

+1 to this. I think the first step is to create solid L2s for everything so a user could create their own Vpc from scratch. We could then build more utility on top of that.

File: text/0507-subnets.md [hunk 11]
```diff
@@ -0,0 +1,437 @@
+# Full control over VPC and subnet configuration
+
+* **Original Author(s):**: @otaviomacedo
+* **Tracking Issue**: #507
+* **API Bar Raiser**: @{BAR_RAISER_USER}
+
+New L2 constructs to allow for greater control over VPC design. Among other
```
corymhall, line 7: I think it would be interesting to see a breakdown of all the L1s that are used in creating a Vpc  and which ones we do not have L2 coverage for. To me this breaks down into 3 steps

1. Show how to create the current default `Vpc` with a new complete set of L2s.
2. Show how to create a complicated non-default Vpc with the new set of L2s.
3. Can we make the above two easier with opinionated abstractions

corymhall (changes requested):

text/0507-subnets.md
mrpackethead: I think it might be quite a short list.

- the vpc itself. ( its kind of unforuantate that ec2.Vpc was called Vpc, becuase a vpc in cdk and a vpc in the rest of amazon are two differnet thigns )
- subnets
- route tables.

text/0507-subnets.md
```diff
@@ -0,0 +1,430 @@
...
+```
+
+---
+
+Ticking the box below indicates that the public API of this RFC has been
+signed-off by the API bar raiser (the `api-approved` label was applied to the
+RFC pull request):
+
+```
+[ ] Signed-off by API Bar Raiser @corymhall
```
corymhall, line 329:
```suggestion
[X] Signed-off by API Bar Raiser @corymhall
```

corymhall (approved):

Issue #510, opened 2023-06-06
DyanmoDB Global Table L2 Construct


vinayak-kukreja:
## Description

DynamoDB Global Table lets the user to create a table that can be replicated to other regions. An update to any of the replicas is propagated to other replicas. This enhances availability and reduces latency to write/access table data from different regions.
This feature would enable customers to use an L2 to provision Global Table with CDK. 

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @vinayak-kukreja 
| Author(s)           | @vinayak-kukreja 
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | @aws/aws-cdk-team 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #511, opened 2023-06-14
RFC 510: DynamoDB Global Table L2

vinayak-kukreja:
This is a request for comments about DynamoDB GlobalTable L2 construct. See https://github.com/aws/aws-cdk-rfcs/issues/510 for additional details. 

APIs are signed off by @rix0rrr and @madeline-k .

---


text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+```
+
+##### Provisioned Mode
+
+* __Table level configuration with auto scaling__
+
+```typescript
+  new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    billingMode: BillingMode.PROVISIONED,
```
rix0rrr, line 93: Looks like you decided not to do enum-like classes. Can I ask why not?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+    writeWithAutoScaling: {
```
rix0rrr, line 94: `withAutoScaling`? 

There is no non-autoscaling read/write, so seems like you don't have to provide this?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+  });
+```
+
+__NOTE:__
+
+* Provisioned mode for write capacity could only be used with autoscaling. There is no way to provision
+an exact write capacity unit in global tables.
+* Write capacity could not be configured on a per replica basis. You can only modify read capacity per
+replica.
+* You can either define `readCapacity` or `readWithAutoScaling` but not both at the same time.
```
rix0rrr, line 122: That would be an excellent candidate for an enum-like class then?

```ts
  read: Capacity.fixed(50),
  read: Capacity.autoscaled({ max: 50 })
```

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+    writeWithAutoScaling: {
+      max: 50,
+    },
+    readCapactity: 50,
+  });
+```
+
+__NOTE:__
+
+* Provisioned mode for write capacity could only be used with autoscaling. There is no way to provision
```
rix0rrr, line 118:
```suggestion
* Provisioned mode for write capacity can only be used with autoscaling. There is no way to provision
```

etc

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+an exact write capacity unit in global tables.
+* Write capacity could not be configured on a per replica basis. You can only modify read capacity per
```
rix0rrr, line 120: I would phrase this as "replicas are read-only, so write capacity only applies to the primary table" (or whatever terminology DDB uses for this). Maybe put the fact that replicas are read-only up top as well, that's good to know for people.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+  globalTable.addGlobalSecondaryIndex({
+    indexName: 'FooTableGsi',
+    partitionKey: { name: 'Bar', type: AttributeType.STRING },
+  });
+```
+
+__NOTE:__
+
+* You can create up to 20 global secondary indexes.
+* You can only create or delete one global secondary index in a single stack operation.
```
rix0rrr, line 154: WTF!!!! 

Is this true? 

When I have a table with 5 GSIs, and I want to deploy it to a new region (or create a new replica), is that not possible?

Because if so, we need to make provisions for that so that your definition can be the same but we temporarily turn off GSIs or something so CloudFormation doesn't attempt to deploy them.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+table. There is some configuration that could be specified per replica.
+
+```typescript
+  const globalTable = new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    sortKey: { name: 'Bar', type: AttributeType.STRING },
+    globalSecondaryIndex: [
+      {
+        indexName: 'FooTableGsi',
+        partitionKey: { name: 'Bar', type: AttributeType.STRING },
```
rix0rrr, line 136: What about provisioned capacity of GSIs?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+```typescript
+  new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    encryption: TableEncryption.AWS_MANAGED,
+  });
+```
+
+__NOTE__:
+
+* If the encryption mode is CUSTOMER_MANAGED, then you would need to specify a KMS key for each replica.
```
rix0rrr, line 208: A different KMS key for each region for each replica? What does this look like in practice?

I think cross-stack references will become a problem here, no?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+```typescript
+  new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    stream: StreamViewType.NEW_IMAGE,
+  });
+```
+
+#### Time To Live
+
+You can enable time to live for an attribute of the global table. This would be the same for each replica.
```
rix0rrr, line 232: Use one sentence to explain what this feature does.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+  new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    timeToLiveAttribute: 'Foo',
+  });
+```
+
+#### Replicas
+
+You can define replicas for your global table. By default, if this configuration is undefined, then the
+table is deployed to the region where stack is being deployed.
```
rix0rrr, line 245: This sentence is not clear to me. If my `replicas` array has one element, does that lead to a `table + replica`, or just a `table` (replacing all the properties at the top level) ?

I think I'd prefer the first behavior?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+```typescript
+  globalTable.addReplica({
+    region: 'us-east-1',
+  });
+```
+
+__NOTE:__
+
+* A replica needs to be deployed to the region where stack is deployed to. We do this automatically if
+user has not defined the region in replica.
```
rix0rrr, line 293: I would say we always add the current region automatically, and the user only specifies additional replication regions in their input.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+* `region` --> Needs to be specified
+* `contributorInsightsEnabled` --> Gets copied over from table level props if defined.
+* `deletionProtection` --> Gets copied over from table level props if defined.
+* `pointInTimeRecovery` --> Gets copied over from table level props if defined.
+* `tableClass` --> Gets copied over from table level props if defined.
+* `tags` --> Gets copied over from table level props if defined.
+* `readWithAutoScaling` --> Gets copied over from table level props if defined.
+* `readCapactity` --> Gets copied over from table level props if defined.
+* `encryptionKey` --> This needs to be defined per replica
```
rix0rrr, line 276: This might be tricky to get. We need to get the ARN of a key in a different region.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+        region: 'us-west-2',
+      },
+    ],
+  });
+
+  const lambdaRole = new iam.Role(stack, 'Role', {
+    assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
+  });
+
+  globalTable.grantReadData(lambdaRole, ['us-west-2']);
```
rix0rrr, line 313: You didn't explain this. How does it work?

Also, `lambdaRole` will typically live in a different stack, won't it?

Alternatively, we could do something like this:

```ts
const replica: ITable = globalTable.replica('us-west-2');
replica.grantReadData(lambdaRole);
```

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+  const globalTable = new GlobalTable(stack, 'FooTable', {
+    partitionKey: { name: 'Foo', type: AttributeType.STRING },
+    replicas: [
+      {
+        region: 'us-west-2',
+      },
+    ],
+  });
+
+  const metric = globalTable.metricConsumedWriteCapacityUnits({
```
rix0rrr, line 330: Doesn't this make it impossible for `GlobalTable` to implement `ITable` ?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+ either table name, table arn or table attributes to import it.
+
+```typescript
+GlobalTable.fromTableName(stack, 'FooTableId', 'FooTable');
+```
+
+### NOTE
+
+* Be careful while naming your global table. If there is a name conflict in a region with a DynamoDB table,
+that could lead to removal of that table with the global table replica.
```
rix0rrr, line 347: WAIT WTF AGAIN

This needs to be up top when you talk about naming.

Also the scenario isn't quite clear to me.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+## Public FAQ
+
+### What are we launching today?
+
+We are launching L2 support for DynamoDB Global Table feature.
+
+### Why should I use this feature?
+
+You should use this feature if you would like to provision a DynamoDB table that has the capability
```
rix0rrr, line 369: You would also need to use this feature if:

- You want to import tables using *CloudFormation import*
- You want to use drift detection
- You want to successfully create replicated GSIs for an autoscaled table (our current custom resource doesn't do that properly)

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+We currently offer two solution to provision a global table,
+
+1. Custom CloudFormation Resource within Table construct
+2. L1 for global table resource
+
+This means the customer is not blocked to use global tables. Adding L2 support would take up developer
+time and effort and, would also add to maintenance load for the CDK team.
+
+### What is the technical solution (design) of this feature?
```
rix0rrr, line 407: Please focus on the high level decisions, no need to paste entire source files in here.

The most interesting to me are:

- The design of specifying replicas itself
- Throughput specifications for table, replicas, GSIs
- Per-replica KMS keys (because of the nature of those KMS keys).
- How tables are referenced/used in downstream (replicated) stacks.

 Please include realistic examples that have multiple stacks that deploy to different regions.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+#### Kms key not created for customer managed keys
+
+In the proposed solution, if a user specifies customer managed key as
+their choice on encryption, then they
+would need to specify a key for each replica.
+
+Unlike the Table construct, I am not creating the KMS keys for the customers.
+I believe customer is making a conscious decision of using such encryption
+and should add relevant keys.
```
rix0rrr, line 1623: Probably good from a cost perspective as well.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+To mitigate customer pain in this scenario, we can use a
+[multi region kms key](https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-overview.html)
+ and provision that if none is present.
```
rix0rrr, line 1627: Good find! However, the API for this is tricky -- we'll only be able to get the ordering right using support stacks.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+### What are the drawbacks of this solution?
+
+#### Code Redundancy and maintainance
+
+This solution uses a lot of code that is shared with the Table construct. Even if we try to maximize the
+code shared between Table and Global Table constructs, there would still be some repeated code in each of
+these. This could lead to added maintenance load in the long run since an update to the repeated code in
+one of construct probably would need to be reflected in other as well. If such an update is missed in one
```
rix0rrr, line 1636: Very acceptable. We are `@deprecat`ing the old `Table` construct as soon as this is out. I don't care.

There are some nasty struct inheritances in the old table design. Make sure not to copy those over.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1742 @@
...
+
+#### Deprecating CloudFormation Custom Resource Global Table Support
+
+Since L2 support would now be added that is using the CloudFormation resource, there is no need of the
+custom resource solution. This should be deprecated and users must be informed of this change.
+
+#### RFC for migration path from Table construct to GlobalTable construct
+
+There would be potential users who would like to move from Table construct to GlobalTable construct in
+their CDK app. We need to come up with a solution to support this use case.
```
rix0rrr, line 1653: This will be a blog post describing the steps to go through. At this point there's no need for design anymore, doesn't have to be an RFC.

text/0510-dynamodb-global-table.md
vinayak-kukreja: I see. Could you tell me what you mean by support stacks here?

text/0510-dynamodb-global-table.md
vinayak-kukreja: I was looking into it, and looks like if we can represent a replica as an ITable, then that would help with cross stack references too. Am I understanding this correctly?

text/0510-dynamodb-global-table.md
vinayak-kukreja: Looks like this was just for resources managed by CFN template. :)

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1179 @@
...
+either table name, table arn or table attributes to import it.
+
+```typescript
+GlobalTable.fromTableName(stack, 'FooTableId', 'FooTable');
+```
+
+### NOTE
+
+* We only support version [2019.11.21](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V2.html)
+of the global table since CloudFormation just has support for the same.
```
rix0rrr, line 432: Just a note: "the same" to refer to something earlier in the sentence is a very Indian English thing to say. It's not common in international English.

I'll leave it to you if you want to change it or not, you can be proud of your heritage and let it shine through. But be sure to do it consciously 🤗

rix0rrr (commented):
I like it! There are three big things left in my mind:

- I think a GSI probably shouldn't have a default capacity, we should make the user specify it? At least the default for a replica's GSI capacity should be the base GSI, not the replica table.
- I'd like one final thought to be spent on using enums vs enum-like classes for billing mode (and articulated why you prefer one over the other). I can also see why you might not want to.
- I'd like an example for encryption, it's not clear to me now whether it's a central property or a per-replica property

text/0510-dynamodb-global-table.md
vinayak-kukreja: Ahh, I see. I really appreciate your feedback. I was not aware of it, will update this. :)

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1179 @@
...
+      indexName: 'UniqueGsiName',
+      partitionKey: {
+        name: 'FooRangeKey',
+        type: AttributeType.STRING,
+      },
+    }],
+    replicas: [{
+      region: 'us-east-1',
+      read: Capacity.autoscaled({ max: 70 }),
+      globalSecondaryIndexOptions: [
```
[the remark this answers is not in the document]
vinayak-kukreja: > Was there a conscious decision between a list and a map?

Yes, I was just trying to keep it similar to how GSIs were defined. We can also have it like,

```
{
	'UniqueGsiName': {
		read: Capacity.fixed(55),
		contributorInsightsEnabled: false,
	}
}
```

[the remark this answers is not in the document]
vinayak-kukreja: `read` and `contributorInsightsEnabled` are the only two props that can be configured for the GSI per replica. 

CFN resource: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-globaltable-replicaglobalsecondaryindexspecification.html

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1179 @@
...
+```typescript
+new GlobalTable(tableStack, 'GlobalTable', {
+  tableName: 'FooTable',
+  partitionKey: {
+    name: 'FooHashKey',
+    type: AttributeType.STRING,
+  },
+  billingMode: BillingMode.PROVISIONED,
+  write: Capacity.autoscaled({ max: 70 }),
+  read: Capacity.fixed(20),
```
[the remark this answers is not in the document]
vinayak-kukreja: I see, taking a look.

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1179 @@
...
+
+It is also multi-active database, that means there is no primary table and all the tables created as called
+as replicas and all replicas support reads and writes. Writes to a replica are eventually propagated to other
+replicas where conflicts are resolved by 'last writer wins'.
+
+#### Read and Write Capacity
+
+Global tables by default have `BillingMode.PAY_PER_REQUEST`. If you choose the `BillingMode.PROVISIONED` mode,
+then you will need to specify read and write capacities. If these values are specified at global table level, then
+those values are used for each replica and its global secondary indexes(GSIs).
```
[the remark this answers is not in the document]
vinayak-kukreja: For the GSI capacity, it does make sense to not copy over the read capacity from table, but I believe copying write capacity of table would make sense. WDYT? 

DynamoDB [recommends](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html#GSI.ThroughputConsiderations) the write capacity of GSI to be either same or higher than the table write capacity. 
```
To avoid potential throttling, the provisioned write capacity for a global secondary index should be equal or greater than the write capacity of the base table because new updates write to both the base table and global secondary index. 
```

text/0510-dynamodb-global-table.md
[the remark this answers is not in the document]
vinayak-kukreja: This makes sense to me. One of the reason for keeping it the same was to keep it similar to the table API to minimize user changes. But this would not be that big of a churn to update.

It does make sense, since it is a stronger contract and we do not need `read` and `write` values for `BillingMode.ondemand` which the user can still pass in with the API I had proposed. That can be validated and thrown an error for, but what you are proposing would take away the need for it. :)

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1179 @@
...
+  sortKey: { name: 'Foo', type: AttributeType.STRING },
+});
+```
+
+__NOTE:__
+
+* You need a sort key to define a local secondary index.
+* You can create up to five local secondary indexes.
+
+#### Encryption
```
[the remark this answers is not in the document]
rix0rrr: Again this could be an enum-like class: `

```ts
encryption: TableEncryption.regionKeyArns({
  'us-east-1': 'arn:kms:...',
  'us-east-2': 'arn:kms:...',
});
```

text/0510-dynamodb-global-table.md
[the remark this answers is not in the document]
rix0rrr:
```
{
	'UniqueGsiName': {
		read: Capacity.fixed(55),
		contributorInsightsEnabled: false,
	}
}
```

I kind of enjoy the above. Feels like less ceremony to me. No?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1258 @@
...
+`encryptionKey` property of each replica.
+
+```typescript
+new GlobalTable(tableStack, 'FooTable', {
+  tableName: 'FooGlobalTable',
+  partitionKey: {
+    name: 'FooHashKey',
+    type: AttributeType.STRING,
+  },
+  encryption: GlobalTableEncryption.KEY_ARNS,
```
rix0rrr, line 332: Don't forget the keyARN for the table itself. Which probably shouldn't be an ARN but an `IKey`.

text/0510-dynamodb-global-table.md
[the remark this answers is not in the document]
rix0rrr: Taking into account that the table itself also needs a key, and that should probably be an `IKey` for single-region consistency:

```ts
encryption: TableEncryption.customerManagedKey(tableKey, {
  replicaKeyArns: {
    'us-east-1': 'arn:kms:...',
    'us-east-2': 'arn:kms:...',
  }
});
```

Perhaps?

text/0510-dynamodb-global-table.md
```diff
@@ -0,0 +1,1258 @@
...
+    this.globalTable = new GlobalTable(this, 'FooTable', {
+      tableName: 'FooGlobalTable',
+      partitionKey: {
+        name: 'FooHashKey',
+        type: AttributeType.STRING,
+      },
+      replicas: [
+        {
+          region: 'us-west-2',
+          encryptionKey: 'FooKmsKeyArn',
```
rix0rrr, line 373: Encryption is not really relevant for this example, so if you remove it the example will be shorter and easier to read.

rix0rrr (approved):
Almost there! A couple of final details to think about, but they're very minor. I like this, after you consider the last couple of points I think this is good to merge.

text/0510-dynamodb-global-table.md
vinayak-kukreja: Will update this.

text/0510-dynamodb-global-table.md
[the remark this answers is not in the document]
vinayak-kukreja: Hey, I have just updated the `Replica GSI options` from list to map. I can see this can also be done for `GSI, Replica and LSI`. Let me know if that is what you would prefer. :) 

For me, it makes sense to have them as list since it conveys intent like, `list of global secondary indexes`, or `list of replicas`, but I am fine either way.

rix0rrr (approved):

PR #515, opened 2023-06-29
RFC 513: Application Specific Staging Resources

rix0rrr:
Retrospective RFC #513 

Not the best write-up, but it'll give us a location to start discussing.

[Rendered version](https://github.com/aws/aws-cdk-rfcs/blob/huijbers/app-specific-bootstrapping/text/0513-app-specific-staging.md)

---


text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+* We will introduce a new Stack Synthesizer, called `AppStagingSynthesizer`.
+* This synthesizer will create a support stack with the bucket, and an ECR repository per Docker image.
+* Assets will have a dependency on the support stack. This is a new concept that doesn’t currently exist because assets
+  are an orchestration artifact that looks independent like stacks are, but they aren't really: in practice the orchestration
+  ignores everything except stacks, and treats assets as being part of a stack.
+    * Docker assets may still be built before the first deployment (although for proper caching we need the repository
+      to exist first), but will only be uploaded when it’s their time in the orchestration workflow.
+* For a minimal diff these resources could have fixed names, but we could add support for Stack Outputs and assets could
+  have support for Parameters, so that we can thread generated bucket and repository names through the system. For now,
+  we will do fixed names for the staging resources.
```
corymhall, line 140: Is this going to cause issues for cleaning up resources? If I deploy a test app and then delete it am I going to have a bunch of left over repos and buckets that I have to clean up before I can deploy again?

text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+```ts
+import { AppStagingSynthesizer } from '@aws-cdk/app-staging-synthesizer';
+
+const app = new App({
+  defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({
+    appId: 'my-app-id', // put a unique id here
+    deploymentIdentities: DeploymentIdentities.defaultBootstrapRoles({ bootstrapRegion: 'us-east-1' }),
+
+    // How long to keep File and Docker assets around for rollbacks (without requiring resynth)
+    deployTimeFileAssetLifetime: Duration.days(100),
```
corymhall, line 155: Could this be configured at the Stage/Stack level? I would probably want to be aggressive in non-prod environments.

text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+the way the staging resources are structured, they now allow the use of lifecycle rules, keeping costs down for
+running CDK applications over a long period of time.
+
+### Why should I use this feature?
+
+You should use this feature if you:
+
+- Want to take advantage of lifecycle rules on asset staging resources;
+- Do not use ECR and don't want to see the SecurityHub warning that tells you you have an empty ECR repository;
+- Need to deploy to multiple regions in a set of accounts and want to use StackSets to bootstrap the accounts;
```
corymhall, line 228: Why can't I do this today?

text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+- Want to deploy an application and remove it and be sure that the assets have been cleaned up as well;
```
corymhall, line 229: Are the assets cleaned up?

text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+- We have to clearly explain the concept of Synthesizers, the account contract, and Bootstrapping, along with the choices
+  users have and how they should navigate them in the Developer Guide.
+- Customization by subclassing is possible, but we will probably have to selectively expose some protected helper
+  functions to make it more convenient. We will do that when feature requests start coming in.
+- After a tryout period, we will move the synthesizer into the core library and document it as a possible alternative
+  in the developer guide.
+
+### Are there any open issues that need to be addressed later?
+
+- The template for the staging resources stack must be small enough to fit into a CloudFormation API call, which means
```
corymhall, line 255: ohh chicken/egg issue. I can definitely see this overflow happening.

text/0513-app-specific-staging.md
```diff
@@ -0,0 +1,281 @@
...
+staging resources from the roles. Roles will still be bootstrapped (if used), but staging resources will not.
+
+* Staging resources will be created on a per-CDK app basis. We will create one S3 bucket with different object prefixes
+  for different types of assets (see Appendix A: two types of assets), and an ECR repository per Docker image. Resource
+  access roles can also be created on an as-needed basis. This solves the problems of asset resources of different
+  applications mixing together, and it would also remove the need for garbage collection by allowing use of life cycle
+  rules.
+* Since the roles are now the only things that need to be bootstrapped, that will have a number of advantages:
+    * Bootstrapping will be faster since the heavy resource of a KMS key is no longer involved.
+    * Because roles are a global resource, every account now only needs to be bootstrapped once. First of all the lack
```
corymhall, line 106: Can you add a section to this RFC that goes over the changes needed to the bootstrap template? How will users opt-in to the new bootstrap template? Is there a migration path?

text/0513-app-specific-staging.md
rix0rrr: Hmm. Interesting call. Technically, yes, because the synthesizer can be configured at the app level but also at the stack level; so you can instantiate one with different parameters. However, you'd have to make sure you either use a different appid or don't deploy prod stacks to the same account and region as the dev stacks (otherwise they'd still be sharing resources).

text/0513-app-specific-staging.md
rix0rrr: You can but all Organizations-powered stacksets deploy in one region. So if you want:

- Acct 1, region A
- Acct 2, region B

You would need to add automatic stacksetting for region A and region B, leading you to bootstrap the full Cartesian product of `(1, A), (1, B), (2, A), (2, B)`.

mrgrain (dismissed):

mrgrain (dismissed):

mrgrain (approved):

Issue #487, opened 2023-03-10
New L2 Construct for Step Functions Map State in Distributed Mode


beck3905:
## Description

Step Functions recently released capabilities for a `Map` state in `DISTRIBUTED` mode.  This new capability comes with several new complex parameters.  It seems that adding those parameters as options to the existing `Map` state construct would result in a pretty complicated UX.  So I propose creating a new L2 construct with several supporting classes and interfaces for the various parameters of a `Map` state in `DISTRIBUTED` mode.  This L2 construct would extend the existing `Map` L2 construct.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @beck3905
| Author(s)           | @beck3905
| API Bar Raiser      | @MrArnoldPalmer 
| Stakeholders        | @alias, @alias, @alias

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

mrpackethead:
@evgenyka , who is the bar-raiser?

MrArnoldPalmer:
@mrpackethead I am, taking a look at the proposal.

awsmjs:
Closing this ticket. We believe the functionality is beneficial, but does not intersect with the core framework and should be vended and maintained separately.

PR #529, opened 2023-08-14
RFC 487: Add Distributed Map State L2 Construct

beck3905:
This is a request for comments about Step Functions Distribute Map State CDK L2 Construct. See #487 for
additional details. 

APIs are signed off by @{BAR_RAISER}.

---

### DIFF
--- a/text/0487-distributed-map-construct.md
+++ b/text/0487-distributed-map-construct.md
@@ -1,0 +1,278 @@
+# RFC - Step Functions Distribute Map State CDK L2 Construct
+
+* **Original Author(s):**: @beck3905
+* **Tracking Issue**: #487
+* **API Bar Raiser**: @{BAR_RAISER_USER}
+
+Users will be able to create Map states with the new DISTRIBUTED mode in their Step Functions state machine definitions.
+
+## Working Backwards
+
+The following is the README for the Distributed Map State L2 Construct.
+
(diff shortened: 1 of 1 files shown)


MrArnoldPalmer (changes requested):
Good start, thank you for writing this @beck3905. The readme section overall looks good and it seems most step functions users will be able to get started and navigate basic usage from the documentation available there.

What I would like to see added here are more info on the decision points the current design takes. A couple of things I'm thinking:
1. With alternatives, a natural question is why create a separate class instead of adding a new `mode` prop that defaults to `inline` but could be set to `distributed`, similar to the states language setup. My instinct is that a separate class is easier, more clear, and fits better with the other APIs in CDK. If so do we have any examples of why that all is true?
2. It is mentioned that there may be deviations to what users of amazon states language would expect, what are those differences and why is the deviation worth it. How do we make sure users know about it, does the current documentation cover it?
3. What, if anything, makes the current design easier to use than writing raw JSON definitions? What invariants are we able to enforce during synth?

Additionally, when designing something that is going to be put into `aws-cdk-lib` (as opposed to within an alpha module), what steps have we taken to ensure that we will be able to support this API without breaking changes going forward. Have we made an effort to ensure that we cover use cases that may be uncommon that the service interface and/or cloudformation may support.

Issue #605, opened 2023-12-20
Rewrite EKS L2 construct (EKSv2)


evgenyka:
## Description
When an EKS cluster is created, the only role that has access to the cluster itself (e.g running kubectl commands) is the role that created the cluster. When using CloudFormation, this would be the CloudFormation execution role. Since this role isn’t assumable by anyone, it effectively means it is impossible to connect to the cluster post creation.
In CDK, we workaround this issue by implementing the L2 using **custom resources**, instead of L1s. This allows us to **create** the role that creates the cluster (i.e invokes `eks.CreateCluster` API), and subsequently use this role to grant additional (user defined) roles permissions on the cluster.
The EKS team added a [new feature](https://aws.amazon.com/blogs/containers/a-deep-dive-into-simplified-amazon-eks-access-management-controls/) that allows more control over cluster access. Now it would be possible for CloudFormation to specify a list of roles to be granted access to the cluster, in addition to the role that creates the cluster.
The RFC is to create a new EKS L2 construct and drop the custom resource implementation in favor of the native L1. 

This is going to incur a breaking change that will require cluster replacement (because type will change from Custom::AWSCDK-EKS-Cluster to AWS::EKS::Cluster). Given a breaking change is inevitable, we can decide also to make some additional breaking changes in the API that make it more ergonomic and aligned with the new cluster implementation.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @alias
| Author(s)           | @alias, @alias, @alias
| API Bar Raiser      | @alias
| Stakeholders        | @alias, @alias, @alias

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [ ] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

[...]

PR #646, opened 2024-10-29
RFC605: EKS L2 Rewrite

xazhao:
This is a request for comments about Rewrite EKS L2 Construct. See #605 for additional details.

APIs are signed off by @iliapolo.

---


text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+  outputConfigCommand: true,
+});
+```
+
+After
+
+```
+const cluster = new eks.Cluster(this, 'Cluster', {
+  version: eks.KubernetesVersion.V1_31,
+  outputInfo: true,
```
iliapolo, line 237: I'm tempted to just let the user create `CfnOutput`s if they so wish. We just make sure the cluster exposes its information as properties. What do you think?

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+### Kubectl Handler Configuration
+
+KubectlProvider is a lambda function that CDK deploys alongside the EKS cluster
+in order to execute kubectl commands against the cluster.
+
+A common scenarios is that users create a CDK app that deploys the EKS cluster,
+which is then imported in other apps in order to deploy resources onto the
+cluster.
+
+Difference
```
iliapolo, line 250: Please extend this to show how resources can be deployed to imported clusters, not just how to import the kubectl provider. The ergonomics around this experience is very awkward and I would like to improve it. 

Right now, to deploy resources onto an imported cluster, we can do:

```ts
const clusterWithKubectlHandler = eks.Cluster.fromAttributes(this, 'Cluster', {
  kubectlFunctionArn: '',
});

clusterWithKubectlHandler.addManifest(...) // will reuse the existing kubectl function

const clusterWithoutKubectlHandler = eks.Cluster.fromAttributes(this, 'Cluster', {
});

// will create a new kubectl function that can deploy to the cluster
// but in this case, additional attributes must be provided when importing the cluster (e.g the role)
// this is confusing. 
clusterWithoutKubectlHandler.addManifest(...) 
```

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+
+```
+const kubectlProvider = eks.KubectlProvider.fromKubectlProviderArn(this, 'KubectlProvider', {
+  functionArn: // Required. ARN of the original kubectl function
+});
+```
+
+Following parameters are removed:
+
+- kubectlRoleArn
```
iliapolo, line 272: So you're suggesting we bake this into the provider itself? I like it - just want to make sure.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+```
+new eks.Cluster(this, 'MyCluster', {
+  version: eks.KubernetesVersion.V1_31,
+  kubectlLayer: new KubectlV31Layer(this, 'kubectl'),
+  kubectlProviderOptions: {
+    memory: Size.gigabytes(4),
+    environment: {
+      'http_proxy': 'http://proxy.myproxy.com',
+    },
+    role: iam.Role.fromRoleArn(
```
iliapolo, line 329: Will this have a default? is it the role we will use to connect to the cluster?

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+      this,
+      'MyRole',
+      'arn:aws:iam::123456789012:role/lambda-role'
+    );
+});
+```
+
+## Migration Path
+
+Note: We can't guarantee it's a safe migration.
```
iliapolo, line 339: Why not?

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+
+Due to the fact that switching from a custom resource
+(Custom::AWSCDK-EKS-Cluster) to a native L1 (AWS::EKS::Cluster) resource
+requires cluster replacement, CDK users who need to preserve their cluster will
+have to take additional actions.
```
iliapolo, line 344: We will need a blog post about this I think. Also - lets consider creating a dedicated migration tool for customers to use, this is too many manual operations.

iliapolo: Please add explanations to each of these steps.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+The new EKS module provides faster deployment, less complexity, less cost and
+more features (e.g. isolated VPC and escape hatching).
+
+### What's the future plan for existing `aws-eks` module?
+
+- When the new alpha module is published, `aws-eks` module will enter
+  `maintenance` mode which means we will only work on bugs on `aws-eks` module.
+  New features will only be added to the new `aws-eksv2-alpha` module. (Note:
+  this is the general guideline and we might be flexible on this)
+- When the new alpha module is stabilized, `aws-eks` module will transition into
```
iliapolo, line 380: Will we still address critical bugs? When will we stop?

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+  a deprecation phase. This implies that customers should plan to migrate their
+  workloads to the new module. While they can continue using the old module for
+  the time being, CDK team will prioritize new features/bug fix on the new
+  module
+
+## Internal FAQ
+
+### Why are we doing this?
+
+This feature has been highly requested by the community since Feb 2023. The
```
iliapolo, line 390: Show evidence. Lets also put the list of issues that are going to get resolved with this change.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+The migration for customers is not easy and we can't guarantee it's a safe
+migration without down time.
+
+### Is this a breaking change?
+
+Yes it's breaking change hence it's put into a new alpha module. A few other
+breaking changes are shipped together to make it more ergonomic and aligned with
+the new cluster implementation.
+
+### What is the high-level project plan?
```
iliapolo, line 405: Incorporate the migration path work here as well - it will take quite some time to write a tool/blog-post and test it out.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+- [ ] Gather feedback on the RFC
+- [ ] Get bar raiser to sign off on RFC
+- [ ] Create the new eksv2 alpha module and implementation
+- [ ] Make pull request to aws-cdk repository
+- [ ] Iterate and respond to PR feedback
+- [ ] Merge new module
+
+### Are there any open issues that need to be addressed later?
+
+TBD
```
iliapolo, line 417: Stabilization of this module should be prioritized after the bake time.

iliapolo (changes requested):

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+
+```
+new eks.Cluster(this, 'MyCluster', {
+  version: eks.KubernetesVersion.V1_31,
+  kubectlMemory: Size.gigabytes(4),
+  kubectlLayer: new KubectlV31Layer(this, 'kubectl'),
+  kubectlEnvironment: {
+    'http_proxy': 'http://proxy.myproxy.com',
+  },
+  kubectlRole: iam.Role.fromRoleArn(
```
iliapolo, line 310: Mention if `kubectlRole` is required or not. In either case, mention what are its requirements, i.e:

- Can be assumed by lambda
- Has master credentials to the cluster

text/0605-eks-rewrite.md
iliapolo: Mention that `kubectlRole` will be granted admin permissions on the cluster.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+3. Set removal policy to RETAIN on the existing cluster (and manifests) and
+   deploy.
+4. Remove cluster definition from their CDK app and deploy
+5. Add new cluster definition using the new constructs(EKSV2).
+6. Follow cdk import to import the existing cluster as the new definition.
+   1. All relevant EKS resources support import.
+   2. AWS::EKS::Cluster
+   3. AWS::EKS::FargateProfile
+   4. AWS::EKS::Nodegroup
+7. Add Manifests.
```
iliapolo, line 360: Lets set a high bar for the migration experience. This will be the biggest pain-point for users.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,476 @@
...
+Yes it's breaking change hence it's put into a new alpha module. A few other
+breaking changes are shipped together to make it more ergonomic and aligned with
+the new cluster implementation.
+
+### What is the high-level project plan?
+
+- [ ] Publish the RFC
+- [ ] Gather feedback on the RFC
+- [ ] Get bar raiser to sign off on RFC
+- [ ] Create the new eksv2 alpha module and implementation
```
iliapolo, line 410: We don't necessarily have to publish migration instructions here, but we should test the migration is even possible.

text/0605-eks-rewrite.md
xazhao: Add different scenarios for importing cluster/kubectl handler

text/0605-eks-rewrite.md
xazhao: `kubectlRoleArn` is not needed because `role` will be granted with admin permissions to the cluster.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,448 @@
...
+resource approach for creating EKS clusters and Fargate Profiles. Beyond the
+resource types change, the module introduces several other breaking changes
+designed to better align the L2 construct with its updated implementation. These
+modifications collectively enhance the module's functionality and integration
+within the AWS ecosystem, providing a more robust and streamlined solution for
+managing Elastic Kubernetes Service (EKS) resources.
+
+This RFC primarily addresses the distinctions between the new module and the
+original EKS L2 construct. Comprehensive use cases and examples will be
+available in the README file of the forthcoming `eks-alpha-v2` module.
```
adamjkeller, line 18: eks-v2-alpha or do we want to reference the actual module name aws-eksv2-alpha?

xazhao: `eks-v2-alpha` looks more clean to me. I'm open to either way though.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,448 @@
...
+const kubectlProvider = eks.KubectlProvider.fromKubectlProviderArn(this, 'KubectlProvider', {
+  functionArn: ''
+});
+const cluster = eks.Cluster.fromClusterAttributes(this, 'MyCluster', {
+  clusterName: 'my-cluster-name',
+  kubectlProvider: kubectlProvider
+});
+```
+
+## Migration Guide
```
QuantumNeuralCoder, line 259: I think we need answers to the following questions

Whats the feasibility of the migration? 
Whats possible whats breaking?
How much effort will we. need to upgrade a construct written in v1 and port it to v2.

text/0605-eks-rewrite.md
xazhao: Discussed offline and agreed we will include a general migration guide in the RFC.

text/0605-eks-rewrite.md
xazhao: Added.

text/0605-eks-rewrite.md
xazhao: Added a list of Github issues that will be resolved with the new module.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,500 @@
...
+This RFC primarily addresses the distinctions between the new module and the
+original EKS L2 construct. Comprehensive use cases and examples will be
+available in the README file of the forthcoming `eks-alpha-v2` module.
+
+Compared to the original EKS module, it has following major changes:
+
+- Use native L1 `AWS::EKS::Cluster` resource to replace custom resource `Custom::AWSCDK-EKS-Cluster`
+- Use native L1 `AWS::EKS::FargateProfile` resource to replace custom resource `Custom::AWSCDK-EKS-FargateProfile`
+- `Kubectl Handler` will not be created by default. It will only be created if users specify it.
+- Deprecate `AwsAuth` construct. Permissions to the cluster will be managed by Access Entry.
```
iliapolo, line 25:
```suggestion
- Remove `AwsAuth` construct. Permissions to the cluster will be managed by Access Entry.
```

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,500 @@
...
+ +--------------------+
+```
+
+In a nutshell:
+
+- EKS Cluster - The cluster endpoint created by EKS.
+- Managed Node Group - EC2 worker nodes managed by EKS.
+- Fargate Profile - Fargate worker nodes managed by EKS.
+- Auto Scaling Group - EC2 worker nodes managed by the user.
+- Kubectl Handler (Optional) - Lambda function for invoking kubectl commands on the
```
iliapolo, line 89:
```suggestion
- Kubectl Handler (Optional) - Custom resource (i.e Lambda Function) for invoking kubectl commands on the
```

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,500 @@
...
+
+const cluster = eks.Cluster.fromClusterAttributes(this, 'MyCluster', {
+  clusterName: 'my-cluster-name',
+  kubectlProvider: kubectlProvider
+});
+
+cluster.addManifest();
+```
+
+2. Import the cluster and create a new kubectl Handler
```
iliapolo, line 244: Do we need a separate `fromClusterWithKubectlProvider` function for this use case? 

How about:

```ts
const clusterName = 'my-cluster-name';
const kubectlProvider = new eks.KubectlProvider(this, 'KubectlProvier', {
  clusterName,
})

const cluster = eks.Cluster.fromClusterAttributes(this, 'MyCluster', {
  clusterName,
  kubectlProvider
});
```

This way we just have one `fromXXX` function for cluster import. Do you see any problems with this approach?

xazhao: Are you saying `kubectlProviderOptions` will not be exposed in `fromClusterAttributes()`?

That makes sense to me and I like it. Will update the RFC.

text/0605-eks-rewrite.md
xazhao: @dsilbergleithcu-godaddy Yes. As long as the cluster can be imported to CFN, I don't see a blocker of importing them into CDK and manage them with CDK.

xazhao: @QuantumNeuralCoder 
1. The feasibility of the migration depends on how complicated the stack is. The step that takes most time is to use `cdk import` to import the new resource.
2. The only step that could fail is `cdk import`. Users should use `cdk diff` to verify all diffs before moving forward.
3. These are manual steps but I don't think that will take lots of time. Should be able to migrate in 1 day or 2.

markusl:
Is [EKS Auto Mode](https://aws.amazon.com/eks/auto-mode/) supported in the new construct?

xazhao:
> Is [EKS Auto Mode](https://aws.amazon.com/eks/auto-mode/) supported in the new construct?

Yes because `EKS Auto Mode` is natively supported in `CfnCluster` L1 resource, there's no blocker to support it in the new construct. We will decide a proper API for EKS Auto Mode and it will be added to the construct as a new feature.

markusl:
> > Is [EKS Auto Mode](https://aws.amazon.com/eks/auto-mode/) supported in the new construct?
> 
> Yes because `EKS Auto Mode` is natively supported in `CfnCluster` L1 resource, there's no blocker to support it in the new construct. We will decide a proper API for EKS Auto Mode and it will be added to the construct as a new feature.

Thanks! This sounds good. Is there any estimate for the release with Auto Mode?

xazhao:
> > > Is [EKS Auto Mode](https://aws.amazon.com/eks/auto-mode/) supported in the new construct?
> > 
> > 
> > Yes because `EKS Auto Mode` is natively supported in `CfnCluster` L1 resource, there's no blocker to support it in the new construct. We will decide a proper API for EKS Auto Mode and it will be added to the construct as a new feature.
> 
> Thanks! This sounds good. Is there any estimate for the release with Auto Mode?

The current estimate is 2025 Q1.

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,437 @@
...
+permissions to an IAM role/user.
+
+As a result, `AwsAuth` construct in the previous EKS module will not be provided in the new module.
+`grant()` functions are introduced to replace the awsAuth. It’s implemented using
+Access Entry.
+
+Grant Admin Access to an IAM role
+
+```
+cluster.grantAdmin('adminAccess', roleArn, eks.AccessScopeType.CLUSTER);
```
iliapolo, line 197: Can we make the second argument here accept an [IGrantable](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-iam/lib/principals.ts#L14) and not a string?

iliapolo: Does this function return something? an `AccessEntry` maybe?

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+You can also use general `grantAccess()` to attach a policy to an IAM role/user.
+See https://docs.aws.amazon.com/eks/latest/userguide/access-policies.html for all access policies
+
+```
+# A general grant function is also provided
```
iliapolo, line 204:
```suggestion
# A general grant function is also provided, where you can explicitly set policies.
```

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,437 @@
...
+
+This imported cluster is not associated with a `Kubectl Handler`. It means we won't be able to
+invoke `addManifest()` function on the cluster.
+
+2. Import the cluster and the kubectl Handler
+
+```
+const kubectlProvider = eks.KubectlProvider.fromKubectlProviderArn(this, 'KubectlProvider', {
+  functionArn: ''
+});
```
iliapolo, line 235: This should either:

```ts
const kubectlProvider = eks.KubectlProvider.fromKubectlProviderArn(this, 'KubectlProvider', 'function-arn');
```

or 

```ts
const kubectlProvider = eks.KubectlProvider.fromKubectlProviderAttributes(this, 'KubectlProvider', {
  functionArn: ''
});
```

text/0605-eks-rewrite.md
```diff
@@ -0,0 +1,437 @@
...
+    - `AWS::EKS::Nodegroup`
+    - `AWS::EKS::AccessEntry`
+
+6. **After `cdk import`, running `cdk diff` to see if there's any unexpected changes.**
+
+## Public FAQ
+
+### What are we launching today?
+
+We’re launching a new EKS module `aws-eks-v2-alpha`. It's a rewrite of existing
```
iliapolo, line 308:
```suggestion
We’re launching a new EKS alpha module `aws-eks-v2-alpha`. It's a rewrite of existing
```

text/0605-eks-rewrite.md
xazhao: This API is basically the current `grantAccess()` with hardcode values: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_eks.Cluster.html#grantwbraccessid-principal-accesspolicies

> Can we make the second argument here accept an [IGrantable](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-iam/lib/principals.ts#L14) and not a string?

Yes I agree.

> Does this function return something? an `AccessEntry` maybe?

It's the same as current API which is `void`.

> Is the access scope argument mandatory? what would be the API for granting admin on a specific namespace?


Actually I think we should remove the third argument to make `grantAdmin()` easier to use. So something like:
```
cluster.grantAdmin('adminAccess', role);
```
It will add `AmazonEKSClusterAdminPolicy` to the role. There's also AmazonEKSAdminPolicy which can be applied to a namespace. Have 2 `grantAdmin` is a bit confusing. Customers can use the general grant method for granting admin on a specific namespace. We can add more`grantXXX()` later.

text/0605-eks-rewrite.md
xazhao: Changed it to 
```
const kubectlProvider = eks.KubectlProvider.fromKubectlProviderArn(this, 'KubectlProvider', 'function-arn');
```

PR #654, opened 2024-11-12
RFC300: Programmatic Toolkit

mrgrain:
This is a request for comments about Programmatic Toolkit. See #300  for
additional details. 

[Rendered Version](https://github.com/aws/aws-cdk-rfcs/blob/mrgrain/0300-programmatic-toolkit/text/0300-programmatic-toolkit.md)

---


text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+// Define the CloudExecutable
+const cx = CloudExecutable.fromCloudAssemblyProducer(async (context) => {
+  const app = new cdk.App({ context });
+  const stack = new cdk.Stack(app);
+  return app.synth();
+});
+
+// Synth and store a cached CloudExecutable, this won't run synth again
+const cxCache = await cdk.synth(cx);
```
rix0rrr, line 40: Is `cxCache` the same type as `cx`? Or implements the same interface?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+const cdk = new Toolkit()
+```
+
+You can customize the functionality of the toolkit.
+The main options are provided through `ioHost` and `awsSdkProvider`.
+
+```ts
+const cdk = new Toolkit({
+  ioHost: new MyIoHost(),
+  awsSdkProvider: new MyAwsSdkProvider(),
```
rix0rrr, line 63: I think it's probably fine to do something more simple here, and only provide configuration for the SDK.

It's still possible to do what we always do: if we feel we need more control, add an optional `sdkProvider?:` later and use runtime checks to make it mutually exclusive with configuration properties.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+```
+
+##### Changes and Breaking Changes to messages and requests
+
+Messages and requests play an important role in structuring interactions with the Toolkit.
+Because messages will be emitted in many places, they will change frequently.
+An integrator needs to take great care when handling messages and basing implementation on them.
+A list of all messages and requests is available at [TODO: link to auto-generated list of messages].
+
+For a given message code, the following changes are considered breaking:
```
rix0rrr, line 99: I think before you start classifying into breaking and non-breaking, a statement of intent around backwards compatibility would be welcome.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+Message receivers need to implement the `IIoHost` interface.
+The `IoHost` is responsible for handling all logging and interactions:
+
+```ts
+interface IIoHost {
+  /**
+   * Notifies the host of a message.
+   * The caller waits until the notification completes.
+   */
+  async notify(msg: IoMessage<T>);
```
rix0rrr, line 129: The `T` in `IoMessage<T>` is unbound, and so doesn't do anything. Effectively you're saying `IoMessage<any>`.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+  
+  /**
+   * Notifies the host of a message that requires a response.
+   * 
+   * If the host does not return a response the suggested
+   * default response from the input message will be used.
+   */
+  async requestResponse(msg: IoRequest<T,U>): U | undefined;
```
rix0rrr, line 137: Same here.

```
 async requestResponse(msg: IoRequest<any, any>): any | undefined;
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+}
+```
+
+For every message, the toolkit calls `notify` or `requestResponse` is called respectively.
```
rix0rrr, line 141:
```suggestion
For every message, the toolkit calls `notify` or `requestResponse`, respectively.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+The a generic `IoEmitter` is available to provide a familiar event emitter interface.
```
rix0rrr, line 143:
```suggestion
A generic `IoEmitter` is available to provide a familiar event emitter interface.
```

rix0rrr: I actually don't like `EventEmitter` as an API, since it is a single object that  allows both listening (good) and trigger (not good: the user isn't in control of triggering, the toolkit is)

Do we have `EventListener` as well?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+The toolkit emits *messages* and *requests* to structure interactions.
+A *request* is a *message* that requires the receiver to respond, otherwise the toolkit will continue with a default response.
+
+For every emitted message, the toolkit calls the `notify` or `requestResponse` method of an `IoHost` (more on this below).
+Every call is `await`ed for its completion, allowing to receiver to perform asynchronous operations like HTTP calls.
+It's okay for the receiver not do anything with the input and return immediately.
+The type definition of messages and requests is as follows:
+
+```ts
+interface IoMessage<T> {
```
rix0rrr, line 78: I think this needs to be a tagged union.

```typescript
type IoMessage = 
  | { code: 'deploy' } & TypedIoMessage<{ arn: string }>
  | { code 'message' } & TypedIoMessage<string>
  | ...
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+For every message, the toolkit calls `notify` or `requestResponse` is called respectively.
+
+The a generic `IoEmitter` is available to provide a familiar event emitter interface.
+This implementation will do nothing for messages and requests that don't have a registered listener, i.e. the default response will be used as required.
+
+```ts
+const io = new IoEmitter();
+io.on('deploy', async (msg: IoMessage) => { /* do something */ });
+io.once('deploy:confirm-security', async (msg: IoRequest<boolean>): boolean => {
```
rix0rrr, line 149: I suppose this makes me think of DOM events, where we could use methods on the `Event` object to modify it. That's not too bad either? 

```ts
io.once('deploy:confirm-security', async (event) => {
  event.preventDefault();
});
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+  // reject the deployment
+  return false;
+})
+```
+
+The `CommandLineIoHost` is available, implementing all the interaction behavior for the AWS CDK CLI [TODO: this might end up in a separate package]:
+
+```ts
+const io = new CdkCliHost({
+  isInteractive: true
```
rix0rrr, line 159: What is the effect of this going to be?

My suspicion is that `Toolkit` itself is going to have to have a notion of interactivity already. Then it seems simpler to just say:

```ts
new Toolkit({
  interactive: false
  ioHost: new JustLogToTheScreen(),
});
```

And the implementation of JustLogToTheScreen

```ts
class JustLogToTheScreen {
  async function requestResponse() {
    throw new Error('I cannot do that Jim');
  }
}
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+A standard `SdkProviderForCdk` is available that includes all features of the AWS CDK CLI [TODO: this might end up in a separate package].
+
+#### Cloud Executables
+
+A Cloud Executable is the primitive that defines a deployable AWS CDK application.
+AWS CDK apps might need to be synthesized multiple times with additional context values before they are ready.
+This concept is what the Cloud Executable represents.
+
+The most flexible way to obtain a Cloud Executable is from a function returning a `CloudAssembly`.
+Providing a function allows your implementation to integrate with the [Programmatic Toolkit]'s multi-pass requirement.
```
rix0rrr, line 179:
```suggestion
Providing a function allows your implementation to integrate with the [Programmatic Toolkit]'s automatic context fetching system.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+The function will be called one or more times with the known context values of the current iteration.
+It must then use these values to synthesize and return a `CloudAssembly`.
+For all features (e.g. lookups) to work correctly, the `App()` must be instantiated with the received context values.
+Since it is not possible to update the context of an app, it must be created inside the function scope.
```
rix0rrr, line 183:
```suggestion
Since it is not possible to update the context of a `cdk.App`, it must be created inside the function scope.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+#### Cloud Executables
+
+A Cloud Executable is the primitive that defines a deployable AWS CDK application.
+AWS CDK apps might need to be synthesized multiple times with additional context values before they are ready.
+This concept is what the Cloud Executable represents.
+
+The most flexible way to obtain a Cloud Executable is from a function returning a `CloudAssembly`.
+Providing a function allows your implementation to integrate with the [Programmatic Toolkit]'s multi-pass requirement.
+The function will be called one or more times with the known context values of the current iteration.
+It must then use these values to synthesize and return a `CloudAssembly`.
```
rix0rrr, line 181:
```suggestion
It must then use these values to synthesize and return a `CloudAssembly`.
If the Cloud Assembly contains metadata that indicates that context information is missing, the toolkit will perform lookup AWS API calls using the appropriate credentials, and invoke the cloud executable again.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+    const stack = new cdk.Stack(app);
+    return app.synth();
+  }
+);
+```
+
+Existing applications, can also be used by simply passing an AWS CDK app directory (containing a `cdk.json` file):
+
+```ts
+const cx: ICloudExecutable = CloudExecutable.fromCdkApp("/path/to/cdk/app");
```
rix0rrr, line 200: There's also a mode where we do a `CloudExecutable` from an external executable without the use of a `cdk.json` (pass the info that's normally in `cdk.json` directly)

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+It returns a list of stack deployments.
+
+#### Errors
+
+All errors are thrown as exceptions and the integrator must be prepared to handle them.
+Expected errors are structured data and extend the `IoMessage` interface.
+
+```ts
+interface ToolkitError extends IoMessage {
```
rix0rrr, line 236: I would expect

```ts
class ToolkitError extends Error { ... }
```

What's the value in extending `IoMessage`?

Do we really need `level`, `time`, `command` ... ? (`code` of course is very useful)

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+To assist integrators with detecting the type of error, the following helper methods are available.
+Even though errors are typed, you should not rely on `instanceof` checks because it can behave unexpectedly when working with multiple copies of the same package. [TODO: insert reference link]
+
+```ts
+try {
+  await cdk.deploy(...);
+} catch (e: unknown) {
+  if (ToolkitError.isToolkitError(error)) {
+    // handle toolkit errors
+  } else if (ToolkitError.isUserError(e)) {
```
rix0rrr, line 254: What does this mean? You just said if we throw anything other than a `ToolkitError` it's a bug.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+    // handle user code errors
+  }
+  
+  if (ToolkitError.isCdkError(e)) {
```
rix0rrr, line 258: What is a CDK Error?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+```ts
+const cdk = new Toolkit({
+  ioHost: new MyIoHost(),
+  awsSdkProvider: new MyAwsSdkProvider(),
+})
+```
+
+#### Messages and Requests
+
+The toolkit emits *messages* and *requests* to structure interactions.
```
otaviomacedo, line 69: An example of a sequence of messages back and forth would be nice.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+```ts
+interface RecoverableError extends IoRequest<T, U> {
+  level: 'error';
+  defaultResponse: {
+    retry: boolean; // indicates if the block should retried
+    // ... other response values
+  };
+  data: {
+    isRecoverable: true;
```
rix0rrr, line 284: I wouldn't have expected this to be part of `data`, but part of the top-level struct?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+```
+
+When `IoHost.requestResponse()` is called with a recoverable error, the integrator may choose to return `retry: true` to indicate the block should be retried.
+A single retry is allowed, if the block fails again with the same error it will be raised as an exception.
+Retries are not guaranteed and integrators must handle the case where a requested retry is not executed.
+Recoverable errors are rare as they require special programming and not always possible.
+
+##### Synth-time Errors
+
+There is a subtle semantic difference between errors that originate from the [Programmatic Toolkit] and from a Cloud Executable, i.e. from a user's app:
```
rix0rrr, line 297: I disagree, that's not subtle at all! 😆

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+Errors from the [Programmatic Toolkit] are typically misconfigurations and maybe fixable by an integrator,
+while errors from a Cloud Executable are usually code problems and only fixable by the user.
+
+The [Programmatic Toolkit] emits all synth-time errors as recoverable errors.
```
rix0rrr, line 301: This surprises me. How do you expect an integrator to recover from this?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+```ts
+const io = new CdkCliHost({
+  isInteractive: true
+})
+```
+
+##### SdkProvider
+
+  [Alternative: Make this `CredentialsProvider` or `SdkConfigProvider` instead]
```
otaviomacedo, line 165: This is an opportunity to try to redesign the SDK abstraction, to allow for less traumatic version updates.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+#### IoHost
+
+This is the central piece of the proposal.
+Based on tenet T5, we are aiming to separate actions from interactions.
+This is important because programmatic use is inherently different to interactive use.
+
+A good example for this is `cdk deploy` and the `--rollback` / `--no-rollback` options.
+Typically `--no-rollback` is used to increase iteration speed during development and a production deployment would run with `--rollback`.
+However there are situations in which a rollback is still preferable.
```
rix0rrr, line 469:
```suggestion
However there are situations in which deploying in rollback mode is necessary, or even an explicit rollback needs to be performed before a deployment can continue.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+Both together make up the interface of `IoHost`.
+See the readme above for more details.
+
+#### Messages and Requests
+
+The proposed `IoHost` design requires us to commit to a stable set of messages and requests send between [Programmatic Toolkit] and integrator.
+A detailed definition of what would constitute a breaking change is included in the readme above.
+
+To support integrations, we will need to publish a list of all messages.
+We will use static code analysis to automatically compile and publish this list without adding additional maintenance burden on to the team.
```
rix0rrr, line 495: Any thoughts to how that would work, or TBD?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+#### SdkProvider
+
+The main purpose of this feature is to expose SDK configuration to an integrator.
+We will be providing our default implementation as a drop-in solution for CLI-like environments.
+However other possible scenarios like Electron or even web apps will require much different setups.
```
rix0rrr, line 501: Or the VSCode IDE plugin

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+
+The [Programmatic Toolkit] aims to solve these limitations by providing a proper programmatic interface to CDK's functionality,
+sallowing integrators to have full control over all output and combine CDK actions with custom steps in their applications.
+With this project, we are changing the paradigm.
+Programmatic usage will become an equally considered use case, in fact the proposed implementation will require it to be considered first.
+
+### Why should we _not_ do this?
+
+There are requirements that cannot be satisfied with the current available options.
+Specifically hooks, control over logging, structured errors and typed returns are not possible.
```
sumupitchayan, line 430: Hooks = CFN hooks?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,745 @@
...
+The proposed design aims to resolve these tensions.* Here are some explicit call-outs:
+
+*T3 TypeScript vs T4 jsii ecosystem* - The tension is that participating in the jsii ecosystem usually means that all features must be jsii compatible.
+It implies a lo-fi solutions compared to native TypeScript.
+Both tenets are aimed at different use cases.
+A tool absorbing CDK actions management will realistically be build in TypeScript and should be able to make the full use of it.
+A jsii package will benefit the scripting community and use in Notebooks.
+There are naturally some TypeScript features that will be impossible to implement for jsii, e.g. Error types or custom SDK providers.
+
+*T2 Real use cases vs T4 jsii ecosystem* - The tension is that there is little evidence that users want a jsii version of this.
```
sumupitchayan, line 673: Is there a way to find evidence of customer demand for this?

text/0300-programmatic-toolkit.md
mrgrain: done

text/0300-programmatic-toolkit.md
mrgrain: In JS, `EventListener`s are just tuples of event name and callback function.

To be clear, I'm not proposing we are fully implementing or extending Node's [EventEmitter](https://nodejs.org/docs/latest/api/events.html#class-eventemitter). I'm borrowing the familiar interface of `on`, `once`, `off` etc. 

`IoHost` already is an EventEmitter (in the sense of Node's class) since that's what `notify` and `requestResponse` are doing.

text/0300-programmatic-toolkit.md
mrgrain: Hmm, I also like this. The main difference is that it prevents the default action from happening. I.e. it would not confirm the security-check. But we still need a way to replace the default action with something else.

So then we would also have `event.setResponse(false)` or similar. This doesn't feel any better:
```ts
io.once('deploy:confirm-security', async (event<boolean>) => {
  event.preventDefault();
  event.setResponse(false);
});
```

It could be just this though. Maybe that's more fitting to the EventEmitter style.

```ts
io.once('deploy:confirm-security', async (event<boolean>) => {
  event.setResponse(false);
});
```

DOM events also deal with hierarchy. So they are more complicated than what we need.

mrgrain: I guess in general we might be closer to the observer pattern except that we are enforcing a single observer (though at least for messages an IoHost could have multiple observers). I suppose the `IoHost` really assumes the role of an EventBus or CommandBus and leaves the rest to the implementor. 

CommandBus because we not only have informative messages, but also sometimes "command" the IoHost to make a decision. Obviously we are limiting what commands are received by it and don't allow it to handle complex commands like `deploy`. Suppose it could though. 🤔

text/0300-programmatic-toolkit.md
mrgrain: `Toolkit` captures interactivity through `requestResponse()`. We could short circuit this by providing a flag, but then we are removing the opportunity for an integrator to make a decision in a non-interactive way.

A CLI tool however has a notion of "can ask for user input" and "cannot ask for user input" that is fairly well defined (Is `CI` set? Is TTY available?). This example was an attempt to capture that, but I don't think its too relevant to the RFC. We might have `CdkCliHost` with a flag or we might have a `CdkCliHost` and an `JustLogToTheScreenCliHost` host.

I can adjust this example to make this more clear.

text/0300-programmatic-toolkit.md
mrgrain: Mostly if an integrator is also generating the code, e.g. a visual builder tool. But I think we can use try-catch for this use case for now.

text/0300-programmatic-toolkit.md
mrgrain: Fixed!

text/0300-programmatic-toolkit.md
mrgrain: No as in Event hooks, see requirements table.

text/0300-programmatic-toolkit.md
mrgrain: Yes. I don't think that needs to go into the readme though.

text/0300-programmatic-toolkit.md
mrgrain: Make it easier for integrators. `level` can be argued. `time` is free, `command` I think is important for provenance and if we do the setup right also pretty much free.

mrgrain: But this also works

```ts
class ToolkitError extends Error implements IoMessage { ... }
```

text/0300-programmatic-toolkit.md
mrgrain: I'm looking for a way to differentiate between 
- errors from the Toolkit
- expected errors from the App (e.g. validation)

And then also offer a way to catch both. This part is optional and can be handled by an OR statement as well.

guysqr:
This is going to be a really useful capability that will enable many interesting and valuable use cases, including being able to bake CDK into an Electron app and compose, deploy and destroy CDK apps entirely in code. This will put the benefits of the CDK into the hands of a whole new audience - those who can't or won't install command line tools.

I think it would be good to remove any interdependency between this and the consuming app around SDK, so consider if rather than passing an SdkProvider we could just pass a SDKv3 credentials provider.

Also, one of the challenges I saw when trying to use the CDK programatically in the past is how it sends CFN outputs to STDOUT and STDERR - would be ideal if we wanted to consume the CFN stack creation events via CDK to have a way to tap into that feed.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+### Why should we _not_ do this?
+
+There are requirements that cannot be satisfied with the current available options.
+Specifically event hooks, control over logging, structured errors and typed returns are not possible (refer to the appendix for details).
+
+Implementing the [Programmatic Toolkit] proposal will make it more difficult to add new features to the CLI.
+It creates a new level of abstraction, that needs to be considered going forward.
+The complexity of these abstractions might discourage community contributions and increase the time to build new features.
+
+We are only aware of a small number of interested integrators,
```
iliapolo, line 404: This statement resonates with me, and raises some concerns. To alleviate them, we should present alternative solutions to the new specific requirements we have.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+
+The proposed `IoHost` design requires us to commit to a stable set of messages and requests send between [Programmatic Toolkit] and integrator.
+A detailed definition of what would constitute a breaking change is included in the readme above.
+
+To support integrations, we will need to publish a list of all messages.
+We will use static code analysis to automatically compile and publish this list without adding additional maintenance burden on to the team.
+
+#### Synth-time Errors
+
+This includes making changes to all errors thrown from the construct library.
```
iliapolo, line 467: Can you elaborate? what changes?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+
+The existing plugin system is a superset of the SDK v3's `AwsCredentialIdentityProvider` type as it introduces a mode (reading or writing).
+The public contract will subsume and formalize this.
+
+#### Readme
+
+The `AwsSdkProvider` defines how the toolkit interacts with AWS services.
+A default provider is available.
+You can provide a custom class implementing the `IAwsSdkProvider` interface.
+A standard `SdkProviderForCdk` is available that includes all features of the AWS CDK CLI [TODO: this might end up in a separate package].
```
iliapolo, line 513: I don't follow. Is this part of the previous section? Also it doesn't sound like an alternate solution, but rather a possible future extension?

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+
+#### Event stream
+
+> Streams are non blocking by design.
```
iliapolo, line 517:
```suggestion
> Disregarded. Streams are non blocking by design.
```

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+We mitigate this by providing tooling that helps contributors to define different inputs and defaults declaratively and generate (parts of) the code for it.
+Additional test tooling will help writing the same test cases onces for different input modes.
+
+#### A lot of refactoring
+
+The proposal requires to move a lot of code around, mostly into new packages.
+It also implies refactoring of existing interfaces to match the new API design.
+Both of these things are prone to human errors.
+
+Mitigation for this is test coverage and separating out PRs that are purely moving code, from actual changes.
```
iliapolo, line 594: We probably want to coordinate rolling this out only after we address the coverage gaps we detected.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+- Additional actions can be migrated to the [Programmatic Toolkit]
+- The `IoHost` can be extended to include other I/O tasks like filesystem interactions.
+  This would allow implementations that are truly, in-memory like in browser environments.
+- The Toolkit setup can allow additional configuration, like providing a custom SDK or authentication configuration.
+
+### Is this a breaking change?
+
+No.
+
+### What alternative solutions did you consider?
```
iliapolo, line 487: What about the option of not providing an event API at all, and instead just focus on the objects the CLI returns, and make sure they contain all the information necessary for customers to take actions?

That is:

```ts
// no hooks available on the `cdk` instance
const deployment = cdk.deploy(...);

// take the output and do something
const stacks = deployment.listStacks();

....
```

I'm not really advocating for this approach, but I want to understand exactly what use-cases this solution doesn't address.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+| FR11 | Event hooks                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
iliapolo, line 737: Can you map these requirements to specific use-cases? that is, specific actions we know customers want to take inside events. 

This will help us understand if the use case can be addressed in some other way.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+});
+
+// Synth and store a cached CloudExecutable, this won't run synth again
+const cxCache = await cdk.synth(cx);
+
+// Deploy our pre-synth'ed app
+const deployment = await cdk.deploy(cxCache, { stacks: ['MyTestStack'] });
+
+// Returns a list of ARNs of the deployment
+return deployment.listResources().filter(r => r.arn);
```
misterjoshua, line 45: Does this method also allow access to resource attributes like `AWS::Lambda::Url`'s `FunctionUrl`? I could see this being very helpful for automated testing.

misterjoshua, line 45: Is there the possibility of filtering resources here by node path? This might make it easier to locate specific resources for a system under test.

mrgrain: This is a great suggestion! Might not be in the first version, but should be easy to add.

text/0300-programmatic-toolkit.md
mrgrain: Excluded for now.

text/0300-programmatic-toolkit.md
mrgrain: Explained. Will make a note.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+For every emitted message, the toolkit calls the `notify` or `requestResponse` method of an `IoHost` (more on this below).
+Every call is `await`ed for its completion, allowing to receiver to perform asynchronous operations like HTTP calls.
+It's okay for the receiver not do anything with the input and return immediately.
+The type definition of messages and requests is as follows:
+
+```ts
+interface IoMessage<T> {
+  time: string;
+  level: 'error' | 'warn' | 'info' | 'debug';
+  action: 'synth' | 'list' | 'deploy' ... ;
```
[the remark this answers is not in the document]
mrgrain: Meant to be the invoked command, i.e. "what is the toolkit currently doing".

`code` would be used for sub actions.

text/0300-programmatic-toolkit.md
mrgrain: Unfortunately I don't think that's easily possible as CFN stack APIs [don't include this information](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ListStackResources.html). You'd need to use a Stack output or make the request to the service API manually. But at least that should be much easier.

misterjoshua: Stack outputs would work fine, but how would that look work under this proposal? Is there a separate `deployment.listOutputs()` or something like that?

corymhall (commented):
Thanks for all the work on this toolkit, it's been super useful in the [pulumi-cdk](https://github.com/pulumi/pulumi-cdk) project!

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,791 @@
...
+When implementing a custom Cloud Executable in code, an integrator may choose to handle synth-time errors in the toolkit code or in the Cloud Executable.
+
+```ts
+const cx: ICloudExecutable = CloudExecutable.fromCloudAssemblyProducer(
+  async (context: Record<string, any>): cxapi.CloudAssembly => {
+    try {
+      const app = new cdk.App({ context });
+      const stack = new cdk.Stack(app);
+      return app.synth();
+    } catch(e: unknown) {
```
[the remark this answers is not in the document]
mrgrain: They will all be typed. The `e: unknown` is a TypeScript quirk. I can fix the example to be more clear.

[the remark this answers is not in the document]
mrgrain: Context lookups will NOT be emitted from within this though. They are attached to the Toolkit and not the app.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,791 @@
...
+
+For every emitted message, the toolkit calls the `notify` or `requestResponse` method of an `IoHost` (more on this below).
+Every call is `await`ed for its completion, allowing to receiver to perform asynchronous operations like HTTP calls.
+It's okay for the receiver not do anything with the input and return immediately.
+The type definition of messages and requests is as follows:
+
+```ts
+interface IoMessage<T> {
+  time: string;
+  level: 'error' | 'warn' | 'info' | 'debug';
```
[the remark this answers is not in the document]
mrgrain: All logs are send. The log level is a concern outside of the Toolkit.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+```ts
+const io = new IoEmitter();
+io.on('deploy', async (msg: IoMessage) => { /* do something */ });
+io.once('deploy:confirm-security', async (msg: IoRequest<boolean>): boolean => {
+  // reject the deployment
+  return false;
+})
+```
+
+The `CdkCommandLineIoHost` is available, implementing all the interaction behavior for the AWS CDK CLI [TODO: this might end up in a separate package]:
```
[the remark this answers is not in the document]
mrgrain: I think that'd be a future feature of the AWS CDK implementation, not the [Programmatic Toolkit]. I don't think there's anything in this proposal that's blocking us from doing something like that.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+```ts
+const cdk = new Toolkit()
+```
+
+You can customize the functionality of the toolkit.
+Currently only `ioHost` is available for customization:
+
+```ts
+const cdk = new Toolkit({
+  ioHost: new MyIoHost(), // optional
```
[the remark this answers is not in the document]
mrgrain: Actually none. Fixed. We should however provide our `CdkCommandLineIoHost`

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+    const stack = new cdk.Stack(app);
+    return app.synth();
+  }
+);
+```
+
+Existing applications, can also be used by simply passing an AWS CDK app directory (containing a `cdk.json` file):
+
+```ts
+const cx: ICloudExecutable = CloudExecutable.fromCdkApp("/path/to/cdk/app");
```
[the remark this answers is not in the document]
mrgrain: Very simple API like `public exec(): Promise<CloudAssembly>`

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+```ts
+const cx: ICloudExecutable = CloudExecutable.fromCdkOut("/path/to/cdk.out");
+```
+
+#### Actions - synth
+
+The `synth` action takes a Cloud Executable and synthesizes it in to a `CloudAssembly`.
+
+```ts
+const cx: ICloudExecutable =  await cdk.synth(cx, { stacks: ['MyTestStack'] });
```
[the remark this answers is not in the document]
mrgrain: `CloudAssembly` is a data type, `ICloudExecutable` a behavioral type to create a `CloudAssembly`.

All CDK actions currently accept the behavioral type and implicitly create the assembly in the background. That makes `ICloudExecutable` somewhat a superset of `CloudAssembly` because you can always do something like

```ts
class StaticCloudExecutable implements ICloudExecutable {
  constructor(private readonly assmebly: CloudAssembly) {}

  public exec() {
    return this.assembly;
  }
}
```

Which seems simple enough.

If we don't do this, we have the options to:
- Make every action accept both `ICloudExecutable` and `CloudAssembly`
- Implement a dynamic `CloudAssembly` dynamic so that a call to `tryGetArtifact(id)` would first do what ever needs to be done to create the `CloudAssembly`.

I've opted for the most option that is imo the most explicit.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+It returns a list of stack deployments.
+
+#### Errors
+
+All errors are thrown as exceptions and the integrator must be prepared to handle them.
+Expected errors are structured data and extend the `IoMessage` interface.
+
+```ts
+interface ToolkitError extends IoMessage {
+  source: 'toolkit' | 'user';
```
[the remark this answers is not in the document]
mrgrain: This seems like naming to me. This RFC proposes there are two kinds of errors:
- Errors originating from the CDK APP that are definitely in the scope of the user to fix. No change to the toolkit caller code can fix these.
- Errors originating from the Toolkit that are possibly resolvable by the user either by changing input or configuration.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+  await cdk.synth(cx);
+} catch (err: unknown) {
+  // handle app errors here
+}
+```
+
+#### Dispose
+
+The `Toolkit` class implements an `AsyncDisposable` Resource according to the [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) feature in ECMAScript.
+This means that any resources used by the [Programmatic Toolkit] are automatically cleaned-up.
```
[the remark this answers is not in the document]
mrgrain: Mainly temporary files.

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+Additionally the interface `cli-lib-alpha` offers is not optimized for programmatic use, further limiting its usefulness.
+
+The [Programmatic Toolkit] aims to solve these limitations by providing a proper programmatic interface to CDK's functionality,
+sallowing integrators to have full control over all output and combine CDK actions with custom steps in their applications.
+With this project, we are changing the paradigm.
+Programmatic usage will become an equally considered use case, in fact the proposed implementation will require it to be considered first.
+
+### Why should we _not_ do this?
+
+There are requirements that cannot be satisfied with the current available options.
```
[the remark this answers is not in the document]
mrgrain: Agreed. This is confusing. removed.

text/0300-programmatic-toolkit.md
mrgrain: We have specific requirements. Alternatives were considered. This sections is why we shouldn't do this and the answer here is implying that the technical complexity may not outweigh the benefits. However I'm clearly proposing we should do this and in order to satisfy requirements we have to do something, and I think it should be this proposal.

Maybe I'm missing your point though.

text/0300-programmatic-toolkit.md
mrgrain: added

text/0300-programmatic-toolkit.md
mrgrain: Sure. The currently proposed alternative is to not offer a configuration.

text/0300-programmatic-toolkit.md
mrgrain: added an alternative

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,776 @@
...
+   * If the host does not return a response the suggested
+   * default response from the input message will be used.
+   */
+  requestResponse<T, U>(msg: IoRequest<T,U>): Promise<U | undefined>;
+}
+```
+
+For every message, the toolkit calls `notify` or `requestResponse`, respectively.
+
+A generic `IoEmitter` is available to provide a familiar event emitter interface.
```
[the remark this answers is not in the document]
rix0rrr: `IoHost` as specced is at least jsii compatible. `.on` and `.once()` less so, or they need to look very different than what JS slingers are used to.

text/0300-programmatic-toolkit.md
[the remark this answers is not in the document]
rix0rrr: I like what you did. The name is maybe just setting some expectations. `ICloudAssemblySource` ?

rix0rrr (approved):

text/0300-programmatic-toolkit.md
```diff
@@ -0,0 +1,863 @@
...
+msg("Diff of Stack B", diffB)
+msg("Deploying Stack A")
+msg("Stack event for Stack A", event)
+msg("Stack event for Stack A", event)
+msg("Stack event for Stack A", event)
+msg("Stack A deployment completed", stackA)
+msg("Deploying Stack B")
+msg("Stack event for Stack B", event)
+msg("Stack event for Stack B", event)
+msg("Stack B deployment completed", stackB)
```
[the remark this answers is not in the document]
mrgrain: Yes, that's a good point. The progress monitor will probably be inside the CliIoHost and we need to make sure these messages contain enough information.

text/0300-programmatic-toolkit.md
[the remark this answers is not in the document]
mrgrain: Oh I do like `ICloudAssemblySource` 🤔

text/0300-programmatic-toolkit.md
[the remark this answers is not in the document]
mrgrain: > At a certain level, won't this be annoying?

Yeah I can see that. I was hoping we can do something global here, but might not actually work if two commands are executed in parallel (in which case the "current command" approach doesn't work either.

The field exists for the receiver, i.e. the IoHost to know what command an event is attached to. If we cannot find a good solution we might have to punt it.

rix0rrr (approved):

PR #657, opened 2024-11-18
RFC 655: Enhanced L1s

aaythapa:
This is a request for comments about enhancing L1s in CDK. See #655 for additional details. 

APIs are signed off by @alias.

---

### DIFF
--- a/text/0655-enhanced-l1.md
+++ b/text/0655-enhanced-l1.md
@@ -1,0 +1,237 @@
+# Priority-Ordered Aspect Invocation
+
+* **Original Author(s):**: @aaythapa
+* **Tracking Issue**: #655
+* **API Bar Raiser**: @alias
+
+## Background
+
+In the CDK, Level 1 constructs or "L1s" represent constructs that directly correspond to AWS CloudFormation resources. These constructs are automatically generated primarily from the CloudFormation resource provider schema and provide a one-to-one mapping with the corresponding CloudFormation resource. Since they are automatically generated these constructs are updated on a ~weekly basis with new features and services as they are released in CloudFormation.
+
+## Problem
+
+A current gap with the CDK L1 constructs is their disconnect and feature gaps compared to CDK L2 constructs. L2s offer more sophisticated interfaces for AWS services, with built-in defaults, validation, enum definitions, and more. Additionally, L2s can easily reference other L2 constructs wherever a resource is required, using object references rather than attribute references. This provides developers with a richer and more intuitive experience.
+
+In contrast, L1s constructs rely solely on primitive types for properties. Developers must hard-code all string and number properties, even when only specific values are permitted. There's is also no built-in validation with L1s. Many properties in CloudFormation must follow specific patterns if they're a string or fall within certain ranges if they're number, but developers using L1s only discover these constraints when their applications fail to deploy.
+
+Furthermore, CDK L1s and L2s are not interoperable. L2 constructs are equipped with resource interfaces (e.g., IBucket, IRole, IKey, etc.), allowing them to be referenced seamlessly in other L2s. L1s lack these resource interfaces and can only be referenced using attribute . This becomes problematic when an L2 construct expects a resource interface rather than an attribute string, preventing L1s from being referenced within L2s.
+
+For example when working with an L2 Lambda function, it's straightforward to pass in an L2 IAM role.
+
+```
+const lambdaRoleL2 = new iam.Role(this, 'lambdaRole', {
+    ...
+});
+const lambdaFunctionL2 = new lambda.Function(this, 'lambdaFunction', {
+    ...
+    role: lambdaRole,
+    ...
+});
+```
+
+In this case, the developer doesn't need to perform additional steps or know specific attributes to pass. However, if they attempt to pass an L1 IAM role, it becomes more complex. They'd need to manually create an `I<Resource>` object using the L1 role's attribute.
+```
+const lambdaRoleL1 = new iam.CfnRole(this, 'lambdaRole', {
+    ...
+});
+const lambdaFunctionL2 = new lambda.Function(this, 'lambdaFunction', {
+    ...
+    role: iam.Role.fromRoleArn(this, 'importedRole', lambdaRoleL1.attrArn),
+    ...
+});
+```
+This disconnect between L1s and L2s creates challenges for developers trying to use both in the same CDK application. Bridging this gap in a maintainable and backward-compatible way would significantly enhance the utility of L1s in CDK.
+
+## Proposed Solution
+
+### Adding enums and validation
+
+#### Getting the data 
+
+The `aws-service-spec` currently integrates various data sources (listed [here](https://github.com/cdklabs/awscdk-service-spec?tab=readme-ov-file#data-sources)), combining them into a single JSON file that serves as the source of truth for L1 generation. This JSON file is then published to NPM. The CDK repository uses this file in conjunction with the `spec2cdk` script ([here](https://github.com/aws/aws-cdk/tree/main/tools/%40aws-cdk/spec2cdk)) to generate L1s. While we can enhance the CDK L1s with enums and validations by adjusting our generation code, we first need to ensure the necessary data for this generation is available.
+
+Within our current list of data sources we can find some enums and validation information. The registry schema, which forms the base of our source of truth, already includes some enums and validation information. We just need to modify the service spec to incorporate this information into the published JSON file. We can then expand our data sources to improve coverage. `cfn-lint` is a command line tool that helps validate CFN templates for syntax and best practices. It contains a schema that builds on the CFN Resource Provider Schemas by adding additional information from various SDKs, then repackages these schemas for distribution. We can include the information provided by `cfn-lint` into our source of truth to achieve better coverage. 
+
+In cases where a schema doesn’t contain the required information, it doesn’t matter whether we retrieve this data from a JSON file or by executing an AWS CLI command. For instance this PR updates the enum values available for a property based on the output of an AWS CLI command defined in the docs for that property. We can build workflows that run these commands and update specific properties with new data. These workflows should support all necessary AWS CLI commands and allow flexible specification of which command to run and which property to update.
+
+#### Representing enums in L1s
+
+To represent enums in L1s we can follow the same implementation as enums in L2s which is to create a new type in the generation process. This will allow developers to use the static enum variables from the schema or pass in an hard coded value if needed.
+
+#### Runtime class will be generated
+```
+export class Runtime {
+    public static readonly PYTHON310 = new Runtime('python3.10');
+    public static readonly PYTHON311 = new Runtime('python3.11');
+    ...
+    ...
+    public readonly name: string
+    protected constructor(name: string) {
+        this.name = name;
+    }
+    public static of(value: string) {
+        return new Runtime(value);
+    }
+}
+...
+runtime: Runtime.of('python3.10') # Allowed.
+# OR
+runtime: Runtime.PYTHON310 # Allowed
+```
+
+This implementation offers greater flexibility compared to using built-in enum types as custom values would not be allowed. This ensures that developers aren’t restricted when a CFN enum value is available but not yet included in the CDK.
+
+### Validating properties
+
+Currently, L1 constructs in the CDK perform minimal validation on properties, mostly ensuring type correctness. We can extend these validations by leveraging additional data from our sources. Depending on the property type, various validation rules can be applied, as defined in the schema:
+
+1. String
+    1. `minLength`: min length the string can be
+    2. `maxLength`: max length the string can be
+    3. `pattern`: the pattern the string must match (commonly used for arns)
+    4. `enum`: the enum values the string can be, this likely won’t be part of the validation as we’ll have an enum type
+    5. `const`: the constant the string must be
+2. Array
+    1. `minItems`: min number of items in the array
+    2. `maxItems`: max number of items in the array
+3. Integer
+    1. `minimum`
+    2. `maximum`
+
+
+e.x Right now `Description` property of Lambda Function currently undergoes string validation
+```
+errors.collect(cdk.propertyValidator("description", cdk.validateString)(properties.description));
+```
+In the Registry Schema we have a maxLength value of `256` for `Description`. We can extend the validation to include this:
+e.x
+```
+// add a validator function
+errors.collect(cdk.propertyValidator("description", CfnFunctionDescriptionPropertyValidator)(properties.description));
+...
+
+// corresponding validator function
+function CfnFunctionDescriptionPropertyValidator(prop: any): cdk.ValidationResult {
+  ...
+  const maxLength = 256
+  if (prop.length > maxLength) {
+    errors.collect(new cdk.ValidationResult(<Error message>));  
+  }
+  ...
+}
+```
+### Delivering in a backwards compatible way
+
+The challenge lies in delivering these changes in a backward-compatible way. We can't simply overwrite the current types, nor can we add the new types as unions with the existing ones. Doing so would break backward compatibility by violating the original type guarantees. For example, changing a property type from `string` to `string | number` would remove the certainty that the property is always a string, which could prevent developers from safely performing string-specific operations. Additionally, unions could create complications in languages other than TypeScript, such as Java, C#, or Go, where unions might be represented as objects, leading to a loss of typing information.
+
+The best approach is to deprecate the older types/files and generate new ones with distinct names. There are several ways we could implement this:
+
+1. Add new properties to existing L1 files: Differentiate the new properties by adding a suffix, e.g., `runtime` vs. `runtime_v2`. This follows an established pattern in the CDK, where certain properties have versioned variants (e.x here)
+    1. Drawback: The UX may be confusing as developers have to decide which version of a property they want to use
+2. Create new L1 files in the existing library: Here, the suffix would be added to the file name rather than the properties, e.g., `aws_lambda` vs. `aws_lambda_v2`.
+    1. Drawback: This would overload the current library and may be redundant in some cases as not all properties have a _v2 so we would essentially be copying over the same thing to the new `_v2` file. For libraries with `_v2` files there may be very minimal changes.
+3. Generate a new library with the new L1s: The suffix would be applied to the library name itself, e.g., `aws-cdk-lib` vs. `aws-cdk-lib-v2`.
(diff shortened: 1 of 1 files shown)


text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+    2. `maxLength`: max length the string can be
+    3. `pattern`: the pattern the string must match (commonly used for arns)
+    4. `enum`: the enum values the string can be, this likely won’t be part of the validation as we’ll have an enum type
+    5. `const`: the constant the string must be
+2. Array
+    1. `minItems`: min number of items in the array
+    2. `maxItems`: max number of items in the array
+3. Integer
+    1. `minimum`
+    2. `maximum`
```
mikejgray, line 99: This will be a huge help. It's immensely frustrating to have so many checks in your CDK code only to have a simple validation check fail in Cloudformation. Thank you for recommending this!

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+const func = new lambda.CfnFunction(this, 'test', {
+      code: {
+        ...
+      },
+      handler: 'index.handler',
+      role: r.attrArn,
+      runtime_v2: lambda.CfnFunction.Runtime.PYTHON310, // new V2 property
+      // if both runtime and runtime_v2 are set we can throw an error
+      timeout: 30
+    })
```
mikejgray, line 150: I can't think of a better way to manage the backwards compatibility issues here, but I'm not crazy about the possibility of having `runtime_v2` defined but not `runtime`. Can't think of specific examples but I get the impression validation becomes a little more difficult - if you must have a property defined, now you have to account for that property and its _v2 variant, which could introduce issues for certain use cases.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+    super();
+    this.bucket = s3.Bucket.fromCFnBucket(bucket);
+}
+...
+```
+Note that we won’t be able to change `I<Resource>` to `ICfn<Resource>` everywhere. Public properties with the type `I<Resource>` can’t be changed as developers might be using an `I<Resource>` specific property. We can work on deprecating these properties and creating new `_v2` version of the property that will accept the CFN interface
+
+### POC
+
+There were some internal experiments done around this idea before. We concluded that generating `ICfn<Resource>` is fairly straight forward but actually backporting the interface to our existing L2 API is manual and runs the risk of accidentally creating breaking changes e.x turning a public property `IBucket` into `ICfnBucket` (since users could be referencing `IBucket` -specific properties) or renaming a parameter to an API can cause breaking changes in Python. If we move forward with this we'll continue the work from the experiments while keeping the findings in mind.
```
mikejgray, line 236: Cautiously optimistic about this approach, and definitely curious about the findings.

mikejgray (commented):
It's exciting to see big changes being proposed in the CDK! Thank you for the work on this one.

johnf:
I love everything being proposed here. This will bring significant improvements!

Except for the `_v2` suggestion. I understand why this approach is being suggested, but it doesn't feel right to me.

I think this adds cognitive load to the developer, I have to now think about wether to use _v2 or not. Also, does this stay around forever, or does it get dumped in a future change? DO we end up with _v3 down the track?

**My suggestion:** Bite the bullet, make a breaking change and bump the major version.

A possible approach to help, if we release today
* release aws-cdk-lib 3.174.0 and 2.174.0.
* Keep the versions in sync
* The only difference would be v3 has the `prop_v2` logic written out to `prop`
* So this should be automatable

Some alternatives I can think of
* Make a breaking change and release a codemod to help everyone upgrade their code
* Typescript magic: pass in something like cdk_type_version: 2 into every prop, which gives me the new types - in the future, we could make this go away. I suspect there is some Typescript magic that might allow us to set something globally and have this still work

rehanvdm:
I love the idea of having Interfaces for the CFN L1s and improving the L1-L2 interoperability. 

I agree with @johnf, it has been a few years since a major release and there are many Flags in cdk.json and @deprecated functions. Including this big change, it justifies a new major release for me.

mrgrain:
At re:Invent 2024 I had the opportunity to discuss this with a number of AWS Heroes: @hoegertn @Osterjour @pgarbe This is a write-up of this discussion from my perspective. Please add to it as required.

It was questioned if enums are necessary in L1s as they don't seem to add much value and take L1s further away from what the CFN is. It was also noted that the use of `_v2 ` is confusing and annoying. In particular deploying this pattern for enums will make it a regular occurrence across the L1 library, as opposed to using the suffix only in situations for where an upstream type change is breaking. In response to this it was noted that old properties would be marked as deprecated and that it is an IDE and documentation concern to ensure that a user is not flooded with multiple of the same props to choose from.

The discussion then dived into what L1s are representing, what they should be representing and what they could be representing in future. There was consensus that L1 are currently roughly equal to CFN, however it was not clear what exactly that means. As an example Validations are brought up which are currently not part of L1s but having them would make them actually closer to what CFN will do - at least from some perspective. Another example was IAM grants. While the group agreed that this is not currently part of CFN, it would be very useful if grants can be auto-generated from a data source.

Based on this, the idea was proposed that L1s could be everything that can be auto generated (i.e. not hand written) from a data source. However maybe this could still be separate layer and the proposed features in the RFC and new features like grants could be introduced as a new layer between the current L1 and L2. However there were concerns that this was also confusing.

End of discussion.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+3. Generate a new library with the new L1s: The suffix would be applied to the library name itself, e.g., `aws-cdk-lib` vs. `aws-cdk-lib-v2`.
+    1. Drawback: While this would provide a clean break between the old L1s and the new we would have to maintain both versions of the library, adding more overhead. It would also present an issue every time we have new type changes, we’d have to make multiple `_v<number>` libraries
+
+### Recommended approach
+
+Adding new property to the existing L1s would be the recommended approach. The suffix of `_v2` should make it clear which property to use and alleviate any confusion. We can also deprecate the older properties to signal to developers to use the new one. By deprecating the older properties rather than removing them, we ensure that existing code continues to function without breaking. Developers can still use the original properties, while being encouraged to adopt the new "_v2" version at their own pace.
```
tmokmss, line 138: I love the more thorough validations added with the enhanced L1s!

One question, how does this work when the original prop is required? For example in [`CfnTableProps`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.CfnTableProps.html), the `keySchema` property is required, something like below:

```ts
interface CfnTableProps {
  keySchema: FooType
}
```

After enhancing L1, we cannot make both original and v2 prop required, because users only need to set one.

```ts
// NOT WORK
interface CfnTableProps {
  keySchema: FooType
  keySchema_v2: FooType
}
```

So I think we have to make the interface something like this:

```ts
interface CfnTableProps {
  keySchema?: FooType
  keySchema_v2?: FooType
}
```

But imho this is actually worse than the original L1, because it lost the information that keySchema is a required property. (of course we can do runtime validation, but type check is faster.)

This is another reason why I'm not a big fan of option 1. I like option 3 better with the upgrade of aws-cdk-lib to v3 (and possibly aws-cdk v3) as @johnf said. With cdk v3, we can also remove all the feature flags that have been added since the v2 release, which has been a burden for construct library maintainers to validate their code with various possible flags.

go-to-k:
This is an excellent suggestion. It would be especially nice to have the built-in validation in L1 by default.

And most of my opinions have already been mentioned above. Introducing V2 for each parameter could lead to complexity (props become bloated, required properties become optional, etc.).

Also, many users will have created a mechanism to check against L1 resource properties using Aspects, and I'm concerned that the addition of this v2 property will break that (existing) mechanism.

Therefore, I agree with the opinion to upgrade with CDK V3 as they say.

text/0655-enhanced-l1.md
aaythapa: You can look at this branch for a deeper look on this POC: https://github.com/aws/aws-cdk/tree/conroy/generate

text/0655-enhanced-l1.md
aaythapa: This is a good example, my thought was we could do runtime validation (making sure one of these is set) but as you mentioned type check is faster. I'll add this example to the cons for option 1

aaythapa:
Thank you all for the feedback everyone!

The general consensus seems to be that the features are valuable and would make a great addition to the L1s. However, the proposed implementation (`_v2` suffix on the new properties), compromises UX while aiming to maintain backward compatibility, and most prefer moving forward with CDK V3 instead.

I'll share this feedback with the team and provide updates here as they come. In the meantime, please feel free to share any additional comments or suggestions. Thanks again

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+    1. Drawback: The UX may be confusing as developers have to decide which version of a property they want to use
+2. Create new L1 files in the existing library: Here, the suffix would be added to the file name rather than the properties, e.g., `aws_lambda` vs. `aws_lambda_v2`.
+    1. Drawback: This would overload the current library and may be redundant in some cases as not all properties have a _v2 so we would essentially be copying over the same thing to the new `_v2` file. For libraries with `_v2` files there may be very minimal changes.
+3. Generate a new library with the new L1s: The suffix would be applied to the library name itself, e.g., `aws-cdk-lib` vs. `aws-cdk-lib-v2`.
+    1. Drawback: While this would provide a clean break between the old L1s and the new we would have to maintain both versions of the library, adding more overhead. It would also present an issue every time we have new type changes, we’d have to make multiple `_v<number>` libraries
+
```
eladb, line 135: For options 1 + 2, don't forget to add "@derpreacted" tags to the old elements.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+The best approach is to deprecate the older types/files and generate new ones with distinct names. There are several ways we could implement this:
+
+1. Add new properties to existing L1 files: Differentiate the new properties by adding a suffix, e.g., `runtime` vs. `runtime_v2`. This follows an established pattern in the CDK, where certain properties have versioned variants (e.x here)
+    1. Drawback: The UX may be confusing as developers have to decide which version of a property they want to use
+2. Create new L1 files in the existing library: Here, the suffix would be added to the file name rather than the properties, e.g., `aws_lambda` vs. `aws_lambda_v2`.
+    1. Drawback: This would overload the current library and may be redundant in some cases as not all properties have a _v2 so we would essentially be copying over the same thing to the new `_v2` file. For libraries with `_v2` files there may be very minimal changes.
+3. Generate a new library with the new L1s: The suffix would be applied to the library name itself, e.g., `aws-cdk-lib` vs. `aws-cdk-lib-v2`.
```
eladb, line 133: If there's enough "meat" for CDKv3, then I would recommend choosing this approach, but bear in mind that a major version is a big project (v2 tool us almost two years to release).

It is also okay to punt this particular enhancement until you are ready to release v3 given its breaking nature.

text/0655-enhanced-l1.md
eladb: Damn this stuff is hard!

For option 1, I agree that the loss of strong typing (turning required properties to optional) is a major drawback.

But option 1 is tricky for another reason: you will likely want to deprecate the legacy properties and keep only the modern versions. In the above example, it means that you will need to delete `keySchema` and only leave `keySchema_v2` which is not a good outcome. If you end up modernizing `keySchema`, you'll need to remove `keySchema_v2` and then you will penalize the users who migrated to the modern version earlier by breaking or deprecating their code.

I would say the best approach would be to go with option 3 and couple this feature with a new major version (v3) of the CDK, which will allow you to break the existing APIs. When we planned v2 we had a GitHub issue where we accumulated these major breaking changes, and at some point we got to a critical mass that justified a new major version.

If v3 is too far away and there's a very strong push to introduce this capabilities sooner, I think option 2 is a reasonable tradeoff albeit there's a chance it will cause the library to be too large.

I am wondering if perhaps as an intermediate step (before a major version is released), you can just add better documentation and runtime (synth-time) checks that the value is valid. This will go a long way to shift-left errors without breaking. In TypeScript, you could even use a type union like `"foo" | "bar" | "baz"` to improve typings without impacting JSII languages.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+}
+```
+Any location that accepted `IBucket` would now accept `ICfnBucket`
+```
+export interface CodePipelineProps {
+    ...
+    readonly artifactBucket?: s3.ICfnBucket;
+}
+```
+If those locations use a functionality from the L2 interface, we can do an upconversion to an `IBucket` by implementing an API that converts an `ICfn<Resource>` into an `I<Resource>`. We do a similar operation when a resource that isn’t created in the current CDK stack is referenced. This needs to done to maintain backwards compatibility and not break existing customers.
```
eladb, line 214: I am wondering if perhaps it would be sufficient to just add `fromCfnXxx` to all L2s instead of juggling all of these interfaces.

These conversion static methods would be straightforward to implement:

```js
public static fromCfnBucket(bucket: CfnBucket): IBucket {
  return this.fromBucketArn(bucket.attrBucketArn);
}
```

Indeed, it won't allow users to directly pass an L1 everywhere an `IBucket` is needed but it feels like this would be good enough to improve the ergonomics without too many breaking changes.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+
+Within our current list of data sources we can find some enums and validation information. The registry schema, which forms the base of our source of truth, already includes some enums and validation information. We just need to modify the service spec to incorporate this information into the published JSON file. We can then expand our data sources to improve coverage. `cfn-lint` is a command line tool that helps validate CFN templates for syntax and best practices. It contains a schema that builds on the CFN Resource Provider Schemas by adding additional information from various SDKs, then repackages these schemas for distribution. We can include the information provided by `cfn-lint` into our source of truth to achieve better coverage.
```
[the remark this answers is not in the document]
jfuss: `cfn-lint` already has a process built that pulls the data from the AWS SDK (specifically boto3). Idea is to not re-implement that ourselves. CFN Lint has similar data requirements/usage, so (imo) makes more sense to re-use. Ideally this data should just be in the CFN schemas per service directly, which is something we can work with `cfn-lint` and CFN on directly. That is outside of this, but helps with the full picture.

text/0655-enhanced-l1.md
```diff
@@ -0,0 +1,237 @@
...
+
+In cases where a schema doesn’t contain the required information, it doesn’t matter whether we retrieve this data from a JSON file or by executing an AWS CLI command. For instance this PR updates the enum values available for a property based on the output of an AWS CLI command defined in the docs for that property. We can build workflows that run these commands and update specific properties with new data. These workflows should support all necessary AWS CLI commands and allow flexible specification of which command to run and which property to update.
```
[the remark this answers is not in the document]
jfuss: This is (more or less) what `cfn-lint` is doing (see above comment) but packaging it as CFN schemas that we can just leverage instead of doing ourselves.

text/0655-enhanced-l1.md
aaythapa: Bucket L2 already has a `fromCfnBucket` that implements what you mentioned but I'm not sure how many other L2s have something similar or what the effort to add the missing ones would be.

text/0655-enhanced-l1.md
aaythapa: > If you end up modernizing keySchema, you'll need to remove keySchema_v2 and then you will penalize the users who migrated to the modern version earlier by breaking or deprecating their code.

By modernizing `keySchema` do you mean if/when we go to CDK V3 we'd have to go back to `keySchema`? 

> I am wondering if perhaps as an intermediate step (before a major version is released), you can just add better documentation and runtime (synth-time) checks that the value is valid.

Updating the docs is a great idea and the implementation should be fairly straight forward. For the runtime validation, is it possible to do without introducing a new property and without breaking compatibility?

> When we planned v2 we had a GitHub issue where we accumulated these major breaking changes?

Do you have a link to this github issue? Couldn't find it through a quick search 😅

text/0655-enhanced-l1.md
eladb: It should be easy to enforce the `fromCfnXxx` static methods using [awslint](https://github.com/aws/aws-cdk/tree/main/packages/awslint) so it's semi-manual.

text/0655-enhanced-l1.md
```diff
@@ -44,8 +44,18 @@ This disconnect between L1s and L2s creates challenges for developers trying to
 
 ## Proposed Solution
 
+To resolve these issues in L1s we're proposing enhancing the current L1s by
+1. Expanding the current list of data sources from which we generate L1s.
```
eladb, line 48: Nit: item 1 here is not really "solving any problem". It's the underlying mechanism you are using to implement the solutions described below. Probably not needed.

aaythapa: agreed

mrpackethead:
> It was questioned if enums are necessary in L1s as they don't seem to add much value and take L1s further away from what the CFN is. It was also noted that the use of `_v2 ` is confusing and annoying. In particular deploying this pattern for enums will make it a regular occurrence across the L1 library, as opposed to using the suffix only in situations when an upstream type change is breaking.

I may have missed the point here, but to me, having enums where there is a defined list of possible values, is very valuable. Its an almost every day occurrence having to look at the docs, to check on valid values. 

Why would this not be valuable?

text/0655-enhanced-l1.md
eladb: > > If you end up modernizing keySchema, you'll need to remove keySchema_v2 and then you will penalize the users who migrated to the modern version earlier by breaking or deprecating their code.
> 
> 
> 
> By modernizing `keySchema` do you mean if/when we go to CDK V3 we'd have to go back to `keySchema`? 
> 
> 

Yes. In the long run that's the "right" name.

> 
> > I am wondering if perhaps as an intermediate step (before a major version is released), you can just add better documentation and runtime (synth-time) checks that the value is valid.
> 
> 
> 
> Updating the docs is a great idea and the implementation should be fairly straight forward. For the runtime validation, is it possible to do without introducing a new property and without breaking compatibility?
> 
> 


I believe it should be possible to only add runtime validations instead of introducing new types.

> 
> > When we planned v2 we had a GitHub issue where we accumulated these major breaking changes?
> 
> 
> 
> Do you have a link to this github issue? Couldn't find it through a quick search 😅

https://github.com/aws/aws-cdk-rfcs/blob/main/text/0079-cdk-2.0.md

That's the relevant RFC

Issue #693, opened 2025-01-28
Set Default Values to Construct Properties using PropertyInjector


pcheungamz:
## Description

We would like to propose a framework for organizations to set default values to Construct Properties such as BucketProps.

Key Components:
* IPropertyInjector - has `constructFqn` and and `inject` function that takes in the original BucketProps and returns a BucketProps.
* PropertyInjectors - A collection of `IPropertyInjectors`.  There is only one `IPropertyInjector` for a Construct.  We can attach PropertyInjectors to any Construct, but the typical use case is to set `PropertyInjectors` on `App`, `Stage`, or `Stack`.
* applyInjectors - Constructs such as `Bucket` will call applyInjectors in their constructors to get the new `BucketProps`.  It calls `findInjectorsFromConstruct` to find the correct injector to apply to the properties.
* findInjectorsFromConstruct - This starts from the Construct's parent Node and walk up the Tree to find an `IPropertyInjector` applicable to this Construct.
* Constructs that support Property Injection will need a FQN and call applyInjectors.

With this PropertyInjector Framework, AWS and SDO Security orgs can set default properties for builders that meet their security bars.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @pcheungamz
| Author(s)           | @pcheungamz
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | @pcheungamz, @manan-pancholi

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #699, opened 2025-02-13
RFC 0693: Property injectors

pcheungamz:
This is a request for comments about Set Default Values to Construct Properties using PropertyInjector. See #693 for
additional details. 

APIs are signed off by @{rix0rrr}.

---

### DIFF
--- a/text/0693-property-injection.md
+++ b/text/0693-property-injection.md
@@ -203,7 +203,7 @@ PropertyInjectors can be attached to any scope, but the typical use case is to a
 
 In [Scope Tree Traversal](#scope-tree-traversal), we will discuss how to find the correct injector when they are specified in app, stack, etc.
 
-### Update Construct constructors
+### Update Construct constructors - Alternate Option
 
 Below are changes we will make to S3 Bucket.  We need to add `PROPERTY_INJECTION_ID` and in the constructor, call `applyInjectors`.
 
@@ -229,9 +229,10 @@ export class Bucket extends BucketBase {
 `applyInjectors` finds the injector associated with `Bucket.PROPERTY_INJECTION_ID` by calling `findInjectorsFromConstruct`,
 and once the injector is found, it applies the changes to the props.  See the next section for how we walk up the scope tree to find the injector.
 
-### An Alternate Way to Update Construct constructors using Decorators
+### Update Construct constructors - Preferred Option
 
-**Note: We are still studying the feasibility of this approach.  It is blocked by a JSII Error.**
+**This option is now possible thanks to
+[this commit](https://github.com/aws/aws-cdk/commit/b48a5ad0d0c96c80252aae3ff32df41c1fb89099)**
 
 We first define a Decorator called `propertyInjectionDecorator`.
 
@@ -524,3 +525,87 @@ const bucketA = new Bucket(function, 'test-mybucket', {
 
 `bucketA` will get BucketInjector `b3` applied to it, regardless of what Bucket Injector was defined in Stack or App.
 The typical use case is to attach IPropertyInjectors to App, Stage, and Stack.  So we have added `propertyInjectors` to their props for ease of use.
+
+### Can I overwrite a default value in an Injector with undefined?
+
+Using this Injector as an example:
+
+```ts
+class MyBucketPropsInjector implements IPropertyInjector {
+  public readonly constructUniqueId: string;
+
+  constructor() {
+    this.constructUniqueId = Bucket.PROPERTY_INJECTION_ID;
+  }
+
+  public inject(originalProps: BucketProps, _context: InjectionContext): BucketProps {
+    return {
+      blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
+      enforceSSL: true,
+      ...originalProps,
+    };
+  }
+}
+```
+
+TypeScript clients can use:
+
+```ts
+new Bucket(this, 'My-test-bucket', {
+  blockPublicAccess: undefined,
+});
+```
+
+to create a Bucket with `blockPublicAccess: undefined`.
+
+However, `undefined` is not support in Python and Java.  If Python client set `blockPublicAccess` to `None`,
+it woudl be as if `blockPublicAccess` was never passed in, so the Injector would use `BLOCK_ALL` as default.
+To get around this, orgs can allow Dev Team to subclass the Injector and specify `blockPublicAccess: undefined`.
+
+One example is:
+
+```ts
+class MyBucketPropsInjector implements IPropertyInjector {
+  public readonly constructUniqueId: string;
+
+  constructor() {
+    this.constructUniqueId = Bucket.PROPERTY_INJECTION_ID;
+  }
+
+  protected getDefaultBlockPublicAccess() {
+    return BlockPublicAccess.BLOCK_ALL;
+  }
+
+  public inject(originalProps: BucketProps, _context: InjectionContext): BucketProps {
+    return {
+      blockPublicAccess: getDefaultBlockPublicAccess(),
+      enforceSSL: true,
+      ...originalProps,
+    };
+  }
+}
+```
+
+```ts
+export MyTeamInjector extends MyBucketPropsInjector {
+  protected getDefaultBlockPublicAccess() {
+    return undefined;
+  }
+}
+```
+
+The Python client uses `MyTeamInjector`, then they can specify `blockPublicAccess` at Bucket creation time with
+
+```py
+bucket = s3.Bucket(
+    self, "MyBucket",
+    block_public_access=None,
+)
+```
+
+```py
+bucket = s3.Bucket(
+    self, "MyBucket",
+    block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
+)
+```


text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+
+Why do we need Blueprint?
+
+Let's say our org wants to prevent publically accessible S3 Buckets.  You can extend the Bucket class, set `blockPublicAccess: BlockPublicAccess.BLOCK_ALL` and tell all development teams to use our new class.  Or we can also require all development teams to use `blockPublicAccess: BlockPublicAccess.BLOCK_ALL` in their code.  With Blueprint, we no longer have to subclass Bucket and it is easy for development teams to use.
+
+What are the key pieces of Blueprint?
+
+`IPropertyInjector` - An IPropertyInjector defines a way to inject additional properties that are not specified in the props.  It is specific to one Construct and operates on that Construct’s properties.  This will be called `Injector` for short.
+
+`propertyInjectors` - A collection of Injectors attached to App, Stage, or Stack.  Any matching Construct created in this App, Stage, or Stack will have the Injector applied to it.
```
moelasmar, line 17: I believe we also agreed on that the injectors can be also attached to any construct. It can be attached to any node in the nodes tree.

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+    this.constructFqn = Bucket.UNIQUE_FQN;
+  }
+
+  inject(originalProps: BucketProps, _context: InjectionContext): BucketProps {
+    return {
+      blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
+      enforceSSL: true,
+      ...originalProps,
+    };
+  }
```
moelasmar, line 114: can we let customers provide only the default values they want to apply, and let us make the inject function part of CDK workflow that it controls. So customers will do something like this 
```
getDefaults(): BucketProps {
	return {
		blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
		enforceSSL: true,
	};
}
```
then in CDK side in inject we can call `getDefaults` function. This give us more flexibility in future if we want to add more logic, and also from customer POV, they only need to provide default values.

moelasmar (approved):

text/0693-property-injection.md
pcheungamz: Yes, I have an example of that in the FAQ.  The common way is to attach them to App, Stage, or Stack.

text/0693-property-injection.md
pcheungamz: Why can't the builders provide the default values directly in their originalProps?  This way, `inject` won't overwrite any of their provided values.

moelasmar: ok, yeah I missed this point. thanks @pcheungamz

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
+# Blueprint via Property Injection
```
rix0rrr, line 1: This document is proposing "blueprints" and "property injection" at the same time. 

I think the reader would benefit from a small definition of both terms and how they relate.

text/0693-property-injection.md
rix0rrr: Please write something to clarify that here, it confuses me every time I read it. You can try verbiage like the following:

> A collection of injectors attached to the construct tree. Injectors can be attached to any construct, but in practice we expect most of them will be attached to `App`, `Stage` or `Stack`.

That acknowledges both the genericity of the proposed solution as well as the expected use cases.

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+An Org sets the standards and default value, and it is the responsibility of the development teams to adhere to that.  We also recognize that there are situations where the development team might need to override the standard.  In these situations, we value the autonomy of the development team by giving them the flexibility to override.
+
+In the Working Backwards example below, the Org defines standard defaults for development teams to use, but development teams can override the defaults if they need to.  Development teams can also write their own Injectors and share them across teams.  
+
+## Working Backwards
+
+### Development Team Experience
+
+Development teams can start using Blueprint by attaching propertyInjectors to App.
+In the example below, this dev team wants use the Property Injectors provided by its org for S3 Buckets and Lambda Fucntions.
```
rix0rrr, line 30:
```suggestion
In the example below, this dev team wants use the Property Injectors provided by its org for S3 Buckets and Lambda Functions. 
```

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+
+See [Scope Tree Traversal](#scope-tree-traversal) for more information.
+
+### Organization Level Standardization
+
+As an org with many dev teams, we can standardize how AWS Resources are created by providing dev teams with Injectors.
+
+AWS CDK will expose this `IPropertyInjector` interface for orgs to implement:
+
+```ts
```
rix0rrr, line 67: Leave out this code block I think. I understand the desire to be complete and accurate, but we're purely talking about the perspective of the implementor here. Let's show a representative code snippet they would write themselves. That means we don't need to include show the library code that powers their implementation (yet).

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+```
+
+Here is a simple PropertyInjector to make sure the S3 Bucket is not publically accessible and SSL is used.
+
+```ts
+class MyBucketPropsInjector implements IPropertyInjector {
+  public readonly constructFqn: string;
+
+  constructor() {
+    this.constructFqn = Bucket.UNIQUE_FQN;
```
rix0rrr, line 105:
```suggestion
    this.constructFqn = Bucket.PROPERTY_INJECTION_FQN;
```

And I'm not even sure that I like the term "fqn" here. Users shouldn't care what's in there, so it shouldn't matter if it's an FQN. For all intents and purposes it's an opaque token... but I also don't want to use the term "token" because that already means something in CDK-land.

How about `PROPERTY_INJECTION_ID` or something like that?

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  inject(originalProps: BucketProps, _context: InjectionContext): BucketProps {
```
rix0rrr, line 108:
```suggestion
  public inject(originalProps: BucketProps, _context: InjectionContext): BucketProps {
```

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+      enforceSSL: true,
+      ...originalProps,
+    };
+  }
+}
+```
+
+Notes:
+* We will add Bucket.UNIQUE_FQN to unique identify this Construct.  This is a new Property that we will add to Bucket and other supported Constructs.
+* This TypeScript code will set blockPublicAccess to BlockPublicAccess.BLOCK_ALL and enforceSSL to true if these properties are not specified in originalProps.  This implementation allows the dev teams to override the org's recommended defaults.
```
rix0rrr, line 120: Also include a sentence here making explicit that it's the exmaple implementation of the injector that leads to dev teams being able to overwrite the defaults. The injector can be written differently to explicitly always overwrite dev team decisions.

text/0693-property-injection.md
```diff
@@ -0,0 +1,468 @@
...
+
+Development teams can start using Blueprint by attaching propertyInjectors to App.
+In the example below, this dev team wants use the Property Injectors provided by its org for S3 Buckets and Lambda Fucntions. 
+
+```ts
+import { BucketPropsInjector, FunctionPropsInjector } from '@my-org/standard-props-injectors';
+...
+const app = new App({
+  ...
+  propertyInjectors: [
```
rix0rrr, line 37: Show the `PropertyInjectors.of()` syntax here as an equivalent alternative solution.

text/0693-property-injection.md
pcheungamz: I try to define that in line 7.
```
Blueprint is a way for orgs to enforce standards by setting default values to CDK Constructs.  The Blueprint standards can be shared and applied to many development teams.  This is done via injecting default property values at creation time.
```

What is a better way to communicate this?

text/0693-property-injection.md
pcheungamz: Thanks

text/0693-property-injection.md
pcheungamz: I was trying to be generic since FQN can be useful elsewhere, but I see you point.

I like `PROPERTY_INJECTION_ID`.  Will change that.  thanks

text/0693-property-injection.md
pcheungamz: ok thanks.

text/0693-property-injection.md
pcheungamz: ok

text/0693-property-injection.md
```diff
@@ -0,0 +1,526 @@
...
+  }
+}
+```
+
+Notes:
+
+* We will add Bucket.PROPERTY_INJECTION_ID to unique identify this Construct.
+This is a new Property that we will add to Bucket and other supported Constructs.
+* This TypeScript code will set blockPublicAccess to BlockPublicAccess.BLOCK_ALL and enforceSSL to true
+if these properties are not specified in originalProps.
```
alexpulver, line 119: What happens if the properties are specified? It would be great to have a warning in case both the code and injector define a value for the property. Perhaps also the ability to explicitly suppress the warning.

text/0693-property-injection.md
```diff
@@ -0,0 +1,526 @@
+# Property Injection Implementation of Blueprints
+
+* **Original Author(s):**: @pcheungamz
+* **Tracking Issue**: [#693](https://github.com/aws/aws-cdk-rfcs/issues/693)
+* **API Bar Raiser**: @rix0rrr
+
+`Property injection` is a new mechanism that makes it possible to control the properties that are used to instantiate a construct via an out-of-band mechanism.
```
alexpulver, line 7: Is there some alignment story with tools like [cdk-nag](https://github.com/cdklabs/cdk-nag)?

text/0693-property-injection.md
rix0rrr: Do keep in mind that the current design will not make it possible for the property injector to add warnings *on* the construct (because when the injector executes, the construct won't exist yet!), so the best it can do is add warnings on its parent scope.

text/0693-property-injection.md
pcheungamz: After reading through the link you provided, I have a basic understand of cdk-nag.  I think cdk-nag and Property Injectors are both trying to prevent builders from creating AWS resources incorrectly.  cdk-nag will warn the builders about non-compliant resource creation, but does not provide remediation.  Property Injectors will create resources with the correct properties.

text/0693-property-injection.md
```diff
@@ -0,0 +1,526 @@
...
+Constructs explicitly opt in to being injectable; all L2 constructs in the AWS CDK standard library will be property injectable.
+
+Property injection allows the implementation of `Blueprints`:
+a collection of property injectors vended inside a large organization to make it easy for application builders to apply
+organizational best practices when configuring constructs.
+"Blueprints" is a branding term, and they do not imply or require any new technical capability besides property injection.
+Blueprints by themselves are not a compliance enforcement mechanism;
+instead they are a mechanism to make it easier for developers to hit compliance targets that are already being enforced via other means.
+
+**Why do we need Blueprints?**
```
dbartholomae, line 17: How do Property Injectors interact with Aspects? When should I use property injectors to modify resources and when aspects? Based on the RFC so far I don't get yet why I should use injectors at all instead of just modifying the resources via Aspects.

text/0693-property-injection.md
```diff
@@ -0,0 +1,526 @@
...
+```
+
+In this case, Constructs created in betaStack will use `BetaBucketPropsInjector` instead of `BucketPropsInjector`.
+This is useful when Beta resources have different requirements.
+
+See [Scope Tree Traversal](#scope-tree-traversal) for more information.
+
+### Organization Level Standardization
+
+As an org with many dev teams, we can standardize how AWS Resources are created by providing dev teams with Injectors.
```
dbartholomae, line 92: If the teams still need to actitvely use the injector, then what's the benefit compared to just giving them a "SecureBucket" class to build upon? In both cases, the org still needs to rely on the devs to actually use the injector/construct, but with a SecureBucket at least the code would be easier to debug (anyone wondering why the bucket isn't publicly accessible needs to look just at the bucket, not go through every single Stack/Stage/App layer above to verify whether an injector exists).

dbartholomae (commented):
With this, I would need to check every layer above a construct for possible injectors to understand why behaviour is different. On the other hand, I don't really get anything compared to just having a custom construct library, as injectors still require teams to actually apply them and then not to overwrite the defaults. And if I really need to make sure that e.g. buckets are not set to public, I could instead add an Asepct that modifies all buckets to ensure that they are not public.
So for me, I don't really understand yet what this achieves that would justify the added complexity, especially since it breaks locality of code changes.

text/0693-property-injection.md
pcheungamz: One disadvantage of subclassing is that it breaks other tools like `cdk-nag`.  `cdk-nag` checks for Bucket, but not SecureBucket subclass.  Since `Property Injector` is native to Bucke, `cdk-nag` can work with it.

pcheungamz: Another tradeoff to consider when writing your own SecureBucket is the maintenance burden that comes with it.  You will constantly need to sync your package that contains SecureBucket with AWS-CDK.  AWS-CDK is currently at 2.182.0 and there is usually a new minor version every week.  My org currently uses a subclass solution and we are very familiar with this weekly sync'ing.  This is why we are moving away from a subclass solution.

pcheungamz: I also want to quote
```
Finally, keep in mind that writing your own "L2+" constructs might prevent your developers from taking advantage of AWS CDK packages such as [AWS Solutions Constructs](https://docs.aws.amazon.com/solutions/latest/constructs/welcome.html) or third-party constructs from Construct Hub. These packages are typically built on standard AWS CDK constructs and won't be able to use your wrapper constructs.
```
from this doc: https://docs.aws.amazon.com/cdk/v2/guide/best-practices.html

Issue #162, opened 2020-05-21
CDK Refactoring Tools


Gtofig:
At the moment, when I move constructs around in CDK, this results in logical IDs changing and resources being replaced during deployment, even stateful ones. This causes major pain, especially if it's a production service. 

I would like CDK to offer a simpler way of refactoring code without having to worry about resource replacement.

### Use Case

Refactoring code is a natural part of development, including CDK development. I don't want to have to worry about breaking production due to moving constructs around.

### Proposed Solution

Option 1) Work with CloudFormation team to support resource logical ID renaming without replacement (maybe use CDK metadata as a key to identify that it's the same resource)

Option 2) Make it easier to hardcode logical IDs in CDK. At the moment it only works via L1 resources, but most constructs we deal with are not L1, therefore it becomes painful to manually set logical IDs. Perhaps Construct could expose a method to hardcode prefix for logical IDs

### Other

* [ ] :wave: I may be able to implement this feature request
* [X] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

Gtofig:
> Are you familiar with [`renameLogicalId`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#rename-wbr-logical-wbr-idoldid-newid)?

Yes, but this only works at L1 level, so if I have a high level construct which I move around, it becomes more complicated. Perhaps you have a pattern in mind for this?

tdclark:
It also seems like a big ask to have to drop down to `renameLogicalId` every time I want to move/refactor a construct around my CDK app. Also, how would this function for composed constructs? If I have a construct that I vend to multiple apps, how would `renameLogicalId` help me move around some of the sub-constructs within the construct and not have impact on the users of the construct?

tdclark:
Any updates here? It seems untenable to not allow folks to refactor CDK code after the first deployment without causing breakages.

rix0rrr:
I think the better solution is to allow remapping construct paths at any point in the tree.

As long as that's a hierarchical operation we should be fine.

njlynch:
Use case:

Creating and introducing L2s where only L1s previously existed. Doing this and retaining the logical IDs (preventing replacement) is difficult, and can only be done with the right circumstances.

**Example:**
```ts
class Foo extends Resource implements IFoo {
  constructor(...) {
    new CfnFooHelper(this, 'FooHelper', { /* ... */ });
[...]

PR #705, opened 2025-02-19
CDK refactoring support

otaviomacedo:
This is a request for comments about CDK Refactoring Support. See #162 for
additional details. 

---

### DIFF
--- a/text/0162-refactoring-support.md
+++ b/text/0162-refactoring-support.md
@@ -23,12 +23,12 @@ different CDK applications into a single reusable construct (often called an L3
 construct). Introducing a new node for the L3 construct in the construct tree
 will rename the logical IDs of the resources in that subtree.
 
-Additionally, you might need to move resources within the tree for better
-readability or between stacks to isolate concerns. Accidental renames have also
-caused issues for customers in the past. Perhaps even worse, if you depend on a
-third-party construct library, you are not in control of the logical IDs of
-those resources. If the library changes the logical IDs from one version to
-another, you will be affected without any action on your part.
+Also, you might need to move resources within the tree for better readability or
+between stacks to isolate concerns. Accidental renames have also caused issues
+for customers in the past. Perhaps even worse, if you depend on a third-party
+construct library, you are not in control of the logical IDs of those resources.
+If the library changes the logical IDs from one version to another, you will be
+affected without any action on your part.
 
 To address all these problems, the CDK CLI now automatically detects these
 cases, and refactors the stack on your behalf, using the new CloudFormation
@@ -76,7 +76,7 @@ The refactored construct tree looks like this:
 Even though none of the resources have changed, their paths have
 (from `MyStack/Bucket/Resource` to `Web/Website/Origin/Resource` etc.) Since the
 CDK computes the logical IDs of the resources from their path in the tree, all
-three resources will have different logical IDs changed.
+three resources will have their logical IDs changed.
 
 If you run `cdk deploy` now, by default the CLI will detect these changes and
 present you with a selection prompt:
@@ -343,7 +343,7 @@ better experience by automatically detecting these cases, and interacting with
 the user when necessary.
 
 Another alternative is to use aliases, using [Pulumi's model]
-[pulumi-aliases] as inspiration. This feature would be similar to the  
+[pulumi-aliases] as inspiration. This feature would be similar to the
 `renameLogicalId` function, but operating on a higher level of abstraction, by
 taking into account the construct tree and construct IDs. And, just like
 `renameLogicalId`, it could be perceived as a workaround. However, we are open
--- a/text/0162-refactoring-support.md
+++ b/text/0162-refactoring-support.md
@@ -5,27 +5,27 @@
 - **API Bar Raiser**: @{BAR_RAISER_USER}
 
 An improvement to the CDK CLI `deploy` command, to protect against replacement
-of resources that had their logical IDs modified. The CLI now detects these
-cases automatically, and refactors the stack before the actual deployment.
+of resources when they change location. The CLI now detects this case, and
+automatically refactors the stack before the actual deployment.
 
 ## Working Backwards
 
 AWS CloudFormation identifies resources by their logical ID. As a consequence,
 if you change the logical ID of a resource after it has been deployed,
 CloudFormation will create a new resource with the new logical ID and possibly
-delete the old one. In any case, for stateful resources, this may cause
-interruption of service or data loss, or both.
+delete the old one. For stateful resources, this may cause interruption of
+service or data loss, or both.
 
-Historically, the advice for developers has been to avoid changing logical IDs
-of resources. However, this is not always possible, or goes against good
-engineering practices. For example, you may have duplicated code across
-different CDK applications, and you want to consolidate it into a single
-reusable construct (usually referred to as an L3 construct). The very
-introduction of a new node for the L3 construct in the construct tree will lead
-to the renaming of the logical IDs of the resources in that subtree. You may
-also need to move resources around in the tree to make it more readable, or even
-between stacks to better isolate concerns. Not to mention accidental renames,
-which have also impacted customers in the past.
+Historically, our advice for developers has been to avoid changing logical IDs
+of resources. In practice, however, this is not always possible, or goes against
+good engineering practices. For example, you may have duplicated code across
+different CDK applications which you want to consolidate into a single reusable
+construct (usually referred to as an L3 construct). The very introduction of a
+new node for the L3 construct in the construct tree will lead to the renaming of
+the logical IDs of the resources in that subtree. You may also need to move
+resources around in the tree to make it more readable, or even between stacks to
+better isolate concerns. Not to mention accidental renames, which have also
+impacted customers in the past.
 
 To address all these problems, the CDK CLI now automatically detects these
 cases, and refactors the stack on your behalf, using the new CloudFormation
@@ -55,12 +55,12 @@ account:
 - Rename the bucket from `Bucket` to the more descriptive name `Origin`.
 - Create a new L3 construct called `Website` that groups the bucket and the
   distribution, to make this pattern reusable in different applications.
-- Move the `Website` construct to a new stack called `Web`, for better
-  separation of concerns.
+- Move the web-related constructs (now under the `Website` L3) construct to a
+  new stack called `Web`, for better separation of concerns.
 - Rename the original stack to `Service`, to better reflect its new specific
   role in the application.
 
-The construct tree now looks like this:
+The refactored construct tree looks like this:
 
     App
     ├ Web
@@ -76,14 +76,32 @@ CDK computes the logical IDs of the resources based on their path in the tree,
 all three resources will have different logical IDs in the synthesized template,
 compared to what is already deployed.
 
-If you run `cdk deploy` now, by default the CLI will detect this change and
-automatically refactor the stacks. You will see the following output:
+If you run `cdk deploy` now, by default the CLI will detect these changes and
+present you with a selection prompt:
+
+    The following resources were moved or renamed:
+
+    ┌───────────────────────────────┬───────────────────────────────┬──────────────────────────┐
+    │ Resource Type                 │ Old Logical ID                │ New Logical ID           │
+    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
+    │ AWS::S3::Bucket               │ MyStack.Queue1A4198146        │ Web.Bucket843D52FF       │
+    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
+    │ AWS::CloudFront::Distribution │ MyStack.DistributionE3BB089E  │ Web.Distribution7142E1F1 │
+    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
+    │ AWS::Lambda::Function         │ MyStack.FunctionA5EA2BD8      │ Service.Function8F0BB69B │
+    └───────────────────────────────┴───────────────────────────────┴──────────────────────────┘
+
+    ? What do you want to do? (Use arrow keys)
+    ❯ Execute the refactor and deploy
+      Deploy without refactoring (will cause resource replacement)
+      Quit
+
+If you choose to refactor and deploy, the CLI will show the progress as the
+refactor is executed:
 
     Refactoring...
     Creating stack refactor...
 
-     Prg | Time       | Status               | Resource Type                 | Old Logical ID               | New Logical ID               
-     ----|------------|----------------------|-------------------------------|------------------------------|-----------------------------
      0/3 | 2:03:17 PM | REFACTOR_IN_PROGRESS | AWS::S3::Bucket               | MyStack.Bucket5766466B       | Web.Bucket843D52FF           
      0/3 | 2:03:17 PM | REFACTOR_IN_PROGRESS | AWS::CloudFront::Distribution | MyStack.DistributionE3BB089E | Web.Distribution7142E1F1     
      1/3 | 2:03:18 PM | REFACTOR_COMPLETE    | AWS::S3::Bucket               | MyStack.Bucket5766466B       | Web.Bucket843D52FF           
@@ -93,22 +111,24 @@ automatically refactor the stacks. You will see the following output:
     
     ✅  Stack refactor complete
 
-In case you do want to replace the resources, you can override this default
-behavior and skip the refactoring, by passing the `--skip-refactoring` flag to
-the CLI, or by configuring this setting in the `cdk.json` file:
+You can configure the refactoring behavior by passing the `--refactoring` flag
+to the CLI, or by configuring this setting in the `cdk.json` file:
 
 ```json
 {
   "app": "...",
-  "skipRefactoring": true
+  "refactoring": "EXECUTE_AND_DEPLOY"
 }
 ```
 
 ### Ambiguity
 
-In some cases, the CLI may not be able to automatically refactor the stack. To
-understand this, consider the following example, where there are two _identical_
-resources, called `Queue1` and `Queue2`, in the same stack:
+In the unlikely event that there are two or more _identical_ resources (that is,
+resources with the same properties, except for the `Metadata` field), and you
+rename or move two of them at the same time, the CLI will not be able to
+automatically determine which resource should be replaced by which. For example,
+suppose you have two identical queues, named `Queue1` and `Queue2`, in the same
+stack:
 
     App
     └ Stack
@@ -125,7 +145,7 @@ If they get renamed to, let's say, `Queue3` and `Queue4`,
 Then the CLI will not be able to establish a 1:1 mapping between the old and new
 names. In this case, it will ask you to confirm the changes:
 
-    Resource Name Changes
+    Ambiguous Resource Name Changes
     ┌───┬──────────────────────┐
     │   │ Resource             │
     ├───┼──────────────────────┤
@@ -144,13 +164,14 @@ names. In this case, it will ask you to confirm the changes:
 
     Do you wish to deploy these changes (y/n)?
 
-To skip this prompt, pass the `--ignore-ambiguous-renaming` flag to the CLI, or
-configure this setting in the `cdk.json` file:
+To skip this prompt and go straight to deployment, pass the
+`--ignore-ambiguous-refactoring` flag to the CLI, or configure this setting in
+the `cdk.json` file:
 
 ```json
 {
   "app": "...",
-  "ignoreAmbiguousRenaming": true
+  "ignoreAmbiguousRefactoring": true
 }
 ```
 
@@ -189,9 +210,10 @@ the RFC pull request):
 
 ### What are we launching today?
 
-An improvement to the CDK CLI `deploy` command, to protect against replacement
-of resources that had their logical IDs modified. The CLI now detects these
-cases automatically, and refactors the stack before the actual deployment.
+A new developer experience for CDK users, that allows them to change the
+location of a construct (stack plus logical ID) without causing resource
+replacement. This new experience is available in the CDK CLI `deploy`
+and `refactor` commands, as well as a CDK trigger.
 
 ### Why should I use this feature?
 
@@ -206,12 +228,12 @@ stack refactoring support:
 
 ### Can the CLI help me resolve ambiguity when refactoring resources?
 
-Not at the moment. One of the tenets behind this feature is that it should work
-in any environment, including CI/CD pipelines, where there is no user to answer
-questions. Although we could easily extend this feature to include ambiguity
-resolution for the interactive case, it wouldn't transfer well to the
-non-interactive case. If you are interested in an in-depth explanation of the
-problem and a possible solution, check Appendix A.
+Not at the moment. One of the constraints we imposed on this feature is that it
+should work in any environment, including CI/CD pipelines, where there is no
+user to answer questions. Although we could easily extend this feature to
+include ambiguity resolution for the interactive case, it wouldn't transfer well
+to the non-interactive case. If you are interested in an in-depth explanation of
+the problem and a possible solution, check Appendix A.
 
 ### What if I can't use the CDK CLI in my pipeline?
 
@@ -225,12 +247,19 @@ narrow set of permissions.
 
 Alternatively, you can use this feature via a [trigger](#trigger).
 
+### What if the deployment fails?
+
+After refactoring the stack, the CLI will proceed with the deployment
+(assuming that is your choice). If the deployment fails, and CloudFormation
+rolls it back, the CLI will execute a second refactor, in reverse, to bring the
+resources back to their original locations.
+
 ## Internal FAQ
 
 ### Why are we doing this?
 
-This feature was initially requested back in May 2020. It is one of the top 5
-most up-voted RFCs, with 94 thumbs-up. Despite being almost 5 years old, it has
+This feature was initially requested in May 2020. It is one of the top 5 most
+up-voted RFCs, with 94 thumbs-up. Despite being almost 5 years old, it has
 continued to be highly discussed and debated by CDK customers and the community.
 The solution we currently provide, the `renameLogicalId` method, is perceived by
 customers as a workaround, adding unnecessary cognitive load for developers.
@@ -247,31 +276,31 @@ The main attraction of this feature is also, in a way, its greatest deterrent:
 that the refactoring happens automatically, and that the CLI makes decisions
 that the user may not even be aware they need to make (accidental renames, for
 example). This may cause anxiety for some users, who might not understand what
-exactly happened after a successful deployment that had refactoring involved. We
-can mitigate this risk with good documentation and careful interaction design.
+exactly is happening, and what the consequences are. We can mitigate this risk
+with good documentation and careful interaction design.
 
 ### What is the technical solution (design) of this feature?
 
 High level description of the algorithm:
 
-First, it lists all the stacks: both local and deployed. Then it builds an index
-of all resources from all stacks. This index maps the _content_ address
-(physical ID or digest) of each resource to all the _location_ addresses (stack
-name + logical ID) they can be found in. Resources that have different locations
-in new stacks compared to the old ones are considered to have been moved. For
-each of those, it creates a mapping from the old location to the new one.
+First, list all the stacks: both local and deployed. Then build an index of all
+resources from all stacks. This index maps the _content_ address (physical ID or
+digest) of each resource to all the _location_ addresses (stack name + logical
+ID) they can be found in. Resources that have different locations in new stacks
+compared to the old ones are considered to have been moved. For each of those,
+it creates a mapping from the old location to the new one.
 
 Since the CloudFormation API expects not only the mappings, but also the
 templates in their final states, we need to compute those as well. This is done
 by applying all the mappings locally, essentially emulating what CloudFormation
 will eventually do. For example, if a mapping says that a resource has moved
-from stack A with name Foo, to stack B with name Bar, we will remove Foo from
-the template for stack A, and add a new resource called Bar to the template for
-stack B.
+from stack `A` with name `Foo`, to stack `B` with name `Bar`, we will remove
+`Foo` from the template for stack `A`, and add a new resource called `Bar` to
+the template for stack `B`.
 
 At this point, if there is any ambiguity (more than one source or more than one
-destination for a given mapping), it stops and asks the user whether they want
-to proceed.
+destination for a given mapping), stop and ask the user whether they want to
+proceed.
 
 Assuming there were no ambiguities, or the user wants to proceed anyway, it is
 ready to call the API to actually perform the refactor, using the mappings and
@@ -289,19 +318,29 @@ No.
 
 ### What alternative solutions did you consider?
 
-The most straightforward alternative is to implement a wrapper around the
+The most straightforward alternative is to implement a simple wrapper around the
 CloudFormation API, and have the user provide all the parameters: which
 resources to move from which stack to which stack. But the CDK CLI can provide a
 better experience by automatically detecting these cases, and interacting with
 the user when necessary.
 
-Another alternative is to use aliases, following [Pulumi's model]
-[pulumi-aliases], for example. This feature would be similar to the  
+Another alternative is to use aliases, using [Pulumi's model]
+[pulumi-aliases] as inspiration. This feature would be similar to the  
 `renameLogicalId` function, but operating on a higher level of abstraction, by
 taking into account the construct tree and construct IDs. But, just like
 `renameLogicalId`, it could be perceived as a workaround, that adds scar tissue
-to the code every time refactoring is done. However, we could revisit this
-decision if enough customers indicate their preference for it in this RFC.
+to the code at every refactor. However, we could revisit this decision if enough
+customers indicate their preference for it in this RFC.
+
+A possible variation of the solution presented in this RFC is to do something
+similar to resource lookup: for every environment where the application could be
+deployed to, the CLI would have configured in a file what refactors have to be
+done. The entries to this file could be automatically generated by some CLI
+command, run individually in each environment. But this solution requires a lot
+more work and coordination among the parties involved (developers,
+administrators, security engineers, etc.), and is more error-prone: failure to
+record a refactor in the file could lead to inconsistencies between the
+environments, and even unintended resource replacements.
 
 ### What are the drawbacks of this solution?
 
@@ -317,15 +356,15 @@ This improved deployment experience actually consists of two separate steps,
 behind the scenes: refactoring followed by deployment. And the whole workflow is
 controlled by the CLI. As a result, this is not an atomic operation: it is
 possible that the refactoring step succeeds, but before the CLI has a chance to
-deploy, it gets interrupted (computer crash, network failures, etc.) In this
-case, the user will be left with a stack that is neither in the original state
-nor in the desired state.
+deploy the changes, it gets interrupted (computer crash, network failures, etc.)
+In this case, the user will be left with a stack that is neither in the original
+state nor in the desired state.
 
 In particular, the logical ID won't match the CDK construct path, stored in the
 resource's metadata. This has consequences for the CloudFormation console, which
 will show a Tree view that is not consistent with the Flat view.
 
-Some possible solutions to consider, from more specific to more general:
+Possible solutions to consider, from more specific to more general:
 
 - CloudFormation to ignore changes in the `Metadata[aws:cdk:path]` resource
   attribute in refactor operations.
@@ -491,8 +530,17 @@ stacks in an environment, it could also be used for other purposes. For example,
 development tools could use the history to provide a "time machine" feature,
 that allows developers to see the state of the infrastructure at any point in
 time. CloudFormation itself could build on that, and provide a way to roll back
-or forward to another state. Potentially, this could also help with drift
-resolution (or prevention).
+or forward to another state.
+
+Another problem that could be solved with the historical model is the infamous
+"deadly embrace", where a consumer stack depends on a producer stack via
+CloudFormation Exports, and you want to remove the use from the consumer. At the
+moment, customers have to use the `stack.exportValue(...)` method, and do two
+deployments. The history would give the CLI all the information it needs to do
+this without user intervention.
+
+Potentially, this could also help with drift resolution (or prevention), if
+CloudFormation itself starts using the history internally.
 
 [pulumi-aliases]: https://www.pulumi.com/docs/iac/concepts/options/aliases/
 
--- a/text/0162-refactoring-support.md
+++ b/text/0162-refactoring-support.md
@@ -58,7 +58,7 @@ account:
 - Rename the bucket from `Bucket` to the more descriptive name `Origin`.
 - Create a new L3 construct called `Website` that groups the bucket and the
   distribution, to make this pattern reusable in different applications.
-- Move the web-related constructs (now under the `Website` L3) construct to a
+- Move the web-related constructs (now under the `Website` L3) to a
   new stack called `Web`, for better separation of concerns.
 - Rename the original stack to `Service`, to better reflect its new specific
   role in the application.
--- a/text/0162-refactoring-support.md
+++ b/text/0162-refactoring-support.md
@@ -115,6 +115,17 @@ refactor is executed:
     
     ✅  Stack refactor complete
 
+If you want to execute only the automatic refactoring, use the `cdk refactor`
+command. The behavior is basically the same as with `cdk deploy`: it will detect
+whether there are refactors to be made, ask for confirmation if necessary
+(depending on the flag values), and refactor the stacks involved. But it will
+stop there and not proceed with the deployment. If you only want to see what
+changes would be made, use the `--dry-run` flag.
+
+To perform refactoring, the CLI needs new permissions in the bootstrap stack.
+Before using this feature, run `cdk bootstrap` for every target environment, to
+add these new permissions.
+
 Please note that the same CDK application can have multiple stacks for different
 environments. In that case, the CLI will group the stacks by environment and
 perform the refactoring separately in each one. So, although you can move
@@ -122,69 +133,78 @@ resources between stacks, both stacks involved in the move must be in the same
 environment. Trying to move resources across environments will result in an
 error.
 
-If you want to execute only the automatic refactoring, use the `cdk refactor`
-command. The behavior is basically the same as with `cdk deploy`: it will detect
-whether there are refactors to be made, ask for confirmation if necessary (
-depending on the flag values), and refactor the stacks involved. But it will
-stop there and not proceed with the deployment. If you only want to see what
-changes would be made, use the `--dry-run` flag.
-
 ### Settings
 
-By default, refactoring is disabled on deployments. To override this behavior,
-you can either use a command line flag (`--refactor-action`) or set the
-`refactorAction` field in the `cdk.json` file. The flag/setting can have one of
-the following values:
+The behavior of the refactoring feature can be controlled by a few settings.
 
-- `CONFIRM`: this enables the behavior described in the previous section, where
-  the CLI shows the changes and asks for confirmation. This only works in
-  interactive mode (TTY).
-- `REFACTOR`: automatically detects resource moves, executes the refactors, and
-  deploys the stacks.
-- `SKIP`: goes straight to deployment, skipping refactoring. This is the default
-  behavior.
-- `QUIT`: process stops, returning a non-zero exit status.
+For both `deploy` and `refactor`:
 
-The last three values are the same options as the prompt.
+- `--export-mapping=<FILE>`: writes the mapping to a file. The file can be used
+  later to apply the same refactors to other environments.
+- `--import-mapping=<FILE>`: use the mapping from a file, instead of computing
+  it. The file can be generated using the `--export-mapping` option.
+- `--dry-run`: shows on stdout what would happen, but doesn't execute.
+- `--unstable=refactor`: enables the feature. If the flag is not set, the
+  command fails with an error message explaining that the feature is
+  experimental.
 
-If there are ambiguities, the CLI will look for a mapping file to resolve them.
-If the file does not exist or doesn't apply, the CLI will fail. Otherwise, it
-will apply the mappings and proceed with the deployment.
+For `deploy` only:
 
-These options are summarized in the following flowchart:
+- `--refactoring-action=[ACTION]`: the action to take in case there is a
+  refactor to be made. Possible values for `ACTION` are:
+    - `confirm`: ask the user what to do. This is the scenario described in the
+      **How it works** section.
+    - `refactor`: automatically refactor and deploy.
+    - `quit`: stop with a non-zero exit code.
+    - `skip`: deploy without refactoring. This is the default value.
 
-```mermaid
-flowchart TD
-    map{Mapping file}
-    amb{Ambiguities}
-    conf[/"Refactor action\n (cli flag or config)"/]
-    ask[/"Ask user"/]
-    comm>"Same options as the 'Refactor action' flag (except CONFIRM). 
-    Omitted to avoid cluttering the diagram."]
-    tty{is TTY}
-    DEPLOY([Deploy only])
-    APPLY([Apply mapping and deploy])
-    FAIL([Fail])
-    AUTO([Auto refactor and deploy])
+All these settings are also available in the `cdk.json` file:
 
-    conf -->|null or SKIP| DEPLOY
-    conf -->|CONFIRM| tty
-    map -->|No| FAIL
-  map -->|Yes| APPLY
-  conf -->|QUIT| FAIL
-    conf -->|REFACTOR| amb
-    amb -->|No| AUTO
-    amb -->|YES| map
-    tty -->|Yes| ask
-    tty -->|No| FAIL
-    ask --> comm
+```json
+{
+  "app": "...",
+  "refactor": {
+    "refactoringAction": "confirm",
+    "exportMapping": "output.json",
+    "dryRun": true
+  }
+}
 ```
 
-Also note that this feature is still experimental, so you have to pass
-`--unstable=refactor` flag to confirm you are aware of this to both `deploy` 
-and `refactor` commands.
+### Explicit mapping
 
-### Rollbacks
+Although the CLI will automatically refactor only what is unambiguous, you may
+still need to have more control over the refactoring process. Companies usually
+have stringent policies on how changes are made to the production environment,
+for example. One such policy is that every change must be explicitly declared in
+some sort of code, including refactors.
+
+In a situation like this, you can import and export mapping files. Here is how
+it works: at development time, you made a change that the CLI detected as a
+refactor. Since you want that refactor to be propagated to other environments,
+you export the mapping file, with the command
+`cdk refactor --export-mapping=file.json`.
+
+There are at least two possible paths from here, depending on the company's
+policies:
+
+1. You send the exported file to an operations team, who will review it. If
+   approved, they manually run the command
+   `cdk refactor --import-mapping=file.json` on every protected environment in
+   advance (i.e., before your changes get deployed to those environments). When
+   you import a mapping, the CLI won't try to detect refactors.
+2. The `--apply-mapping` option is also available for the `deploy` command. So
+   you can commit the mapping file to version control, and configure the CLI to
+   use apply it on every deployment. This is a more convenient option, because
+   it requires less coordination between different roles, but the mapping file
+   must be removed from the repository once the refactor has been applied.
+   Otherwise, the next deployment will fail.
+
+In general, if the protected environment is not in the same state as the
+environment where the mapping was generated, the `refactor --apply-mapping`
+command will fail.
+
+### Rollback
 
 After refactoring the stack, the CLI will proceed with the deployment
 (assuming that is your choice). If the deployment fails, and CloudFormation
@@ -251,6 +271,11 @@ names. In this case, it will show you the ambiguity, and stop the deployment:
     If you want to take advantage of automatic resource refactoring, avoid 
     renaming or moving multiple identical resources at the same time.
 
+    If you want to provide an explicit mapping, use the --import-mapping option.
+
+As the message suggests, you can use the `--import-mapping` option as a way to
+resolve ambiguities.
+
 ### Programmatic access
 
 The same refactoring feature is also available in the CDK toolkit library:
@@ -281,7 +306,6 @@ try {
   }
 }
 ```
-
 ---
 
 
@@ -315,15 +339,6 @@ stack refactoring support:
 - Moving constructs between different stacks.
 - Renaming stacks.
 - Upgrading dependencies on construct libraries.
-
-### Can the CLI help me resolve ambiguity when refactoring resources?
-
-Not at the moment. One of the constraints we imposed on this feature is that it
-should work in any environment, including CI/CD pipelines, where there is no
-user to answer questions. Although we could easily extend this feature to
-include ambiguity resolution for the interactive case, it wouldn't transfer well
-to the non-interactive case. If you are interested in an in-depth explanation of
-the problem and a possible solution, check Appendix B.
 
 ## Internal FAQ
 
@@ -398,7 +413,11 @@ templates computed previously.
 
 ### Is this a breaking change?
 
-No.
+No. By default, the CLI will skip refactoring on deployment. The user must
+explicitly enable it by passing a value other than `skip` to the
+`--refactoring-action` (or `refactoringAction` in `cdk.json`) option. Also, this
+feature will initially be launched as experimental, and the user must
+acknowledge this by passing the `--unstable=refactor` flag.
 
 ### What alternative solutions did you consider?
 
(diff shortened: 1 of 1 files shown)


text/0162-refactoring-support.md
```diff
@@ -0,0 +1,451 @@
...
+(account and region). Given that one CDK app can have stacks for multiple
+environments, the CLI will group the stacks by environment and perform the
+refactoring separately in each one. Trying to move resources between stacks that
+belong in different environments will result in an error.
+
+### Is this a breaking change?
+
+No.
+
+### What alternative solutions did you consider?
```
fitz-vivodyne, line 256: Worth considering the way that Pulumi handles refactoring via Aliases as a way to allow the developer to explicitly create associations between old/new resources. 
* [docs](https://www.pulumi.com/docs/iac/concepts/options/aliases/)
* [blog](https://www.pulumi.com/blog/cumundi-guest-post/)

I've had positive experiences refactoring Pulumi code in the past, although the ecosystem is a bit simpler and most resources are closer to L1 constructs.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+better experience by automatically detecting these cases, and interacting with
+the user when necessary.
+
+Another alternative is to use aliases, using [Pulumi's model]
+[pulumi-aliases] as inspiration. This feature would be similar to the  
+`renameLogicalId` function, but operating on a higher level of abstraction, by
+taking into account the construct tree and construct IDs. But, just like
+`renameLogicalId`, it could be perceived as a workaround, that adds scar tissue
+to the code at every refactor. However, we could revisit this decision if enough
+customers indicate their preference for it in this RFC.
```
fitz-vivodyne, line 333: If you're doing it the way Pulumi does it, you're not necessarily accumulating scar tissue, which implies permanence.
Once Pulumi code is run with the aliasing in place, the resource would permanently be modified to use the new resource ID and the alias can be safely removed.

This is more akin to a temporary feature flag that can be cleaned up after completion vs a permanent addition to the code base.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+{
+  "app": "...",
+  "refactoring": "EXECUTE_AND_DEPLOY"
+}
+```
+
+### Ambiguity
+
+In the unlikely event that there are two or more _identical_ resources (that is,
+resources with the same properties, except for the `Metadata` field), and you
```
jiayiwang7, line 127: same properties do you mean same property field and value combinations?
e.g. if i have a stack of lambda function with loggroup pointing to a logical id of a log group resource. and i refactor this entire stack, will cfn refactoring api handle it?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+command in the main CDK CLI. This command uses assumes a specific role, with a
+narrow set of permissions.
+
+Alternatively, you can use this feature via a [trigger](#trigger).
+
+### What if the deployment fails?
+
+After refactoring the stack, the CLI will proceed with the deployment
+(assuming that is your choice). If the deployment fails, and CloudFormation
+rolls it back, the CLI will execute a second refactor, in reverse, to bring the
```
jiayiwang7, line 254: if the deployment fails and CFN rolls it back, why does CDK CLI need to execute a second refactor in reverse? is it because the location of resources in tree is already updated in CDK causing a drift in cfn?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+  organization, or to create reusable components.
+- Moving constructs between different stacks.
+- Renaming stacks.
+
+### Can the CLI help me resolve ambiguity when refactoring resources?
+
+Not at the moment. One of the constraints we imposed on this feature is that it
+should work in any environment, including CI/CD pipelines, where there is no
+user to answer questions. Although we could easily extend this feature to
+include ambiguity resolution for the interactive case, it wouldn't transfer well
```
SimonCMoore, line 234: The core capability is identifying logical ids that changed and generating the refactor for them at run time. Could we add a capability to generate on a developer machine a mapping file that could be used during ci/cd deployment such that the developer can confirm mappings in advance and commit the mapping file to be picked up during CI/CD?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+passing the flag `autoRefactor` to your application:
+
+```typescript
+const app = new App({
+  autoRefactor: true,
+  ...
+});
+```
+
+If this is enabled, the CLI will skip the refactoring step during deployment,
```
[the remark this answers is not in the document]
SimonCMoore: I am unclear how it could skip the refactor step and not end up deleting the old resources and recreating them as new resources with the new logicalid. Doesn't the underlying CFN implementation require refactoring to occur before attempting to deploy the updated CDK application with the new logicalids. Whereas triggers are executed after I believe

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,549 @@
...
+include ambiguity resolution for the interactive case, it wouldn't transfer well
+to the non-interactive case. If you are interested in an in-depth explanation of
+the problem and a possible solution, check Appendix A.
+
+### What if I can't use the CDK CLI in my pipeline?
+
+Some customers have their own mechanisms for deploying stacks to AWS, that don't
+use the CDK CLI. If that is your case, there is still a way you can use this
+feature. The refactoring logic will also be released as a Node.js library and a
+standalone CLI tool (names TBD). If you can incorporate any of these into your
```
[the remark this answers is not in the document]
otaviomacedo: Removed. We probably don't need this.

text/0162-refactoring-support.md
otaviomacedo: Because, from the user's point of view, the deployment is atomic: either all of it happens (refactor + deployment), or neither of them. If the deployment fails, CloudFormation will only roll back the deployment. For CloudFormation, the refactoring was an independent operation that happened before and was successful. So, renaming in reverse is the CDK CLI way of completing the rollback, from the user's perspective.

text/0162-refactoring-support.md
otaviomacedo: Created an appendix to explain more precisely what it means for two resources to be equivalent.

text/0162-refactoring-support.md
otaviomacedo: We can, but I don't feel comfortable with this. It's just too hard for users to get this right. To create a mapping file, you need to predict what the current state of the environment will be. But when it comes the time to apply this mapping, if the state is not what you predicted, the result is undefined.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+
+If there are refactors to be done, this is considered an error, because we can't
+guarantee that the refactor makes sense (let alone that this was the developer's
+intent).
+
+#### The future
+
+There is still some work to be done to prove this system works in practice. But
+assuming it does, we could use it to expand the scope to which the automatic
+refactoring applies. Consider the case in which you want to rename a certain
```
jiayiwang7, line 534: ah so the proposed in-scope refactoring solution will treat this case (rename a resource + update properties of the resource) a replacement not refactoring? would we print out a warning message for this case?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+
+Before that, let's define a digest function, `d`:
+
+    d(resource) = hash(type + physicalId)                       , if physicalId is defined
+                = hash(type + properties + dependencies.map(d)) , otherwise
+
+where `hash` is a cryptographic hash function. In other words, if a resource has
+a physical ID, its type and physical ID uniquely identify that resource. So we
+compute the hash from these two fields. Otherwise, the hash is computed from its
+type, its own properties (that is, excluding properties that refer to other
```
jiayiwang7, line 573: ah makes sense! so base on this, the digest of a lambda function with property log group referring to another resource will remain the same after refactoring, as the log group property is excluded in the calculation.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+state, for that matter), condition 2 is satisfied. But remember that the scope
+of the mapping file is the difference between two consecutive revisions. If the
+developer's local branch is multiple commits ahead of the revision that was
+deployed to production, the source state in production is not the same as the
+one in the mapping file, violating condition 1.
+
+#### Making history
+
+An improvement we can make is to turn this into an event sourcing system.
+Instead of storing a single mapping between two states, we store the whole
```
jiayiwang7, line 452: i assume the whole history needs to be stored in a persistent layer (S3?) that all parties using cdk can access?

text/0162-refactoring-support.md
otaviomacedo: The digest of the log group is included in the calculation. But since logical IDs are never part of the digest, we can freely change the logical ID of the log group without affecting the digests of the resources that depend on it.

text/0162-refactoring-support.md
otaviomacedo: Yes, the history would have to be stored in two places: in Git, and in a persistent on AWS. S3 seems the most straightforward choice, but it could also be in DynamoDB, etc.

text/0162-refactoring-support.md
otaviomacedo: >ah so the proposed in-scope refactoring solution will treat this case (rename a resource + update properties of the resource) a replacement not refactoring?

Yes.

>would we print out a warning message for this case?

We could do that, but this case is not as clear-cut as it sounds. Unlike the in-scope case, there is no unambiguous way to differentiate between _refactor_ and _delete + creation_. We would need to define some distance function between two resources, and empirically come up with a threshold for that distance. If there are pairs whose distance is below the threshold, we would print a warning. But it's possible, and something I'm interested in exploring, anyway. This could be a phase 3.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+This improved deployment experience actually consists of two separate steps,
+behind the scenes: refactoring followed by deployment. And the whole workflow is
+controlled by the CLI. As a result, this is not an atomic operation: it is
+possible that the refactoring step succeeds, but before the CLI has a chance to
+deploy the changes, it gets interrupted (computer crash, network failures, etc.)
+In this case, the user will be left with a stack that is neither in the original
+state nor in the desired state.
+
+In particular, the logical ID won't match the CDK construct path, stored in the
+resource's metadata. This has consequences for the CloudFormation console, which
```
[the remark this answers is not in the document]
otaviomacedo: >can user rerun the cdk commands and expect success?

Yes, in the next run, there will be nothing to refactor, so it goes straight to deployment. If the deployment succeeds, the consistency is restored.

>is this the only user-facing impact?

This is the most obvious one. But there is a more esoteric case I can think of:

1. Developer renames a resource from "A" to "B", and also adds some new ones.
2. Refactor succeeds.
3. Deployment fails and the stack is rolled back.
4. Before it has a chance to revert the refactor, the CLI gets interrupted.
5. Developer decides to abandon that change, and reverts the code to the
   previous state (before the change described in step 1). Now, in the code, the
   resource is called "A", but in the deployed stack, it is called "B".
6. Some time later, some other developer, unaware of this discrepancy, makes a
   change to the content of resource "A", believing this will lead to an update.
7. The CDK CLI won't detect this as a rename (because of the content change),
   and will thus proceed to deployment, leading to the replacement of resource "B".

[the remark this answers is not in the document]
otaviomacedo: Good point. But that works to our advantage, then, doesn't it? It's a warning to developers that it's probably not safe to deploy. But, of course, they can still ignore it.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+
+If you run `cdk deploy` now, by default the CLI will detect these changes and
+present you with a selection prompt:
+
+    The following resources were moved or renamed:
+
+    ┌───────────────────────────────┬───────────────────────────────┬──────────────────────────┐
+    │ Resource Type                 │ Old Logical ID                │ New Logical ID           │
+    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
+    │ AWS::S3::Bucket               │ MyStack.Queue1A4198146        │ Web.Bucket843D52FF       │
```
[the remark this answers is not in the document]
otaviomacedo: Thanks! Yes, this was a typo.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+    If you want to take advantage of automatic resource refactoring, avoid 
+    renaming or moving multiple identical resources at the same time.
+
+    If these changes were intentional, and you want to proceed with the
+    resource replacements, please confirm below.
+
+    Do you wish to deploy these changes (y/n)?
+
+To skip this prompt and go straight to deployment, pass the
+`--ignore-ambiguous-refactoring` flag to the CLI, or configure this setting in
```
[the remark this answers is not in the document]
otaviomacedo: I don't like this flag, either, but for another reason: it's a second flag related to refactoring, that may be confusing for users. So it seems there are enough arguments against it. I'll remove the flag and the command line prompt, and just fail in case of ambiguities.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+  "app": "...",
+  "ignoreAmbiguousRefactoring": true
+}
+```
+
+If you want to execute only the automatic refactoring, use the `cdk 
+refactor` command. The behavior is basically the same as with `cdk deploy`: it
+will detect whether there are refactors to be made, ask for confirmation if
+necessary (depending on the flag values), and refactor the stacks involved. But
+it will stop there and not proceed with the deployment.
```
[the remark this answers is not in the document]
otaviomacedo: That was the intention, but I missed it 🤦 Thanks!

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+### What if I can't use the CDK CLI in my pipeline?
+
+Some customers have their own mechanisms for deploying stacks to AWS, that don't
+use the CDK CLI. If that is your case, there is still a way you can use this
+feature: the refactoring logic will also be released in the CDK toolkit library,
+that can be used programmatically by your own tools. If you can incorporate it
+into your deployment tooling, you will have the same functionality as the
+`refactor` command in the CDK CLI.
+
+### What if the deployment fails?
```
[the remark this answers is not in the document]
otaviomacedo: The refactoring of all resources in all stacks are executed in a single operation. So this much, at least, is atomic.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+First, list all the stacks: both local and deployed. Then build an index of all
+resources from all stacks. This index maps the _content_ address (physical ID or
+digest) of each resource to all the _location_ addresses (stack name + logical
+ID) they can be found in. Resources that have different locations in new stacks
+compared to the old ones are considered to have been moved. For each of those,
+it creates a mapping from the old location to the new one.
+
+Since the CloudFormation API expects not only the mappings, but also the
+templates in their final states, we need to compute those as well. This is done
+by applying all the mappings locally, essentially emulating what CloudFormation
```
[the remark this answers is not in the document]
otaviomacedo: Not exactly. Because the synthesis may also have additions, deletions and updates. These cannot be part of the refactor API call.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+High-level tasks:
+
+- Implement resource equivalence without physical ID.
+- Implement the computation of mappings.
+- Implement the display of the mappings to the user.
+- Implement ambiguity detection.
+- Implement the display of the ambiguities to the user.
+- Add physical ID to resource equivalence.
+
+#### Phase 2 (application)
```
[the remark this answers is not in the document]
otaviomacedo: I hadn't thought about that. The splitting into these phases was more to de-risk the project and lead to smaller PRs. But I guess we can plan a bake time, as well. How long would be good, do you reckon?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+with good documentation and careful interaction design.
+
+### What is the technical solution (design) of this feature?
+
+High level description of the algorithm:
+
+First, list all the stacks: both local and deployed. Then build an index of all
+resources from all stacks. This index maps the _content_ address (physical ID or
+digest) of each resource to all the _location_ addresses (stack name + logical
+ID) they can be found in. Resources that have different locations in new stacks
```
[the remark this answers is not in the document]
otaviomacedo: Clarified.

text/0162-refactoring-support.md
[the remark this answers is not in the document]
iliapolo: Ah thats cool.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+    
+    ✅  Stack refactor complete
+
+You can configure the refactoring behavior by passing the `--refactoring` flag
+to the CLI, or by configuring this setting in the `cdk.json` file:
+
+```json
+{
+  "app": "...",
+  "refactoring": "EXECUTE_AND_DEPLOY"
```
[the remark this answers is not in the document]
otaviomacedo: I don't think it makes sense to configure this per stack. A refactor operation applies to a set of stacks at the same time.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
+# CDK Refactoring Support
+
+- **Original Author(s):**: @otaviomacedo
+- **Tracking Issue**: #162
+- **API Bar Raiser**: @rix0rrr
+
+An improvement to the CDK CLI and toolkit library, to protect against
```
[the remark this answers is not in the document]
mrgrain:
```suggestion
An improvement to the AWS CDK Toolkit (both CLI and Library), to protect against
```

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+replacement of resources when they change location. The CLI now detects this
```
mrgrain, line 8: nit, also everywhere following etc.
```suggestion
replacement of resources when they change location. The Toolkit now detects this
```

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+of resources. In practice, however, this is not always possible, or goes against
+good engineering practices. For example, you may have duplicated code across
+different CDK applications which you want to consolidate into a single reusable
+construct (usually referred to as an L3 construct). The very introduction of a
+new node for the L3 construct in the construct tree will lead to the renaming of
+the logical IDs of the resources in that subtree. You may also need to move
+resources around in the tree to make it more readable, or even between stacks to
+better isolate concerns. Not to mention accidental renames, which have also
+impacted customers in the past.
+
```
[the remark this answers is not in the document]
mrgrain: Mentioning automatic cross-stack references would also be good. But know this paragraph is getting long so maybe an appendix.

text/0162-refactoring-support.md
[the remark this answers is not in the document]
iliapolo: Oh right. feels somewhat tricky to implement. Can you add a high level overview on how we're gonna do this?

text/0162-refactoring-support.md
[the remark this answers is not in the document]
iliapolo: 3 months? unless we have good engagement before that and feel confident.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+    └ MyStack
+      ├ Bucket
+      ├ Distribution
+      └ Function
+
+Now suppose you make the following changes, after having deployed it to your AWS
+account:
+
+- Rename the bucket from `Bucket` to the more descriptive name `Origin`.
+- Create a new L3 construct called `Website` that groups the bucket and the
```
mrgrain, line 59: Thinking out loud: Using "L3" in this context might be more confusing than it helps.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+  distribution, to make this pattern reusable in different applications.
+- Move the web-related constructs (now under the `Website` L3) construct to a
```
mrgrain, line 61:
```suggestion
- Move the web-related constructs (now under the `Website` L3) to a
```

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+    │    └ Distribution
+    └ Service
+      └ Function
+
+Even though none of the resources have changed, their paths have
+(from `MyStack/Bucket/Resource` to `Web/Website/Origin/Resource` etc.) Since the
+CDK computes the logical IDs of the resources from their path in the tree, all
+three resources will have their logical IDs changed.
+
+If you run `cdk deploy` now, by default the CLI will detect these changes and
```
kaizencc, line 81: nit: the word 'now' confuses me. i initially took it to mean what currently happens before the CDK refactoring feature. but after reading i realize it means "now, after the refactoring feature". its not technically wrong, but i was confused for a bit.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+```json
+{
+  "app": "...",
+  "refactoring": "EXECUTE_AND_DEPLOY"
+}
+```
+
+Please note that the same CDK application can have multiple stacks for different
+environments. In that case, the CLI will group the stacks by environment and
+perform the refactoring separately in each one. Trying to move resources between
```
mrgrain, line 128: This makes sense, but your example above makes it look like moving resources between stacks is possible.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+     0/3 | 2:03:17 PM | REFACTOR_IN_PROGRESS | AWS::S3::Bucket               | MyStack.Bucket5766466B       | Web.Bucket843D52FF           
+     0/3 | 2:03:17 PM | REFACTOR_IN_PROGRESS | AWS::CloudFront::Distribution | MyStack.DistributionE3BB089E | Web.Distribution7142E1F1     
+     1/3 | 2:03:18 PM | REFACTOR_COMPLETE    | AWS::S3::Bucket               | MyStack.Bucket5766466B       | Web.Bucket843D52FF           
+     1/3 | 2:03:18 PM | REFACTOR_IN_PROGRESS | AWS::Lambda::Function         | MyStack.FunctionA5EA2BD8     | Service.Function8F0BB69B     
+     2/3 | 2:03:19 PM | REFACTOR_COMPLETE    | AWS::CloudFront::Distribution | MyStack.DistributionE3BB089E | Web.DistributionE3BB089E 
+     3/3 | 2:03:20 PM | REFACTOR_COMPLETE    | AWS::Lambda::Function         | MyStack.FunctionA5EA2BD8     | Service.FunctionA5EA2BD8     
+    
+    ✅  Stack refactor complete
+
+You can configure the refactoring behavior by passing the `--refactoring` flag
```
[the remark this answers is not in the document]
kaizencc: i'm interpreting this to mean that you can configure refactoring behavior in lieu of the prompt, and the values you can send in to the `--refactoring` command is `EXECUTE_AND_DEPLOY`, or `DEPLOY_WITHOUT_REFACTORING`. but i think i'm filling in the blanks to get to that understanding; i'd rather the RFC be clear about the options.

I also assume that this is primarily a way to bypass the prompting of the user, which would benefit the toolkit-lib, is that right?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+
+Even though none of the resources have changed, their paths have
+(from `MyStack/Bucket/Resource` to `Web/Website/Origin/Resource` etc.) Since the
+CDK computes the logical IDs of the resources from their path in the tree, all
+three resources will have their logical IDs changed.
+
+If you run `cdk deploy` now, by default the CLI will detect these changes and
+present you with a selection prompt:
+
+    The following resources were moved or renamed:
```
iankhou, line 84: This describes behavior after implementation. It would be helpful to also include an example of behavior before the implementation (what the CDK currently does), to better illustrate to users the problem we are solving.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+Please note that the same CDK application can have multiple stacks for different
+environments. In that case, the CLI will group the stacks by environment and
+perform the refactoring separately in each one. Trying to move resources between
+stacks that belong in different environments will result in an error.
+
+### Rollbacks
+
+After refactoring the stack, the CLI will proceed with the deployment
+(assuming that is your choice). If the deployment fails, and CloudFormation
+rolls it back, the CLI will execute a second refactor, in reverse, to bring the
```
mrgrain, line 135: why would CFN roll back?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+resources back to their original locations.
```
mrgrain, line 136: We will need a flag to not do that. Or maybe the existing `--no-rollback` will do.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+### Programmatic access
+
+The same refactoring feature is also available in the CDK toolkit library:
+
+```typescript
+declare const toolkit: Toolkit;
+declare const cx: ICloudAssemblySource;
+
+// To execute possible refactors as part of the deploy operation:
+await toolkit.deploy(cx, {
```
iankhou, line 212: How does the toolkit library handle the question of equivalent resources? Will it simply fail?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+resources back to their original locations.
+
+### Ambiguity
+
+Imagine a person walking down the street, and someone takes two snapshots of
+them, a few seconds apart. When you look at the snapshots, you can tell that
+it's the _same person_, but at different locations, and not two different
+people. What leads you this conclusion? The fact that the person's features
+(face, clothes, height, etc.) are exactly the same in both photographs. But if,
+instead of one person walking, you have two identical siblings, wearing the same
```
mrgrain, line 145:
```suggestion
instead of one person walking, you have two identical twins, wearing the same
```

text/0162-refactoring-support.md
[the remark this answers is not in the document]
otaviomacedo: Works for me.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+
+If they get renamed to, let's say, `Queue3` and `Queue4`,
+
+    App
+    └ Stack
+      ├ Queue3
+      └ Queue4
+
+then the CLI will not be able to establish a 1:1 mapping between the old and new
+names. In this case, it will show you the ambiguity, and stop the deployment:
```
mrgrain, line 178: I still think we should allow the user to provide a resolution file in this case (or any case really) that explicitly maps these changes.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+[ ] Signed-off by API Bar Raiser @rix0rrr
+
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+A new developer experience for CDK users, that allows them to change the
+location of a construct (stack plus logical ID) without causing resource
```
mrgrain, line 237: Now I am really confused. Are we supporting moving stacks or not?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+top of that API to bring a seamless experience to CDK users.
+
+### Why should we _not_ do this?
+
+The main attraction of this feature is also, in a way, its greatest deterrent:
+that the refactoring happens automatically, and that the CLI makes decisions
+that the user may not even be aware they need to make (accidental renames, for
+example). This may cause anxiety for some users, who might not understand what
+exactly is happening, and what the consequences are. We can mitigate this risk
+with good documentation and careful interaction design.
```
mrgrain, line 285: Nice!

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,622 @@
...
+destination for a given mapping), stop and ask the user whether they want to
+proceed.
+
+Assuming there were no ambiguities, or the user wants to proceed anyway, it is
+ready to call the API to actually perform the refactor, using the mappings and
+templates computed previously.
+
+### Is this a breaking change?
+
+No.
```
mrgrain, line 335: Call-out: We need to implement this carefully for `cdk deploy` otherwise it could accidentally break Cx.

text/0162-refactoring-support.md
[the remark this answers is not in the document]
kaizencc: is this a good use case for the `--unstable` flag?

text/0162-refactoring-support.md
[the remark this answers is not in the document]
kaizencc: refactoring is a boolean command?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,586 @@
...
+### B. Equivalence between resources
+
+To detect which resources should be refactored, we need to indentify which
+resources have only changed their location, but have remained "the same", in
+some sense. This can be made precise by defining an [equivalence relation] on
+the set of resources.
+
+Before that, let's define a digest function, `d`:
+
+    d(resource) = hash(type + physicalId)                       , if physicalId is defined
```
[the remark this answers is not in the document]
otaviomacedo: Oh, I see. The word "stable" is creating confusion. I meant to say that they are stable under renaming. The naive approach would be to hash every property. But then the digest would change if any of its dependencies were renamed. The proposed function here is immune to that. But, of course, it's not stable under property changes. So we couldn't use it as the construct ID. I'll rephrase this.

text/0162-refactoring-support.md
otaviomacedo: Good catch!

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+
+For `deploy` only:
+
+- `--refactoring-action=[ACTION]`: the action to take in case there is a
+  refactor to be made. Possible values for `ACTION` are:
+    - `confirm`: ask the user what to do. This is the scenario described in the
+      **How it works** section.
+    - `refactor`: automatically refactor and deploy.
+    - `quit`: stop with a non-zero exit code.
+    - `skip`: deploy without refactoring. This is the default value.
```
bdoyle0182, line 165: Will this always be the default? I would expect most use cases not want this to be the default and either default to refactor or confirm; and choosing this as the default is to not affect existing behavior. Would the plan be to change the default to refactor in the next major version of cdk cli and this functionality is no longer experimental

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+explicitly enable it by passing a value other than `skip` to the
+`--refactoring-action` (or `refactoringAction` in `cdk.json`) option. Also, this
+feature will initially be launched as experimental, and the user must
+acknowledge this by passing the `--unstable=refactor` flag.
+
+### What alternative solutions did you consider?
+
+The most straightforward alternative is to implement a simple wrapper around the
+CloudFormation API, and have the user provide all the parameters: which
+resources to move from which stack to which stack. But the CDK CLI can provide a
```
bdoyle0182, line 433: +1, a simple wrapper around the cloud formation stack refactor api does not solve the developer pain problem over the current solution. It might be cleaner than pinning logical ids so refactor can actually occur rather than have a forever mapping in repo, but it doesn't reduce developer cost in terms of management and cognitive load in any way if the cdk doesn't attempt to detect and perform refactors on its own.

text/0162-refactoring-support.md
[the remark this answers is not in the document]
bdoyle0182: You're example was exactly what came to mind to me for risk here.

7. occurring may sound like a super niche case, but in practice this kind of thing happening on a large team with CD pipelines is fairly regular occurrence. It's concerning when the outcome is potentially catastrophic. 

How about the cli being smarter to detect the current cloud formation status? Something like if refactor is turned on and the cli sees that the current stack state is Update Rollback X, then check if refactor rollback is needed before continuing with new deploy?

Another option would be something like refactor plan being stored as state somewhere (cdk asset bucket?) and this state only gets deleted if the stack update completes or the stack update fails and the refactor rollback completes. If the deploy gets interrupted, on next deploy the cdk would detect the refactor state and complete the rollback before continuing with next deploy.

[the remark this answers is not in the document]
SimonCMoore: Is refactoring itself atomic, could some resources be refactored and others not be refactored? Could such a state be recovered from?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+    d(resource) = hash(type + physicalId)                       , if physicalId was defined by the user
+                = hash(type + properties + dependencies.map(d)) , otherwise
+
+where `hash` is a cryptographic hash function. In other words, if a resource has
+a physical ID, we can use the physical ID plus its type to uniquely identify
+that resource. In this case, the digest can be computed from these two fields
+alone. A corollary is that such resources can be renamed and have their
+properties updated at the same time, and still be considered equivalent.
+
+Otherwise, the digest is computed from its type, its own properties (that is,
```
SimonCMoore, line 539: Is this CFN type? So refactoring from one CDK construct to another that generated the same CFN type and properties would be possible?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+  say, a `executeChangeSetAndRefactor()`).
+
+Since all these options depend on changes on the CloudFormation side, and this
+edge case is unlikely to happen, we are going to address it later.
+
+## Appendix
+
+### A. Equivalence between resources
+
+To detect which resources should be refactored, we need to indentify which
```
SimonCMoore, line 523: So we don't check for resources that will be deleted and ones that will be created and find some near equivalency? Could we generate a warning, e.g. Lambda A is being deleted Lambda B is being created, one property difference is preventing refactoring.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+place, and the resource properties as its "personal" features. If a resource has
+different properties before and after a potential deployment, the best we can do
+is assume they are different resources. If the properties are the same from one
+point in time to the other, they very likely are the same resource (although the
+developer still has the last word on this). But if there are two resources that
+have the same properties in the same stack, they are like the twin siblings
+case: there's no way to tell them which is which in case they both move to a
+different stack or get renamed.
+
+Indistinguishable resources in this sense are said to be _equivalent_ (see
```
SimonCMoore, line 245: In software terms does it matter? if we use one twin or the other, they are identical and for software that would make them interchangeable.  Perhaps what we are missing is what might make the CLI see them as identical when in fact they are not and thus not interchangeable?  Perhaps some examples, I am guessing maybe some stateful resource might have same properties but actually be in a different state?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+location of a construct (stack plus logical ID) without causing resource
+replacement. This new experience is available in the CDK CLI `deploy`
+and `refactor` commands, as well as in the toolkit library.
+
+### Why should I use this feature?
+
+If you ever find yourself doing one of the following, you will benefit from
+stack refactoring support:
+
+- Renaming constructs, either intentionally or by mistake.
```
SimonCMoore, line 343: What about moving from nearly equivalent constructs, perhaps an org created lambda construct to the cdk lib lambda construct and vice versa (also see my comment asking if we can detect near equivalence, in this case maybe the org version sets a default the user was not aware of, if we could make them aware they could change their code to explicitly match the default or maybe we could allow them to refactor and change the default)

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+
+To address all these problems, the CDK CLI now automatically detects these
+cases, and refactors the stack on your behalf, using the new CloudFormation
+stack refactoring API. This brings more flexibility for developers, and reduces
+the risk of accidental changes that lead to resource renaming.
+
+### How it works
+
+When you run `cdk deploy`, the CLI will compare the templates in the cloud
+assembly with the templates in the deployed stack. If it detects that a resource
```
shikha372, line 41: how will we detect these changes, is it only gonna be based on property level differences?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,606 @@
...
+
+Resources that have different locations before and after, are considered to have
+been moved. For each of those, create a mapping from the source (the currently
+deployed location) to the destination (the new location, in the local template).
+Example:
+
+    // Keys are the content address of the resources
+    // Values are the location addresses
+    { 
+      "5e19886121239b7a": { // Moved and renamed -> goes to mapping
```
shikha372, line 392: will it throw ambiguity in case the rename in logicalID is within the same stack?

otaviomacedo: No, rename in the same stack is just a special case.

text/0162-refactoring-support.md
otaviomacedo: Hmm, I suppose you're right. There are two potential issues related to ambiguity:

- **Resource state**: for this, ambiguity is only a problem if there are other resources pointing to the ambiguous ones, and we mix them up. For example, there is a pair of twin buckets, and a Lambda function pointing to one of them before the refactoring, and to the other one after. But this is totally preventable by using the structure of the resource graphs themselves ("the mapping should be an isomorphism"). I hadn't appreciated that before. Thanks for bringing this up.
- **Naming**: if there is more than one way to refactor, we can end up with logical IDs that don't match the meaning of the resources they name. Suppose the twin buckets are called `FinancialData` and `CatImages`, which get renamed to `FinancialReports` and `CatPhotos`. A human can tell which should be renamed to which. The program can't, so it may do it the wrong way around. No harm will be done to the data or the bucket configurations, but it's still an annoyance to have a bucket called `FinancialReports` in CloudFormation, but when you look inside, you see cat images.

The second case can't be reliably solved automatically. On the other hand, I guess we can think of it as just a special case of the main workflow, in which the developer may exercise control over the mapping, via mapping files and CLI interaction. I am going to drop this section, and just add a note there.

text/0162-refactoring-support.md
otaviomacedo: I am very much interested in that generalization: replacing the notion of equivalence with the notion of distance between two resources. Equivalent resources would then be a special case (distance = 0). The challenge is to come up with a workable distance function, and probably also a threshold. But given the interest (this is also related to Shikha's question), I can add this to the scope, but address in a third phase.

text/0162-refactoring-support.md
otaviomacedo: Yes to both questions.

text/0162-refactoring-support.md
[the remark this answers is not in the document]
otaviomacedo: @bdoyle0182 that's a good suggestion. Thanks!

@SimonCMoore yes, refactoring itself is atomic.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,665 @@
...
+- Renaming stacks.
+- Upgrading dependencies on construct libraries.
+
+### What if I incorrectly refactor a stack?
+
+The CLI offers mechanisms to prevent accidental refactoring, such as dry runs,
+interactive mode, and importing and exporting of mapping files. Nevertheless,
+mistakes can happen. The first thing to be aware of is that refactors don't
+affect the resources themselves. The worst that can happen is you ending up with
+incorrect resource IDs in CloudFormation. The second thing is that refactors can
```
rix0rrr, line 311: You say this like it's not a big deal, but even though the refactor itself won't be too bad, the next deployment *can* mess things up pretty badly if you're not paying attention.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,665 @@
...
+be always be reverted. Simply edit the mapping file if you have one (or create
+one from scratch if you don't), and apply it to the affected stacks.
```
rix0rrr, line 313: Is this something we should do automatically? Generate a `refactor-rollback-mapping.json` ?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,665 @@
...
+        "before": ["stack1/logicalId3"]
+      },
+      "07266c4dd0146e8a": { // Unchanged -> ignored 
+        "before": ["stack1/logicalId4"]
+        "after": ["stack1/logicalId4"]
+      }
+    }
+
+Since the CloudFormation API expects not only the mappings, but also the
+templates in their final states, we need to compute those as well. This is done
```
rix0rrr, line 520: Is that not the output of a regular `synth` already?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,665 @@
...
+single step because it would create a dangling reference (and therefore
+CloudFormation would reject it). The solution is also similar to what we suggest
+for solving the deadly embrace, but is done automatically here:
+
+1. Update the value of the output in `Producer1` to the hard-coded physical ID
+   of the bucket.
+2. Call the refactor API to move the bucket from `Producer1` to `Producer2`,
+   preserving the old output in `Producer1` while creating a new output in
+   `Producer2`, that references the bucket.
+3. Update the `Fn::ImportValue` in `Consumer` to point to the new output in
```
rix0rrr, line 620: So 3 deployments, right?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,665 @@
...
+
+- **A distance function**. To generalize the notion of strict equality to a more
+  flexible notion of similarity, we need to come up with a distance function,
+  that is, a function that takes two resources as input, and produces a number
+  as output. The higher the number, the more different the resources are.
+- **A threshold for the distance**. This is a cut-off value that we can use to
+  decide whether two resources are similar enough to be candidates for
+  refactoring. It will depend on the distance function we choose and also on
+  experimentation.
+- **Graph isomorphism**. Since we will not have a digest anymore to quickly find
```
rix0rrr, line 648: All very true, but I think we can probably wing it with a little user feedback and be *mostly* fine.

Take objects of the same type, order them by distance, ask the user to pick the mapping. Bob's your uncle 😎

text/0162-refactoring-support.md
otaviomacedo: Ok, two questions then:

1. Can you elaborate? What is the mess up scenario you have in mind?
2. What should we do? Always fail on ambiguous cases? (I'm open to considering that)

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+  deploy`: it will detect whether there are refactors to be made, ask for
+  confirmation if necessary (depending on the flag values), and refactor the
+  stacks involved. But it will stop there and not proceed with the deployment.
+  If you only want to see what changes would be made, use the `--dry-run` flag.
+- If you pass any filters to the `deploy` command, the refactor will work on
+  those stacks plus any other stacks the refactor touches. For example, if you
+  choose to only deploy stack A, and a resource was moved from stack A to stack
+  B, the refactor will involve both stacks. But if there was a rename in, let's
+  say, stack C, it will not be refactored. The same set of filters is available
+  for the `refactor` command.
```
mrgrain, line 135: there's also an `--exclusive` flag. This might need defining.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+have stringent policies on how changes are made to the production environment,
+for example. It could be that your company requires that every change must be
+explicitly declared in some sort of code, including refactors (as opposed to
+computed on-the-fly).
+
+In a situation like this, you can import and export mapping files. Here is how
+it works: at development time, you made a change that the CLI detected as a
+refactor. Since you want that refactor to be propagated to other environments,
+you export the mapping file, with the command
+`cdk refactor --export-mapping=file.json`.
```
mrgrain, line 159: 🚲 🛖 probably should align this closer to the existing options for import
```
  --record-resource-mapping, -r  If specified, CDK will generate a mapping of
                                 existing physical resources to CDK resources to
                                 be imported as. The mapping will be written in
                                 the given file path. No actual import operation
                                 will be performed                      [string]
  --resource-mapping, -m         If specified, CDK will use the given file to
                                 map physical resources to CDK resources for
                                 import, instead of interactively asking the
                                 user. Can be run from scripts          [string]
```

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+    │ AWS::S3::Bucket               │ MyStack/Bucket/Resource       │ Web/Website/Origin/Resource       │
+    ├───────────────────────────────┼───────────────────────────────┼───────────────────────────────────┤
+    │ AWS::CloudFront::Distribution │ MyStack/Distribution/Resource │ Web/Website/Distribution/Resource │
+    ├───────────────────────────────┼───────────────────────────────┼───────────────────────────────────┤
+    │ AWS::Lambda::Function         │ MyStack/Function/Resource     │ Service/Function/Resource         │
+    └───────────────────────────────┴───────────────────────────────┴───────────────────────────────────┘
+
+    ? What do you want to do? (Use arrow keys)
+    ❯ Execute the refactor and deploy
+      Deploy without refactoring (will cause resource replacement)
```
iliapolo, line 104: Where do we stand on this? 

```suggestion
      Deploy without refactoring (will cause resource replacement)
      Create a `refactor.json` file describing the expected operations in detail
```

text/0162-refactoring-support.md
otaviomacedo: Sounds good to me.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+
+For `deploy` only:
+
+- `--refactoring-action=[ACTION]`: the action to take in case there is a
+  refactor to be made. Possible values for `ACTION` are:
+    - `confirm`: ask the user what to do. This is the scenario described in the
+      **How it works** section.
+    - `refactor`: automatically refactor and deploy.
+    - `quit`: stop with a non-zero exit code.
+    - `skip`: deploy without refactoring. This is the default value.
```
mrgrain, line 206: 🚲 🛖 

seems overly complicated for the default causes

```
--refactor
--no-refactor
```

Do we need `quit`? We don't have this for `--rollback` either.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+2. The `--apply-mapping` option is also available for the `deploy` command. So
+   you can commit the mapping file to version control, and configure your
+   pipeline to use it on every deployment. This is a more convenient option, as
+   it requires less coordination between different roles. Every time a refactor
+   is applied, a record of it is stored in the environment. This is to prevent
+   the same refactor from being applied multiple times.
+
+In general, if the protected environment is not in the same state as the
+environment where the mapping was generated, the `refactor --apply-mapping`
+command will fail.
```
iliapolo, line 177: How will it fail? what are the implications of failing? This seems like a scary touch point we need to clarify.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+  financial reports will happen to be called `CustomerData` in CloudFormation.
+  Its physical ID and properties remain unchanged, though.
+- When you actually want to replace resources, despite them having the same
+  properties before and after the deployment. In this case, you can remove the
+  unwanted mapping from the mapping file.
+
+We recommend that you always use the `confirm` option at development time. This
+way, you can see what changes the CLI is going to make, and decide whether they
+are correct. This gives you a chance to modify any incorrect mappings (by, for
+example, exporting the mapping, editing it, and importing it again).
```
mrgrain, line 245: I tend to agree.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+  it. This file can be generated using the `--export-mapping` option.
+- `--unstable=refactor`: used to acknowledge that this feature is experimental.
+  If the flag is not set, and the CLI would try to perform some refactor, the
+  command fails with an error message explaining why.
+
+For `deploy` only:
+
+- `--refactoring-action=[ACTION]`: the action to take in case there is a
+  refactor to be made. Possible values for `ACTION` are:
+    - `confirm`: ask the user what to do. This is the scenario described in the
```
iankhou, line 202: For `confirm`, how would we provide this programmatically? Or would it be a CLI-only option?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,677 @@
...
+
+Possible solutions to consider, from more specific to more general:
+
+- CloudFormation to ignore changes in the `Metadata[aws:cdk:path]` resource
+  attribute in refactor operations.
+- CloudFormation to allow resource additions and deletions in refactor
+  operations.
+- Two-phase commit. The CLI could create the refactor and the changeset, and
+  then have a new command to execute both in a single atomic operation (let's
+  say, a `executeChangeSetAndRefactor()`).
```
mrgrain, line 457: this is never going to be really atomic, although much closer than what we currently propose

otaviomacedo: As atomic as CloudFormation can be, I suppose.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+`MyStack/Bucket/Resource` to `Web/Website/Origin/Resource` etc.) Since the CDK
+computes the logical IDs of the resources from their path in the tree, all three
+resources will have their logical IDs changed.
+
+From your perspective, the changes above are mere moves and renames, but what
+CloudFormation sees is the deletion of old resources and the creation of new
+ones. By opting in to the CDK refactoring support, the CDK CLI will work on your
+behalf to notify CloudFormation of your intention.
+
+To execute a refactor, run `cdk refactor --unstable=refactor`. The CLI will show
```
GavinZZ, line 81: what does the `--unstable` flag mean? and why do we name it `--unstable`?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+    "dryRun": true
+  }
+}
+```
+
+### Rollback
+
+After refactoring the stack, the CLI will proceed with the deployment
+(assuming that is your choice). If the deployment fails, and CloudFormation
+rolls it back, the CLI will execute a second refactor, in reverse, to bring the
```
GavinZZ, line 216: I didn't get this part. With my experiments with CFN refactor, if an errors happens, CFN rolls it back automatically to the original state. Why does CDK CLI need to execute a second refactor?

text/0162-refactoring-support.md
otaviomacedo: This is a generic flag we use in the CLI to indicate that the feature is unstable. By passing that flag, the user is acknowledging that.

text/0162-refactoring-support.md
otaviomacedo: Yes, if the refactor call fails, it will be rolled back automatically. But what this paragraph is describing is the case where the _deployment_ (that immediately follows a refactor) fails.

text/0162-refactoring-support.md
GavinZZ: ahh I see, thank you for the clarification!

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+     0/3 | 2:03:17 PM | REFACTOR_IN_PROGRESS | AWS::CloudFront::Distribution | MyStack/Distribution/Resource | Web/Website/Distribution/Resource     
+     1/3 | 2:03:18 PM | REFACTOR_COMPLETE    | AWS::S3::Bucket               | MyStack/Bucket/Resource       | Web/Website/Origin/Resource           
+     1/3 | 2:03:18 PM | REFACTOR_IN_PROGRESS | AWS::Lambda::Function         | MyStack/Function/Resource     | Service/Function/Resource     
+     2/3 | 2:03:19 PM | REFACTOR_COMPLETE    | AWS::CloudFront::Distribution | MyStack/Distribution/Resource | Web/Website/Distribution/Resource 
+     3/3 | 2:03:20 PM | REFACTOR_COMPLETE    | AWS::Lambda::Function         | MyStack/Function/Resource     | Service/Function/Resource     
+    
+    ✅  Stack refactor complete
+
+You can also refactor the resources as part of a deployment, by running `cdk 
+deploy --refactoring-action=refactor --unstable=refactor`. You will be shown the
```
shikha372, line 113: are there any other actions that they can specify with --refactoring-action ??

otaviomacedo: For now, only "refactor" and "skip".

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+accomplish this, you can remove the unwanted mapping from the mapping file.
+
+### Settings
+
+You have a few settings available to control the behavior of the refactoring
+feature.
+
+For both `deploy` and `refactor`:
+
+- `--record-resource-mapping=<FILE>`: writes the mapping to a file. The file can
```
SimonCMoore, line 180: I am unclear if when I use this it only writes the mapping file out OR it does the remapping and writes a file out (to apply elsewhere)?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+`FinancialData` and `CustomerData`, and you want to rename them to
+`FinancialReports` and `CustomerInfo`, respectively. But the CLI could end up
+doing the opposite, so that, after the refactor, the bucket that contains
+financial reports would be called `CustomerData` in CloudFormation, and vice
+versa.
+
+Even though all the resources involved in this scenario would remain unchanged,
+we have decided to err on the side of caution and not perform the refactor. If,
+as the result of computing a refactor, the CLI detects such a case of ambiguity,
+it will fail with an error message explaining the situation.
```
SimonCMoore, line 307: I am unclear if we could output a potential mapping file with --record-resource-mapping, edit it and then use that to do the correct mapping, of whether that will also stop before creating the output file? (maybe the output file should have some warnings about the ambiguity/needs human review).

I assume the ambiguity can be resolved by supplying the mapping file (however obtained), please make that clear.

text/0162-refactoring-support.md
5d: do you mean the second refactor here is undoing the resource mappings?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+
+### Settings
+
+You have a few settings available to control the behavior of the refactoring
+feature.
+
+For both `deploy` and `refactor`:
+
+- `--record-resource-mapping=<FILE>`: writes the mapping to a file. The file can
+  be used later to apply the same refactors in other environments.
```
samson-keung, line 181: I was under the impression that resource mappings are environment specific. So how can it a resource mapping be applied to other environments? Do you have an example of how you envision the file JSON structure would be?

text/0162-refactoring-support.md
samson-keung: I would expect that if the CFN refactor call succeeded, then it means that actual deployment would also succeed. What would be the scenarios that the CFN refactor call succeeded but not the deployment that follows?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,702 @@
...
+
+### Limitations and failure modes
+
+#### Pipelines with version superseding
+
+As we have seen, there are two ways to apply refactors: automatically, by
+consuming a mapping file, or interactively, when the CLI computes the refactor
+to be made, and the user is asked whether to proceed. If you want to use this
+feature in a CI/CD pipeline, so that the refactors are applied automatically,
+you must use commit the mapping file along with your code and use it on each
```
5d, line 248: Will the mapping file be used in the next refactoring, or is it generated without considering the history? What happens if a customer refactors the same resource twice?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,829 @@
...
+In the CDK construct library there are some constructs that rely on
+CloudFormation's behavior to force the replacement of resources. API Gateway's
+`Deployment` and Lambda's `Version` are two examples of this strategy. By
+default, the CLI will detect these changes as refactors and not replacements. We
+have already seen a mechanism to override this: skip files. But it would be too
+inconvenient to have to maintain skip files and remembering to update it every
+time the logical ID changes.
+
+To solve this, CDK constructs can be automatically excluded by calling the new
+method `Stack.skipRefactoring(constructToBeSkipped)`. By calling this method,
```
otaviomacedo, line 274: @rix0rrr is this a good way to do it?

rix0rrr: I think it will be some form of metadata.

```ts
Annotations.of(constructToBeSkipped).add(SKIP_REFACTORING, true);
```

Or something like that? Could be abstracted via a `Stack` method, sure.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,829 @@
...
+`deploy` and `refactor` commands.
+
+This is a regular JSON file, that you can write yourself. But to make it
+convenient for you, every time the CLI computes a mapping, it will automatically
+generate a refactor file at the root of your CDK application, with a timestamp
+in the name. Example: `.refactor-2025-04-03_12-05-49.json`. You can use these
+files as a starting point to edit the mapping, combine multiple mappings into
+one, split mappings into multiple files, etc.
+
+#### Skip file
```
rix0rrr, line 179: Isn't this better in one file together with the refactorings?

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,829 @@
...
+If you still want to perform a refactor, you have to explicitly provide the
+mapping. There are two ways to do this: you can either create a refactor file
+from scratch, or you can pass another command line option to only override the
+ambiguous part of the mapping.
+
+For example, if the CLI detects that there are two resources in `StackA`, `X`
+and `Y`, that are identical, and both were renamed to `Y` and `Z`, you can tell
+it exactly which mapping you want:
+
+    cdk refactor --override='{"StackA.X": "StackA.Z", "StackA.Y": "StackA.W"}'
```
rix0rrr, line 355: Can we make this slightly nicer to type for humans?

```suggestion
    cdk refactor --map StackA.X:StackA.Z --map StackA.Y:StackA.W
```

rix0rrr (approved):
Final comments

text/0162-refactoring-support.md
otaviomacedo: That was my initial idea. But that file would be used to:

1. Override the result of a computed mapping (the skip part).
2. Preclude mapping computation altogether (the refactor part).

These two functions are mutually exclusive. It would be confusing to have one file with both.

text/0162-refactoring-support.md
otaviomacedo: I like this. Thanks!

rix0rrr (approved):

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,858 @@
...
+
+- If you pass any filters to the `deploy` command, the refactor will work on
+  those stacks plus any other stacks the refactor touches. For example, if you
+  choose to only deploy stack A, and a resource was moved from stack A to stack
+  B, the refactor will involve both stacks. But if there was a rename in, let's
+  say, stack C, it will not be refactored. The same set of filters is available
+  for the `refactor` command.
+- To perform refactoring, the CLI needs new permissions in the bootstrap stack.
+  Before using this feature, run `cdk bootstrap` for every target environment,
+  to add these new permissions.
```
dsilbergleithcu-godaddy, line 152: We need to make sure these changes are available in the experimental module for `app-staging-synthesizer` which allows for bootstrap-less CDK.  We've adopted it at GoDaddy and need it to be supported with this feature.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,858 @@
...
+- `--force`: go ahead with refactoring, without prompting the user.
+
+All these settings are also available in the `cdk.json` file:
+
+```json
+{
+  "app": "...",
+  "refactor": {
+    "refactoringAction": "refactor",
+    "dryRun": true
```
dsilbergleithcu-godaddy, line 245: We rely on snapshot testing EXTENSIVELY at GoDaddy.  We encourage it as a best practice, and ship our starter-template with automatic snapshot testing built in. 

We should be able to specify the refactor file that we want to use in the `cdk.json`.  This would allow:

1) Our snapshot test suite to consume the file and apply it, updating our snapshots and giving us a critical preview of what we will be deploying during a very sensitive deployment.
2) Commit, tracking, and review, of any and all refactoring decisions across the larger team.
3) CICD workflows to run mostly unchanged (we might need to add the unstable flag and let that run at all times).  When a refactor is needed, we could disable `dryRun` and specify a file to use.

text/0162-refactoring-support.md
```diff
@@ -0,0 +1,858 @@
...
+have already seen a mechanism to override this: skip files. But it would be too
+inconvenient to have to maintain skip files and remembering to update it every
+time the logical ID changes.
+
+To solve this, CDK constructs can be automatically excluded by calling the new
+method `Stack.skipRefactoring(constructToBeSkipped)`. By calling this method,
+you are signalling to the CLI, via the cloud assembly, that the corresponding
+CloudFormation resource should (logically) be added to the skip file. Normally,
+this method will be called from the constructor of the class itself:
+`Stack.skipRefactoring(this)`.
```
dsilbergleithcu-godaddy, line 294: This is great. I wonder how you'll make sure that devs KNOW that a particular resource needs to have this enabled.

One thing we see (which is both a wonder and a chore with CDK) is that users often don't understand what their CDK is actually doing. Abstracting away some of the plumbing causes them to not understand how the water flows.

What happens if somebody FORGETS to add this flag, and the refactor goes through? What state is their deployment in? Will there be explicit guidance published (when this feature rolls out) about the kinds of constructs where we need to be careful?

dsilbergleithcu-godaddy (commented):
I love it.  I'm excited to see this feature implemented as it will provide tremendous value to GoDaddy's engineers, and the broader community.  Stack Refactoring from CloudFormation is a game changer, and figuring out how to get CDK on board only opens doors.

text/0162-refactoring-support.md
mrgrain: @dsilbergleithcu-godaddy Thanks for the insight. I'm curious about your snapshot testing setup. Specifically how you are using a specific app's `cdk.json` with them.

Issue #710, opened 2025-03-11
Node.js Version Support Policy for AWS CDK


adamjkeller:
## Description

This RFC proposes the introduction of a formal policy in the AWS Cloud Development Kit (CDK) for discontinuing support of Node.js versions six months after the Node.js community [officially](https://github.com/nodejs/Release#release-schedule) designates them as End-of-Life (EOL). This ensures users have ample transition time and that the AWS CDK remains aligned with actively supported, secure Node.js releases.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @adamjkeller 
| Author(s)           | @adamjkeller 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #712, opened 2025-03-11
RFC 710: Formalizing a Node Deprecation Strategy for AWS CDK

adamjkeller:
This is a request for comments about formalizing a deprecation strategy for Node.js in the AWS CDK. See https://github.com/aws/aws-cdk-rfcs/issues/710 for
additional details. 

[Rendered Version](https://github.com/adamjkeller/aws-cdk-rfcs/blob/nodejs-strategy/text/0710-node-deprecation-strategy.md)

---

### DIFF
--- a/text/0710-node-deprecation-strategy.md
+++ b/text/0710-node-deprecation-strategy.md
@@ -97,6 +97,6 @@ In this proposal, we formalize a policy for all Node.js versions going forward.
 
 ## Conclusion
 
-This RFC seeks to introduce a formal Node.js version support policy for AWS CDK: the AWS CDK stops supporting a Node.js version six months after that version’s official EOL date. This provides a predictable, security-forward posture that balances end-user needs for stability with the AWS CDK’s continued modernization.
+This RFC seeks to introduce a formal Node.js version support policy for AWS CDK, which is for the AWS CDK to stop supporting a Node.js version six months after that version’s official EOL date. This provides a predictable, security-forward posture that balances end-user needs for stability with the AWS CDK’s continued modernization.
 
 Feedback is welcome! If you have any questions or concerns, please comment on this issue and if needed we can refine the proposal.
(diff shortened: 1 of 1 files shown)


johnf:
A couple of thoughts

Is there a good reason to support non-LTS versions? No one should be using these in production. It would halve the number of releases that need support and move the deprecation cycle to yearly. 

I recommend aligning with Node.js's EOL. Node.js provides 30 months, which is effectively 2 years plus 6 months.
By aligning, we discourage users from running unsupported versions of Node in production. Deprecation notices can be sent two years out, which still gives a six-month warning. If a deprecation policy exists, then users immediately have notice of when support will end, so we shouldn't need to give them extra time.

It could be handy to link to https://endoflife.date/nodejs or https://nodejs.org/en/about/previous-releases as a reference.

jiayiwang7:
> Is there a good reason to support non-LTS versions? 

You're right that most production users should be on LTS releases. Our primary focus is indeed on LTS versions, though we've found some developers use non-LTS versions in development environments or for testing new features. Supporting those non-LTS versions can be important for those targeted use cases.

> I recommend aligning with Node.js's EOL. Node.js provides 30 months, which is effectively 2 years plus 6 months.

The reason we plan to maintain the EOL + 6 months approach as originally proposed is because it aligns with the support model used by other AWS services, including the AWS SDK for JavaScript, which follows a similar EOL + 6 months pattern (as referenced in the RFC). This consistency across AWS developer tools provides a more predictable experience for our users.

The 6-month grace period serves several purposes:
1. It gives enterprise customers additional time to complete their upgrade cycles
2. It maintains consistency with other AWS developer tools
3. It provides a balanced approach between security and practical migration timelines

> It could be handy to link to https://endoflife.date/nodejs or https://nodejs.org/en/about/previous-releases as a reference.

Appreciate the suggestion. We'll definitely include these references in our future version documentation page to help users track upcoming EOL dates.

text/0710-node-deprecation-strategy.md
```diff
@@ -0,0 +1,115 @@
...
+* Most AWS CDK users who stay on current LTS releases will remain unaffected by this policy.
+* Users on older Node.js versions are encouraged to upgrade as soon as possible to avoid issues and missed security patches.
+
+## Conclusion
+
+This RFC seeks to introduce a formal Node.js version support policy for AWS CDK, which is for the AWS CDK to stop supporting a Node.js version 
+six months after that version's official EOL date. This provides a predictable, security-forward posture that balances end-user needs for 
+stability with the AWS CDK's continued modernization.
+
+Feedback is welcome! If you have any questions or concerns, please comment on this issue and if needed we can refine the proposal.
```
jiayiwang7, line 115:
```suggestion
Feedback is welcome! If you have any questions or concerns, please comment on this issue and if needed we can refine the proposal.

```

jiayiwang7 (approved):

Issue #715, opened 2025-03-28
CDK Construct Update Support


GavinZZ:
## Description

As CDK evolves, new features, improvements, and fixes are introduced, sometimes requiring changes that are not backward-compatible. To manage these changes while minimizing disruptions, one of the method CDK uses is to introduce a new version of construct or a new module. Examples are:
* New construct version – Provides a redesigned construct in the same module with improved behavior while coexisting with the previous version. i.e. `aws-cdk-lib.aws-dynamodb.Table` to `aws-cdk-lib.aws-dynamodb.TableV2`
* New module version – Releases a new module version with significant changes without affecting the existing one. `aws-cdk-lib.aws-ec2` to `@aws-cdk.aws-ec2-alpha` which rewrites VPC related constructs.

For CDK users, migrating to the new construct or module can be more complex and time-consuming, often necessitating resource replacements or removals that could result in production downtime or other complications. 

The primary goal of this RFC is to outline a migration approach for new construct and new module version that aims to simplify the migration process for users, automating certain tasks where possible, and offering guidance and tools to support the transition. The migration approach must ensure a reliable and controlled process with zero downtime, avoiding unintended resource replacements, enabling safe rollbacks, and preserving configuration consistency.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @CDK team
| Author(s)           | @GavinZZ 
| API Bar Raiser      | @iliapolo 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

mrgrain:
[...]

PR #717, opened 2025-04-01
RFC 0715: CDK Construct Upgrade Support

GavinZZ:
This is a request for comments about CDK Migration Support. See https://github.com/aws/aws-cdk-rfcs/issues/715 for
additional details.

---


text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,412 @@
+# CDK Migration Support
```
SimonCMoore, line 1: CDK construct upgrade support?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,413 @@
...
+
+1. In-place Migration - This approach requires users to refactor current stack with CDK code to
+use V2 constructs and make sure the resource configuration matches by following the migration guide.
+Users then needs to specify the `refactor.json` file for CloudFormation stack refactoring feature to
+map logical ID changes. Then apply CloudFormation stack refactor feature by calling
+`aws cloudformation execute-stack-refactor --stack-refactor-id <id>` to proceed migration.
+
+2. Retain-Remove-Import Migration - This approach expects users to have `RemovalPolicy.RETAIN` set
+on the V1 construct in the stack. Users also need to configure table construct so that the custom resource
+deletion does not delete the stateful resources. If not already set, deploy the retain policy change. Then
```
godwingrs22, line 78: nit: Since Dynamo table construct is one of the example for this strategy, it would be good to mention like as an example.  The description of this strategy immediately jumps into DynamoDB table construct specifics without indicating it's an example.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+Arguments:
+
+- CDK stack ID (Optional)
+    The construct ID of the CDK stack from your app to synthesize.
+
+Options:
+
+- --migration-id (Required)
```
godwingrs22, line 239: Just a suggestion. How about renaming as `--target`?

Eg:
`cdk construct-upgrade --target dynamodb`

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,413 @@
...
+    specify the migration target. Valid values are [vpc, dynamodb]
+- --dry-run (Optional)
+    This flag is default to false. When the flag is set to true, the CLI acts as dry-run
+    and will only trigger validation. The output of the validations will be presented to users
+    in the log and be written into `upgrade.context.json` file.
+- --strict (Optional)  
+    Treats validation warnings as errors.  
+- --context-file (Optional)
+    specify custom context file instead of default `upgrade.context.json`
```
godwingrs22, line 248: I assume it is a validation file that stores validation result. In that case can we rename this flag to be consistent with the file actual purpose (`--validation-file`)

GavinZZ:
Thanks @godwingrs22 for the review. I've addressed them and updated the RFC accordingly!

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
+# CDK Construct Upgrade Support
+
+- **Original Author(s):**: @GavinZZ
+- **Tracking Issue**: #715
+- **API Bar Raiser**: @iliapolo
+
+An improvement to the CDK CLI and CDK construct library, to support developers in
```
iliapolo, line 7: "An improvement" is a generic term - be more specific.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+    in the log and be written into `construct-upgrade.context.json` file.
+- --strict (Optional)  
+    Treats validation warnings as errors.  
+- --validation-file (Optional)
+    specify custom context file instead of default `construct-upgrade.context.json`
+- --accept-violations (Optional)
+    Specify this flag to override and ignore any failures or errors in the validations. Use
+    at users’ own risk.
+- --abort (Optional)
+    Abort the migration and discard the construct-upgrade.context.json context file.
```
iliapolo, line 263: What does it mean to discard? delete the file?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+code migration process themselves.
+
+##### Phase 1 (dry-run)
+
+This phase will ensure through the automated migration tool that user’s existing CDK stack and
+environment area ready for migration, flagging warnings and blockers like CloudFormation drift.
+Additionally, validate that the migrated code is safe to deploy without causing downtime or replacement
+through CloudFromation template comparison and change set analysis.
+
+High-level tasks:
```
iliapolo, line 393: Seems like these validations are the main added value of this feature. Can you please elaborate on them a bit? give examples for each? 

Also - have you considered putting them inside `cdk diff`? at the end of the day, they are all based on template comparison, which is what `cdk diff` does already.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+
+1. Drift Detection  
+2. Environment Validation
+3. Resource Mapping Validation
+4. Dependency Graph Validation
+5. CloudFormation Change Set Analysis
+6. [In-Place Migration Only] CloudFormation stack refactoring Validation for Migration Path
+7. [Retain-Remove-Import Migration Only] Retain Policy Validation
+8. Any other module specific validations
```
iliapolo, line 402: Like what?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+`cdk construct-upgrade` CLI command and it will automatically deduce from `construct-upgrade.context.json` file
+if validations are complete to execution and deployment step.
+2. If `--dry-run` option is not specified, the CLI command will perform validation and automatically proceed
+with execution and deployment if validations are successful.
+
+Alternatively, users can use `cdk deploy` with construct upgrade option to validate their construct upgrade
+changes and deploy the stacks `cdk deploy --validate-construct-upgrade`.
+
+In the initial release of CDK construct upgrade tool, we target to support VPC related constructs migration
+from `aws-cdk-lib/aws-ec2` to `@aws-cdk/aws-ec2-alpha` as well as `Table` to `TableV2` construct
```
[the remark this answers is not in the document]
GavinZZ: The CLI will be generic but the guide is limited to these two constructs. let me update this.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+The next step is to run the CDK CLI command `cdk construct-upgrade`. This command
+will trigger the validation process to make sure the changes are valid and safe to deploy.
+
+```sh
+cdk construct-upgrade --unstable=migrate --target dynamodb --dry-run
+```
+
+CDK will do a dry-run validation on current stack status, compare the current and new CFN
+template files, analyze dependency changes, analyze CloudFormation change set, validate retain
+policies (if applies), validate stack refactor JSON file, and etc.
```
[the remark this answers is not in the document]
GavinZZ: yes

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+Users can choose to run `cdk construct-upgrade` CLI command with:
+
+1. `--dry-run` option to perform validation only and generate the the validation JSON file. Users can re-run
+`cdk construct-upgrade` CLI command and it will automatically deduce from `construct-upgrade.context.json` file
+if validations are complete to execution and deployment step.
+2. If `--dry-run` option is not specified, the CLI command will perform validation and automatically proceed
+with execution and deployment if validations are successful.
+
+Alternatively, users can use `cdk deploy` with construct upgrade option to validate their construct upgrade
+changes and deploy the stacks `cdk deploy --validate-construct-upgrade`.
```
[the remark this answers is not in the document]
GavinZZ: updated

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+Enterprise developers often lack production AWS credentials, making pre-deployment
+validations (e.g., drift detection, change set analysis) impossible in production environments.
+
+At development time, here is how it works for common enterprise scenarios. Developers have upgrade
+their CDK stack code to use V2 construct. They can use `dry-run` feature which will allow users to
+run the validations on the test/dev account without deploying.
+
+There are at least two possible paths forward, depending on the circumstances:
+
+1. Developers can send the output validation file to an operations team, who will review it. If
```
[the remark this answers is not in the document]
GavinZZ: They can choose to omit it but the validation file were jsut a sample output from a non-production environment. They can first review it to then decide whether they want to go through the effort to validate on production environments.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+
+Arguments:
+
+- CDK stack ID (Optional)
+    The construct ID of the CDK stack from your app to synthesize.
+
+Options:
+
+- --target (Required)
+    specify the migration target. Valid values are [vpc, dynamodb]
```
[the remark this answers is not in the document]
GavinZZ: Good idea!

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+mistakes can happen.
+
+If you're migrating using In-Place migration approach, then
+CloudFormation stack refactoring feature will perform additional validations and
+rollback automatically. If you accidentally migrate a stack, you can run the CLI with
+inverse mapping file `refactor.json` to bring the resources back to their original state.
+
+If you're migrating using Retain-Remove-Import migration approach, then CloudFormation
+deployment will rollback the stack for you automatically if things go wrong. If you
+want to switch back to V1 construct from V2 cosntruct, you can do this by retaining
```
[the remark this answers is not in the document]
GavinZZ: I can't see the suggestion. let's discuss this tmr.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+```
+
+CDK will do a dry-run validation on current stack status, compare the current and new CFN
+template files, analyze dependency changes, analyze CloudFormation change set, validate retain
+policies (if applies), validate stack refactor JSON file, and etc.
+
+The output of the validation will be shown in the CLI output as well as in a context file,
+`construct-upgrade.context.json` file. A sample output is shown below:
+
+```json
```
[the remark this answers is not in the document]
GavinZZ: Let's discuss this tomorrow to see if we still need this.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+
+- **Original Author(s):**: @GavinZZ
+- **Tracking Issue**: #715
+- **API Bar Raiser**: @iliapolo
+
+An improvement to the CDK CLI and CDK construct library, to support developers in
+migrating their CDK applications from supported older constructs or modules to the
+latest constructs/modules.
+
+## Working Backwards
```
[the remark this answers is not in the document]
rix0rrr: To add on: working backwards implies an "artifact from the future".

This is either the README, or the blog post, that tells users "if you have problem X, you can now do Y to solve it. See how easy that is?"

Show the code that a user would write, the commands they would run, and the feedback they would see.

In this doc, there are probably 2 angles to show light on:

- The end user experience
- The construct author experience

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+
+* The `@aws-cdk/aws-ec2-alpha` alpha module introduces the modernized `VpcV2` construct,
+designed to succeed the original Vpc in `aws-cdk-lib/aws-ec2`
+* The `TableV2` construct in `aws-cdk-lib/aws-dynamodb` replaces legacy Table implementations
+while retaining API compatibility
+
+These updates unlock enhanced capabilities. Migration between versions, however, often requires
+code changes that could trigger CloudFormation resource replacements (e.g., NAT gateways being
+recreated, DynamoDB global tables being recreated). For stateful production resources, this can
+lead to availability risks and operational complexity.
```
rix0rrr, line 28: These are weasel words. Be specific about the problems.

"Recreation of resources leads to data loss for stateful resources, and can lead to loss of availability for others."

I'm not sure what the "operational complexity" is referring to. What is complex about data loss?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+
+Note that each migration guide follows a standardized structure that includes dependency changes, behavioral
+analysis, risk assessments, and rollback instructions. This ensures consistency, clarity, and safety across
+all migrations. See [Appendix A](#appendix-a-migration-guide) for a detailed breakdown of the proposed migration
+guide format.
+
+With the initial release of the CDK construct upgrade tool, developers use the migration guide to
+update their CDK code, replacing legacy constructs (e.g., Table) with modern equivalents (e.g., TableV2).
+Developers have the flexibility to implement changes either manually or with the assistance of AI-powered
+code transformation tools (e.g., LLMs).
```
rix0rrr, line 55: Would the migration guide contain the prompts for the LLMs?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+
+After updating the code to use the new construct, the CDK CLI command `cdk construct-upgrade` validates the
+existing CDK stack and environment, ensuring they are ready for migration. The tool flags any warnings or
+blockers, such as CloudFormation drift, and cross-compares the old and new CloudFormation templates.
```
rix0rrr, line 59: Can you be specific about the types of changes we would flag here?

Does this basically amount to `cdk diff` shows no diff (or at least: no replacement)... perhaps that diff requirement is softened a little. Perhaps we accept some changes that we know are safe.

If so, wouldn't these kinds of safety checks make sense as part of `cdk diff` proper as well? They don't seem to need to be locked behind a new subcommand.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+- If the `--dry-run` option is not specified, the CLI command will perform validation and automatically
+proceed with execution and deployment if the validations are successful.
+
+Alternatively, users can use `cdk deploy --construct-upgrade` with construct upgrade option to validate
+their construct upgrade changes before deployment starts.
+
+In the initial release of the CDK construct upgrade tool, the CLI supports migration of VPC-related constructs
+from `aws-cdk-lib/aws-ec2` to `@aws-cdk/aws-ec2-alpha` and `Table` to `TableV2` in `aws-cdk-lib/aws-dynamodb`.
+While the tool is designed to handle a wide range of construct migrations, the migration guides for these two
+upgrades will be the first available, with additional guides planned for future releases.
```
rix0rrr, line 76: Whatever mechanism we employ for this should be open and pluggable, right? Worth calling out whether it is or isn't.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+CloudFormation stack.
+
+2. Retain-Remove-Import Migration - This approach is used when the underlying resource has significant changes,
+such as transitioning from a custom resource to a native L1 resource, as seen in the migration from `Table` to
+`TableV2`. For this type of migration, users need to ensure that `RemovalPolicy.RETAIN` is set on the V1 construct
+in the stack to prevent the deletion of the resource during the migration. The `RemovalPolicy.RETAIN` ensures that
+the existing resource is retained by CloudFormation, avoiding accidental deletion when the CDK stack is updated.
+
+If the retain policy is not already set, users must first deploy the retain policy change to ensure that
+CloudFormation will retain the resource during the migration process. It's a good practice to keep the stateful
```
rix0rrr, line 107: Totally something we could automate.

For example, it is easily possible to

- Download the current template
- Patch the current template, changing/adding `DeletionPolicy`s wherever necessary
- Deploy that template

At no point did the user have to change their code to change those `DeletionPolicy`s; we can do that for them.


-------

I think (2) is a different, less-safe form of (1) Stack Refactoring that we can offer, but it's an operation that slots into the same location in the workflow I think.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+resources retained by default. Once this step is completed, users should follow the migration guide and modify
+their CDK stack code to remove the old construct and replace it with the new one (e.g., replacing the legacy
+`Table` with the new `TableV2` construct).
+
+The next step is to run `cdk deploy --import-existing-resources`. The `--import-existing-resources` flag is used
```
rix0rrr, line 112: Tell me we automate this as well?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,423 @@
...
+```
+
+CDK provides a DynamoDB table migration guide that guides users on the code migration
+processes. Assuming users follow the migration guide correctly and update their CDK code
+to use `TableV2` construct.
+
+In case of DynamoDB Table migration, the new CDK code would become:
+
+```ts
+const table = new TableV2(this, 'MyGlobalTable', {
```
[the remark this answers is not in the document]
rix0rrr: Alternatively: if you keep the construct ID the same, and CDK notices the resource type change `AWS::DynamoDB::Table` -> `AWS::DynamoDB::GlobalTable`, we can look up that change in our collection of migration guide strategies, and automatically execute a (2) Retain-Delete-Import workflow.

text/0715-construct-upgrade-support.md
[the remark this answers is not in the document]
rix0rrr: What is this file used for?

Wouldn't it make more sense to have an "execution plan" in this file, if you want a file?

A list of preconditions to check and a list of actions to execute?

text/0715-construct-upgrade-support.md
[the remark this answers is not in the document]
rix0rrr: But the validation only contains `{ "result": "PASS" }`. I can make a JSON file myself that says that.

Does that prove anything? And is it necessary to make any kind of operation work?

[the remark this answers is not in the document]
rix0rrr: Anyway, strategy (1) is: "some other privileged team runs the migration", right? 

It's worthwhile thinking about what the full workflow looks like and spelling that out:

- The dev team has a new version of the source code
- They can't commit it because that would start CI/CD and that would be undesirable because a manual action needs to be performed in every prod region first
- Two options
  1. They zip up the source code and send it to the ops team. The ops team runs the migration in every region. Afterwards the dev team commits the latest  version of their source code.
  2. They disable the CI pipeline, commit the source code. The ops team runs the migration in every region. Afterwards the dev team unblocks the CI pipeline.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,514 @@
...
+run the validations on the test/dev account without deploying.
+
+There are at least two possible paths forward, depending on the circumstances:
+
+1. Developers can send the output validation file to an operations team, who will review it. If
+operations team approve the validation file, developers or operation team engineers can proceed
+with `cdk construct-upgrade --dry-run` CLI command to run the validations on every protected
+environment in advance (i.e., before your changes get deployed to those environments) and
+follow by `cdk construct-upgrade` to deploy the changes to every protected enviornments.
+2. configure your pipeline to run the new CDK CLI command `cdk construct-upgrade` and then commit
```
rix0rrr, line 279: Option (2) works for Jenkins pipelines etc. It wouldn't work for CDK Pipelines-pipelines (and internal Amazon Pipelines) which do a mass synth at the start, and then use pure CloudFormation actions at every stage to do deployments. Deployments do not involve the CLI.

Stack Refactoring has the same problem, and it's also not solved there.

I think we can punt this and say this is a difficult problem that we don't have a good solution for yet. I don't think we're going to solve this here.

------

Even if we did have the CLI involved, at the top you said the command would have to be:

```
cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2
```

What that means is effectively the workflow would have to be:

1. When you need to do an upgrade, block the CI system
2. Edit the CI system to add the *specific* `cdk construct-upgrade` command to the CI system
3. Commit the change
4. Unblock the CI system
5. Wait for it to  run to completion, make sure not to commit anything new while it is running
6. At the end, edit the CI system again to take away the specific upgrade command

For every change.

That seems quite fiddly, and not a great experience at all.

And would the app builders even have permission to edit the CI system commands? It seems that if they would not have access to production systems, chances are they also wouldn't have access to add arbitrary commands to the CI system, since doing that would open up exactly that security hole that that rule was intended to close.

text/0715-construct-upgrade-support.md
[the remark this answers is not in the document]
GavinZZ: Make sense, thanks for the clarification. Let me update this

text/0715-construct-upgrade-support.md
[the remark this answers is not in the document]
GavinZZ: Good point, we can update the CDK side to recognize this change!

text/0715-construct-upgrade-support.md
GavinZZ: Yes

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+```
+
+> Note that directly deploying the above CDK would cause CloudFormation
+> to delete your existing table and create a new one, due to differences in logical IDs
+> and resource types.
+
+The next step is to run the CDK CLI command `cdk construct-upgrade`. This command
+will trigger the validation process to make sure the changes are valid and safe to deploy.
+This migration will use the `Retain-Remove-Import` migration strategy because the underlying
+resource type has changed from using a custom resource to native CloudFormation resource.
```
iliapolo, line 342: I think you mean from using `AWS::DynamoDB::Table` to using `AWS::DynamoDB::GlobalTable`?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+causing downtime or replacement through CloudFromation template comparison and change set analysis. Here is a
+complete list of validations that CDK performs:
+
+#### Drift Detection
+
+One of the validation workflow is to trigger CloudFormation drift detection for the CDK stacks. This feature
+is generic and can be applied to all CDK users. It will be an added functionality in the `cdk diff` command
+through the option `cdk-diff --detect-drift`.
+
+The new `cdk construct-upgrade` CLI command will implicitly call the `cdk diff --detect-drift` command as part of
```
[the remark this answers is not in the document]
GavinZZ: You're right that occasional, non-significant drift can exist, and we’ve considered that in the design. Currently, CloudFormation drift detection operates at the stack level—there’s no direct API support for performing drift detection selectively at the resource level without first initiating it at the stack level.

That’s why the --detect-drift flag is designed to be generic within cdk diff. It surfaces stack-level drift detection results in the same spirit as how cdk diff shows change sets—informative, not prescriptive.

For more granular, construct-specific validations—including deeper analysis of potential drift—we intend to surface those through a separate construct upgrade command. This separation ensures cdk diff remains lightweight and non-restrictive while enabling richer validation workflows when needed.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+Validations will ensure that user’s existing CDK stack and environment area ready for migration, flagging
+blockers like CloudFormation drift. Additionally, validate that the migrated code is safe to deploy without
+causing downtime or replacement through CloudFromation template comparison and change set analysis. Here is a
+complete list of validations that CDK performs:
+
+#### Drift Detection
+
+One of the validation workflow is to trigger CloudFormation drift detection for the CDK stacks. This feature
+is generic and can be applied to all CDK users. It will be an added functionality in the `cdk diff` command
+through the option `cdk-diff --detect-drift`.
```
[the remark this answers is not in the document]
GavinZZ: I updated the description. I think this will not and should not fail as `cdk diff` is meant to provide information about changes and is designed to be status-agnostic.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+
+#### CloudFormation Change Set Analysis
+
+Another validation is to validate CloudFormation change set diff. The `cdk diff` command already trigger
+CloudFormation change set generation, providing high-level details such as "add", "delete", "update", etc.
+However, this validation will enhance the cdk diff functionality by adding the `cdk diff --full` option,
+which shows the full change set diff, including detailed information on each resource change.
+
+The new `cdk construct-upgrade CLI` command will implicitly call `cdk diff --full` to retrieve the comprehensive
+change set diff. This allows for more granular analysis of the changes, ensuring that unintended modifications
```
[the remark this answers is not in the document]
GavinZZ: I updated this section entirely. Please review!

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+are detected, and that actions like "Modify", "Add" and "Remove" are properly validated for correctness. The tool
+will analyze the changes to ensure that each "Remove" action has a corresponding "Add" of similar types, and
```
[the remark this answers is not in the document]
GavinZZ: Totally agree, I updated the entire change set section to make it more clear.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+
+This validation is specific to the context of construct upgrade framework and will be implemented as part of
+`cdk construct-upgrade` CLI command.
+
+#### Custom validations
+
+Custom validations ensure that the construct upgrade process adheres to best practices with custom
+configurations. For example, when migrating `Table` to `TableV2` for DynamoDB, the CDK team provides
+validations to ensure that custom resource deletions do not affect replica tables, that `TableName`
+properties are correctly set, and that `DeletionProtection` is enabled for safety.
```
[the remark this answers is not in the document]
GavinZZ: I updated this section. I think you raised a good point that this should ideally be enabled by default on construct side itself. However, in my mind, the best and most important selling point for this tool is validation and safety so we should still include the check in the V2 construct.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,772 @@
...
+command. This command triggers the stack refactoring operation, which helps to retain the existing resources
+and applies the necessary updates to the infrastructure.
+
+2. Retain-Remove-Import Migration - This approach is used when the underlying resource has significant changes,
+such as transitioning from a custom resource to a native L1 resource, as seen in the migration from `Table` to
+`TableV2`. For this type of migration, users need to ensure that `RemovalPolicy.RETAIN` is set on the V1 construct
+in the stack to prevent the deletion of the resource during the migration. The `RemovalPolicy.RETAIN` ensures that
+the existing resource is retained by CloudFormation, avoiding accidental deletion when the CDK stack is updated.
+
+If the retain policy is not already set, users must first deploy the retain policy change to ensure that
```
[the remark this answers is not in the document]
GavinZZ: Yes I spoke with Rico offline on this point and we have different opinions on this approach. For summary, I am a bit hesitant to implicitly deploy the user stack by setting the retain policy and deploy the change without user consent. 

My concern with automatically deploying the RETAIN change is:
1. Resources like Table should by default have RETAIN and this really only applies when users explicitly specify DELETE retain policy and I wonder why they would do this for production environment that needs migration...
2. Implicit updating their template seems something sketchy to me and needs security review
3. Even if we will change it back in the end, this makes the deployment non-atomic and multi-steps. I don't know if the complexity covers the benefits

The following is Rico's feedback on my concerns:
> Resources should have RETAIN already.

Well that's easy then isn't it, then both the automatic and manual step are a no-op :smile: .
> Implicit updating their template seems something sketchy

You can find this sketchy, but we already do it today and I'm pretty sure we're also planning to do it for the refactoring feature :smile: (though I need to double-check that). For a trivial transform, it seems preferable to the alternative of making the user do the changes, especially if the changes are ephemeral.
> this makes the deployment non-atomic and multi-steps.

And every deployment is always multi-step, only made "atomic" by the fact that someone keeps a record of the steps and knows how to do a rollforward or rollback if necessary. Happily for us, oftentimes that is CloudFormation so we don't need to put in the effort. But someone has to, and it could also be us if we choose it to be.

Rico: "In any case, I don't think we have to necessarily agree on this to move forward. I just want to have stated my case that UX and "do what I want" is of paramount importance to".

I think the workaround I mentioned is to support the following mentioned in the RFC
```
Prompt the user with a message like:

```sh
Detected AWS::DynamoDB::Table with logical ID 'MyTable' does not have RETAIN set as the deletion policy.
CDK will deploy your stack with the deletion policy set to RETAIN on 'MyTable'.

Would you like to proceed (y/n)?
```

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1035 @@
...
+
+```sh
+# Dry run the validations without deploy the construct upgrade changes
+cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2 --dry-run
+
+# This will validate and deploy the construct upgrade changes
+cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2
+
+# Alternatively, users can use a generic `cdk deploy` with construct upgrade option
+cdk deploy --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2
```
godwingrs22, line 85: Seems `--unstable=construct-upgrade` being proposed in both `cdk deploy` and `cdk construct-upgrade` commands. While this makes sense for `cdk deploy` since it's an existing stable command and we're adding new experimental functionality, it seems redundant for the `cdk construct-upgrade` command which is itself new and experimental. Do you think it is good to remove the unstable option for the construct-upgrade command? Also the term `construct-upgrade` appears repetitive when used both as a command name and as an option value (`cdk construct-upgrade` and `--unstable=construct-upgrade`). This creates a verbose and potentially confusing user experience.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+    },
+    ......
+]
+```
+
+The validation begins by verifying that every source LogicalId in `refactor.json` corresponds to a resource
+in the deployed V1 template, while every target LogicalId maps to a resource in the synthesized V2 template.
+
+Orphaned entries — such as unmapped logical IDs or resources missing from either template — will block the upgrade.
+Resource types must also align: for example, a resource in V1 template of type `AWS::EC2::VPC` can only map to
```
iliapolo, line 305: How do we determine which types are compatible to which types? is this construct specific logic?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+prompt the user:
+
+```sh
+Detected path metadata in the synthesized CloudFormation template, which is not compatible with CloudFormation
+stack refactoring. CDK will deploy your stack with path metadata disabled using: cdk deploy --no-path-metadata
+
+Would you like to proceed? (y/n)
+```
+
+If the user confirms by selecting "y", the tool will proceed to run `cdk deploy` with `--no-path-metadata` flag
```
iliapolo, line 363: is `--no-path-metadata` an existing capability we have? will it just remove path metadata forever until a user deploys again without this flag?

text/0715-construct-upgrade-support.md
GavinZZ: I think this is typically how CLI handles the unstable CLI and options. I've asked the same question for the refactor RFC and Otavio mentioned that this is required.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+Sample output of `cdk diff --detect-drift` command is as follows:
+
+```s
+Drifts
+[~] AWS::Lambda::Function MyFunction drifted
+[~] AWS::DynamoDB::Table MyTable drifted
+
+Resources
+[-] AWS::DynamoDB::Table MyTable orphan
+[+] AWS::DynamoDB::GlobalTable MyTable import
```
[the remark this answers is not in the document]
GavinZZ: This is intentional as `cdk diff --detect-drift` should include the current `cdk diff` output along with the additional drift output.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+            "StackName": "<stack-name>",
+            "LogicalResourceId": "vpcamazonProvided373FDA5C"
+        }
+    },
+    ......
+]
+```
+
+The validation begins by verifying that every source LogicalId in `refactor.json` corresponds to a resource
+in the deployed V1 template, while every target LogicalId maps to a resource in the synthesized V2 template.
```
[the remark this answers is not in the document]
GavinZZ: It does not. However, the error msg is not clear and I think we can do a preemptive check here to early exit and provide better error msgs.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+Another validation is to validate CloudFormation change set diff. The `cdk diff` command implicitly calls
+CloudFormation change set creation, providing high-level details such as "add", "delete", "modify", "import",
+and etc. like the following:
+
+```s
+[-] AWS::DynamoDB::Table MyTable orphan
+[+] AWS::DynamoDB::GlobalTable MyTable
+```
+
+To build on this, we are introducing a new `--json` flag to the `cdk diff` command like `cdk diff --json` which
```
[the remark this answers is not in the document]
GavinZZ: Good point, I was testing with CDK CLI and that's why I added this. I agree that this is not needed.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1131 @@
...
+
+> Note that auxiliary resources—such as IAM roles, policies, or custom resources may be added or removed as part of
+> construct upgrade for this strategy, and will omitted during the validations.
+
+For `In-Place Migration` strategy, `construct-upgrade` CLI command will retrieve the JSON format change set, ensure
+no unintended modifications are present and validate that each `Remove` action on main resources has a corresponding
+`Add` of the same resource type. Additionally validate that the `beforeContext` and `afterContext` values for each
+`Remove` and `Add` pair are identical, as CloudFormation stack refactors prohibit configuration changes.
+
+For the `Retain-Remove-Import Migration` strategy, we will introduce an additional option `--import` in the
```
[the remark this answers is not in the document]
GavinZZ: Updated to be clearer

text/0715-construct-upgrade-support.md
GavinZZ: I think this is a general context info that we have (not specific to this validation step). For supported construct upgrade, i.e. VPC or Table, we should have a generic context information on the relevant main resources and we have to map main resource type to the same main resource type in two templates.

text/0715-construct-upgrade-support.md
GavinZZ: Yes it's an existing behaviour. I think this is the same type of behaviour as if we deploy the changes by setting RETAIN policy.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1274 @@
...
+            "StackName": "<stack-name>",
+            "LogicalResourceId": "vpcamazonProvided373FDA5C"
+        }
+    },
+    ......
+]
+```
+
+The validation begins by verifying that every source LogicalId in `refactor.json` corresponds to a resource
+in the deployed V1 template, while every target LogicalId maps to a resource in the synthesized V2 template.
```
iliapolo, line 310: What about mappings that the user may have forgotten? For example, the app has 2 subnets, but the `refactor.json` file only mentions 1 subnet. Will this be blocked by CFN?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1274 @@
...
+> This step is still needed to validate the inferred `refactor.json` file by CDK refactor. Depending on
+> the CDK refactor implementation, we may omit this validation if it's already performed in CDK refactor;
+> or we can move the validation logics to be a generic validation on CDK refactor side.
+
+#### [In-Place Migration Only] Path Metadata Validation
+
+In In-Place migrations, we leverage the CloudFormation stack refactoring feature, which imposes a limitation:
+no changes to resource configurations are allowed, including updates to the Metadata section of any resource.
+
+A common issue arises when users migrate a construct across modules (i.e. `@aws-cdk.aws-ec2-alpha.VpcV2` vs
```
iliapolo, line 333: Like we discussed, this only depends on the logical ID changing, not the module.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1274 @@
...
+
+The `cdk diff` command will be extended to support a new `--custom-rule` option, which accepts a
+JavaScript file implementing a predefined `Rules` interface. These custom rule files allow both
+the CDK team and end-users to define and enforce additional rules during the diff process, enabling safer
+and more opinionated construct upgrade paths. When the `--custom-rule` option is specified, the CLI
+will `require()` the input file, instantiate the `Rules` object, and invoke the custom rules specified.
+An example is provided below for the custom rule set up and usage.
+
+```js
+export interface Rules {
```
iliapolo, line 421:
```suggestion
export interface DiffAnalysis {
```

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1274 @@
...
+deploy the change.
+
+#### Custom validations
+
+Custom validations ensure that the construct upgrade process adheres to best practices with custom
+configurations. For example, when migrating `Table` to `TableV2` for DynamoDB, the CDK team provides
+validations to ensure that custom resource deletions do not affect replica tables, that `TableName` and
+`SkipReplicaDeletion` properties are correctly set, and that `DeletionProtection` is enabled for safety.
+
+The `cdk diff` command will be extended to support a new `--custom-rule` option, which accepts a
```
iliapolo, line 413:
```suggestion
The `cdk diff` command will be extended to support a new `--custom-analysis` option, which accepts a
```

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1274 @@
...
+  init?: (host: IPluginHost) => void;
+}
+```
+
+An example of users' provided custom validation file is as follows:
+
+```ts
+import { IRules, Rules } from '@aws-cdk/cli-custom-rules';
+
+class DemoValidation {
```
iliapolo, line 442: Shouldn't this implement some interface as well? how would the CLI know to call the `execute` method?

PR #724, opened 2025-05-01
RFC 0715: CDK Construct Upgrade Guide

GavinZZ:
This is a request for comments about CDK Construct Upgrade Framework See https://github.com/aws/aws-cdk-rfcs/issues/715 for additional details.

---

### DIFF
--- a/text/0715-construct-upgrade-support.md
+++ b/text/0715-construct-upgrade-support.md
@@ -1,0 +1,1251 @@
+# CDK Construct Upgrade Support
+
+- **Original Author(s):**: @GavinZZ
+- **Tracking Issue**: #715
+- **API Bar Raiser**: @iliapolo
+
+A construct upgrade framework in the CDK CLI that helps developers upgrade from deployed and
+older construct libraries to modern, fully supported alternatives that incorporate the latest
+features and best practices.
+
+## Motivation
+
+As the AWS Cloud Development Kit (CDK) evolves, we continuously deliver new features,
+performance optimizations, and reliability improvements. To maintain backward compatibility
+while enabling forward progress, the CDK team uses structured versioning strategies for critical updates.
+
+A common pattern involves releasing versioned constructs or modules (e.g., V2 variants)
+that coexist with their predecessors. For example:
+
+* The `@aws-cdk/aws-ec2-alpha` alpha module introduces the modernized `VpcV2`, `SubnetV2`, and other
+new construct, designed to succeed the original Vpc related constructs in `aws-cdk-lib/aws-ec2`.
+* The `TableV2` construct in `aws-cdk-lib/aws-dynamodb` replaces legacy Table implementations
+while retaining API compatibility.
+
+These updates unlock enhanced capabilities. Construct upgrade between variants, however, often requires
+code changes that could trigger CloudFormation resource replacements (e.g., NAT gateways being
+recreated, DynamoDB global tables being recreated). Recreation of resources leads to data loss for
+stateful resources, and can lead to loss of availability for others.
+
+To address this, CDK team develops a construct upgrade framework designed to simplify adoption while
+minimizing risks. This approach combines guided documentation, automated validations, and deployment
+safeguards to ensure a controlled transition.
+
+## Proposed Solution
+
+In a high-level overview, for each supported migration, the construct upgrade framework provides:
+
+- A comprehensive, step-by-step migration guide enabling users to transition from legacy constructs/modules
+to their modern equivalents. Migration guide follows a standardized structure that includes dependency changes,
+behavioral analysis, risk assessments, and rollback instructions. This ensures consistency, clarity, and safety
+across construct migration. See [Appendix A](#appendix-a-migration-guide) for a detailed breakdown of the proposed
+migration guide format.
+
+- A new CDK CLI command `cdk construct-upgrade --unstable=construct-upgrade` to validate code changes and
+execute deployments safely. For more details on the CLI usage, please visit [CLI Settings](#settings)
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade [stacks] \
+  --target [id] \
+  [OPTIONS]
+```
+
+With the initial release of the CDK construct upgrade tool, developers would use the migration guide to
+update their CDK code, replacing legacy constructs (e.g., Table) with modern equivalents (e.g., TableV2).
+
+After updating the code to use the new construct, users would use the CDK CLI command `cdk construct-upgrade --unstable=construct-upgrade`
+to validates the existing CDK stack and environment, ensuring they are ready for migration. The tool flags any
+blockers, such as CloudFormation drift. Additionally, it analyzes the CloudFormation change set and validates
+any library-specific configurations. See [Construct Upgrade Validations](#construct-upgrade-validations) for a detailed breakdown of the validations.
+
+In the initial release of the CDK construct upgrade tool, VPC-related constructs upgrade from `aws-cdk-lib/aws-ec2`
+to `@aws-cdk/aws-ec2-alpha` and `Table` to `TableV2` in `aws-cdk-lib/aws-dynamodb` come with enhanced
+construct-specific validations and comprehensive migration guides provided by the CDK team to ensure
+safe deployment. These two construct upgrades are so-called `fully supported` construct upgrade.
+
+We aim to gradually increase the number of `fully supported` constructs. The tool is also designed to be open
+and pluggable, allowing users to support other construct migrations by incorporating their own custom
+validations. This flexibility enables users to extend the tool to cover additional constructs that may not
+yet be officially supported with custom validations, see [Custom Validations](#custom-validations-optional-future-enhancement).
+
+## Walkthrough of Construct Upgrade Example: DynamoDB Table Construct Upgrade
+
+Below is a common upgrade scenario to demonstrate how the CDK Construct Upgrade framework helps developers
+safely migrate their infrastructure. Consider a DynamoDB table deployed to CloudFormation using the legacy
+`Table` construct:
+
+```ts
+import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
+
+export class DemoStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    const table = new dynamodb.Table(this, 'MyTable', {
+      partitionKey: {
+        name: 'PK',
+        type: dynamodb.AttributeType.STRING,
+      },
+      replicaRemovalPolicy: RemovalPolicy.Retain,
+      replicationRegions: ['us-west-2'],
+    });
+  }
+}
+```
+
+CDK team provides a comprehensive migration guide that includes construct property mappings
+(e.g., replicationRegions → replicas), code snippets, common migration scenarios, and etc.
+To upgrade to TableV2 construct, first update your CDK code following the migration guide:
+
+```ts
+import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
+
+export class DemoStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    const table = new dynamodb.TableV2(this, 'MyTable', {
+      partitionKey: {
+        name: 'PK',
+        type: dynamodb.AttributeType.STRING,
+      },
+      tableName: 'DemoStack-MyTable794EDED1-11W4MR8VZ0UPE',
+      replicas: [
+        {
+          region: 'us-west-2',
+        }
+      ]
+    });
+  }
+}
+```
+
+> Note: the above CDK stack requires `tableName` property to be specified and this is a
+> known limitation to construct upgrade framework. For more detail, see
+> [Import Strategy Limitation](#limitation-retain-remove-import-migration-strategy).
+
+Now, validate the construct upgrade using the CDK CLI:
+
+> Note that the flag `--unstable=construct-upgrade` is necessary to indicate that
+> it's an experiemntal feature that users are using until feature stablized.
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2
+```
+
+CDK runs the set of validations required and produce the following report:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+✅ Change set analysis complete
+
+Do you wish to proceed with this upgrade? [y/N]
+```
+
+If you answer `y` for yes, the CLI will proceed to deploy and show the progress as the deployment
+is executed:
+
+```s
+Performing construct upgrade...
+
+2:03:17 PM | IMPORT_IN_PROGRESS   | AWS::DynamoDB::GlobalTable          | MyStack/MyGlobalTable35835197E/Resource
+2:03:17 PM | IMPORT_COMPLETE      | AWS::DynamoDB::GlobalTable          | MyStack/MyGlobalTable35835197E/Resource
+2:03:19 PM | DELETE_IN_PROGRESS   | Custom::DynamoDBReplica             | MyStack/MyTableReplicauswest285A33668/Resource
+2:03:19 PM | DELETE_IN_PROGRESS   | AWS::IAM::ManagedPolicy             | MyStack/MyTableManagedPolicyXXXXXXX/Resource
+2:03:19 PM | DELETE_IN_PROGRESS   | AWS::CloudFormation::NestedStack    | MyStack/ReplicaProviderNestedStack/Resource
+2:03:19 PM | DELETE_IN_PROGRESS   | AWS::DynamoDB::Table                | MyStack/MyTable794EDED1/Resource
+2:03:21 PM | DELETE_COMPLETE      | Custom::DynamoDBReplica             | MyStack/MyTableReplicauswest285A33668/Resource
+2:03:22 PM | DELETE_COMPLETE      | AWS::IAM::ManagedPolicy             | MyStack/MyTableManagedPolicyXXXXXXX/Resource
+2:03:22 PM | DELETE_COMPLETE      | AWS::CloudFormation::NestedStack    | MyStack/ReplicaProviderNestedStack/Resource
+2:03:19 PM | DELETE_COMPLETE      | AWS::DynamoDB::Table                | MyStack/MyTable794EDED1/Resource
+
+✅  Stack deployment complete
+```
+
+> For CI/CD use cases, you can streamline deployments by using the `--require-approval never`
+> option with the CDK CLI to bypass manual approval prompts.
+
+### Drift Detection Failure Example
+
+If a CloudFormation drift is detected, the CLI will show:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+❌ CloudFormation drift detected
+  │ Resource: AWS::DynamoDB::Table MyTable794EDED1
+  │ Actual changes found:
+  └   • BillingMode: "PAY_PER_REQUEST" (expected: "PROVISIONED")
+✅ Change set analysis complete
+
+Error: Construct upgrade validation failed
+Please resolve the drift before proceeding with the upgrade.
+For more information about handling drift, visit: 'https://docs.aws.amazon.com/cdk/v2/guide/drift'
+```
+
+### Change Set Validation Failure Example
+
+**Example 1:** If the change set analysis detects the new table is not being imported (by setting the `TableName` property
+on the construct), the CLI will show:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+❌ Change set analysis failed
+  │ Resources: MyGlobalTable794EDED1 (AWS::DynamoDB::GlobalTable)
+  │ Change Set:
+  └  • Action: 'Add' (expected: 'Import')
+
+Error: Construct upgrade validation failed
+```
+
+**Example 2:** If the change set analysis detects the existing table does not have policy action `Retain` (by setting
+the `RemovalPolicy` on existing table construct), the CLI will show:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+❌ Change set analysis failed
+  │ Resource: MyTable794EDED1 (AWS::DynamoDB::Table)
+  └   • RemovalPolicy: 'Delete' (expected: 'Retain')
+
+Error: Construct upgrade validation failed
+```
+
+**Example 3:** If the change set analysis detects the Replica custom resource does not have policy action `Retain`
+(by setting the `ReplicaRemovalPolicy` on existing table construct), the CLI will show:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+❌ Change set analysis failed
+  │ Resources: MyTableReplicauswest285A33668 (Custom::DynamoDBReplica)
+  └   • RemovalPolicy: 'Delete' (expected: 'Retain')
+
+Error: Construct upgrade validation failed
+```
+
+### Custom Validations (Optional Future Enhancement)
+
+While the currently supported constructs don't require custom validations, future upgrades
+(e.g., for AWS::EKS resources) may benefit from this feature. Additionally, this functionality would allow
+customers to incorporate enterprise-specific requirements into the validation process.
+
+For instance, users upgrading to TableV2 might want to ensure their existing table construct has
+`DeletionProtection` enabled. To accommodate such scenarios, the `cdk construct-upgrade` CLI command would
+support custom analysis files as input.  For more details about the setup and file format for custom
+validations, refer to the [Custom Validations](#custom-validations-optional-future-enhancement) section.
+
+Example usage:
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2 \ 
+  --custom-analysis table-custom-analysis.ts
+```
+
+If the custom validation detects that `DeletionProtection` is not enabled on existing table construct,
+the CLI would display:
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+✅ Change set analysis complete
+❌ Custom validation failed
+  │ Resource: MyTable794EDED1 (AWS::DynamoDB::Table)
+  └   • DeletionProtection: 'false' (expected: 'true')
+
+Error: Construct upgrade validation failed
+```
+
+This feature would enhance the flexibility of the construct upgrade process, allowing teams to enforce
+specific policies or best practices during migrations.
+
+> If you're interested in seeing a full example of VPC V2 example, please checkout
+> [Appendix B: VPC V2 Upgrade Example](#appendix-b-vpc-related-migration-example).
+
+### User Experience: Visual Workflow
+
+```
+┌───────────────────────────────────────────────────────────┐
+|               Manual Workflow from Users                  |
+|                                                           |
+|                ┌─────────────────────┐                    |
+|                │  Follow Migration   │                    |
+|                │        Guide        │                    |
+|                └─────────────────────┘                    |
+|                           │                               |
+|            ┌──────────────┴──────────────┐                |
+|            ▼                             ▼                |
+| ┌─────────────────────┐       ┌─────────────────────┐     |
+| │   Human refactor    │       │Other refactor method│     |
+| │                     │       │      like LLM       │     |
+| └─────────────────────┘       └─────────────────────┘     |
+|            │                             │                |
+|            └──────────────┬──────────────┘                |
+|                           |                               |
+└───────────────────────────|───────────────────────────────┘
+                            |
+┌───────────────────────────|───────────────────────────────┐
+|                Automated Workflow by CDK CLI              | 
+|                           |                               |
+|                           ▼                               |
+|                ┌─────────────────────┐                    |
+|                │ Validate CFN drift  │                    |
+|                └─────────────────────┘                    |
+|                           |                               |
+|                           ▼                               |
+|                ┌─────────────────────┐                    |
+|                │   CFN Change set    |                    |
+|                |       analysis      │                    |
+|                └─────────────────────┘                    |               
+|                           |                               |
+|                 (Optional)|──────────────┐                |
+|                           |              |                |
+|                           ▼              |                |
+|                ┌─────────────────────┐   |                |
+|                │ Custom Validations  |   |                |
+|                └─────────────────────┘   |                │
+|                           │              |                |
+|                           │──────────────┘                |
+|                           │                               |
+|           ┌───────────────┴───────────────┐               |
+|           ▼                               ▼               |
+|┌───────────────────────┐         ┌───────────────────────┐|   
+|│  Stack Refactoring    |         | Retain-Remove-Import  ||   
+||       Strategy        |         │        Strategy       ||   
+||           │           |         |            │          ||    
+||           ▼           |         |            ▼          ||    
+|| ┌───────────────────┐ |         |┌─────────────────────┐||                                 
+|| │ Execute Stack     | |         |│     Execute         |||  
+|| |    refactoring    | |         || Retain-Remove-Import||| 
+|| └───────────────────┘ |         |└─────────────────────┘|| 
+|└───────────────────────┘         └───────────────────────┘|
+|           │                               │               | 
+|           └───────────────────────────────┘               |
+|                          |                                |
+|                          ▼                                |
+|               ┌─────────────────────┐                     |
+|               │        Done         │                     |
+|               └─────────────────────┘                     |
+└───────────────────────────────────────────────────────────┘
+```
+
+## Settings
+
+You have a few settings available to control the behavior of the new CDK CLI command
+for construct upgrade feature.
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade [stacks] \
+  --target [id] \
+  [OPTIONS]
+```
+
+Arguments:
+
+- CDK stack ID (Optional)
+    The construct ID of the CDK stack from your app to synthesize.
+
+Options:
+
+- --target (Required):
+  The target construct and module you want to upgrade to. For example, `aws-cdk-lib.aws-dynamodb.TableV2` or `@aws-cdk/aws-ec2.VpcV2`
+  For fully supported construct upgrades, we can short hand it to `TableV2` or `VpcV2`.
+- --dry-run (Optional)
+  This flag is default to false. When the flag is set to true, the CLI acts as dry-run and will only trigger validation.
+- --require-approval (Optional)
+  Specify what security-sensitive changes require manual approval.
+  - any-change – Manual approval required for any change to the stack.
+  - broadening – Manual approval required if changes involve a broadening of permissions or security group rules.
+  - never – Approval is not required.
+  Valid values: [any-change, broadening, never]
+  Default value: broadening
+- --skip-validation (Optional)
+  Default to false. Use as users' own risk to by pass the validations step.
+- --custom-analysis (Optional)
+  When user provides a custom JavaScript file for validation, more will be explained in [Custom Validations](#custom-validations-optional-future-enhancement)
+- Other global options like --quiet, --help, etc
+
+## Construct Upgrade Strategies
+
+The CDK construct upgrade framework employs two different strategies for migrating resources, automatically
+selecting the appropriate approach based on the `--target` construct. CDK CLI stores and uses a context variable
+to determine the mapping between upgrade targets and their corresponding strategies.
+
+### In-place Migration
+
+This strategy is suitable for constructs where changes do not involve custom resource
+removal or introduction of new underlying resource types. It leverages the CloudFormation stack refactoring
+feature to maintain resource continuity.
+
+Currently, users must specify a `refactor.json` file that maps logical ID changes from the old construct to the new one.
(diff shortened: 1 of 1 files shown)


GavinZZ:
Linking to the old RFC PR: https://github.com/aws/aws-cdk-rfcs/pull/717.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1121 @@
...
+A common pattern involves releasing versioned constructs or modules (e.g., V2 variants)
+that coexist with their predecessors. For example:
+
+* The `@aws-cdk/aws-ec2-alpha` alpha module introduces the modernized `VpcV2`, `SubnetV2`, and other
+new construct, designed to succeed the original Vpc related constructs in `aws-cdk-lib/aws-ec2`.
+* The `TableV2` construct in `aws-cdk-lib/aws-dynamodb` replaces legacy Table implementations
+while retaining API compatibility.
+
+These updates unlock enhanced capabilities. Construct upgrade between variants, however, often requires
+code changes that could trigger CloudFormation resource replacements (e.g., NAT gateways being
```
bdoyle0182, line 26: Internally with our own wrapper CDK we are discussing the idea of a semantic versioning schema of four entries X.X.X.X where you have major infrastructure deployment breaking change, major code breaking change, minor, patch. This allows us to decouple refactors from live service breaking change. Especially in a world with code assistants making breaking library code changes far less disruptive if non-breaking to the live resource. Just wanted to throw this out there as an idea to discuss for the aws-cdk team

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1124 @@
...
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+✅ Change set analysis complete
+❌ Retain policies validation failed
+  │ Resource: MyTable794EDED1 (AWS::DynamoDB::Table)
+  └   • RemovalPolicy: 'Delete' (expected: 'Retain')
```
iliapolo, line 228: If the removal policy is indeed `Delete`, won't the changeset already tell us that by marking it as a removal? or will the changeset always mark it as removal, not taking into account the policy? 

I'm just wondering if we really need this explicit validation, or can we extract the same information from the changeset.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1124 @@
...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+✅ Change set analysis complete
+✅ Retain policies configured
+❌ Construct-specific validations failed
+  │ Resources: MyGlobalTable794EDED1 (AWS::DynamoDB::GlobalTable)  │ 
+  │ Property 'TableName' is required for import but was not specified
```
iliapolo, line 249: If `TableName` is not configured, then change set analysis will fail - do we really need this validation then? will we even get here? It seems more like a cause rather than its own validation. 

Also please add the `SkipReplicationDeletion` validation.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1124 @@
...
+to `@aws-cdk/aws-ec2-alpha` and `Table` to `TableV2` in `aws-cdk-lib/aws-dynamodb` come with enhanced
+construct-specific validations and comprehensive migration guides provided by the CDK team to ensure
+safe deployment. These two construct upgrades are so-called `fully supported` construct upgrade.
+
+We aim to gradually increase the number of `fully supported` constructs. The tool is also designed to be open
+and pluggable, allowing users to support other construct migrations by incorporating their own custom
+validations. This flexibility enables users to extend the tool to cover additional constructs that may not
+yet be officially supported with custom validations, see [Custom Validations](#custom-validations).
+
+## Walkthrough of Construct Upgrade Example
```
iliapolo, line 71: The example is great. I think it would also be very helpful to add the VPC example as well, since it follows the other flow (refactor).

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1124 @@
...
+export class DemoStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    const table = new dynamodb.TableV2(this, 'MyTable', {
+      partitionKey: {
+        name: 'PK',
+        type: dynamodb.AttributeType.STRING,
+      },
+      tableName: 'DemoStack-MyTable794EDED1-11W4MR8VZ0UPE',
```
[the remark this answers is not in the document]
GavinZZ: This is actually not possible. The `resourcesToImport` features does not work unfortunately in this case because what it does requires the stack to be identical except the newly created resource that will be imported. What this means is that the user workflow will be torn into multiple step:
1. Users add a new TableV2 construct
2. Users run cdk import/cdk deploy (with migrate.json to use resourcesToImport feature).
3. Once import is successful, users remove the TableV1 construct
4. Deploy the change

This now becomes a multi-step and non-automated step with user intevention in between. I do not like this approach and I don't think it's worth the pain. I outlined a few possible workarounds instead to not hardcode the value in the stack.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+```ts
+const tableNames = JSON.parse(fs.readFileSync('./table-names.json', 'utf8'));
+
+const table = new dynamodb.TableV2(this, 'MyTable', {
+  // ... other properties ...
+  tableName: tableName['prod'],
+});
+```
+
+> Important: The tableName property is required for the upgrade process and cannot be removed afterward, as CloudFormation will
```
iliapolo, line 803: We spoke about possibly enhancing the `deploy` command to support a mapping file, just like `import` does:

```console
cdk deploy --resource-mapping /path/to/mapping.json
``` 

Doing so will allow omitting the `tableName` property from the CDK code itself, and thus not introduce this limitation. Does this paragraph mean to say that `cdk deploy` cannot be enhanced like that? If so - i'm not sure I see why. Can you elaborate?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+CDK stores a JSON formatted context file in the CDK CLI codebase that maps targets to their upgrade configurations:
+
+```json
+{
+  "aws-cdk-lib.aws_dynamodb.TableV2": {
+    // Valid values are 'Import' and 'Refactor'
+    "strategy": "Import",
+    
+    // Resource type prefixes to check in drift detection and retain policy validation
+    "source": ["AWS::DynamoDB", "Custom::ReplicationDeletion"],
```
iliapolo, line 1137: Can we be more accurate here?

```suggestion
    "retain": ["AWS::DynamoDB::Table", "Custom::ReplicationDeletion"],
```

Also, this list is relevant only for the `aws-cdk-lib.aws_dynamodb.Table` construct. It means we are making an assumption that the upgrade is always from `Table` to `TableV2`. 

This holds true because those are the only two we have **right now**, but once a future `TableV3` is introduced, this context becomes ambiguous:

```json
{
  "aws-cdk-lib.aws_dynamodb.TableV3": {...}
}
```

What does the above context refer to? is it `Table -> TableV3` OR `Tablev2 -> TableV3`. Nothing guarantees that the same information is sufficient for both. Since this is an implementation detail, we can evolve it however we want to fit that use case, so it doesn't need to be addressed now; but it would be good to see how such context will support upgrade when more than 2 variants are available.

#### Thought for the future

This list is essentially **the list of stateful resources** this construct encompasses. It can and should be at some point available as static data that the CLI has access to. This will allow `cdk deploy` to optionally block removal of stateful resources regardless of whether an upgrade or a regular deploy is performed.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+    // Resource types to verify import action in change set analysis
+    "target": ["AWS::DynamoDB::GlobalTable"]
```
iliapolo, line 1140:
```suggestion
    "import": ["AWS::DynamoDB::GlobalTable"]
```

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+```s
+Performing construct upgrade validation...
+
+⚠️  Note: This is an experimental feature
+
+Validations:
+✅ No CloudFormation drift detected
+❌ Change set analysis failed
+  │ Resources: MyTableReplicauswest285A33668 (Custom::DynamoDBReplica)
```
iliapolo, line 242: Did you validate that retaining the custom resource while removing the backing lambda function is possible? CloudFormation doesn't reject such a change?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+- Blocking the CI system while adding the `cdk construct-upgrade` command
+- Committing the change to trigger the upgrade validation
+- Unblocking the CI system and preventing new commits during the process
+- Removing the upgrade command from CI pipeline configuration after completion
+
+This process is cumbersome and typically requires administrative permissions to modify the CI system. Given
+these challenges, we recommend using Ops-led deployment for teams with complex pipeline setups until a more
+elegant solution is developed in the future updates.
+
+## Rollback
```
iliapolo, line 722: What about if the user wants to go back to the previous variant? not because it failed - just because they changed their minds. 

Will running `cdk construct-upgrade aws-cdk-lib.aws_dynamodb.Table` work for the reverse migration?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+}
+```
+
+Now, validate the construct upgrade using the CDK CLI:
+
+> Note that the flag `--unstable=construct-upgrade` is necessary to indicate that
+> it's an experiemntal feature that users are using until feature stablized.
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade --target @aws-cdk.aws-ec2-alpha.VpcV2
```
iliapolo, line 1042: Its somewhat confusing to use `@aws-cdk.aws-ec2-alpha.VpcV2` as the target here because the code change actually introduces many more new concstructs:

- ec2alpha.Route
- ec2alpha.SubnetV2
- ...

Can we think of a different value here that more accurately describes the migration? Maybe:

```suggestion
cdk construct-upgrade --unstable=construct-upgrade --target @aws-cdk.aws-ec2-alpha
```

?

iliapolo:
## Overall Thoughts

I think the benefit of this proposal outweighs its cost and we should drive this to completion. 

### Benefit

Correctly deploying a CDK application that underwent a destructive change is not a trivial task. 
Depending on the change itself, the user must either perform a CloudFormation refactor, or carefully execute a `retain-import-deploy` workflow. Mistakes can happen on two fronts:

- Incorrect code changes.
- Incorrect deployment action.

The new `cdk construct-upgrade` command alleviates a lot of the concern to deploy such changes. The validations themselves are generic enough to provide long term value for future construct variants, and will help the team roll them out more frequently. 

> We should start working on a support policy for the old variants. Otherwise, we might end up having to support multiple variants in parallel; and that will have a significant maintenance cost.

### Cost

From a coding perspective, the changes required from the CLI boil down to:

- Implement drift detection (good feature to have anyway)
- Enhance `cdk diff` to detect resource imports.
- Change set analysis to detect unintended resource removal.
- Change set analysis to detect expected resource imports.
- New `construct-upgrade` command that invokes existing CDK capabilities.
- Context file

These are all straightforward to implement, and mostly involve either stitching code together, or adding simple checks onto information the CLI already has. A few focused weeks should be enough.

From a long term maintenance perspective, I believe this change does not incur much additional cost:

- Construct specific validations are not currently required. For future variants, I believe we can convert most potentially specific validation, to a generic change set check.
- The new `construct-upgrade` command will, for the most part, invoke existing CDK features and will not create a large branching in the code.
- Compared to the effort required into putting together the migration guide, the changes required to the CLI are small. 

   > Note that the migration something is something we must provide anyway, for every new variant.

### Adoption

I have my doubts about whether this feature will increase user adoption of new construct variants. In my experience, most users will delay undertaking breaking change migrations until the old construct is out of support. For those users who do decide to migrate though, this feature will be beneficial, and thats a good enough reason to do it.

## RFC State

Until now, reviews have mostly been focused on determining required scope and functionality. We managed to significantly reduce the cost while not compromising on UX too much. Contentious points have been addressed and agreed on.

Merging the RFC however still requires some more work:

- Deeper API review. We can also postpone this to implementation time though.
- High level project plan is out of date.
- Rollback section needs more details.

Depending on people allocation we have for the project, we can start implementing some functionality while still working on finalizing the RFC for merge.

## Future Enhancements

Once a new construct variant is published, there will most likely be users who try to upgrade to it without following the migration guide.

> Because the guide is something the user has to know about, but nothing guarantees they will.

Ideally, the validations proposed here could be deeply integrated into the regular `cdk deploy` command, and not require users to be aware of additional resources. The main blocker for this right now is the operation specific context users must provide (e.g `--target`) in order for us to know which changes are expected, and which aren't. 

We could potentially brainstorm a way for `cdk deploy` to detect that based on a deeper (and more complex) change set analysis. After we roll this first phase out and get some customer feedback and real life examples, we could consider this approach as a way of graduating the feature out of the `unstable` phase.

text/0715-construct-upgrade-support.md
GavinZZ: It's possible but this will cause the whole workflow to be very complex.



The resourcesToImport features does not work unfortunately in this case because what it does requires the stack to be identical except the newly created resource that will be imported. What this means is that the user workflow will be torn into multiple step:

    Users add a new TableV2 construct
    Users run cdk import/cdk deploy (with migrate.json to use resourcesToImport feature).
    Once import is successful, users remove the TableV1 construct
    Deploy the change

This now becomes a multi-step and non-automated step with user intevention in between. I do not like this approach and I don't think it's worth the pain. I outlined a few possible workarounds instead to not hardcode the value in the stack.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+With the initial release of the CDK construct upgrade tool, developers would use the migration guide to
+update their CDK code, replacing legacy constructs (e.g., Table) with modern equivalents (e.g., TableV2).
+
+After updating the code to use the new construct, users would use the CDK CLI command `cdk construct-upgrade --unstable=construct-upgrade`
+to validates the existing CDK stack and environment, ensuring they are ready for migration. The tool flags any
+blockers, such as CloudFormation drift. Additionally, it analyzes the CloudFormation change set and validates
+any library-specific configurations. See [Construct Upgrade Validations](#construct-upgrade-validations) for a detailed breakdown of the validations.
+
+In the initial release of the CDK construct upgrade tool, VPC-related constructs upgrade from `aws-cdk-lib/aws-ec2`
+to `@aws-cdk/aws-ec2-alpha` and `Table` to `TableV2` in `aws-cdk-lib/aws-dynamodb` come with enhanced
```
rix0rrr, line 62: Is moving from a stable construct to an alpha construct really an "upgrade"?

text/0715-construct-upgrade-support.md
GavinZZ: Yes, it will work smoothly. I can add some more content to Rollback section to explain this. The same can also be applied to the refactor approach.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+Validations:
+✅ No CloudFormation drift detected
+❌ Change set analysis failed
+  │ Resources: MyTableReplicauswest285A33668 (Custom::DynamoDBReplica)
+  └   • RemovalPolicy: 'Delete' (expected: 'Retain')
+
+Error: Construct upgrade validation failed
+```
+
+### Custom Validations (Optional Future Enhancement)
```
otaviomacedo, line 248: Isn't this already covered by the support to policy-as-code in the CLI?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+For instance, users upgrading to TableV2 might want to ensure their existing table construct has
+`DeletionProtection` enabled. To accommodate such scenarios, the `cdk construct-upgrade` CLI command would
+support custom analysis files as input.  For more details about the setup and file format for custom
+validations, refer to the [Custom Validations](#custom-validations-optional-future-enhancement) section.
+
+Example usage:
+
+```sh
+cdk construct-upgrade --unstable=construct-upgrade --target aws-cdk-lib.aws-dynamodb.TableV2 \ 
+  --custom-analysis table-custom-analysis.ts
```
rix0rrr, line 263: Woah! People are going to submit source code?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+- --target (Required):
+  The target construct and module you want to upgrade to. For example, `aws-cdk-lib.aws-dynamodb.TableV2` or `@aws-cdk/aws-ec2.VpcV2`
+  For fully supported construct upgrades, we can short hand it to `TableV2` or `VpcV2`.
+- --dry-run (Optional)
+  This flag is default to false. When the flag is set to true, the CLI acts as dry-run and will only trigger validation.
+- --require-approval (Optional)
+  Specify what security-sensitive changes require manual approval.
+  - any-change – Manual approval required for any change to the stack.
+  - broadening – Manual approval required if changes involve a broadening of permissions or security group rules.
```
otaviomacedo, line 383: Do we expect this case to happen?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+This strategy is suitable for constructs where changes do not involve custom resource
+removal or introduction of new underlying resource types. It leverages the CloudFormation stack refactoring
+feature to maintain resource continuity.
+
+Currently, users must specify a `refactor.json` file that maps logical ID changes from the old construct to the new one.
+If construct upgrade validations are successful, the command continues to execute and deploy the CDK stack using CloudFormation
+stack refactor feature, which helps retain existing resources while applying the necessary infrastructure updates.
+
+> Future Enhancement: CDK team is developing a [built-in refactor feature](https://github.com/aws/aws-cdk-rfcs/pull/705)
```
otaviomacedo, line 409: The RFC has already been merged:

```suggestion
> Future Enhancement: CDK team is developing a [built-in refactor feature](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0162-refactoring-support.md)
```

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+|                           |                               |
+|                           ▼                               |
+|                ┌─────────────────────┐                    |
+|                │ Validate CFN drift  │                    |
+|                └─────────────────────┘                    |
+|                           |                               |
+|                           ▼                               |
+|                ┌─────────────────────┐                    |
+|                │   CFN Change set    |                    |
+|                |       analysis      │                    |
```
rix0rrr, line 323: Not explicitly set, but we create the change set here I suppose?

A change set that would just try to deploy the new template? And in the case of the DynamoDB table migration, we would see:

* **Delete** `AWS::DynamoDB::Table` with dynamic name
* **Add** `AWS::DynamoDB::GlobalTable` with name `DemoStack-MyTable794EDED1-11W4MR8VZ0UPE`

And our validation would say "yes that is what we want to see" ?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+Arguments:
+
+- CDK stack ID (Optional)
+    The construct ID of the CDK stack from your app to synthesize.
+
+Options:
+
+- --target (Required):
+  The target construct and module you want to upgrade to. For example, `aws-cdk-lib.aws-dynamodb.TableV2` or `@aws-cdk/aws-ec2.VpcV2`
+  For fully supported construct upgrades, we can short hand it to `TableV2` or `VpcV2`.
```
rix0rrr, line 377: Shouldn't this identify an instance (construct)? Why does it identify a class?

text/0715-construct-upgrade-support.md
GavinZZ: We already have a plan to stablize this before end of Q2, so hopefully it's a stablized module when the framework actually launches.
But IMO, I think this is a good point being raised that and I think people in general won't upgrade to an experimental module given that it may not be backward compatible.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+- --dry-run (Optional)
+  This flag is default to false. When the flag is set to true, the CLI acts as dry-run and will only trigger validation.
+- --require-approval (Optional)
```
rix0rrr, line 380: Nit, but this flag is poorly named. It should have been called `--require-confirmation`. We should deprecate the old flag, add the new one, and don't add any more copies of the silly name.

text/0715-construct-upgrade-support.md
GavinZZ: A change set does not deploy the new template until we execute it.

In the case of the DynamoDB table migration, we would see:

- Remove AWS::DynamoDB::Table (with policyAction saying 'Retain')
- Add AWS::DynamoDB::GlobalTable with name DemoStack-MyTable794EDED1-11W4MR8VZ0UPE and the corresponding action is 'Import'

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+Error: Construct upgrade validation failed
+```
+
+This feature would enhance the flexibility of the construct upgrade process, allowing teams to enforce
+specific policies or best practices during migrations.
+
+> If you're interested in seeing a full example of VPC V2 example, please checkout
+> [Appendix B: VPC V2 Upgrade Example](#appendix-b-vpc-related-migration-example).
+
+### User Experience: Visual Workflow
```
SimonCMoore, line 290: If you are going to show a visual with named blocks, I think you need to explain what each block is before or provide a key with links to where the explanation is. Ensure explanation covers both why it is needed and how it works.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+### Retain-Remove-Import Migration
+
+This strategy is employed when significant changes occur in the underlying resource, such as transitioning from
+a custom resource to a native L1 resource or resource property (e.g., Table to TableV2 migration). The CDK construct upgrade
+framework handles this process through several automated steps:
+
+- Validation of `RemovalPolicy.RETAIN` on the original construct to prevent accidental resource deletion.
+- Execution of the `cdk construct-upgrade` command, which performs necessary validations.
+- If validations pass, automatic triggering of deployment using CDK's import feature to link existing resources
```
rix0rrr, line 421: Missing:

- A deployment that removes the resources from the template.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+- Present a confirmation prompt requiring explicit user approval to proceed (unless `--require-approval never` is specified).
+
+This enhancement would help prevent accidental resource replacements by guiding users toward the proper construct upgrade process.
+
+### Limitation: In-Place Migration Strategy
+
+This strategy utilizes the CDK/CloudFormation refactoring feature and inherits its limitations. It experiences
+the [same limitations](https://github.com/aws/aws-cdk-rfcs/blob/otaviom/refactoring-support/text/0162-refactoring-support.md#pipelines-with-version-superseding)
+as described in the CDK Refactoring RFC. Primarily, these limitations affect pipelines with version superseding,
+potentially causing issues in complex deployment scenarios.
```
vishaalmehrishi, line 775: Would be great to understand the complex deployment scenarios and how likely they are to cause issues.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+The new `cdk construct-upgrade` CLI command will implicitly call the `cdk diff --detect-drift` command as part of
+its validation process. To avoid being overly restrictive, the validation will primarily focus on the main resources
+change. The main resources can be determined based on the `--target` option, more can be found in
+[Appendix C: Main Resources](#appendix-c-main-resources-computation).
+
+Drift detection is important to ensure construct upgrade safety by preventing accidental deployment issues when
+the CDK construct state does not match the actual deployed resource state. Without detecting drift, upgrades may
+unintentionally remove manual configurations made through the AWS Console, potentially triggering resource
+replacements, data loss, or availability downtime.
```
rix0rrr, line 466: Note that the only way to "resolve" drift manually is to manually undo those manual configurations via the AWS Console.

Changing your template to match reality may or may not do the right thing (it may even fail to deploy), and so it's unreliable and we cannot recommend it.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+```s
+[-] AWS::DynamoDB::Table MyTable orphan
+[+] AWS::DynamoDB::GlobalTable MyTable add
+```
+
+For `In-Place Migration` strategy, change set analysis is not required as CloudFormation stack refactoring enforces that
+the refactor operations don't allow new resource creations, resource deletions, or changes to resource configurations,
+see [Stack refactoring limitations](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stack-refactoring.html#stack-refactoring-limits).
+
+For the `Retain-Remove-Import Migration` strategy, we will introduce an additional option `--import-existing-resources`
```
otaviomacedo, line 508: I find all this part until the end of the section hard to follow. There are a lot of implementation details, but I don't understand what is important from the user's point of view.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+The CDK construct upgrade framework employs two different strategies for migrating resources, automatically
+selecting the appropriate approach based on the `--target` construct. CDK CLI stores and uses a context variable
+to determine the mapping between upgrade targets and their corresponding strategies.
+
+### In-place Migration
+
+This strategy is suitable for constructs where changes do not involve custom resource
+removal or introduction of new underlying resource types. It leverages the CloudFormation stack refactoring
+feature to maintain resource continuity.
```
SimonCMoore, line 403: This could be clearer on the problem it solves, "maintain resource continuity" is a bit vague. Today moving constructs around and similar changes in a CDK application can cause the logical id of the resulting cloud-formation resource to change, resulting in a deletion and recreation of the resources. Stack refactoring enables a logical name change without any change to the deployed resource.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+In these systems, implementing the CI/CD-driven deployment would require:
+
+- Blocking the CI system while adding the `cdk construct-upgrade` command
+- Committing the change to trigger the upgrade validation
+- Unblocking the CI system and preventing new commits during the process
+- Removing the upgrade command from CI pipeline configuration after completion
+
+This process is cumbersome and typically requires administrative permissions to modify the CI system. Given
+these challenges, we recommend using Ops-led deployment for teams with complex pipeline setups until a more
+elegant solution is developed in the future updates.
```
vishaalmehrishi, line 720: I think it's really important to define _complex_ here. It would also be good to understand how many customers have complex setups i.e. does the proposed approach actually work for a majority of customers?

text/0715-construct-upgrade-support.md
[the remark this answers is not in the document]
rix0rrr: This really is a non-starter.

The reason is that this construct or stack immediately becomes bound to a single (account, region) combination, and it becomes undeployable anywhere else.

To do it "properly", this should have a lookup map:

```ts
const tableNames = {
  'us-east-1/11111111': 'DemoStack-MyTable794EDED1-11W4MR8VZ0UPE',
  'us-east-2/222222222': 'DemoStack-MyTable794EDED1-83483583'
  // ...
 };

 const table = new dynamodb.TableV2(this, 'MyTable', {
      partitionKey: {
        name: 'PK',
        type: dynamodb.AttributeType.STRING,
      },
      tableName: tableNames[`${this.region}/${this.account}`], // Won't work for region-agnostic stacks literally but the point still stands
      replicas: [
        {
          region: 'us-west-2',
        }
      ]
    });
```

But that's not really proper.

So we need a good answer to this. Unresolving this comment.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+    });
+    cdk.Tags.of(publicSN.routeTable).add('Name', `${this.stackName}/vpc/public-subnet-test-cyberark1Subnet1`);
+  }
+}
+```
+
+Now, validate the construct upgrade using the CDK CLI:
+
+> Note that the flag `--unstable=construct-upgrade` is necessary to indicate that
+> it's an experiemntal feature that users are using until feature stablized.
```
vishaalmehrishi, line 1039: [Possibly out of scope for the RFC] What does it take for the feature to be stabilised?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+#### Milestone 2
+
+Milestone 2 will be a future release to help automate the code migration process as well.
+This is still under investigation and may or may not be included in the CDK construct upgrade support.
+
+### Will this tool support every future V1 to V2 construct?
+
+The CDK construct upgrade tool is designed to support common and high-impact migration scenarios for widely adopted
+constructs and modules. While we aim to expand coverage over time, we cannot guarantee support for every
+future V1-to-V2 migration path due to variations in technical complexity and resource constraints.
```
ozelalisen, line 922: How will users know which constructs are supported? Will there be clear documentation?

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+✅ No CloudFormation drift detected
+✅ Change set analysis complete
+```
+
+Depending on the deployment workflow and environment constraints, there are two common enterprise
+patterns:
+
+1. CI/CD-driven Deployment:
+For teams with access to CI/CD and production environments, developers commit the upgraded CDK stack
+code to version control. The CI/CD pipeline runs `cdk construct-upgrade` as part of the deployment process.
```
rix0rrr, line 691: Note that this explicitly won't work for deployment pipelines as used by Amazon, since they deploy CloudFormation stacks only and don't run arbitrary CLI commands.

text/0715-construct-upgrade-support.md
GavinZZ: We'll publish documentation around the CLI usage and the `--target` value should enumerate the valid values.

text/0715-construct-upgrade-support.md
mrgrain: Counter opinion: Early adopters will still migrate from current to alpha, if the alpha provides important new features.

text/0715-construct-upgrade-support.md
mrgrain: I like this idea. Would it even be more feasible to implement the construct validations using policy-as-code?

text/0715-construct-upgrade-support.md
mrgrain: We don't have a strict criteria for that, especially not for CLI features. Generally the idea is to give us and users some bake time (at least 3 months, better 6 months) to figure out any issues and papercuts that might require breaking changes to resolve. E.g. renaming or removing options.

text/0715-construct-upgrade-support.md
GavinZZ: This way we allow simultaneous multiple instances of the construct to upgrade at the same time. In general, it's up to users' code change, but if users update multiple Vpcs to VpcV2s, I think it makes sense for the construct upgrade framework to upgrade it all.

text/0715-construct-upgrade-support.md
GavinZZ: Can you help me understand this a bit more on this policy-as-code feature?

text/0715-construct-upgrade-support.md
GavinZZ: The phrasing is inaccurate, it's actually talking about the issue right above where the pipeline can't customize the CLI command to run.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+
+- Display a warning message about the potential construct upgrade scenario and prompt users to use `cdk construct-upgrade`
+  and the corresponding migration guide, or
+- Present a confirmation prompt requiring explicit user approval to proceed (unless `--require-approval never` is specified).
+
+This enhancement would help prevent accidental resource replacements by guiding users toward the proper construct upgrade process.
+
+### Limitation: In-Place Migration Strategy
+
+This strategy utilizes the CDK/CloudFormation refactoring feature and inherits its limitations. It experiences
```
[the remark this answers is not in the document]
GavinZZ: Not sure what you meant here, the alternative is to use CDK import feature in the other strategy.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,1251 @@
...
+A common pattern involves releasing versioned constructs or modules (e.g., V2 variants)
+that coexist with their predecessors. For example:
+
+* The `@aws-cdk/aws-ec2-alpha` alpha module introduces the modernized `VpcV2`, `SubnetV2`, and other
+new construct, designed to succeed the original Vpc related constructs in `aws-cdk-lib/aws-ec2`.
+* The `TableV2` construct in `aws-cdk-lib/aws-dynamodb` replaces legacy Table implementations
+while retaining API compatibility.
+
+These updates unlock enhanced capabilities. Construct upgrade between variants, however, often requires
+code changes that could trigger CloudFormation resource replacements (e.g., NAT gateways being
```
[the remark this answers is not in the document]
GavinZZ: The full workthrough example below should show this, but I can add a small section here.

text/0715-construct-upgrade-support.md
GavinZZ: We don't currently have a way to track the analytics usage of users' CI/CD pipeline setup unfortunately. That's why we provided a workaround approach for a less-ideal solution.

text/0715-construct-upgrade-support.md
GavinZZ: Good question, I think this is more of a question to be answered by the core team. Can someone from core team chime in and help explain the general process to stablize a CLI command?

text/0715-construct-upgrade-support.md
GavinZZ: Thanks Simon, I'll move it to the later section after each individual validations are explained.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,786 @@
...
+}
+```
+
+**Step 2:**: CDK users follow the construct upgrade guide to validate code changes by running the
+CDK CLI commands below:
+
+```s
+# exit with 1 if CloudFormation drift is detected for resource type AWS::DynamoDB::Table
+# exit with 0 otherwise
+$ cdk drift --rejectAWS::DynamoDB::Table
```
rix0rrr, line 103: Why do we need to reject specifically this? Would we not reject on any drift at all?

(I mean, it's fine to have options on the command if a user knows they want to ignore drift... but it shouldn't [need to be] be part of the **expected** flow)

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,786 @@
...
+
+Stack DemoStack
+✅ No CloudFormation drift detected
+
+# exit with 1 if resources types AWS::DynamoDB::Table and Custom::DynamoDBReplica are deleted, OR
+# exit with 1 if resources type AWS::DynamoDB::GlobalTable are not imported
+# exit with 0 otherwise
+$ cdk diff --retain AWS::DynamoDB::Table,Custom::DynamoDBReplica --import AWS::DynamoDB::GlobalTable
```
rix0rrr, line 111: This is an annoying command line to type.

I will once again float my idea of doing `cdk diff --validate-migration=<migration plan name>`.

I think the migration plan should also **print to the user** the command they're expected to run next.

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,786 @@
...
+
+Stack DemoStack
+✅ There were no differences
+```
+
+**Step 3:**: CDK users follow the construct upgrade guide to deploy the changes by running the
+CDK CLI command below:
+
+```s
+$ cdk deploy --import-existing-resources
```
rix0rrr, line 121: How about `cdk refactor --retain-delete-import`, which it definitely is going to have to be for this specific migration.

The workflow is then always 

1. Update code
2. `cdk diff --validate-whatever`
3. `cdk refactor`

text/0715-construct-upgrade-support.md
```diff
@@ -0,0 +1,786 @@
...
+cdk drift --rejectAWS::DynamoDB::Table && cdk diff --retain AWS::DynamoDB::Table,Custom::DynamoDBReplica --import AWS::DynamoDB::GlobalTable
+
+if [ $? -eq 0 ]; then
+    echo "Validation passed. Proceeding with deployment..."
+    cdk deploy --import-existing-resources # `cdk refactor --unstable=refactor` for VPC constructs upgrade
+else
+    echo "Validation failed. Deployment blocked."
+    exit 1
+fi
+```
```
rix0rrr, line 422: Can you be specific about the entire workflow end to end? I think it looks like this:

1. Clear their deployment pipelines, no change may be flowing through it
2. Update their pipeline definition to put the above command in it
3. Deploy a single change, the migration change
4. Wait for the change to clear the pipeline (no other changes may go through it)
5. Update their pipeline definition to remove the above command

Is that right?

----

Now think of a typical AWS pipeline that has ~3 pre-prod stages and >20 prod stages, and various approval workflows. Does the above still sound acceptable?

rix0rrr: I think we should say "CI/CD" is too hard to design right now, we will do it later.

rix0rrr:
Another thought: 

In the presence of `cdk refactor`... is there value to doing:

```
cdk diff --validate ...
cdk refactor
```

? What would `cdk diff` add that `cdk refactor` doesn't already do?

text/0715-construct-upgrade-support.md
GavinZZ: I think CI/CD is one of the core feature for construct upgrade. We don't have to fully support complex setup but we should definitely explain the workflow for simpler setup.

I agree the workflow you mentioned doesn't seem ideal especially when users want to intermix construct upgrade with normal feature release changes. The issue with AWS pipeline is that there's usually a 12 hours bake time in between each stage which makes things worse. In either these two case,s I think we can add a note to the RFC to explain the difficulty and suggest users to use the Ops-Lead workflow instead.

Issue #732, opened 2025-06-06
CDK CLI Telemetry


kaizencc:
## Description

The CDK CLI team currently has no visibility into how the CLI is performing for users. In order to detect accidental regressions more quickly, and make data-driven decisions on what to work on next, the CDK CLI will begin to collect anonymous telemetry on user commands.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @kaizencc 
| Author(s)           | @kaizencc
| API Bar Raiser      | @iliapolo 
| Stakeholders        | @rix0rrr @mrgrain 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

mikejgray:
Generally in favor of this approach to data-driven improvement, but I have concerns about the redaction mechanisms of tracebacks and other error-type messages. While they are optional, it seems that there are too many edge cases to properly redact all customer information, and I hope that these optional sections would default to off.

Without investigating in further detail I don't have specifics about which edge cases to examine in more detail, but mention it as the section I find potentially objectionable.

kaizencc:
Thanks for the feedback thus far! A couple of points to address the feedback here -- everyone is also very welcome to comment on the RFC pull request itself.

We believe that client side telemetry is the right way to achieve our goals of making data-driven product decisions and resolving errors faster. We are trying to raise the bar for how we respond to customer pain points and intentionally are driving a solution where we would be able to proactively solve issues rather than rely on reactionary efforts based on [manual] user reports. 

[...]

PR #734, opened 2025-06-06
initial RFC for telemetry

kaizencc:
This is a request for comments about CDK CLI Telemetry. See #732  for
additional details. 

---

### DIFF
--- a/text/0732-cdk-cli-telemetry.md
+++ b/text/0732-cdk-cli-telemetry.md
@@ -1,0 +1,136 @@
+# CDK CLI Telemetry
+
+* **Original Author(s):**: @conroyka
+* **Tracking Issue**: #732
+* **API Bar Raiser**: @iliapolo
+
+The CDK CLI team currently has no visibility into how the CLI is performing for users.
+In order to detect accidental regressions more quickly, and make data-driven decisions on what to work on next,
+the CDK CLI will begin to collect anonymous telemetry on user commands.
+
+The purpose of this RFC is to iterate in public. We welcome community feedback on the opt-out mechanism,
+on whether you would decide to opt-out,
+or on additional telemetry values that we are missing that have proven helpful in other CLIs. 
+
+## Data Collection
+
+Before we start, two definitions that will help guide this RFC:
+
+* A **session** corresponds to a CLI command execution.
+* An **event** is a collection of metrics related to its **session**.
+
+A **session** has multiple **events**, such as the case with cdk watch. cdk watch is a long-running **session** that can invoke the `synth` and `deploy` **events** multiple times. 
+
+At the end of each **event**, we will send data to the telemetry service. 
+
+### Example
+
+Here’s a snippet of data we will collect for a `deploy` **event** (not a full example, just the important parts):
+
+```json
+{
+  identifiers: {
+    sessionId: "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9", // Uuid generated on each CLI command invocation
+    eventId: "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9:1", // sessionId + an increment for each additional event in the session
+    installationId: "3F1FD23A-58A9-4C0D-8A82-098D6101B322", // Uuid stored on a local file on the developer"s machine
+    accountId: "d445c5a1b10d4f9c90a4b17769aa84d2e5d5c3da642c4acd392c71a46275e6f9", // optional hash of account ID
+    region: "us-east-1", // optional region being deployed to
+  },
+  event: {
+    state: "SUCCESS",
+    eventType: "deploy",
+    command: {
+      path: ["cdk", "deploy", "<REDACTED>"],
+      parameters: {},
+      config: {},
+    },
+  }
+}
+```
+
+Alternatively, if deployment fails, the deploy **event** looks like this:
+
+```json
+{
+  identifiers: {
+    sessionId: "14B36D48-4DFF-47C3-B0E4-D966DD6DB038",
+    eventId: "14B36D48-4DFF-47C3-B0E4-D966DD6DB038:1",
+    installationId: "3F1FD23A-58A9-4C0D-8A82-098D6101B322",
+    accountId: "DDCA70E4-C73A-4B93-B464-BBA7DCCB6B86",
+    region: "us-east-1",
+  },
+  event: {
+    state: "FAIL",
+    eventType: "deploy",
+    command: {
+      path: ["cdk", "deploy", "<REDACTED>"],
+      parameters: {},
+      config: {},
+    },
+  },
+  error: {
+    name: "ToolkitError",
+    message: "Error: Asset <REDACTED> upload failed: AccessDenied: Access Denied",
+    trace: "    at AssetPublishing.publishAsset (<PATH>/aws-cdk/lib/assets/asset-publishing.js:128:23)
+    at CloudFormationDeployment.publishAssets (<PATH>aws-cdk/lib/api/cloudformation-deployment.js:295:41)
+    at CloudFormationStackArtifact.prepareForDeployment (<PATH>aws-cdk/lib/api/cdk-toolkit.js:517:12)"
+    logs: "Deploying stack <REDACTED>
+    IAM Statement Changes
+    ┌───┬─────────────────────────┬────────┬─────────────────────────┬─────────────────────────┬───────────┐
+    │   │ Resource                │ Effect │ Action                  │ Principal               │ Condition │
+    ├───┼─────────────────────────┼────────┼─────────────────────────┼─────────────────────────┼───────────┤
+    │ + │ <REDACTED>              │ Allow  │ sts:AssumeRole          │ Service:lambda.amazonaw │           │
+    │   │                         │        │                         │ s.com                   │           │
+    └───┴─────────────────────────┴────────┴─────────────────────────┴─────────────────────────┴───────────┘
+    (NOTE: There may be security-related changes not in this list. See https://github.com/aws/aws-cdk/issues/1299)
+
+    Bundling asset <PATH>...",
+  }
+}
+```
+
+## Opt-Out
+
+Collecting anonymous data will help the CDK team on two fronts: usage metrics and error debugging.
+Both of these aspects will help us better serve CDK customers in the future.
+We will ensure that we are not collecting [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1)
+or personally-identifiable information. 
+
+You can easily opt-out of sending telemetry data. Customers can set `toolkit-telemetry: false` in their `cdk.json` configuration files. Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App. Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
+
+We will also respect an environment variable, `TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
+
+Alternatively, you can run the CLI command (which disables telemetry for the CDK App):
+
+```bash
+> cdk telemetry --disable
+```
+
+## Metrics
+
+Below is an exhaustive list of metrics we will collect, with the reason for each.
+It’s important to note that all data we collect will be anonymous.
+We are only collecting enumerable values derived from customer input.
+For example, in command, we only collect `boolean` or `enum` values but not free text. Any free text will be recorded as a passed-in `boolean`.
+
+| Metric                       | Description                                                                                                                  | Reason                                                                                                                                                                                                                                                                                                          |
+|------------------------------|------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| CDK CLI Version              | CLI version (x.y.z)                                                                                                          | Collecting this will help us bisect our data by version, which can help us identify problematic versions.                                                                                                                                                                                                       |
+| Installation ID              | Uuid stored on local machines executing CLI commands                                                                         | This serves as an alternative to Account ID to help us get a sense of blast radius. Each unique machine will have their own Installation ID, but since multiple different machines can target the same AWS Account this counts a slightly different idea of a 'user'.                                           |
+| Hashed Account ID (optional) | Account ID, hashed to be anonymous                                                                                           | This will help us get a sense of blast radius when regressions are identified.                                                                                                                                                                                                                                  |
+| Region (optional)            | AWS Region                                                                                                                   | This will help us bisect our data for region-specific issues.                                                                                                                                                                                                                                                   |
+| Timestamp                    | The time (generated by the client) the data is sent to our telemetry endpoint                                                | This will help us generate time series graphs.                                                                                                                                                                                                                                                                  |
+| Event State                  | The "result" of the event. Possible values are 'SUCCESS', 'FAILURE', 'ABORTED'                                               | This will help us track error rates on each CLI command and alert us to potential regressions before they are reported on GitHub.                                                                                                                                                                               |
+| Event Type                   | An identifier for the type of event that produces the data                                                                   | Since we aim to send metrics on a per-event basis, and there could be multiple events in a command execution, this is an enum that identifies what kind of event transpired. Values could be 'synth', 'deploy', etc..                                                                                           |
+| Command Path                 | The command and properties entered into the CLI, with any free text redacted                                                 | For example, `cdk deploy myStack` would be turned into `['cdk', 'deploy', '<REDACTED>']` in the telemetry data. We are committed to redacting any customer content, but this information is critical to determine feature usage and will help us make data-driven decisions on what CLI features to prioritize. |
+| Command Parameters           | The optional parameters entereed into the CLI, with any free text redacted                                                   | Some of our features include parameters, like `cdk deploy --watch`. Again, this information is critical to determine fetaure usage and for us to make data-driven decisions.                                                                                                                                    |
+| Configuration                | Additional parameters that may affect the CLI command (i.e. cdk.json / cdk.context.json values), with any free text redacted | This will help us facilitate debugging by providing a comprehensive view of all possible parameters that may affect the behavior of the CLI.                                                                                                                                                                    |
+| Operating System             | The operating system that is being used                                                                                      | This will help us debug issues that only affect specific operating systems.                                                                                                                                                                                                                                     |
+| CI/CD                        | Whether or not the CLI command is being invoked from a Ci/CD environment                                                     | Data points have a different relevance if they are from a developer's machine for from a CI/CD environment.                                                                                                                                                                                                     |
+| Node Version                 | The node version used in the environment                                                                                     | This will help us debug any node-specific issues that arise.                                                                                                                                                                                                                                                    |
+| Duration Stats               | The length of the event (and potentially, any smaller denominations that are interesting)                                    | We will monitor the duration of events like (but not limited to) synthesis, deploy, etc..                                                                                                                                                                                                                       |
+| Counting Stats               | Various counter metrics derived from your CDK app and CLI execution                                                          | We will monitor how counting stats like (but not limited to) CFN resource count affect the duration of deploy.                                                                                                                                                                                                  |
+| Dependencies                 | Relevant AWS dependencies and their versions                                                                                 | We plan to track dependencies that might affect the CLI result, like `aws-cdk-lib`, `jsii`, `projen`, etc..                                                                                                                                                                                                     |
+| Error Messages (optional)    | The error message returned, if applicable. Customer content redacted                                                         | Capturing the error message will help us aggregate data on which errors are encountered at a greater rate, as well as help us debug what went wrong in individual use cases.                                                                                                                                    |
+| Error Stack Trace (optional) | The stack trace of the error message, if applicable. Customer content redacted                                               | The stack trace will be helpful for individual debugging purposes and is necessary for us to be able to reproduce issues that may arise.                                                                                                                                                                        |
+| Error Logs (optional)        | The logs of a failed CLI command, if applicable. Customer content redacted                                                   | Error logs will also help us debug and reproduce issues that we see in the CLI.                                                                                                                                                                                                                                 |


text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+Both of these aspects will help us better serve CDK customers in the future.
+We will ensure that we are not collecting [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1)
+or personally-identifiable information.
+
+You can easily opt-out of sending telemetry data.
+Customers can set `toolkit-telemetry: false` in their `cdk.json` configuration files.
+Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App.
+Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
+
+We will also respect an environment variable, `TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
```
mrgrain: This should be prefixed with CDK. I don't think we can claim to be the only toolkit out there ;)  
```suggestion
We will also respect an environment variable, `CDK_TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+
+Collecting anonymous data will help the CDK team on two fronts: usage metrics and error debugging.
+Both of these aspects will help us better serve CDK customers in the future.
+We will ensure that we are not collecting [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1)
+or personally-identifiable information.
+
+You can easily opt-out of sending telemetry data.
+Customers can set `toolkit-telemetry: false` in their `cdk.json` configuration files.
+Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App.
+Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
```
mrgrain: Using "toolkit telemetry" over "telemetry": Is the intention here to distinguish between app metadata?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+
+We will also respect an environment variable, `TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
+
+Alternatively, you can run the CLI command (which disables telemetry for the CDK App):
+
+```bash
+> cdk telemetry --disable
+```
```
mrgrain: Looking into the future, I'd be nice to write this to be able to write this as well:

```
cdk deploy --no-toolkit-telemetry
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+Customers can set `toolkit-telemetry: false` in their `cdk.json` configuration files.
+Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App.
+Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
+
+We will also respect an environment variable, `TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
+
+Alternatively, you can run the CLI command (which disables telemetry for the CDK App):
+
+```bash
+> cdk telemetry --disable
```
mrgrain: Is there an enable command?

What does `cdk telemetry` do otherwise?

Why is it called "telemetry" here and "toolkit telemetry" elsewhere?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+| Counting Stats               | Various counter metrics derived from your CDK app and CLI execution                                                          | We will monitor how counting stats like (but not limited to) CFN resource count affect the duration of deploy.                                                                                                                                                                                                  |
+| Dependencies                 | Relevant AWS dependencies and their versions                                                                                 | We plan to track dependencies that might affect the CLI result, like `aws-cdk-lib`, `jsii`, `projen`, etc..                                                                                                                                                                                                     |
```
mrgrain:
```suggestion
| Dependencies                 | Relevant AWS dependencies and their versions                                                                                 | We plan to track AWS dependencies that might affect the CLI result, like `aws-cdk-lib`, `jsii`, `projen`, etc..                                                                                                                                                                                                     |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+| Error Messages (optional)    | The error message returned, if applicable. Customer content redacted                                                         | Capturing the error message will help us aggregate data on which errors are encountered at a greater rate, as well as help us debug what went wrong in individual use cases.                                                                                                                                    |
```
otaviomacedo: What counts as customer content?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+> cdk telemetry --disable
+```
+
+## Metrics
+
+Below is an exhaustive list of metrics we will collect, with the reason for each.
+It’s important to note that all data we collect will be anonymous.
+We are only collecting enumerable values derived from customer input.
+For example, in command, we only collect `boolean` or `enum` values but not free text. Any free text will be recorded as a passed-in `boolean`.
```
otaviomacedo: What kind of free text are we dropping?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,140 @@
...
+| Command Path                 | The command and properties entered into the CLI, with any free text redacted                                                 | For example, `cdk deploy myStack` would be turned into `['cdk', 'deploy', '<REDACTED>']` in the telemetry data. We are committed to redacting any customer content, but this information is critical to determine feature usage and will help us make data-driven decisions on what CLI features to prioritize. |
+| Command Parameters           | The optional parameters entereed into the CLI, with any free text redacted                                                   | Some of our features include parameters, like `cdk deploy --watch`. Again, this information is critical to determine fetaure usage and for us to make data-driven decisions.                                                                                                                                    |
```
otaviomacedo:
```suggestion
| Command Parameters           | The optional parameters entered into the CLI, with any free text redacted                                                   | Some of our features include parameters, like `cdk deploy --watch`. Again, this information is critical to determine fetaure usage and for us to make data-driven decisions.                                                                                                                                    |
```

text/0732-cdk-cli-telemetry.md
kaizencc: I have a link earlier in the RFC where i introduce the phrase: https://aws.amazon.com/compliance/data-privacy-faq/#topic-1

> We define customer content as software (including machine images), data, text, audio, video, or images that a customer or any end user transfers to us for processing, storage, or hosting by AWS services in connection with a customer's account, and any computational results that a customer or their end user derives from the foregoing through their use of AWS services. For example, customer content includes content that a customer or their end user stores in Amazon Simple Storage Service (S3). Customer content does not include account information, which we describe below. Customer content also does not include information included in resource identifiers, metadata tags, access controls, rules, usage policies, permissions, and similar items related to the management of AWS resources. AWS recommends that you do not include personally identifying, confidential, or sensitive information in these items. The terms of the [AWS Customer Agreement](https://aws.amazon.com/agreement/) and the [AWS Service Terms](https://aws.amazon.com/service-terms/) apply to your customer content.

text/0732-cdk-cli-telemetry.md
kaizencc: > Is there an enable command?

yes. added a sentence.

> What does cdk telemetry do otherwise?

no-op

> Why is it called "telemetry" here and "toolkit telemetry" elsewhere?

this seems like a marketing question. i like the sound of `cdk telemetry` here over `cdk toolkit-telemetry` and i think it won't be confusing. if you think i should stick to 'toolkit telemetry' everywhere, lemme know!

text/0732-cdk-cli-telemetry.md
kaizencc: as a global variable right? i agree. do you think it's necessary to have this immediately or are the other opt-out methods sufficient for now

text/0732-cdk-cli-telemetry.md
kaizencc: 1) can eventually be part of the toolkit
2) can't just be telemetry to differentiate from metadata resource, and in general feels too vague
3) other options might be cli-telemetry but again, what if we add it to the toolkit

text/0732-cdk-cli-telemetry.md
kaizencc: added an example for this

text/0732-cdk-cli-telemetry.md
mrgrain: I'm okay with that command, but I prefer consistency and right now we call this feature "telemetry" in one place and "toolkit-telemetry" in another.

Maybe we can extend the command to cover _all_ of the various metadatas and telemetry? `cdk telemetry` could be a fun interactive command.

And this could be for the toolkit specifically?
`cdk telemetry --disable toolkit`

--- 

Side bar: Do we now really have 3 different names for this kind of stuff?

text/0732-cdk-cli-telemetry.md
mrgrain: Not needed right now. Just thinking ahead how we can make cdk.json and CLI options more consistent.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,157 @@
...
+Both of these aspects will help us better serve CDK customers in the future.
+We will ensure that we are not collecting [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1)
+or personally-identifiable information.
+
+You can easily opt-out of sending telemetry data.
+Customers can set `toolkit-telemetry: false` in their `cdk.json` configuration files.
+Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App.
+Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
+
+We will also respect an environment variable, `CDK_TOOLKIT_TELEMETRY=false`. If set, this will achieve the same effect of disabling telemetry.
```
mrgrain: Nit: we seem to expect the value to be falsy to disable telemetry. If we accept any value (including true or 1) I'd throw a `DISABLE` in the name.

text/0732-cdk-cli-telemetry.md
kaizencc: i'm fine with `cdk telemetry --disable toolkit` but i am less bullish on our ability to turn `cdk telemetry` into an interactive command eventually. but happy to start with what you've proposed

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Redacting Customer Content
+
+### Redacting CLI Commands
+
+We will redact free text input supplied by the user in the CLI command input.
+We know the available types of CDK CLI parameters and options, so this will be trivial to implement.
+Input values that are boolean or enums are not customer content and not redacted. All other strings are redacted.
+
+Take the following command:
```
mrgrain: I would frame this differently. Redacting kind of implies we are collecting something. 
```suggestion
## Customer Content

We will not collect any customer content. Customer content is defined as [insert link here]. This poses a challenge for certain free text input like CLI arguments and options, because blah blah blah.

### CLI Input

We will only collect known command and option names from CLI input.
Input values that are boolean or enums are not customer content and not redacted.
We will not collect any free text input supplied by the user to CLI.
For arguments and options with string inputs we only collect that the respective argument/option was set.

Take the following command:
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+```bash
+> cdk bootstrap --bootstrap-bucket-name=MyBucket
+```
+
+Since `MyBucket` is free text, it will be redacted in the telemetry data sent to the service. This input will be processed into the following command object:
```
mrgrain:
```suggestion
Since `MyBucket` is free text, we will not collect this as part of telemetry data. This input will be processed into the following command object:
```

mrgrain: apply similar elsewhere

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+```json
+{
+  "command": {
+    "path": ["cdk", "deploy", "<redacted-stack-name>"],
+  }
+}
+```
+
+### Sanitizing Errors
```
mrgrain:
```suggestion
### Errors are sanitized
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+When something goes wrong in the CDK CLI, we will report the error to the telemetry service.
+The error schema looks like this:
+
+```ts
+export type ErrorDetails = {
+  name: string;
+  message?: string; // anonymized error message
+  trace?: string; // anonymized stack trace
+  logs?: string; // anonymized stack logs
```
mrgrain: These probably mean slightly different things
```suggestion
  message?: string; // sanitized error message
  trace?: string; // sanitized stack trace
  logs?: string; // sanitized stack logs
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+
+The error name is an enum — currently either `ToolkitError`, `AuthorizationError`, `AssemblyError`, etc.
+
+The error message, trace, and logs could contain customer input.
```
mrgrain: don't create a new term if we have one already
```suggestion
The error message, trace, and logs could contain customer content.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+An ARN is a string that is prefixed by `arn:`. `arn:aws:s3:::my-bucket` will turn into `<redacted-arn>`.
+
+#### AccountIDs
+
+Anything that is a 12 digit number (that wasn’t already redacted within an ARN) will be redacted. `123456789012` will turn into `<redacted-account-id>`.
+
+#### File Paths
+
+Part of the file path (before node_modules) is customer content that we do not need to store.
```
mrgrain: after `node_modules` can also be customer content, e.g. private packages

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+We will use a regex that includes the following to find and replace file path prefixes:
+
+```
+(file:/+)? -> matches optional file url prefix
+homedir()/process.cwd() -> users home directory or current working directory, replacing \ with /
+[\\w.\\-_@\\\\/]+ -> matches nested directories and file name
```
mrgrain: do these regex consider windows?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+Anything that is a 12 digit number (that wasn’t already redacted within an ARN) will be redacted. `123456789012` will turn into `<redacted-account-id>`.
+
+#### File Paths
+
+Part of the file path (before node_modules) is customer content that we do not need to store.
+We will use a regex that includes the following to find and replace file path prefixes:
+
+```
+(file:/+)? -> matches optional file url prefix
+homedir()/process.cwd() -> users home directory or current working directory, replacing \ with /
```
mrgrain:
```suggestion
homedir()/process.cwd() -> users home directory or current working directory, replacing \ with /
```

Or just replace them with `$HOME` or `$PWD` respectively. Might be less confusing given that `/` is also the root of the FS.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+For example, a file path like `~/MyUser/my/path/to/node_modules/cdk-assets/lib/asset-manifest.js:20:19` will be truncated to `node_modules/cdk-assets/lib/asset-manifest.js:20:19`.
+
+#### Stack Names
+
+Stack Names are arbitrary customer content that we will redact.
+The CDK CLI will know the names of all the Stack Names when the Cloud Assembly is synthesized.
+That means that, for every command that undergoes synthesis, we will have all Stack IDs available.
+The CLI will collect these names and then find and redact them in the error message, stack trace, and logs.
+
+The commands that undergo synthesis are `cdk list`, `cdk diff`, `cdk synth`, and `cdk deploy`.
```
mrgrain: watch as well

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+For example, a file path like `~/MyUser/my/path/to/node_modules/cdk-assets/lib/asset-manifest.js:20:19` will be truncated to `node_modules/cdk-assets/lib/asset-manifest.js:20:19`.
+
+#### Stack Names
+
+Stack Names are arbitrary customer content that we will redact.
+The CDK CLI will know the names of all the Stack Names when the Cloud Assembly is synthesized.
+That means that, for every command that undergoes synthesis, we will have all Stack IDs available.
+The CLI will collect these names and then find and redact them in the error message, stack trace, and logs.
```
mrgrain: Seems less important how exactly we are getting the info.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+The commands that undergo synthesis are `cdk list`, `cdk diff`, `cdk synth`, and `cdk deploy`.
+**For all other commands, since we do not know what could be the Stack Name, we will not collect error messages, traces, or logs.**
+
+For example, the stack name might show up in the following string: `'cdk.out/MyStackName.assets.json'`.
+Once we have synthesized, we will know that MyStackName is a potential stack name and redact.
+The string will become `cdk.out/<redacted-stack-name>.assets.json`.
```
mrgrain: are there potentially multiple versions of this? ID vs name vs nested name etc. Or camelCase vs snake_case.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### Asset Display Names
+
+Asset Display Names are arbitrary customer content that we will redact.
+The CDK CLI will know the names of the Asset Display Names in the `.assets.json`
+file after the Cloud Assembly is synthesized.
+Similar to Stack Names, we will collect and redact these Asset Names and
```
mrgrain:
```suggestion
Similar to Stack Names, we will collect and redact these Asset Names in <insert where>
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
+# CDK CLI Telemetry
+
+* **Original Author(s):**: @conroyka
+* **Tracking Issue**: #732
+* **API Bar Raiser**: @iliapolo
+
+The CDK CLI team currently has no visibility into how the CLI is performing for users.
```
iliapolo: Not entirely true, we **do** have some visibility. 

```suggestion
The CDK CLI team currently has limited visibility into how the CLI is performing for users.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+The purpose of this RFC is to iterate in public. We welcome community feedback on the opt-out mechanism,
+on whether you would decide to opt-out,
+or on additional telemetry values that we are missing that have proven helpful in other CLIs.
+
+## Data Collection
+
+Before we start, two definitions that will help guide this RFC:
+
+* A **session** corresponds to a CLI command execution.
+* An **event** is a collection of metrics related to its **session**.
```
iliapolo: I think it would be more accurate to say:

```suggestion
* An **event** is a specific _action_ taking place within a **session**.
* A **metric** is a specific measure on how the _action_ performed.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+A **session** has multiple **events**, such as the case with `cdk watch`.
+`cdk watch` is a long-running **session** that can invoke the `synth` and `deploy` **events** multiple times.
+
+At the end of each **event**, we will send data to the telemetry service.
```
iliapolo: A few more details. Also - we decide we are going to batch events. 

```suggestion
A **session** can have multiple **events**, and an **event** can have multiple **metrics**. Such as the case with `cdk watch`, which is a long-running **session** that contains the `synth` and `deploy` **events** multiple times.

The CLI will send events to the telemetry service in batches
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+For more information on what gets redacted, see [Redacting Customer Content](#redacting-customer-content).
+
+```json
+{
+  "identifiers": {
+    "sessionId": "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9", // Uuid generated on each CLI command invocation
+    "eventId": "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9:1", // sessionId + an increment for each additional event in the session
+    "installationId": "3F1FD23A-58A9-4C0D-8A82-098D6101B322", // Uuid stored on a local file on the developer"s machine
+    "accountId": "d445c5a1b10d4f9c90a4b17769aa84d2e5d5c3da642c4acd392c71a46275e6f9", // optional hash of account ID
+    "region": "us-east-1", // optional region being deployed to
```
iliapolo, line 41: I believe we said we aren't going to collect these - no?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+## Opt-Out
+
+Collecting anonymous data will help the CDK team on two fronts: usage metrics and error debugging.
+Both of these aspects will help us better serve CDK customers in the future.
+We will ensure that we are not collecting [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1)
+or personally-identifiable information.
+
+You can easily opt-out of sending telemetry data.
```
iliapolo: "easy" is a relative term. 

```suggestion
You can opt-out of sending telemetry data.
```

text/0732-cdk-cli-telemetry.md
iliapolo: I agree that "telemetry" can be confused with the metadata resource, but I think "cli-telemetry" is good. If we do end up moving it to the toolkit, we will essentially have two telemetry systems:

- cli-telemetry
- toolkit-telemetry

Some telemetry will have to remain in the CLI because the toolkit alone will not have knowledge of what CLI command invoked it. 

Also if we move it, than we can support two keys if we want - I prefer clarity right now.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+Setting `toolkit-telemetry: false` in the CDK App level `cdk.json` will disable telemetry in your CDK App.
+Setting `toolkit-telemetry: false` in the `~/.cdk.json` file will disable telemetry across all CDK Apps on the same machine.
+
+We will also respect an environment variable, `CDK_DISABLE_TOOLKIT_TELEMETRY=true`.
+If set, this will achieve the same effect of disabling toolkit telemetry in that environment.
+
+Alternatively, you can run the CLI command (which disables toolkit telemetry for the CDK App):
+
+```bash
+> cdk telemetry --disable toolkit
```
iliapolo: Make a note that this will record an entry into the `cdk.json`. Begs the question - which one? the app specific or the global? need to support both probably. 

Also the command ergonomics is a bit strange to me. It relies on an _option_ to indicate the **action**. Normally options are just, well, options. For example, what would running this mean?

```bash
cdk telemetry
```

?

And what other options do we have apart from `toolkit`? 

```suggestion
> cdk toolkit-telemetry disable/enable --global/--local
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+> cdk bootstrap --bootstrap-bucket-name=MyBucket
+```
+
+Since `MyBucket` is free text, it will be redacted in the telemetry data sent to the service. This input will be processed into the following command object:
+
+```json
+{
+  "command": {
+    "path": ["cdk", "bootstrap"],
+    "parameters": {"bootstrap-bucket-name": "true" },
```
rix0rrr, line 186: I don't enjoy this mix of `string[]` and `{[ key: string]: string }`. Is `path` ever going to contain more than 2 strings, the first of which is going to be `cdk` ? 

I do see the advantage of having `parameters` (`options`?) as a dictionary, because the alternative would be to have to parse `['cdk', 'bootstrap', '--bootstrap-bucket-name=***']`, with and without the `=`, with the short flag alias, and it wouldn't even be parseable if it came from `cdk.json`.

Also, is `"true"` really a string here? Do `"***"` or `"present"` or just `true` ?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+```bash
+> cdk deploy MyStack
+```
+
+`MyStack` is free text, and also redacted. This command input turns into the following telemetry command:
+
+```json
+{
+  "command": {
+    "path": ["cdk", "deploy", "<redacted-stack-name>"],
```
rix0rrr: Why `"true"` for an option but `"<redacted-stack-name>"` for a stack name?

Are we going to have to whatever type of object every positional parameter could refer to, so that we can call it a "stack" name?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+The error name is an enum — currently either `ToolkitError`, `AuthorizationError`, `AssemblyError`, etc.
+
+The error message, trace, and logs could contain customer input.
+We have compiled the following list of possible customer inputs found in these strings and what we will do to redact those inputs:
+
+#### ARNs
+
+An ARN is a string that is prefixed by `arn:`. `arn:aws:s3:::my-bucket` will turn into `<redacted-arn>`.
```
rix0rrr: Do we have to annotate for every string whether it's an ARN or not, so that we can turn it into `<redacted-arn>` ?

(Mmm I suppose we could parse all strings and see if they start with `arn:`...)

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### AccountIDs
+
+Anything that is a 12 digit number (that wasn’t already redacted within an ARN) will be redacted. `123456789012` will turn into `<redacted-account-id>`.
```
rix0rrr: I thought we should be doing what GitHub does: 

```ts
const accountId = this.accountId(...);
output.markAsSecret(accountId);  // Redact the occurrence of this anywhere in printed logs
```

I also see the value of flat out redacting all account IDs, and it *is* simpler...

text/0732-cdk-cli-telemetry.md
rix0rrr: If we see a string named `cdk.out/<redacted-stack-name>.<redacted-stack-name>.json` we will know that the user called their stack `"assets"` 🥸

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### Asset Display Names
+
+Asset Display Names are arbitrary customer content that we will redact.
+The CDK CLI will know the names of the Asset Display Names in the `.assets.json`
+file after the Cloud Assembly is synthesized.
+Similar to Stack Names, we will collect and redact these Asset Names and
+**will not collect error messages, traces, or logs when the cloud assembly is unavailable.**
+
+A log message like `start: Building TelemetryFunction/Code` becomes `start: Building <redacted-asset-name>`.
```
rix0rrr: Still don't understand why we wouldn't replace with `****`.

For something more elaborate: it would be nice if we would replace redacted strings with symbolic names. `$ACCOUNT1`, `$ACCOUNT2`, `$ASSET1`, etc. That way, if there are multiple entities of the same type, we can distinguish them, and if the same entity gets referenced multiple times in the same log, we can cross-reference the occurrences.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+
+There is an equivalent `--enable` command to re-enable telemetry if necessary.
+
+## Metrics
+
+Below is an exhaustive list of metrics we will collect, with the reason for each.
+It’s important to note that all data we collect will be anonymous.
+We are only collecting enumerable values derived from customer input.
+
+The only exception is error reporting. More information on that can be found in [Errors](#sanitizing-errors)
```
iliapolo: Makes it sound like errors may be not anonymous, which is not true. I think you can leave this out from this section and just stick to the table, then explain all of this in the sanitation section.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| CDK CLI Version              | CLI version (x.y.z)                                                                                                          | Collecting this will help us bisect our data by version, which can help us identify problematic versions.                                                                                                                                                                                                       |
+| Installation ID              | Uuid stored on local machines executing CLI commands                                                                         | This serves as an alternative to Account ID to help us get a sense of blast radius. Each unique machine will have their own Installation ID, but since multiple different machines can target the same AWS Account this counts a slightly different idea of a 'user'.                                           |
```
iliapolo: I wouldn't relate this the Account ID - which is also not an **exact** representation of distinct users. 

```suggestion
| Installation ID              | UUID stored on local machines executing CLI commands                                                                         | This helps us estimate the distinct number of users who are either 1) affected by an error, or 2) use a specific CLI feature. Each unique machine will have their own Installation ID.                                           |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Hashed Account ID (optional) | Account ID, hashed to be anonymous                                                                                           | This will help us get a sense of blast radius when regressions are identified.                                                                                                                                                                                                                                  |
```
iliapolo: Is this still relevant?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Timestamp                    | The time (generated by the client) the data is sent to our telemetry endpoint                                                | This will help us generate time series graphs.                                                                                                                                                                                                                                                                  |
+| Event State                  | The "result" of the event. Possible values are 'SUCCESS', 'FAILURE', 'ABORTED'                                               | This will help us track error rates on each CLI command and alert us to potential regressions before they are reported on GitHub.                                                                                                                                                                               |
```
iliapolo:
```suggestion
| Event State                  | The "result" of the event. Possible values are 'SUCCESS', 'FAILURE', 'ABORTED'                                               | This will help us track error rates on each CLI command and alert us to potential regressions before they are reported by customers.                                                                                                                                                                               |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Event Type                   | An identifier for the type of event that produces the data                                                                   | Since we aim to send metrics on a per-event basis, and there could be multiple events in a command execution, this is an enum that identifies what kind of event transpired. Values could be 'synth', 'deploy', etc..                                                                                           |
+| Command Path                 | The command and properties entered into the CLI, with any free text redacted                                                 | For example, `cdk deploy myStack` would be turned into `['cdk', 'deploy', '<REDACTED>']` in the telemetry data. We are committed to redacting any customer content, but this information is critical to determine feature usage and will help us make data-driven decisions on what CLI features to prioritize. |
```
iliapolo:
```suggestion
| Command Path                 | The command and properties entered into the CLI, with any free text redacted                                                 | For example, `cdk deploy myStack` would be turned into `['cdk', 'deploy', '<REDACTED>']` in the telemetry data. We are committed to redacting any customer content. This information is critical to determine feature usage and will help us make data-driven decisions on what CLI features to prioritize. |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Command Parameters           | The optional parameters entered into the CLI, with any free text redacted                                                   | Some of our features include parameters, like `cdk deploy --watch`. Again, this information is critical to determine fetaure usage and for us to make data-driven decisions.                                                                                                                                    |
```
iliapolo: No need to assume an order of read. 

```suggestion
| Command Parameters           | The optional parameters entered into the CLI, with any free text redacted                                                   | Some of our features include parameters, like `cdk deploy --watch`. This information is critical to determine fetaure usage and for us to make data-driven decisions.                                                                                                                                    |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Operating System             | The operating system that is being used                                                                                      | This will help us debug issues that only affect specific operating systems.                                                                                                                                                                                                                                     |
+| CI/CD                        | Whether or not the CLI command is being invoked from a Ci/CD environment                                                     | Data points have a different relevance if they are from a developer's machine for from a CI/CD environment.                                                                                                                                                                                                     |
```
iliapolo:
```suggestion
| CI/CD                        | Whether or not the CLI command is being invoked from a Ci/CD environment                                                     | Helps us prioritize CLI features that better facilitate CI/CD environments.                                                                                                                                                                                                     |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Node Version                 | The node version used in the environment                                                                                     | This will help us debug any node-specific issues that arise.                                                                                                                                                                                                                                                    |
+| Duration Stats               | The length of the event (and potentially, any smaller denominations that are interesting)                                    | We will monitor the duration of events like (but not limited to) synthesis, deploy, etc..                                                                                                                                                                                                                       |
```
iliapolo:
```suggestion
| Timers               | The length of the event (and potentially, any smaller denominations that are interesting)                                    | We will monitor the duration of events like (but not limited to) synthesis, deploy, etc..                                                                                                                                                                                                                       |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Counting Stats               | Various counter metrics derived from your CDK app and CLI execution                                                          | We will monitor how counting stats like (but not limited to) CFN resource count affect the duration of deploy.                                                                                                                                                                                                  |
```
iliapolo:
```suggestion
| Counters               | Various counter metrics derived from your CDK app and CLI execution                                                          | We will monitor how counting stats like (but not limited to) CFN resource count affect the duration of deploy.                                                                                                                                                                                                  |
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Dependencies                 | Relevant AWS dependencies and their versions                                                                                 | We plan to track AWS dependencies that might affect the CLI result, like `aws-cdk-lib`, `jsii`, `projen`, etc..                                                                                                                                                                                                     |
+| Error Messages (optional)    | The error message returned, if applicable. Customer content redacted                                                         | Capturing the error message will help us aggregate data on which errors are encountered at a greater rate, as well as help us debug what went wrong in individual use cases.                                                                                                                                    |
```
iliapolo: How is this optional? can a user opt-out of this or is it optional in that "is only sent when a command fails"? Elaborate on this.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Error Stack Trace (optional) | The stack trace of the error message, if applicable. Customer content redacted                                               | The stack trace will be helpful for individual debugging purposes and is necessary for us to be able to reproduce issues that may arise.                                                                                                                                                                        |
```
iliapolo: Same

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+| Error Logs (optional)        | The logs of a failed CLI command, if applicable. Customer content redacted                                                   | Error logs will also help us debug and reproduce issues that we see in the CLI.                                                                                                                                                                                                                                 |
```
iliapolo: Same

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+{
+  "command": {
+    "path": ["cdk", "deploy", "<redacted-stack-name>"],
+  }
+}
+```
+
+### Sanitizing Errors
+
+A core goal of the CDK CLI Telemetry Service is to provide a more efficient debugging experience for both the CDK core team and customers.
```
iliapolo, line 212: How does telemetry help **customers** debug error?

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,272 @@
...
+This means that instead of back-and-forth on GitHub issues asking for more information — logs, stack trace, etc.
+
```
iliapolo: It will help the team with:

- Identify and resolve issues prior to a customer report.
- Reduce the back-and-forth often needed to resolve an existing customer report.

text/0732-cdk-cli-telemetry.md
iliapolo: True, maybe we include only known paths then? 

- `aws-cdk/**`
- `@aws-cdk/toolkit-lib/**`
- `cdk-assets/**`
- ...

iliapolo: Actually since the CLI is bundled, we probably only need `aws-cdk/**`.

text/0732-cdk-cli-telemetry.md
iliapolo: I think its good to surface the HOW, for transparency and increased customer trust.

text/0732-cdk-cli-telemetry.md
kaizencc: > is `path` ever going to contain more than 2 strings?

contains args too. `cdk deploy Mystack1 Mystack2` generates `['cdk', 'deploy', '<redacted>', '<redacted']`. basically things that can _only_ be supplied via CLI, as options can come from CLI or cdk.json so they go into parameters.

parameters is a dictionary to mirror what we get out of yargs (which takes care of the ` ` vs. `=` conundrum)

> is `"true"` really a string here?

it is a boolean i suppose. supplying a string will turn it into a `true` boolean for us.

text/0732-cdk-cli-telemetry.md
kaizencc: yes the CLI input is strongly typed; we have the information of what the arguments could be (in a correct CLI invocation).

text/0732-cdk-cli-telemetry.md
kaizencc: we will be parsing all strings for `arn:` :)

text/0732-cdk-cli-telemetry.md
kaizencc: is accountId a spooky secret though? or any different than the other things we're redacting? its just customer content. i think we should just go through the simple implementation of just redacting them all.

text/0732-cdk-cli-telemetry.md
kaizencc: We said we are going to collect fully anonymous data. Hashed account IDs and regions are part of that contract in my mind, do you disagree?

text/0732-cdk-cli-telemetry.md
kaizencc: Yes, our current proposal is to include hashed account IDs

text/0732-cdk-cli-telemetry.md
kaizencc: Not optional so i've changed. i mean to say only if an error occurs.

text/0732-cdk-cli-telemetry.md
kaizencc: In addition, its helpful to understand why some errors will be reported and others will be omitted.

text/0732-cdk-cli-telemetry.md
kaizencc: Looking at those regex patterns, they have partial Windows compatibility but aren't fully Windows-aware:

What works for Windows:
• (file:/+)? - This works fine on Windows since file:// URLs use forward slashes regardless of OS
• The homedir()/process.cwd() logic with \ → / replacement is good - normalizing to forward slashes works across platforms
• [\\w.\\-_@\\\\/]+ includes \\ (backslash) in the character class, so it can match Windows paths

Potential Windows issues:
1. Drive letters: The regex doesn't explicitly handle C:, D:, etc. at the start of absolute paths
2. UNC paths: No support for \\server\share network paths  
3. Path separators: While backslashes are included in the character class, there's no logic to handle mixed separators or 
validate proper Windows path structure

For better Windows support, consider:
```
regex
# Handle drive letters
^([A-Za-z]:)?

# Handle UNC paths  
^(\\\\[^\\]+\\[^\\]+)?

# More explicit path separator handling
[/\\]
```

The current patterns will work for many Windows scenarios, especially with the backslash-to-forward-slash normalization, but
they're not comprehensive for all Windows path edge cases. For a CDK context where cross-platform file handling is 
important, you might want to enhance them further.

kaizencc: I think the answer is yes, it considers windows and is probably good enough for our use case. They are copied from amplify, and amplify also has windows users.

text/0732-cdk-cli-telemetry.md
kaizencc: Landing on `cli-telemetry` everywhere

text/0732-cdk-cli-telemetry.md
kaizencc: I can see a potential problem with multiple versions, but also think it is ok. Specifically, there's `stack.displayName` and `stack.stackName`. We are supposed to standardize on `displayName` as output but I guess that may not be enforced.

`stackName` is 
```ts
this.stackName = properties.stackName || artifactId;
```
`displayName` is 
```ts
this.displayName = this.stackName === artifactId
      ? this.hierarchicalId
      : `${this.hierarchicalId} (${this.stackName})`;
```

`hierarchicalId` is
```ts
this.manifest.displayName ?? this.id;
```

That is to say, the `displayName` certainly includes the `stackName` but could have additional information (`hierarchicalId`).

From the cloud assembly we can go into the `manifest.json` file and find the `displayName` of all stacks. This is guaranteed to include the `stackName`, and we can separate it out as well (it's either equal to the `displayName` or inside parentheses). We can then sanitize outputs for both `displayName` _and_ `stackName` (if they are different).

---

All of this probably needs more pairs of eyes to make sure it's ok. But to summarize:

We get `displayName` from the cloud assembly.
We determine what the `stackName` is from the `displayName`.
We sanitize for both, if they are different.

text/0732-cdk-cli-telemetry.md
kaizencc: > Reduce the back-and-forth often needed to resolve an existing customer report.

this should benefit our customers too

text/0732-cdk-cli-telemetry.md
kaizencc: I'll include this for now but we don't have any subcommands in the CLI today and I wonder if that might involve additional work that's not worth it over options `--disable/--enable`

or `cdk enable-cli-telemetry` / `cdk disable-cli-telemetry`

text/0732-cdk-cli-telemetry.md
kaizencc: Good idea. For assets and stacks, I can do this since I'm getting the names from the cloud assembly. For accounts and arns, it's more of a blanket sanitation and i would have to store information on which values I've found in order to count them `$ACCOUNT1` `$ACCOUNT2`, etc. read that as more work than I want to commit to right now, and we can start with `$STACK1` and `$ASSET1` but no counts for other things

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,285 @@
...
+    "installationId": "3F1FD23A-58A9-4C0D-8A82-098D6101B322", // UUID stored on a local file on the developer"s machine
+    "accountId": "d445c5a1b10d4f9c90a4b17769aa84d2e5d5c3da642c4acd392c71a46275e6f9", // optional hash of account ID
+    "region": "us-east-1", // optional region being deployed to
+  },
+  "event": {
+    "state": "SUCCESS",
+    "eventType": "deploy",
+    "command": {
+      "path": ["cdk", "deploy", "$STACK1"],
+      "parameters": {},
```
iankhou, line 48: Should this be a dictionary or an array?

kaizencc: i think they should be dictionary actually. @iankhou sorry to change up the schema slightly but it makes more sense that way imo

iankhou: Yeah makes sense to me!

iliapolo:
Removing approval as there are some additional topics that were surfaced.

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,301 @@
...
+
+```json
+{
+  "command": {
+    "path": ["cdk", "deploy", "$STACK1"],
+  }
+}
+```
+
+### Errors are sanitized
```
iliapolo:
```suggestion
### Errors & Logs are sanitized
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,301 @@
...
+  name: string;
+  message?: string; // sanitized error message
+  stackTrace?: string; // sanitized stack trace
+  logs?: string; // sanitized stack logs
+}
+```
+
+The error name is an enum — currently either `ToolkitError`, `AuthorizationError`, `AssemblyError`, etc.
+
+Below we show how we sanitize possible strings.
```
iliapolo:
```suggestion
Below we show what exactly is being sanitized.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,301 @@
...
+#### UUIDs
+
+Sometimes, a log may include information from an SDK call like this:
+
+```
+(Service: Lambda, Status Code: 400, Request ID: c99252d8-xxxx-xxxx-xxxx-xxxxxxxxxxxx) (SDK Attempt Count: 1)"
+(RequestToken: ab70879f-xxxx-xxxx-xxxx-xxxxxxxxxxxx, HandlerErrorCode: InvalidRequest)
+```
+
+We'll replace UUIDs with `$UUID` by redacting strings that follow the UUID pattern:
```
iliapolo:
```suggestion
We'll replace UUIDs with `$UUID`:
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -0,0 +1,301 @@
...
+
+```
+(Service: Lambda, Status Code: 400, Request ID: $UUID) (SDK Attempt Count: 1)" (RequestToken: $UUID, HandlerErrorCode: InvalidRequest)
+```
+
+#### File Paths
+
+Part of the file path (before `aws-cdk/**`) is customer content that will not record.
```
iliapolo:
```suggestion
Part of the file path (before `aws-cdk/**`) is customer content that we will not record.
```

Issue #750, opened 2025-07-03
Feature Flag Advisor


vivian12300:
## Description

Create a CLI tool for users to modify their feature flag configurations.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @rix0rrr 
| Author(s)           | @vivian12300
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | @rix0rrr

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `status/api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #751, opened 2025-07-03
RFC750: Feature Flag CLI Tool 

vivian12300:
This is a request for comments for the feature flag CLI tool. See #750 for additional details. 

---

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -1,12 +1,12 @@
 name: PR Lint
 
 on:
-  pull_request_target: {}
+  pull_request: {}
   workflow_dispatch: {}
 
 permissions:
   contents: read
-  
+
 jobs:
   markdownlint:
     runs-on: ubuntu-latest
@@ -17,8 +17,7 @@ jobs:
           ref: "${{ github.event.pull_request.merge_commit_sha }}"
           path: checkout_pr
 
-      # Checkout of tools from 'main'; necessary to avoid compromise of GitHub token
-      # (due to pull_request_target)
+      # Checkout of tools from 'main'
       - uses: actions/checkout@v4
         with:
           ref: main
--- a/ACCEPTED.md
+++ b/ACCEPTED.md
@@ -13,7 +13,7 @@
 [162](https://github.com/aws/aws-cdk-rfcs/issues/162)|[CDK Refactoring Tools](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0162-refactoring-support.md)|[@otaviomacedo](https://github.com/otaviomacedo)|👷 implementing
 [670](https://github.com/aws/aws-cdk-rfcs/issues/670)|[AWS CloudWatch Application Signals L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0670-aws-applicationsignals-enablement-l2.md)||👷 implementing
 [673](https://github.com/aws/aws-cdk-rfcs/issues/673)|[AWS CloudWatch Application Signals L2 Constructs for SLO](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0673-aws-applicationsignals-slo-l2.md)||👷 implementing
-[686](https://github.com/aws/aws-cdk-rfcs/issues/686)|[L2 Constructs for Bedrock](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0686-bedrock-l2-construct.md)|[@dineshSajwan](https://github.com/dineshSajwan)|👷 implementing
+[732](https://github.com/aws/aws-cdk-rfcs/issues/732)|[CDK CLI Telemetry](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0732-cdk-cli-telemetry.md)|[@kaizencc](https://github.com/kaizencc)|👷 implementing
 [1](https://github.com/aws/aws-cdk-rfcs/issues/1)|[CDK Watch](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0001-cdk-update.md)||✅ done
 [4](https://github.com/aws/aws-cdk-rfcs/issues/4)|[CDK Testing Toolkit](https://github.com/aws/aws-cdk-rfcs/issues/4)|[@nija-at](https://github.com/nija-at)|✅ done
 [5](https://github.com/aws/aws-cdk-rfcs/issues/5)|[Security-restricted environments](https://github.com/aws/aws-cdk-rfcs/issues/5)||✅ done
@@ -73,6 +73,7 @@
 [605](https://github.com/aws/aws-cdk-rfcs/issues/605)|[Rewrite EKS L2 construct (EKSv2)](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0605-eks-rewrite.md)||✅ done
 [617](https://github.com/aws/aws-cdk-rfcs/issues/617)|[Amazon CloudFront Origin Access Control L2 Construct](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0617-cloudfront-oac-l2.md)||✅ done
 [648](https://github.com/aws/aws-cdk-rfcs/issues/648)|[Priority-Ordered Aspect Invocation](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0648-aspects-priority-ordering.md)|[@sumupitchayan](https://github.com/sumupitchayan)|✅ done
+[686](https://github.com/aws/aws-cdk-rfcs/issues/686)|[L2 Constructs for Bedrock](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0686-bedrock-l2-construct.md)|[@dineshSajwan](https://github.com/dineshSajwan)|✅ done
 [693](https://github.com/aws/aws-cdk-rfcs/issues/693)|[Set Default Values to Construct Properties using PropertyInjector](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0693-property-injection.md)||✅ done
 [710](https://github.com/aws/aws-cdk-rfcs/issues/710)|[Node.js Version Support Policy for AWS CDK](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0710-node-deprecation-strategy.md)||✅ done
 <!--END_TABLE-->
--- a/CLOSED.md
+++ b/CLOSED.md
@@ -30,6 +30,7 @@
 [434](https://github.com/aws/aws-cdk-rfcs/issues/434)|[AWS Ground Station L2 Constructs](https://github.com/aws/aws-cdk-rfcs/issues/434)||👎 rejected
 [446](https://github.com/aws/aws-cdk-rfcs/issues/446)|[Network Firewall L2 Constructs](https://github.com/aws/aws-cdk-rfcs/issues/446)||👎 rejected
 [448](https://github.com/aws/aws-cdk-rfcs/issues/448)|[AWS Compute Optimizer Constructs](https://github.com/aws/aws-cdk-rfcs/issues/448)||👎 rejected
+[456](https://github.com/aws/aws-cdk-rfcs/issues/456)|[L2 ElastiCache support](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0456-elasticache-l2.md)||👎 rejected
 [457](https://github.com/aws/aws-cdk-rfcs/issues/457)|[Create fluent-assertions library to improve consumer test readability](https://github.com/aws/aws-cdk-rfcs/issues/457)|[@MrArnoldPalmer](https://github.com/MrArnoldPalmer)|👎 rejected
 [458](https://github.com/aws/aws-cdk-rfcs/issues/458)|[Service Catalog ProductStack Asset Support](https://github.com/aws/aws-cdk-rfcs/issues/458)||👎 rejected
 [465](https://github.com/aws/aws-cdk-rfcs/issues/465)|[AWS Organizations L2 Construct](https://github.com/aws/aws-cdk-rfcs/issues/465)||👎 rejected
@@ -47,6 +48,7 @@
 [528](https://github.com/aws/aws-cdk-rfcs/issues/528)|[Verified Permissions L2 Constructs](https://github.com/aws/aws-cdk-rfcs/issues/528)||👎 rejected
 [579](https://github.com/aws/aws-cdk-rfcs/issues/579)|[Docker-less dependency resolution (CDK \ SAM)](https://github.com/aws/aws-cdk-rfcs/issues/579)||👎 rejected
 [587](https://github.com/aws/aws-cdk-rfcs/issues/587)|[Amazon Lambda Rust library](https://github.com/aws/aws-cdk-rfcs/issues/587)||👎 rejected
+[715](https://github.com/aws/aws-cdk-rfcs/issues/715)|[CDK Construct Update Support](https://github.com/aws/aws-cdk-rfcs/issues/715)||👎 rejected
 [9](https://github.com/aws/aws-cdk-rfcs/issues/9)|[Master developer guide sources in main repo](https://github.com/aws/aws-cdk-rfcs/issues/9)||🤷 stale
 [10](https://github.com/aws/aws-cdk-rfcs/issues/10)|[New workshop modules](https://github.com/aws/aws-cdk-rfcs/issues/10)||🤷 stale
 [13](https://github.com/aws/aws-cdk-rfcs/issues/13)|[Improvements to Reference docs](https://github.com/aws/aws-cdk-rfcs/issues/13)||🤷 stale
@@ -105,4 +107,5 @@
 [471](https://github.com/aws/aws-cdk-rfcs/issues/471)|[Amazon Managed Grafana L2 Constructs](https://github.com/aws/aws-cdk-rfcs/issues/471)||🤷 stale
 [581](https://github.com/aws/aws-cdk-rfcs/issues/581)|[ARN retrieval function consistency for CDK constructs](https://github.com/aws/aws-cdk-rfcs/issues/581)||🤷 stale
 [586](https://github.com/aws/aws-cdk-rfcs/issues/586)|[Understand Deployment Progress](https://github.com/aws/aws-cdk-rfcs/issues/586)||🤷 stale
+[635](https://github.com/aws/aws-cdk-rfcs/issues/635)|[L2 Constructs for AWS MediaConvert](https://github.com/aws/aws-cdk-rfcs/issues/635)||🤷 stale
 <!--END_TABLE-->
--- a/FULL_INDEX.md
+++ b/FULL_INDEX.md
@@ -133,6 +133,7 @@
 [446](https://github.com/aws/aws-cdk-rfcs/issues/446)|[Network Firewall L2 Constructs](https://github.com/aws/aws-cdk-rfcs/issues/446)||👎 rejected
 [448](https://github.com/aws/aws-cdk-rfcs/issues/448)|[AWS Compute Optimizer Constructs](https://github.com/aws/aws-cdk-rfcs/issues/448)||👎 rejected
 [450](https://github.com/aws/aws-cdk-rfcs/issues/450)|[AWS CDK public roadmap](https://github.com/aws/aws-cdk-rfcs/issues/450)||🤷 stale
+[456](https://github.com/aws/aws-cdk-rfcs/issues/456)|[L2 ElastiCache support](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0456-elasticache-l2.md)||👎 rejected
 [457](https://github.com/aws/aws-cdk-rfcs/issues/457)|[Create fluent-assertions library to improve consumer test readability](https://github.com/aws/aws-cdk-rfcs/issues/457)|[@MrArnoldPalmer](https://github.com/MrArnoldPalmer)|👎 rejected
 [458](https://github.com/aws/aws-cdk-rfcs/issues/458)|[Service Catalog ProductStack Asset Support](https://github.com/aws/aws-cdk-rfcs/issues/458)||👎 rejected
 [460](https://github.com/aws/aws-cdk-rfcs/issues/460)|[Reduce aws-cdk-lib package size](https://github.com/aws/aws-cdk-rfcs/issues/460)||✅ done
@@ -177,21 +178,25 @@
 [627](https://github.com/aws/aws-cdk-rfcs/issues/627)|[L2 constructs for CodeArtifact](https://github.com/aws/aws-cdk-rfcs/issues/627)||💡 proposed
 [629](https://github.com/aws/aws-cdk-rfcs/issues/629)|[Cost Estimation Tools - Reopened](https://github.com/aws/aws-cdk-rfcs/issues/629)||💡 proposed
 [631](https://github.com/aws/aws-cdk-rfcs/issues/631)|[Amazon VPC Lattice L2 Construct](https://github.com/aws/aws-cdk-rfcs/issues/631)||💡 proposed
-[635](https://github.com/aws/aws-cdk-rfcs/issues/635)|[L2 Constructs for AWS MediaConvert](https://github.com/aws/aws-cdk-rfcs/issues/635)||💡 proposed
+[635](https://github.com/aws/aws-cdk-rfcs/issues/635)|[L2 Constructs for AWS MediaConvert](https://github.com/aws/aws-cdk-rfcs/issues/635)||🤷 stale
 [648](https://github.com/aws/aws-cdk-rfcs/issues/648)|[Priority-Ordered Aspect Invocation](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0648-aspects-priority-ordering.md)|[@sumupitchayan](https://github.com/sumupitchayan)|✅ done
 [655](https://github.com/aws/aws-cdk-rfcs/issues/655)|[Enhancements to L1s in CDK](https://github.com/aws/aws-cdk-rfcs/issues/655)||💡 proposed
 [670](https://github.com/aws/aws-cdk-rfcs/issues/670)|[AWS CloudWatch Application Signals L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0670-aws-applicationsignals-enablement-l2.md)||👷 implementing
 [673](https://github.com/aws/aws-cdk-rfcs/issues/673)|[AWS CloudWatch Application Signals L2 Constructs for SLO](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0673-aws-applicationsignals-slo-l2.md)||👷 implementing
 [676](https://github.com/aws/aws-cdk-rfcs/issues/676)|[Proposing a Contributor Council for AWS CDK](https://github.com/aws/aws-cdk-rfcs/issues/676)||💡 proposed
 [683](https://github.com/aws/aws-cdk-rfcs/issues/683)|[L2 Constructs for AWS Event Schemas](https://github.com/aws/aws-cdk-rfcs/issues/683)||💡 proposed
-[686](https://github.com/aws/aws-cdk-rfcs/issues/686)|[L2 Constructs for Bedrock](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0686-bedrock-l2-construct.md)|[@dineshSajwan](https://github.com/dineshSajwan)|👷 implementing
+[686](https://github.com/aws/aws-cdk-rfcs/issues/686)|[L2 Constructs for Bedrock](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0686-bedrock-l2-construct.md)|[@dineshSajwan](https://github.com/dineshSajwan)|✅ done
 [693](https://github.com/aws/aws-cdk-rfcs/issues/693)|[Set Default Values to Construct Properties using PropertyInjector](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0693-property-injection.md)||✅ done
 [695](https://github.com/aws/aws-cdk-rfcs/issues/695)|[AWS S3 directory bucket L2 constructs](https://github.com/aws/aws-cdk-rfcs/issues/695)||💡 proposed
 [703](https://github.com/aws/aws-cdk-rfcs/issues/703)|[apigateway: private domain name L2 construct](https://github.com/aws/aws-cdk-rfcs/issues/703)||💡 proposed
 [707](https://github.com/aws/aws-cdk-rfcs/issues/707)|[CDK Pipelines: Use pipeline ServiceRole as default ActionRole](https://github.com/aws/aws-cdk-rfcs/issues/707)||💡 proposed
 [710](https://github.com/aws/aws-cdk-rfcs/issues/710)|[Node.js Version Support Policy for AWS CDK](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0710-node-deprecation-strategy.md)||✅ done
-[715](https://github.com/aws/aws-cdk-rfcs/issues/715)|[CDK Construct Update Support](https://github.com/aws/aws-cdk-rfcs/issues/715)||💡 proposed
+[715](https://github.com/aws/aws-cdk-rfcs/issues/715)|[CDK Construct Update Support](https://github.com/aws/aws-cdk-rfcs/issues/715)||👎 rejected
 [719](https://github.com/aws/aws-cdk-rfcs/issues/719)|[Native Container Image Tarball Asset Handling](https://github.com/aws/aws-cdk-rfcs/issues/719)||💡 proposed
 [728](https://github.com/aws/aws-cdk-rfcs/issues/728)|[L2 Constructs for AWS Elemental MediaPackageV2](https://github.com/aws/aws-cdk-rfcs/issues/728)||💡 proposed
-[732](https://github.com/aws/aws-cdk-rfcs/issues/732)|[CDK CLI Telemetry](https://github.com/aws/aws-cdk-rfcs/issues/732)|[@kaizencc](https://github.com/kaizencc)|✍️ review
+[732](https://github.com/aws/aws-cdk-rfcs/issues/732)|[CDK CLI Telemetry](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0732-cdk-cli-telemetry.md)|[@kaizencc](https://github.com/kaizencc)|👷 implementing
+[745](https://github.com/aws/aws-cdk-rfcs/issues/745)|[User-Provided cdk init Templates](https://github.com/aws/aws-cdk-rfcs/issues/745)||💡 proposed
+[750](https://github.com/aws/aws-cdk-rfcs/issues/750)|[Feature Flag Advisor](https://github.com/aws/aws-cdk-rfcs/issues/750)||💡 proposed
+[762](https://github.com/aws/aws-cdk-rfcs/issues/762)|[ElastiCache L2 Construct](https://github.com/aws/aws-cdk-rfcs/issues/762)||💡 proposed
+[764](https://github.com/aws/aws-cdk-rfcs/issues/764)|[L2 Constructs for Bedrock AgentCore Memory](https://github.com/aws/aws-cdk-rfcs/issues/764)||💡 proposed
 <!--END_TABLE-->
--- a/PROPOSED.md
+++ b/PROPOSED.md
@@ -21,15 +21,16 @@
 [627](https://github.com/aws/aws-cdk-rfcs/issues/627)|[L2 constructs for CodeArtifact](https://github.com/aws/aws-cdk-rfcs/issues/627)||💡 proposed
 [629](https://github.com/aws/aws-cdk-rfcs/issues/629)|[Cost Estimation Tools - Reopened](https://github.com/aws/aws-cdk-rfcs/issues/629)||💡 proposed
 [631](https://github.com/aws/aws-cdk-rfcs/issues/631)|[Amazon VPC Lattice L2 Construct](https://github.com/aws/aws-cdk-rfcs/issues/631)||💡 proposed
-[635](https://github.com/aws/aws-cdk-rfcs/issues/635)|[L2 Constructs for AWS MediaConvert](https://github.com/aws/aws-cdk-rfcs/issues/635)||💡 proposed
 [655](https://github.com/aws/aws-cdk-rfcs/issues/655)|[Enhancements to L1s in CDK](https://github.com/aws/aws-cdk-rfcs/issues/655)||💡 proposed
 [676](https://github.com/aws/aws-cdk-rfcs/issues/676)|[Proposing a Contributor Council for AWS CDK](https://github.com/aws/aws-cdk-rfcs/issues/676)||💡 proposed
 [683](https://github.com/aws/aws-cdk-rfcs/issues/683)|[L2 Constructs for AWS Event Schemas](https://github.com/aws/aws-cdk-rfcs/issues/683)||💡 proposed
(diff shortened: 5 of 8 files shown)


text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+
+* Original Author(s): [Vivian Chen](https://quip-amazon.com/GGQ9EAAQ3Jw)
+* Tracking Issue: N/A
+* API Bar Raiser: [Rico Huijbers](https://quip-amazon.com/YPP9EA8Gt6Q)
+
+The feature flag CLI tool will inform users of their current feature flag configuration and allow them to switch any/all of them to a different value.
+
+## Working Backwards
+
+### **Help:**
```
rix0rrr: Markdown nit: you don't need to combine `heading` and `bold` styles. Heading alone will suffice.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+The feature flag CLI tool will inform users of their current feature flag configuration and allow them to switch any/all of them to a different value.
+
+## Working Backwards
+
+### **Help:** 
+
+```
+> cdk flags --help
+cdk flags 
+Finds and displays a report of the feature flag configuration and compares users' current values with our recommended values.
```
rix0rrr: Nite: you are talking *to* the user at this point, so `user's` -> `your`.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+ 
+Options:
+    --switch-all          **safely** switch all feature flags to their recommended states
```
rix0rrr: "switch" also sounds slightly vague. 

`--make-all-recommended` ? Slightly long...

rix0rrr: Is there also an unsafe variant for this?

By the way, I think the use case for this operation is fringe. I think `--switch-unconfigured` and its converse, `--record-unconfigured` are the most useful ones.

I wonder if `--safe` should just be an additional flag. Or its inverse: `--allow-changes`.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+Finds and displays a report of the feature flag configuration and compares users' current values with our recommended values.
+ 
+Options:
+    --switch-all          **safely** switch all feature flags to their recommended states 
+    --switch-unconfigured **safely** switch unconfigured feature flags to their recommended states 
+    --toggle=string       toggle the state of a specific flag
+    
+Examples:
+    cdk flags
+    cdk flags --toggle="#FLAGNAME#"
```
rix0rrr: Do a real example here?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+
+The new CLI command `cdk flags` shows you a report comparing your current feature flag configurations and our recommended states. When you run `cdk flags`, a feature flags report is displayed to you in the following format: flags that match the recommended values are at the top of the list, followed by flags that do not match the recommended values and then lastly, flags that are not configured by the user at all. There will be a Menu with options for the user to configure their feature flags. 
+
+```
+> cdk flags 
+    Feature Flag                              Recommended                  User
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - (not configured)
```
rix0rrr: For `- (not configured)`, maybe a shorter variant is `<unset>`.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+  
+  MENU --------------------------------------------------
+  [1] Switch **all** flags to recommended states
+  [2] Switch all **unconfigured** flags to recommended states
+  [3] Switch all unconfigured flags to false state (no impact on application)
```
rix0rrr: "false state" is not really the recommended wording I don't think.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+
+```
+> cdk flags 
+    Feature Flag                              Recommended                  User
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - (not configured)
+  
+  MENU --------------------------------------------------
```
rix0rrr: I think we'd want the menu to look like https://www.npmjs.com/package/enquirer

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+    --toggle=string       toggle the state of a specific flag
+    
+Examples:
+    cdk flags
+    cdk flags --toggle="#FLAGNAME#"
+```
+
+### README:
+
+The new CLI command `cdk flags` shows you a report comparing your current feature flag configurations and our recommended states. When you run `cdk flags`, a feature flags report is displayed to you in the following format: flags that match the recommended values are at the top of the list, followed by flags that do not match the recommended values and then lastly, flags that are not configured by the user at all. There will be a Menu with options for the user to configure their feature flags.
```
rix0rrr: Maybe leave out the "correct" flags altogether by default, hide them behind a `--all/-a` flag ?

rix0rrr: Also describe expected use cases here:

* You can see all your flags.
* If you want to do X, run the command C.
* Afterwards, if you want to do Y, run the command D.
* To get more information on a flag, you can do ...

rix0rrr: Then proceed to show the screens and explain a bit more.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+-      `BucketEncryption``:`` ``None`
++ S3 Bucket:
++    Properties:
++      BucketEncryption: 
++        `ServerSideEncryptionConfiguration``:`
++          ...
+
+Would you like to accept these changes? (Accept/reject)
+> reject
+- S3 Bucket:                                   --
```
rix0rrr: We have a diff formatting library already that we'll probably be using.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
++ S3 Bucket:                             --
++    Properties:                          |    INCOMING VERSION
++      BucketEncryption: None            --
+Resynthesizing... 
+Done.
+```
+
+## What are we trying to solve?
+
+Feature flags are an important tool for adding new bug fixes without affecting the current state of the application. However, feature flags often go unnoticed by the user. In order to change their configurations, users are required to manually update their `cdk.json` file, resulting in a time-consuming task. They can impact a user in many ways. For example, the @aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021 flag changed the default behavior to use the updated version TLSv1.2_2021, which is a more secure and support TLS version for CloudFront. Users who stuck with the previous version may have been exposed to security vulnerabilities and failed compliance tests. To remain Backwards Compatible, newly released feature flags are automatically disabled for existing projects, meaning that those users are required to constantly keep an eye on CDK releases to manually update their feature flags.
```
rix0rrr: Honestly, a fraction of this can go into the introductory paragraph of the `flags` command as well. This is useful context! 

Something like (but not necessarily equal to)

> CDK uses feature flags to roll out changes that may affect existing infrastructure definitions. The `cdk flags` command will show you the available flags, their recommended values, and your currently configured values, along with commands to automatically toggle flags to their recommended states.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Why should we *not* do this?
+
+This tool aims to simplify the feature flag configuration file by asking users if they would like to make the changes from the CLI. However, this creates a risk of users applying our recommendations without fully understanding its impact on their application. For example, a user could unknowingly enable a recommended flag that impacts behavior a user is depending on. To ensure they are informed about these changes as much as possible, any change will result in a comparison of their CloudFormation template and their modified one.
+
+Adding a new CLI tool also increases the CDK CLI complexity, requiring more maintenance and support for this new feature. 
+
+## What alternative solutions did you consider?
+
+Instead of creating a Feature Flag CLI tool, we could expand on `cdk synth`. If a user’s feature flags differ from our recommended values, it would display a warning or suggestion. This reduces the need for another CLI command, but this method risks negatively impacting user experience due to cluttering the `cdk synth` output if many flags differ.
```
rix0rrr: Having `cdk synth` do something flag-related is 100% in the design here, and is not Future Work (and also not really an "alternative")

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+
+Feature flags contain bug fixes and improved default configurations that could potentially help your application. This CLI tool will make enabling feature flags easier and more efficient. 
+
+### Is this a breaking change?
+
+No. Users will be able to view the changes they are making and reject them if necessary. 
+
+### Why not read from `recommended-feature-flags.json` from `aws-cdk-lib` directly?
+
+This would not work for jsii projects as the `recommended-feature-flags.json` file would not transfer over to projects in other CDK-compatible languages.
```
rix0rrr: Yes but why?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+> cdk flags 
+    Feature Flag                              Recommended                  User
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - (not configured)
+  
+  MENU ----------------------------------------------------------------------
+  [1] Switch **all** flags to recommended states 
+  [2] Switch all **unconfigured** flags to recommended states
```
rix0rrr: I'm thinking through the user story here. Again, what is the interaction going to be if they pick this AND decide they don't like the outcome?

Do we say "revert all"?

That's fine, but what's the next step?

rix0rrr (changes requested):

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
+# RFC: Feature Flag Advisor
+
+* Original Author(s): [Vivian Chen](https://quip-amazon.com/GGQ9EAAQ3Jw)
```
otaviomacedo: Nit: use GitHub usernames.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,183 @@
...
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - (not configured)
+  
+  MENU --------------------------------------------------
+  [1] Switch **all** flags to recommended states
+  [2] Switch all **unconfigured** flags to recommended states
+  [3] Switch all unconfigured flags to false state (no impact on application)
+  [4] Toggle **one** flag
```
otaviomacedo: What's the experience if I choose this? We should probably have an arrow pointing to a flag, that you can move up and down. And then the other 3 numeric options.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+> cdk flags 
+    Feature Flag                              Recommended Value            User Value
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  
+  How would you like to proceed?
+   >  Set all flags to recommended values 
+      Set unconfigured flags to recommended values
+      Reset unconfigured flags to defaults (no impact)
```
rix0rrr: "defaults" is good, but it is also confusing. New default or old default? 

```suggestion
      Record current behavior as intentional for all unconfigured flags (no impact)
```

Or something like that... I'm having trouble coming up with better wording.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+      Modify individual flag 
+```
+
+### Expected Use Cases
+
+#### View information about your feature flags
+
+To view a report of all flags, users can run `cdk flags --all`. This will display a report of every feature flag in the following
```
rix0rrr: Can we also run it without `--all` ? What does that do?

----------------------------

> To view a report of all flags, **users can run** `cdk flags --all` (emphasis mine)

We are talking **to** the user in the RFC. So the phrasing should be:

> You can run `cdk flags --all` to see a report of all flags.

Or alternatively:

> If you want to see a report of all flags, run `cdk flags --all`.

Then read that again, and think if that makes sense to you, as a user.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+  * @aws-cdk/...                              true                         - <unset>
+  
+  How would you like to proceed?
+   >  Set all flags to recommended values
+      Set all unconfigured flags to recommended values
+      Reset all unconfigured flags to defaults (no impact)
+      Modify individual flag
+```
+
+To view information about a specific flag, simply run `cdk flags "#FLAGNAME#"`.
```
rix0rrr: "Simply" and "Just" are words you should ban from your technical writing.

They don't add any value, other than indicate that **you** think things are trivial. What's worse, if the person on the other end doesn't understand you, you've now made them feel dumb.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+```
+> cdk flags "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"
+    Feature Flag                                                        Recommended Value            User Value
+    "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"         true                         true
+```
```
rix0rrr: At this point I would show the full explanation. So I think the layout should be different.

Think in terms of a UI as a list + detail pane. We're now past the list, and moved into the detail pane.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+#### I want to modify one feature flag to a different value
+
+To modify one feature flag, first we view its information.
+
+```
+> cdk flags "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"
+    Description: Enable this feature flag to have cloudfront distributions use the security policy TLSv1.2_2021 by default.
+    Recommended Value: true
+    User Value: true
+```
```
rix0rrr, line 285: This is better! But inconsistent with the use of a single-flag display above.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+> 2
+Which feature flag would you like to revert?
+[1] @aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021
+[2] @aws-cdk/aws-s3:BucketEncryption
+[3] Exit
+> 1
+Reverting the change for @aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021...
+Resynthesizing...
+The flag has been reverted to its previous state.
+````
```
rix0rrr: I think we talked about that this requires too much knowledge for users.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+Although this feature will increase visibility to feature flags, we still need to bring visibility to the CLI tool itself. Otherwise,
+CDK users continue to remain unaware of feature flags and their potential benefits. To address this, we will display a notice whenever
+`cdk synth` is run. This notice will be discarded once users have configured all of their flags.
+
+```
+> cdk synth
+Notice: You currently have **10** unconfigured feature flags that could potentially impact
+your CDK application. Run `cdk flags` to learn more. 
+
```
rix0rrr: Why is this not part of the README? Why do we not talk to users about "Hey user, you might see the text `XXXX` pop up on your screen. This is what it means, here is why we're doing it.".

Why is this tucked away under "Additional considerations"?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+### Is this a breaking change?
+
+No. Users will be able to view the changes they are making and reject them if necessary.
+
+### Why not read from `recommended-feature-flags.json` from `aws-cdk-lib` directly?
+
+This would not work for jsii projects as the `recommended-feature-flags.json` file would not transfer over to projects in other
+CDK-compatible languages. Metadata files may not transfer over to projects in languages such as Python and Java and the
+feature flag information needs to be accessible at all times.
```
rix0rrr: "Transfer over" is vague. That can mean anything.

I can understand if you're not familiar enough with the details here to answer this question, but if that's the case then *ask me*, or someone else, to explain it. What you've written now is not convincing.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+    Feature Flag                              Recommended Value            User Value
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  
+  How would you like to proceed?
+   >  Set all flags to recommended values 
+      Set unconfigured flags to recommended values
+      Reset unconfigured flags to defaults (no impact)
+      Modify individual flag
```
otaviomacedo: I'm still curious about what happens if I choose this option: how do I choose which flag do modify?

Also, in general, is there a "save" step? Or does every action automatically save?

I think you can flesh out the experience a bit more.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
+# RFC: Feature Flag Advisor
+
+* Original Author(s): [@vivian12300](https://quip-amazon.com/GGQ9EAAQ3Jw)
+* Tracking Issue: [#750](https://github.com/aws/aws-cdk-rfcs/issues/750)
+* API Bar Raiser: [@rix0rrr](https://quip-amazon.com/YPP9EA8Gt6Q)
+
+CDK uses feature flags to implement changes that could impact current infrastructure. By using these flags, users can add security updates,
```
iliapolo:
```suggestion
CDK uses [feature flags](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0055-feature-flags.md) to implement changes that could impact current infrastructure. By using these flags, users can add security updates,
```

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+new behaviors, or bug patches without altering your application's present state. However, because new feature flags are automatically
```
iliapolo: > without altering your application's present state

But we just said it **does** impact current infrastructure.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+disabled in existing projects to retain backward compatibility, users must manually track and update them as new releases become available.
+The cdk flags command allows you to view all flags, their recommended values, and modify the users' values.
```
iliapolo:
```suggestion
The `cdk flags` command allows you to view all flags, their recommended values, and modify the users' values.
```

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+## Working Backwards
+
+## Help
+
+```
+> cdk flags --help
+cdk flags 
+Finds and displays a report of the feature flag configuration and compares your current values with our recommended values.
```
iliapolo: You are introducing the notion of "recommended" before really explaining what it is. I personally was a bit disoriented at this point, and would benefit from some definitions before diving into the command help. 

- What is the difference between recommended and default?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+## README
+
+The new CLI command `cdk flags` shows you a report of your feature flag configurations that differ from our recommended states. It
+compares your current feature flag configurations and our recommended states. When you run `cdk flags`, a feature flags report is displayed
+to you in the following format: flags that do not match the recommended values and then flags that are not configured by the user at all.
+There will be a Menu with options for the user to configure their feature flags.
+
+```
+> cdk flags
```
iliapolo, line 75: Proposal for a different name, as `flags` is somewhat ambiguous to me.

```suggestion
> cdk features 
```

Or `cdk feature-flags`

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+    --set-default                 set feature flags to their default states
+      --all                          set all feature flags to their default states
+      --unconfigured                 set unconfigured feature flags to their default states
+    "#FLAGNAME#"                  modify a specific flag's state
+      --set-recommended              set specific feature flag to its recommended state
+      --set-default                  set specific feature flag to its default state
+      --set                          set specific feature flag to any state
+    
+Examples:
+    > cdk flags --set-recommended --all
```
iliapolo: Kind of awkward to specify both an _option_ and an _action__ with `--`. This isn't the first time we find ourselves needing subcommands:

`cdk features set --recommended --all`

Maybe this is a good opportunity to see how much it would cost to add this capability? I'm hopinh `yargs` just has this, but we haven't utilized it so far.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+        Feature Flag                              Recommended Value            User Value
+      * @aws-cdk/...                              true                         false
+      * @aws-cdk/...                              true                         false
+      * @aws-cdk/...                              true                         - <unset>
+      Setting all feature flags to the recommended values...
+
+    > cdk flags --set-default --unconfigured
+        Feature Flag                              Recommended Value            User Value
+      * @aws-cdk/...                              true                         - <unset>
```
iliapolo: But what does `<unset>` mean? don't we want to show the default value here also?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+      What value would you like to change "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021" to?
+      > true
+        false
+```
+
+## README
+
+The new CLI command `cdk flags` shows you a report of your feature flag configurations that differ from our recommended states. It
+compares your current feature flag configurations and our recommended states. When you run `cdk flags`, a feature flags report is displayed
+to you in the following format: flags that do not match the recommended values and then flags that are not configured by the user at all.
```
iliapolo: This paragraph contains quite a bit of duplication (you mention "report" and "recommended" twice), which makes it somewhat hard to read. See if you can make this more concise but not repeating yourself.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+      Modify individual flag 
+```
+
+### Expected Use Cases
+
+#### View information about your feature flags
+
+To view a report of all flags, users can run `cdk flags --all`. This will display a report of every feature flag in the following
+order: feature flags with states that match our recommendations, feature flags with states that do not match our recommended values, and
+lastly, unconfigured feature flags.
```
iliapolo: When I have a natural separation of categories like this I prefer using bullets.

- feature flags with states that match our recommendations
- feature flags with states that do not match our recommended values,
-  unconfigured feature flags.

Especially if you want to impose some reading order. (think about it).

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+```
+> cdk flags --all
+    Feature Flag                              Recommended Value            User Value
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  
+  How would you like to proceed?
```
iliapolo: The title of this section says "View" - but now it seems like i'm supposed to take some action? Kind of surprising, and not machine friendly (i.e - cannot be executed in automation). 

Can you think of a way to make this more clear/expected?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+    Feature Flag                                                        Recommended Value            User Value
+    "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"         true                         true
+```
+
+#### Modify your feature flag values
+
+To modify your feature flag configurations, you can use the options `--set-recommended` and `--set-default`. In addition to these
+options, the `"#FLAGNAME#"` option also contains a `--set` suboption for the user to choose the new value.
+
+#### I want to change all my flags to the recommended states
```
iliapolo: Is this supposed to be an example of "Modify your feature flag values"? The same header size makes it seems like you've moved on to some other use case now.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+To view information about a specific flag, simply run `cdk flags "#FLAGNAME#"`.
+
+```
+> cdk flags "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"
+    Feature Flag                                                        Recommended Value            User Value
+    "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021"         true                         true
+```
+
+#### Modify your feature flag values
```
iliapolo: But previously you also showed how to modify my feature flags

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+```
+> cdk flags --set-recommended --all
+    Feature Flag                              Recommended Value            User Value
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  Setting all feature flags to the recommended values...
+
+  Resynthesizing...
```
iliapolo: At this point, did the context file already change? or are we still just proposing the change, not yet committing it to file? 

What if the user sees the proposed change and doesn't like it, how do they go back?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+  +    Properties:
+  +      BucketEncryption: 
+  +        `ServerSideEncryptionConfiguration``:`
+  +          ...
+  
+```
+
+#### I only want to change one feature flag to its recommended state
+
+To simply modify one feature flag to our recommended state, you can run `cdk flags --set-recommended --flagName="#FLAGNAME#"`.
```
iliapolo: Wait where did `--flagName` come from? Wasn't it a positional argument before?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+  +    Properties:
+  +      BucketEncryption: 
+  +        `ServerSideEncryptionConfiguration``:`
+  +          ...
+  
+```
+
+#### I want to modify one feature flag to a different value
+
+To modify one feature flag, first we view its information.
```
iliapolo: > first we view its information.

Do I have to?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+    Description: Enable this feature flag to have cloudfront distributions use the security policy TLSv1.2_2021 by default.
+    Recommended Value: true
+    User Value: true
+```
+
+Additionally, to modify the state, you can run `cdk flags --set --flagName="#FLAGNAME#"`. This will return a list of values for you to choose from.
+
+```
+> cdk flags "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021" --set
+  What value would you like to change "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021" to?
```
iliapolo: Is there a non-interactive way I can apply these changes? Even though the feature will likely mainly used by humans, we try to offer non-interactive versions to everything we do, so that people can create scripts with these capabilities.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+-    Properties:
+-      `BucketEncryption``:`` ``None`
++ S3 Bucket:
++    Properties:
++      BucketEncryption: 
++        `ServerSideEncryptionConfiguration``:`
++          ...
+```
+
+#### I want to revert my changes
```
iliapolo: Ooo this is important information, mention this somewhere on top too.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+keep the changes as shown below. If you choose to revert the changes, you are given options to revert all flags to its previous states
+or just a specific one. There will be a list of flags for you to choose from to revert.
+
+````
+Would you like to accept these changes? (Accept/revert)
+> revert
+Would you like to revert the entire feature flag configuration to its previous state or just revert the specific change(s)?
+[1] Revert all changes -> Restore all flags to their previous state
+[2] Revert only changes to one feature flag -> Undo a specific flag change
+[3] Exit without making any changes -> Keep your current configuration as is
```
iliapolo: What is the state of my current configuration? is it after applying the changes? i.e as I was happy with the result?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+> 2
+Which feature flag would you like to revert?
+[1] @aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021
+[2] @aws-cdk/aws-s3:BucketEncryption
+[3] Exit
+> 1
+Reverting the change for @aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021...
+Resynthesizing...
+The flag has been reverted to its previous state.
```
iliapolo: What if I want to revert 2 feature flags?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+````
+
+## What are we trying to solve?
+
+Feature flags are an important tool for adding new bug fixes without affecting the current state of the application. However,
```
iliapolo: > without affecting the current state of the application

Without? Then why do we need the flag?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+feature flags often go unnoticed by the user. In order to change their configurations, users are required to manually update their
+`cdk.json` file, resulting in a time-consuming task. They can impact a user in many ways. For example, the
+@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021 flag changed the default behavior to use the updated version TLSv1.2_2021,
+which is a more secure and support TLS version for CloudFront. Users who stuck with the previous version may have been exposed to
+security vulnerabilities and failed compliance tests. To remain Backwards Compatible, newly released feature flags are automatically
+disabled for existing projects, meaning that those users are required to constantly keep an eye on CDK releases to manually update their feature flags.
```
iliapolo: > keep an eye on CDK releases

A question comes to mind: Do we have a place that even lists all possible features? is it easy to get to?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+Another solution could be to prompt users to configure their feature flags when they create their CDK app using `cdk init`. This
+increases visibility to the feature flags and can be integrated smoothly within their project creation process. However, this does
+not help existing projects and could be considered an inconvenience and unnecessary by users.
+
+## What is the technical solution (design) of this feature?
+
+To implement this feature, we would first generate a Feature Flag Report artifact to be stored in CloudAssembly. The report is a `Record`
+of individual `FeatureFlag` objects that contain the fields `userValue`, `recommendedValue`, and `summary.` When the `cdk flags` command
+is run, the CLI accesses this artifact stored in `manifest.json` and displays the report to the user. A backup of the previous version of
+`cdk.out`will be stored in memory for future comparison. When the user chooses to modify a specific feature flag to our recommended value,
```
iliapolo: > will be stored in memory for future comparison

So if the CLI crashes during this time, we end up changing their config without acknowledgment? Can we be safer than that?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Increasing visibility to `cdk flags`
+
+Although this feature will increase visibility to feature flags, we still need to bring visibility to the CLI tool itself. Otherwise,
+CDK users continue to remain unaware of feature flags and their potential benefits. To address this, we will display a notice whenever
+`cdk synth` is run. This notice will be discarded once users have configured all of their flags.
+
+```
+> cdk synth
+Notice: You currently have **10** unconfigured feature flags that could potentially impact
```
iliapolo: The phrasing here make it sound like if you **dont** make changes, you might be impacted - was this the intention?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+  * @aws-cdk/...                              true                         - <unset>
+  
+  MENU ----------------------------------------------------------------------
+   >  Set all flags to recommended values
+      Set all unconfigured flags to recommended values
+      Reset all unconfigured flags to defaults (no impact)
+      Modify individual flag
+```
+
+### Ensuring safe configuration changes
```
iliapolo: So, are you actually doing this? or is this for the future? If we are - needs more detail.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+Feature flags contain bug fixes and improved default configurations that could potentially help your application. This CLI tool
+will make enabling feature flags easier and more efficient.
+
+### Is this a breaking change?
+
+No. Users will be able to view the changes they are making and reject them if necessary.
+
+### Why not read from `recommended-feature-flags.json` from `aws-cdk-lib` directly?
+
+This would not work for jsii projects as the `recommended-feature-flags.json` file would not transfer over to projects in other
```
iliapolo: Before diving into the answer, explain a _little_ about `recommended-feature-flags.json` - this is the first time it is mentioned in the entire doc. 

Readers who may not have deep familiarity with the CDK codebase may be interested, and can provide feedback (but not without knowing what it is).

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
+# RFC: Feature Flag Advisor
+
+* Original Author(s): [@vivian12300](https://quip-amazon.com/GGQ9EAAQ3Jw)
+* Tracking Issue: [#750](https://github.com/aws/aws-cdk-rfcs/issues/750)
+* API Bar Raiser: [@rix0rrr](https://quip-amazon.com/YPP9EA8Gt6Q)
```
kaizencc: nit: shouldn't link to quip in public

text/0750-feature-flag-advisor.md
rix0rrr: I asked for an interactive experience by default. I think it more naturally invites exploration.

We should have an unattended interface as well, of course.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+CDK uses feature flags to implement changes that could impact current infrastructure. By using these flags, users can add security updates,
+new behaviors, or bug patches without altering your application's present state. However, because new feature flags are automatically
+disabled in existing projects to retain backward compatibility, users must manually track and update them as new releases become available.
```
kaizencc: imo, ultimately the outcome is that users _do not_ update feature flags for existing projects. this causes a bad experience when they run into an issue that a feature flag would have solved, but have to debug their code until they realize that the solution is simply to flip a feature flag.

text/0750-feature-flag-advisor.md
iliapolo: Also I hope the plan here is to reuse `cdk diff`? because right now the output doesn't quite resemble it.

text/0750-feature-flag-advisor.md
rix0rrr: For your edification: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md

text/0750-feature-flag-advisor.md
kaizencc: two nit comments to add on:

- remove any mention of `new` or time-related words in your README -- in 1 year, the README will still say "new CLI command" but by then `cdk flags` will no longer be new
- you say "the following format" but then proceed to describe a state that is hard to follow. i would expect an example of the format rather than a description of the format.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+    Feature Flag                              Recommended                  User
+    @aws-cdk/...                              true                         true
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  
+  MENU ----------------------------------------------------------------------
+   >  Set all flags to recommended values
+      Set all unconfigured flags to recommended values
+      Reset all unconfigured flags to defaults (no impact)
```
iankhou: I can imagine a user running this just to see what it does/checking status, maybe make the top option (default selected in CLI) do nothing?

That seems to be what "Reset all unconfigured flags to defaults (no impact)" does, but this option name is also kind of verbose.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,343 @@
...
+
+```
+> cdk flags 
+    Feature Flag                              Recommended Value            User Value
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         false
+  * @aws-cdk/...                              true                         - <unset>
+  
+  How would you like to proceed?
+   >  Set all flags to recommended values
```
kaizencc: something i'm confused about re: recommended. does this option guarantee that i will not alter the behavior of my CDK App? how? does recommended mean both "CDK team thinks you should set the flag this way" AND "we know that you are not currently relying on this behavior so it won't mess up your stack"?

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,326 @@
...
+
+### README
+
+Feature flags can be used to introduce bug fixes, security updates, or new behaviors, all while ensuring backwards compatibility with your application. This guide explains how to manage feature flags in your CDK application using the `cdk flags` CLI tool.
+
+To increase visibility to `cdk flags`, you receive a notice whenever `cdk synth` is run on your application similar to the one below. This notice does not interrupt the `cdk synth` command but is visible until all of your feature flags have been configured.
+
+```
+> cdk synth
+Notice: You currently have **10** unconfigured feature flags that may require attention to keep your application up-to-date. Run `cdk flags` to learn more.
```
rix0rrr: In general ❤️, but specifically the word "notice" is something we use to refer to something specific, and reusing it here in a different context may be confusing to users.

I think this message works equally well without that prefix.

rix0rrr: Maybe also recommend `cdk flags -i` for the most friendly experience.

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,326 @@
...
+```
+
+#### Use the interactive menu
+
+```
+> cdk flags -i
+    MENU ---------------------------------------------
+    > Set all flags to recommended values 
+      Set unconfigured flags to recommended values
+      Reset unconfigured flags to their default configuration (no impact)
```
rix0rrr:
```suggestion
      Set unconfigured flags to their implied configuration (record current behavior)
```

?

rix0rrr (commented):
Love it! This is exactly what I was looking for! ❤️

text/0750-feature-flag-advisor.md
```diff
@@ -0,0 +1,398 @@
...
+            ├─ [-] TLSv1.0
+            └─ [+] TLSv1.2_2021
+                    - ...
+    Number of stacks with differences: 1
+  Would you like to accept or reject these changes? (Accept/reject)
+> Accept
+  Resynthesizing...
+```
+
+To achieve the same result, you can also run `cdk flags --set "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021" --value="true"`.
```
otaviomacedo, line 254: If I set a flag with a non-recommended value this way, is `cdk synth` still going to notify that I have a non-recommended value?

PR #753, opened 2025-07-10
chore: updates to cdk cli telemetry RFC

kaizencc:
---

### DIFF
--- a/text/0732-cdk-cli-telemetry.md
+++ b/text/0732-cdk-cli-telemetry.md
@@ -118,18 +118,16 @@ regardless of context values set.
 
 ### CLI Option
 
-There is an existing `--no-version-reporting` global CLI option to opt-out of library metadata collection.
+There is an existing `--no-version-reporting` global CLI option to opt-out of
+library metadata collection for that specific command execution.
+
+> Note: setting `analyticsReporting` at the stack level still takes precedence over the CLI flag for library metadata only.
+
 This flag will be repurposed to include CLI telemetry as well.
 That is, if you are currently opting out of library metadata with `--no-version-reporting` you
 will automatically be opted out of CLI telemetry regardless of context values set.
-
-Note that setting `analyticsReporting` at the stack level still takes precedence over the CLI flag:
-
-```ts
-new cdk.Stack(this, 'MyStack', {
-  analyticsReporting: true, // takes precedence over --no-version-reporting
-});
-```
+If you are not currently using `--no-version-reporting` then you can use it to disable
+CLI telemetry for that specific command execution.
 
 ## Sending Telemetry Data to a Local File
 
--- a/text/0732-cdk-cli-telemetry.md
+++ b/text/0732-cdk-cli-telemetry.md
@@ -125,9 +125,9 @@ library metadata collection for that specific command execution.
 
 This flag will be repurposed to include CLI telemetry as well.
 That is, if you are currently opting out of library metadata with `--no-version-reporting` you
-will automatically be opted out of CLI telemetry regardless of context values set.
+will automatically be opted out of CLI telemetry regardless of context values set (for that specific command execution).
 If you are not currently using `--no-version-reporting` then you can use it to disable
-CLI telemetry for that specific command execution.
+CLI telemetry (for that specific command execution).
 
 ## Sending Telemetry Data to a Local File
 
--- a/text/0732-cdk-cli-telemetry.md
+++ b/text/0732-cdk-cli-telemetry.md
@@ -37,7 +37,6 @@ For more information on what gets sanitized, see [Customer Content](#customer-co
     "sessionId": "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9", // UUID generated on each CLI command invocation
     "eventId": "737EBA96-6A5F-4B1C-BE6D-FD395B10ECE9:1", // sessionId + an increment for each additional event in the session
     "installationId": "3F1FD23A-58A9-4C0D-8A82-098D6101B322", // UUID stored on a local file on the developer"s machine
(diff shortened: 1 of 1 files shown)


text/0732-cdk-cli-telemetry.md
```diff
@@ -222,6 +221,17 @@ It will help the team with:
 The second bullet point should improve the customer experience of reporting bugs that are
 affecting their work.
 
+While error messages, stack traces, and logs are essential to the debugging experience,
+we recognize that it can also contain sensitive information. Therefore we will NOT collect
+this data by default. We will still sanitize these values for customer content and
+personally-identifiable information as described below.
+
+You must opt-in for this data to be automatically sent to the telemetry service.
+Otherwise you will be given a choice to send this "error report" to the telemetry service when an  error occurs in the CLI.
```
kaizencc: im not sure what this will look like right now

text/0732-cdk-cli-telemetry.md
```diff
@@ -115,6 +98,27 @@ Alternatively, you can run a new CDK CLI command:
 
 This will record the action to your local `cdk.json` file, which affects the current CDK App only.
 
+We'll also add a global CLI option you can append to any CDK CLI command: `--disable-telemetry`.
```
iliapolo: `--disable-telemetry` makes it sound like it records something to the context, similarly to how `cli-telemetry --disable` does. It doesn't - so lets use standard negation:

```suggestion
We'll also add a global CLI option you can append to any CDK CLI command: `--no-telemetry`.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -115,6 +98,27 @@ Alternatively, you can run a new CDK CLI command:
...
+If passed in, we will not collect library metadata or CLI telemetry for that command.
+The old `--no-version-reporting` flag is no longer recommended, but will also disable both library
+metadata and CLI telemetry.
```
iliapolo: Add a disclaimer that if the `analyticsReporting` property is set on the stack level, it takes precedence over any CLI flag.

text/0732-cdk-cli-telemetry.md
```diff
@@ -115,6 +98,27 @@ Alternatively, you can run a new CDK CLI command:
 
 This will record the action to your local `cdk.json` file, which affects the current CDK App only.
 
+We'll also add a global CLI option you can append to any CDK CLI command: `--disable-telemetry`.
+If passed in, we will not collect library metadata or CLI telemetry for that command.
+The old `--no-version-reporting` flag is no longer recommended, but will also disable both library
```
iliapolo: Lets mention this prior to the addition of the new `--disable-telemetry` flag because its more important; customers may already be using it to opt-out of library metadata. 

I agree that customers who set `--no-version-reporting` would most likely like to opt-out from CLI telemetry as well, so i'm good with tying the two. But then don't we already have the global flag we want? why add another? If its just for naming consistency, I feel it doesn't belong as a decision in this RFC. Given we aren't going to drop support for `--no-version-reporting`, adding `--disable-telemetry` can be done at any point thereafter. 

I also propose we provide a specific `--no-cli-telemetry`, that will NOT disable library metadata - for customers who just want to opt-out of CLI telemetry (as a complement to `cli-telemetry --disable`)

text/0732-cdk-cli-telemetry.md
iliapolo:
```suggestion
Otherwise you will be given a choice to send an "error report" to the telemetry service when an  error occurs in the CLI.
```

text/0732-cdk-cli-telemetry.md
```diff
@@ -222,6 +221,17 @@ It will help the team with:
...
 
+While error messages, stack traces, and logs are essential to the debugging experience,
+we recognize that it can also contain sensitive information. Therefore we will NOT collect
+this data by default. We will still sanitize these values for customer content and
+personally-identifiable information as described below.
+
+You must opt-in for this data to be automatically sent to the telemetry service.
+Otherwise you will be given a choice to send this "error report" to the telemetry service when an  error occurs in the CLI.
+
+To opt-in, set `cli-telemetry-with-errors: true` in the context of your `cdk.json` or run
```
iliapolo: Meh. Can we say that:

```json
{ 
  "cli-telemetry": "true"
}
```

Is the same as:

```json
{ 
  "cli-telemetry": {
    "usage": true,
    "errors": false,
  }
}
```

And then have `cdk cli-telemetry --enable-errors` set the nested key?

iliapolo (changes requested):
I think we should also:

- Elaborate on Installation ID (how is it generated? where is it stored?)
- Add Session ID & Event ID to the table of metrics. 

Clarify that all these are random numbers, not derived from any personal data.

text/0732-cdk-cli-telemetry.md
kaizencc: I removed `--disable-telemetry` from the RFC entirely. `--no-cli-telemetry` can be added on later but is such a minute detail that i think it doesn't need to be in the RFC

text/0732-cdk-cli-telemetry.md
```diff
@@ -100,6 +83,11 @@ Both of these aspects will help us better serve CDK customers in the future.
 We do not collect [customer content](https://aws.amazon.com/compliance/data-privacy-faq/#topic-1) and we anonymize the telemetry we do collect.
 
 You can opt-out of sending telemetry data.
+Run `cdk cli-telemetry --status` to see if your current telemetry status based on context values
```
iliapolo:
```suggestion
Run `cdk cli-telemetry --status` to see your current telemetry status based on context values
```

PR #797, opened 2025-09-12
RFC 797: Bedrock AgentCore Tools 1P L2 construct

dineshSajwan:
This is a request for Amazon Bedrock AgentCore L2 construct that simplifies the deployment and management of AI agents at scale by wrapping the AgentCore L1 constructs. It provides a high-level, object-oriented approach to creating and managing Amazon Bedrock AgentCore resources. This enables developers to deploy highly effective agents securely using any framework and model. The package will provide constructs for the following primitives:

**AgentCore tools**

See # [Issue 785](https://github.com/aws/aws-cdk-rfcs/issues/785) for
additional details. 

APIs are signed off by @alvazjor

---

### DIFF
--- a/text/0786-bedrock-agentcore-tools-l2.md
+++ b/text/0786-bedrock-agentcore-tools-l2.md
@@ -66,11 +66,10 @@ The Browser construct supports the following network modes:
    - Select whether to run the browser in a virtual private cloud (VPC).
    - By configuring VPC connectivity, you enable secure access to private resources such as databases, internal APIs, and services within your VPC.
 
-While the VPC itself is mandatory, these are optional:
-
-- Subnets - if not provided, CDK will select appropriate subnets from the VPC
-- Security Groups - if not provided, CDK will create a default security group
-- Specific subnet selection criteria - you can let CDK choose automatically
+    While the VPC itself is mandatory, these are optional:
+    - Subnets - if not provided, CDK will select appropriate subnets from the VPC
+    - Security Groups - if not provided, CDK will create a default security group
+    - Specific subnet selection criteria - you can let CDK choose automatically
 
 For more information on VPC connectivity for Amazon Bedrock AgentCore Browser, please refer to the [official documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/agentcore-vpc.html).
 
@@ -122,22 +121,22 @@ const browser = new agentcore.BrowserCustom(this, 'BrowserVpcWithRecording', {
 });
 ```
 
-Browser exposes a [connections](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Connections.html) property. This property returns a
-connections object, which simplifies the process of defining and managing ingress and egress rules for security groups in your AWS CDK applications.
-Instead of directly manipulating security group rules, you interact with the Connections object of a construct, which then translates your connectivity
-requirements into the appropriate security group rules. For instance:
+Browser exposes a [connections](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Connections.html) property. This property returns a connections
+object, which simplifies the process of defining and managing ingress and egress rules for security groups in your AWS CDK applications. Instead of directly
+manipulating security group rules, you interact with the Connections object of a construct, which then translates your connectivity requirements into the
+appropriate security group rules. For instance:
 
 ```typescript fixture=default
-const vpc = new ec2.Vpc(stack, 'testVPC');
+const vpc = new ec2.Vpc(this, 'testVPC');
 
-const browser = new BrowserCustom(stack, 'test-browser', {
+const browser = new agentcore.BrowserCustom(this, 'test-browser', {
   browserCustomName: 'test_browser',
-  networkConfiguration: BrowserNetworkConfiguration.usingVpc(stack, {
+  networkConfiguration: agentcore.BrowserNetworkConfiguration.usingVpc(this, {
     vpc: vpc,
   }),
 });
 
-browser.connections.addSecurityGroup(new ec2.SecurityGroup(stack, 'AdditionalGroup', { vpc }));
+browser.connections.addSecurityGroup(new ec2.SecurityGroup(this, 'AdditionalGroup', { vpc }));
 ```
 
 So security groups can be added after the browser construct creation. You can use methods like allowFrom() and allowTo() to grant ingress access to/egress
@@ -253,7 +252,7 @@ For more information about code interpreter, please refer to the [official docum
 
 ### Code Interpreter Network Modes
 
-The Code Interpreter construct supports two network modes:
+The Code Interpreter construct supports the following network modes:
 
 1. **Public Network Mode** (`CodeInterpreterNetworkMode.usingPublicNetwork()`) - Default
 
@@ -270,11 +269,10 @@ The Code Interpreter construct supports two network modes:
    - Select whether to run the browser in a virtual private cloud (VPC).
    - By configuring VPC connectivity, you enable secure access to private resources such as databases, internal APIs, and services within your VPC.
 
-While the VPC itself is mandatory, these are optional:
-
-- Subnets - if not provided, CDK will select appropriate subnets from the VPC
-- Security Groups - if not provided, CDK will create a default security group
-- Specific subnet selection criteria - you can let CDK choose automatically
+    While the VPC itself is mandatory, these are optional:
+    - Subnets - if not provided, CDK will select appropriate subnets from the VPC
+    - Security Groups - if not provided, CDK will create a default security group
+    - Specific subnet selection criteria - you can let CDK choose automatically
 
 For more information on VPC connectivity for Amazon Bedrock AgentCore Browser, please refer to the [official documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/agentcore-vpc.html).
 
@@ -316,7 +314,7 @@ Instead of directly manipulating security group rules, you interact with the Con
 requirements into the appropriate security group rules. For instance:
 
 ```typescript fixture=default
-const vpc = new ec2.Vpc(stack, 'testVPC');
+const vpc = new ec2.Vpc(this, 'testVPC');
 
 const codeInterpreter = new agentcore.CodeInterpreterCustom(this, "MyCodeInterpreter", {
   codeInterpreterCustomName: "my_sandbox_interpreter",
@@ -326,7 +324,7 @@ const codeInterpreter = new agentcore.CodeInterpreterCustom(this, "MyCodeInterpr
   }),
 });
 
-codeInterpreter.connections.addSecurityGroup(new ec2.SecurityGroup(stack, 'AdditionalGroup', { vpc }));
+codeInterpreter.connections.addSecurityGroup(new ec2.SecurityGroup(this, 'AdditionalGroup', { vpc }));
 ```
 
 So security groups can be added after the browser construct creation. You can use methods like allowFrom() and allowTo() to grant ingress access to/egress


text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1792 @@
...
+Add endpoints with a single `addEndpoint()` call, and benefit from built-in validation that catches configuration errors at compile time rather than deployment.
+
+### Creating a Runtime
+
+#### Option 1: Use an existing image in ECR
+
+Reference an image available within ECR.
+
+```typescript
+import * as path from "path";
```
alvazjor: This line is not needed in this example, lets remove it to avoid confusion

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1792 @@
...
+import {
+  Runtime,
+  NetworkMode,
+} from "@krokoko/agentcore-cdk-constructs/lib/cdk-lib/bedrock-agentcore/runtime/runtime";
```
alvazjor: I know where this is coming from, but lets remember to change them before we remove the draft mode on the RFC

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1792 @@
...
+With Amazon Bedrock AgentCore, developers can accelerate AI agents into production.
+The service provides the scale, reliability, and security critical to real-world deployment.
+
+This construct library facilitates the deployment of AgentCore Runtime, Identity providers, Memory, and Built-in tools.
+The Built-in tools include Browser and Code Interpreter, along with Gateway resources.
+It leverages underlying CloudFormation L1 resources and custom resources to provision these AgentCore features.
+
+For more details please refer here [Amazon Bedrock AgentCore Documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/what-is-bedrock-agentcore.html).
+
+## AgentCore Runtime
```
alvazjor: General comment: 
We are missing here a table of properties and attributes this resource will need/return, but I think it can be added once the L1s are ready.
Additionally, there are some support and helper methods that will target `runtime` directly, but that don't provide a good explanation and use case example on why they are needed, and how can they be used. For example, the code snippets below explaining how to create the runtime are good, but they suddenly also add a endpoint to it, but we don't explain why or what use case will be solved by doing that. Adding additional sections explaining this helper methods, specially for people that are not fully knowledgable on AgentCore specifics, will give a better understanding of how we want this L2 to work and why

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1792 @@
...
+  path.join(__dirname, "path to agent dockerfile directory")
+);
+
+const runtime = new Runtime(this, "MyAgentRuntime", {
+  agentRuntimeName: "myAgent",
+  agentRuntimeArtifact: agentRuntimeArtifact,
+});
+```
+
+### Invoking the Runtime
```
alvazjor: Not sure if we need to add this code snippets here, as they can add confusion since this is not really CDK related. If we want to keep them, lets please add a one paragraph explanation, clarifying that his is just an example how you will use the runtime later, once deployed.

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1792 @@
...
+    payload: JSON.stringify({ prompt: "Your question" }),
+  })
+);
+
+// Process response
+for await (const event of response.response) {
+  console.log(new TextDecoder().decode(event));
+}
+```
+
```
alvazjor: There are no mentions of Runtime versioning. Is this something we don't plan to support initially?

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack - a49314670b41254c6e414fcf03da3769591ba63a

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack - a49314670b41254c6e414fcf03da3769591ba63a

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack , Added some details for versioning and endpoint. - a49314670b41254c6e414fcf03da3769591ba63a

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack, removed this snippet to avoid any confusion.

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Runtime versioning is an implicit behaviour of runtime, added some notes for versioning.  a49314670b41254c6e414fcf03da3769591ba63a

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1866 @@
...
+allowing you to maintain different versions for different environments or gradually roll out updates.
+
+### Runtime Endpoints
+
+Endpoints provide a stable way to invoke specific versions of your agent runtime, enabling controlled deployments across different environments.
+You can create endpoints using the `addEndpoint()` helper method to reference specific versions for staging
+or production environments. For example, you might keep a "production" endpoint on a stable version while testing newer versions
+through a "staging" endpoint. This separation allows you to test changes thoroughly before promoting them
+to production by simply updating the endpoint to point to the newer version.
+The "DEFAULT" endpoint automatically points to the latest version of your agent runtime.
```
alvazjor: This is a good explanation. Can we also add examples how this multi endpoint setup will look like in code? 
I am trying to understand how this method will work. From what I can see, and correct me if I am wrong, `addEndpoint` will assign the endpoint to the latest `runtime`. But lets say I update that runtime at some point in the future, because I modified something in the configuration. How will the original endpoint be preserved and accessed through CDK? What I imagining that will happen is that since the `runtime` changed, then we will end up also creating a new `endpoint` (the "DEFAULT" one) that is targeting the new version of the `runtime` which is ok, but makes me wonder how can people access or modify the old endpoint.

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1866 @@
...
+| `agentRuntimeName` | `string` | Yes | The name of the agent runtime |
+| `agentRuntimeArtifact` | `AgentRuntimeArtifact` | Yes | The artifact configuration for the agent runtime |
+| `executionRole` | `iam.IRole` | No | The IAM role that provides permissions for the agent runtime |
+| `networkConfiguration` | `NetworkConfiguration` | No | Network configuration for the agent runtime |
+| `description` | `string` | No | Optional description for the agent runtime |
+| `protocolConfiguration` | `ProtocolType` | No | Protocol configuration for the agent runtime |
+| `authorizerConfiguration` | `AuthorizerConfigurationRuntime` | No | Authorizer configuration for the agent runtime |
+| `concurrencyConfiguration` | `ConcurrencyConfiguration` | No | Concurrency configuration for the agent runtime |
+| `environmentVariables` | `{ [key: string]: string }` | No | Environment variables for the agent runtime |
+| `clientToken` | `string` | No | A unique identifier to ensure idempotency of the request |
```
alvazjor: I could be wrong, but I think this `clientToken` is needed for the initial custom resource implementation, and wont be needed for the L1s. That property is normally needed in direct API calls, but I dont see anything similar now that the L1s for this resource are available

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1866 @@
...
+## AgentCore Runtime
+
+The AgentCore Runtime construct enables you to deploy containerized agents on Amazon Bedrock AgentCore.
+This L2 construct simplifies runtime creation just pass your ECR repository name
+and the construct handles all the configuration with sensible defaults.
+
+### Runtime Versioning
+
+Amazon Bedrock AgentCore automatically manages runtime versioning to ensure safe deployments and rollback capabilities.
+When you create an agent runtime, AgentCore automatically creates version 1 (V1). Each subsequent update to the
```
alvazjor: Just out of curiosity, can you give a name to the version that is generated, or is it always autogenerated by the service ?

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1866 @@
...
+  description: "A test memory for testing addMemoryStrategy method",
+  expirationDays: Duration.days(90),
+});
+
+// Add strategies after instantiation
+memory.addMemoryStrategy(MemoryStrategy.BUILT_IN_SUMMARIZATION);
+memory.addMemoryStrategy(MemoryStrategy.BUILT_IN_SEMANTIC);
+```
+
+## Browser Tool
```
alvazjor: Lets add some paragraphs explaining what is this construct for, and some usecases explaining how it can be used, The examples below are great, so lets add a bit of explanation here, to help the user understand when they could use each of them

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1866 @@
...
+   - Connects to services using Smithy model definitions
+   - Supports OAUTH and API_KEY credential providers
+   - Ideal for AWS service integrations
+
+3. **Lambda Target** (`LambdaMcpTargetConfiguration`)
+   - Connects to AWS Lambda functions
+   - Supports GATEWAY_IAM_ROLE credential provider only
+   - Ideal for custom serverless function integration
+
+## Code Interpreter
```
alvazjor: Same as in the `Browser tool` section, lets add some paragraphs explaining what is this construct for, and some usecases explaining how it can be used, The examples below are great, so lets add a bit of explanation here, to help the user understand when they could use each of them. And lets bring up the `Network Modes` explanation before the code examples

text/0785-bedrock-agentcore-l2.md
krokoko: Correct, it has been removed from the current implementation which uses cfn resources, we will update the RFC with the changes

text/0785-bedrock-agentcore-l2.md
krokoko: Done in the local repo and will update the rfc

text/0785-bedrock-agentcore-l2.md
dineshSajwan: When you create an endpoint with addEndpoint(), it captures the version at that moment. If you specify a version explicitly, it uses that; otherwise, it uses the runtime's current version. You can use version to point to old endpoint.
I will add some example to RFC.

```
const publicEndpoint = agentRuntime.addEndpoint('public_endpoint_auth', {
      description: 'Public endpoint for AgentCore runtime access',
      version:"2"
    });
```

```
export interface AddEndpointOptions {
  /**
   * Description for the endpoint
   * @default - No description
   */
  readonly description?: string;
  /**
   * Override the runtime version for this endpoint
   * @default - Uses the runtime's version
   */
  readonly version?: string;
}
```

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Yes, It is not in CFN anymore, will remove it. Same with concurrencyConfiguration, will remove that too.

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Yes runtime versions are automatically generated when you update the runtime's configuration.

dineshSajwan: I like this explanation

Screenshot 2025-09-25 at 11 20 58 AM

text/0785-bedrock-agentcore-l2.md
dineshSajwan: On a second thought, do you think we should add a function in l2 construct to list all the old versions of a runtime.
boto3 has this - 
```
response = client.list_agent_runtime_versions(
    agentRuntimeId='string',
    maxResults=123,
    nextToken='string'
)
```

text/0785-bedrock-agentcore-l2.md
alvazjor: This is a very nice table. Does this DEFAULT endpoint exists when the moment you create the runtime? lets say I create the runtime in cdk without any specific endpoint, will this default one be created by the create runtime call? and can customers use it to invoke the runtime? In any case, this is good information and explain well the scenario, lets add it to the notes of the README in the alpha module PR if not there already, it will help customers to understand how they can manage their resources

text/0785-bedrock-agentcore-l2.md
alvazjor: I was confused by this, but now, an also with the table provided above, I get it better. Due to the nature of cdk, listing the version itself in the L2 construct wont work, since this call will be made probably during deployment and at that point you cannot control choosing which version to use, you already made that decision. But lets add this to the documentation both in the JSDoc and in the README file of the module itself, so we let our customers know how to look for existing versions so they can use a diff one than the latest if they need to. But after understanding this, I think this implementation is good enough as it is, we just need to expand the documentation around it

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,1899 @@
...
+
+// Production endpoint - explicitly pinned to a specific version
+const prodEndpoint = runtime.addEndpoint("production", {
+  version: "1",  // prod version 
+  description: "Stable production endpoint"
+});
+
+// Staging endpoint - for testing new versions before production
+const stagingEndpoint = runtime.addEndpoint("staging", {
+  version: "2",  // new version in staging env
```
alvazjor: Reading this code as it is, specially with this comment: `// Initial deployment - Creates Version 1`, it gives me the impression you could already create an endpoint targeting version 2 of the endpoint in one single iteration, but that version doesn't exist yet right? If so, lets split the example into steps: 1st will be to create the runtime and prod endpoint, then explain that a change in the runtime will be done allowing the creation of runtime v2, and then you can add the second endpoint targeting that version.

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Ack.

text/0785-bedrock-agentcore-l2.md
dineshSajwan: Yes, the DEFAULT endpoint gets created with runtime and you can use it to invoke too. 
Screenshot 2025-09-26 at 12 23 05 PM

alvazjor: Ok, makes sense, lets add notes on this, mostly in the readme notes of the alpha module, once the PR of this resource is implemented

text/0785-bedrock-agentcore-l2.md
dineshSajwan: updated the docs.

text/0785-bedrock-agentcore-l2.md
```diff
@@ -0,0 +1,2006 @@
...
+
+```typescript
+const endpoint = new agentcore.RuntimeEndpoint(this, "MyEndpoint", {
+  endpointName: "production",
+  agentRuntimeId: runtimeId,
+  agentRuntimeVersion: "1",
+});
+
+// Later, update to a new version
+endpoint.updateRuntimeVersion("2");
```
alvazjor: I am not sure this method adds value. What is the use case we are trying to solve with it? It seems weird to have a constructor for a resource with a param and then call a method to change that param. Considering that in CDK code everything happens at synth time, is this going to create just 1 resource (targeting v2) or is this going to create 2 endpoints, one per each version?

dineshSajwan: You are right , forgot to remove this, it was there as per the initial plan, but we don't need it now.

Issue #814, opened 2025-10-06
Unified mechanism for abstractions that change constructs after construction


mrgrain:
---
name: "RFC Tracking Issue"
about: "Tracking issue for an RFC"
title: Unified mechanism for abstractions that change constructs after construction
labels: status/proposed
---

## Description

The CDK currently lacks a unified, type-safe mechanism for applying abstractions to constructs after they are created. This forces customers into an "all-or-nothing" choice between sophisticated L2 abstractions and comprehensive AWS coverage, creating significant limitations for enterprise use cases.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @kornherm
| Author(s)           | @kornherm
| API Bar Raiser      | TBD
| Stakeholders        | TBD

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [ ] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [ ] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

### Current Limitations

1. **L1/L2 Incompatibility**: Valuable abstractions are locked within specific L2 implementations and cannot be applied to L1 constructs or third-party constructs
2. **Feature Lock-in**: Using any L2 feature requires accepting all design decisions of that L2, even if only one capability is needed
3. **Day-One Coverage Gap**: New AWS features are available in L1s immediately but lose all sophisticated abstractions, forcing customers to choose between new features and existing abstractions
4. **Customization Barriers**: Enterprise customers report needing to customize 90% of L2 constructs, often rebuilding entire construct libraries

### Real-World Impact

```typescript
// Current problem: Can't use L2 abstractions with L1 constructs
const l1Bucket = new s3.CfnBucket(stack, "Bucket");
// ❌ No way to add auto-delete, encryption, grants, or other L2 features

[...]

PR #824, opened 2025-10-10
RFC 814: CDK Mixins

mrgrain:
This is a request for comments about CDK Mixins: Composable Abstractions for AWS Resources. See #814 for additional details. 

APIs are signed off by @rix0r.

---

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license_

### DIFF
--- a/FULL_INDEX.md
+++ b/FULL_INDEX.md
@@ -203,7 +203,8 @@
 [789](https://github.com/aws/aws-cdk-rfcs/issues/789)|[EC2 Image Builder L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0789-bedrock-agentcore-gateway-l2.md)||👷 implementing
 [792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0792-aws-aiops-investigationgroup-l2.md)||👷 implementing
 [808](https://github.com/aws/aws-cdk-rfcs/issues/808)|[Library Quality Scoring](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0808-library-quality-scoring.md)||✅ done
-[814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||💡 proposed
+[814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||⏰ final comments
 [835](https://github.com/aws/aws-cdk-rfcs/issues/835)|[Native EBS Snapshot Asset Handling](https://github.com/aws/aws-cdk-rfcs/issues/835)||💡 proposed
 [859](https://github.com/aws/aws-cdk-rfcs/issues/859)|[L2 Construct for AWS::ObservabilityAdmin::OrganizationCentralizationRule](https://github.com/aws/aws-cdk-rfcs/issues/859)||💡 proposed
+[867](https://github.com/aws/aws-cdk-rfcs/issues/867)|[S3 Vector L2 CDK Constructs](https://github.com/aws/aws-cdk-rfcs/issues/867)||💡 proposed
 <!--END_TABLE-->
--- a/PROPOSED.md
+++ b/PROPOSED.md
@@ -32,7 +32,8 @@
 [762](https://github.com/aws/aws-cdk-rfcs/issues/762)|[Serverless ElastiCache L2 Construct](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0762-serverless-elasticache-l2-construct.md)|[@otaviomacedo](https://github.com/otaviomacedo)|💡 proposed
 [764](https://github.com/aws/aws-cdk-rfcs/issues/764)|[L2 Constructs for Bedrock AgentCore Memory](https://github.com/aws/aws-cdk-rfcs/issues/764)||💡 proposed
 [785](https://github.com/aws/aws-cdk-rfcs/issues/785)|[Bedrock AgentCore L2 alpha package](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0785-bedrock-agentcore-runtime-l2.md)||💡 proposed
-[814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||💡 proposed
 [835](https://github.com/aws/aws-cdk-rfcs/issues/835)|[Native EBS Snapshot Asset Handling](https://github.com/aws/aws-cdk-rfcs/issues/835)||💡 proposed
 [859](https://github.com/aws/aws-cdk-rfcs/issues/859)|[L2 Construct for AWS::ObservabilityAdmin::OrganizationCentralizationRule](https://github.com/aws/aws-cdk-rfcs/issues/859)||💡 proposed
+[867](https://github.com/aws/aws-cdk-rfcs/issues/867)|[S3 Vector L2 CDK Constructs](https://github.com/aws/aws-cdk-rfcs/issues/867)||💡 proposed
+[814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||⏰ final comments
 <!--END_TABLE-->
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ future state of the libraries and to discover projects for contribution.
 [673](https://github.com/aws/aws-cdk-rfcs/issues/673)|[AWS CloudWatch Application Signals L2 Constructs for SLO](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0673-aws-applicationsignals-slo-l2.md)||👷 implementing
 [789](https://github.com/aws/aws-cdk-rfcs/issues/789)|[EC2 Image Builder L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0789-bedrock-agentcore-gateway-l2.md)||👷 implementing
 [792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0792-aws-aiops-investigationgroup-l2.md)||👷 implementing
+[814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||⏰ final comments
 <!--END_TABLE-->
 
 ## What is an RFC?


text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+bucketAccessLogsMixin.supports(bucket); // true
+bucketAccessLogsMixin.supports(queue); // false
+```
+
+#### Mixins must apply
+
+todo: will fail if encountering construct it cannot deal with. use to ensure now surprises
+
+```ts
+Mixins.of(scope, selector).mustApply(new EncryptionAtRest());
```
gasolima: I felt that it should be the other way around the `.apply()` will fail encountered a construct that's not supported, and another `.mayApply()` function that only maybe will apply.

This feeling come from as a customer expecting `apply()` should apply for everything and if it's not possible throw an error

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+  .with(new AutoDeleteObjects())
+  .with(new ComplianceLogging())
+  .with(new CfnBucketPropsMixin({ 
+    someNewFeatureConfig: { settingOne: "value1" } 
+  }));
+```
+
+#### Cross-Service Mixins
+
+Apply common patterns across different AWS services:
```
gasolima: What different AWS services mean here? what i understand from the example it's only applying for a single service(`CfnLogGroup`)

EDIT: Oh what i'm search for under `**Cross-Service Mixins**: Common patterns across AWS services` section. But not sure what does this section mean

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+```ts
+// Applies the aspect immediately
+Mixins.of(scope).apply(Mixin.fromAspect(new TaggingAspect({ Environment: "prod" })));
+
+// Delays application of the Mixin to the synthesis phase
+Aspects.of(scope).add(Aspect.fromMixin(new EncryptionAtRest()));
+```
+
+Mixins have more features than Aspects. When converting a Mixin to an Aspect, the Mixin will automatically only be applied to supported constructs.
```
gasolima: Does  that mean now we're going to recommend for ppl to not use aspect anymore and go with Mixin?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+#### Error Handling and Edge Cases
+
+Mixins provide comprehensive error handling for common scenarios:
+
+```typescript
+// Validation errors are collected and reported _per Mixin application_:
+const bucket = new s3.CfnBucket(scope, "Bucket")
+  .with(new EncryptionAtRest())
+  .with(new InvalidMixin()); // Throws validation error
```
gasolima: I believe this will only throw validation error, if we implemented `validate` method, right?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+### What are the drawbacks of this solution?
+
+1. **Learning Curve**: Developers need to understand a new abstraction pattern
+2. **API Surface**: More APIs to document and maintain
+
+### Are there any open issues that need to be addressed later?
+
+1. **Performance Impact**: Need to measure runtime overhead of mixin composition
+2. **IDE Support**: Ensure TypeScript language services work well with complex mixin types
```
gasolima: What do we consider as complex mixin types?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+// Apply to constructs matching a pattern
+Mixins.of(
+  scope,
+  ConstructSelector.byId(/.*-prod-.*/) 
+).apply(new ProductionSecurityMixin());
+
+// The default is to apply to all constructs in the scope
+Mixins.of(
+  scope,
+  ConstructSelector.all() // supports depth-first and breadth-first
```
gasolima: What does depth-first and breadth-first mean here?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+  static all(): ConstructSelector {
+    return new AllConstructsSelector();
+  }
+
+  static cfnResource(): ConstructSelector {
+    return new CfnResourceSelector();
+  }
+
+  static resourcesOfType(type: string | Function): ConstructSelector {
```
gasolima: What about `resourcesOfTypes(type: string[] | Function[])`?

EDIT: I feel a bit stupid now is those function coming out of the box from us to the customers or that's an example of how a customer using mixin 😅

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+#### Mixins declare which constructs they support
+
+Mixins typically target specific resource constructs.
+When applied to a construct tree, the mixing will check if a given construct is supported.
+Only when supported applies, otherwise skip.
+
+```ts
+bucketAccessLogsMixin.supports(bucket); // true
+bucketAccessLogsMixin.supports(queue); // false
+```
```
commiterate: Is there a way for users to know when they can safely remove an unused mixin?

`IMixin.supports()` is a runtime check so compilers/transpilers won't show anything at compile time (or via LSP feedback with things like yellow squiggly underlines in your editor of choice).

It seems like the current proposed `Mixins.mustApply()` is the way to detect this, so mixin application being optional sounds like an intended property. I suspect this is because L2 constructs may conditionally create an L1 construct so it's not possible to guarantee a mixin will always be used.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+Mixins.of(
+  scope,
+  ConstructSelector.resourcesOfType("AWS::S3::Bucket")
+).apply(new EncryptionAtRest());
+
+// Apply to constructs matching a pattern
+Mixins.of(
+  scope,
+  ConstructSelector.byId(/.*-prod-.*/) 
+).apply(new ProductionSecurityMixin());
```
commiterate, line 207: If there's an L2 construct which internally generates multiple L1 constructs of the same type, would users have to know how the L1 construct IDs are being generated internally to select the right one?

Or would it be better for users to extend a mixin class and override the `supports()` function to do more complicated matching (e.g. type + value checks).

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+    return construct instanceof s3.CfnBucket;
+  }
+
+  applyTo(bucket: IConstruct): IConstruct {
+    bucket.versioningConfiguration = {
+      status: "Enabled"
+    };
+    return bucket;
+  }
+}
```
commiterate: For a mixin to support both L1 and L2 constructs, do we need something like:

```ts
class EnableVersioning implements IMixin {
  supports(construct: IConstruct): boolean {
    return construct instanceof s3.CfnBucket || construct instanceof s3.Bucket;
  }

  applyTo(bucket: IConstruct): IConstruct {
    if (construct instanceof s3.CfnBucket) {
      // ...
    } else if (construct instanceof s3.Bucket) {
      // ...
    }
  }
}
```

What if the user-defined portions of mixins instead always operate on L1 constructs and any L2 constructs are automatically decomposed to their L1 constructs?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+```typescript
+import * as s3 from "aws-cdk-lib/aws-s3";
+import { Mixins, AutoDeleteObjects, EncryptionAtRest } from "aws-cdk-lib/mixins";
+
+// Base form: apply mixins to any construct
+// calls can be chained
+const bucket = new s3.CfnBucket(scope, "MyBucket");
+Mixins.of(bucket)
+  .apply(new EncryptionAtRest())
```
rix0rrr, line 62: Are we doing variadic functions?

text/0814-cdk-mixins.md
rix0rrr: Hi @commiterate, L2s already contain their L1s, so this already works!

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+// This is what `.with()` is using
+Mixins.of(
+  bucket,
+  ConstructSelector.cfnResource() // provided CfnResource or a CfnResource default child
+).apply(new EncryptionAtRest());
+
+// Apply to all resources of a specific type
+Mixins.of(
+  scope,
+  ConstructSelector.resourcesOfType(s3.CfnBucket)
```
rix0rrr: Does this require passing a class? Because that won't work in jsii.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+Optionally, you may select a subset of constructs
+
+```ts
+import { ConstructSelector } from "aws-cdk-lib/mixins";
+
+// Apply to a given L1 resource or L2 resource construct
+// This is what `.with()` is using
+Mixins.of(
+  bucket,
+  ConstructSelector.cfnResource() // provided CfnResource or a CfnResource default child
```
rix0rrr, line 188: Not sure what this does/adds over `supports()` ?

text/0814-cdk-mixins.md
rix0rrr: Should we even get to specifying/picking traversal order?

(Probably we then also need preorder/postorder 😉 )

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+).apply(new ProductionSecurityMixin());
+
+// The default is to apply to all constructs in the scope
+Mixins.of(
+  scope,
+  ConstructSelector.all() // supports depth-first and breadth-first
+).apply(new ProductionSecurityMixin());
+
+// Option: Alternative syntax
+Mixins.of(scope).select(ConstructSelector.cfnResource())
```
rix0rrr: I like this interface to select a part of the construct tree.

text/0814-cdk-mixins.md
rix0rrr: > will fail encountered a construct that's not supported

I don't think that's what we're proposing. I think we're saying: "Must have applied to `>=1` construct"

It might still track to have `mustApply` be the default...

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+todo: will fail if encountering construct it cannot deal with. use to ensure now surprises
+
+```ts
+Mixins.of(scope, selector).mustApply(new EncryptionAtRest());
+```
+
+#### Validations
+
+Mixins are only validating when applied, never at construction time.
```
rix0rrr: This sentence is not clear to me. I can't parse "validating" here, and "construction time" isn't explained.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+```typescript
+interface IMixin {
+  /** Check if this mixin can be applied to the given construct */
+  supports(construct: IConstruct): boolean;
+  
+  /** Apply the mixin to the construct */
+  applyTo(construct: IConstruct): IConstruct;
+  
+  /** Validate the construct after mixin application */
+  validate?(construct: IConstruct): string[];
```
rix0rrr: This isn't jsii-able.

rix0rrr: `validate()` could be unnecessary -- `Annotations.of().addError()` could do the same job in `applyTo`?

Unless the invocation time for `validate` is different from `applyTo` ?

You need to explain how this is different/useful if we already have `Aspects`.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+  .with(new SecurityCompliance());
+```
+
+#### Mixins and Aspects
+
+Mixins and Aspects are very similar in some regards, but crucial differ in their time of application:
+Mixins are always applied _immediately_, they are a tool of imperative programming.
+On the other hand, Aspects are applied _after_ everything else during the synthesis step, they are declarative.
+
+However because implementing Mixins and Aspects is very similar, they can be converted from each other:
```
rix0rrr: 😈

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+  .with(new EncryptionAtRest())
+  .with(new InvalidMixin()); // Throws validation error
+
+// Graceful handling of unsupported constructs
+Mixins.of(scope)
+  .apply(new EncryptionAtRest()); // Skips unsupported constructs
+
+// Strict application that requires all constructs to match
+Mixins.of(scope)
+  .mustApply(new EncryptionAtRest()); // Throws if any constructs doesn't support the mixin
```
rix0rrr: If there isn't at least one construct that supports the mixin?

Otherwise any grouping construct will fail.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+  .apply(new EncryptionAtRest())
+  .apply(new AutoDeleteObjects());
+```
+
+For convenience, constructs implement the `.with()` method as syntactic sugar:
+
+```typescript
+// Syntactic sugar: chain mixins fluently
+const l1Bucket = new s3.CfnBucket(scope, "MyL1Bucket")
+  .with(new EncryptionAtRest())
```
rix0rrr: How are constructs obtaining the `with()` function? Does it come from `Construct/IConstruct`, or `Resource`? Or do they all re-implement it?

text/0814-cdk-mixins.md
rix0rrr: If the behavior is really "all selected constructs must support this", then basically *every* `Mixins.of()` call needs a selector.

`Bucket` already doesn't support `CfnBucketPropsMixin`... is that the experience we want?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+allowing customers to mix and match sophisticated features without being locked into specific implementations.
+
+## Working Backwards
+
+### README: CDK Mixins
+
+CDK Mixins provide a new, advanced way to add functionality to through composable abstractions.
+Unlike traditional L2 constructs that bundle all features together, Mixins allow you to pick and choose exactly the capabilities you need for constructs.
+
+#### Key Benefits
```
rix0rrr, line 18: We need some discussion of what they do and don't do.

Looks like:

- It can't automatically set required arguments.
- It doesn't do defaults. 

Might be worth discussing up front?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+class EnableVersioning implements IMixin {
+  supports(construct: IConstruct): boolean {
+    return construct instanceof s3.CfnBucket;
+  }
+
+  applyTo(bucket: IConstruct): IConstruct {
+    bucket.versioningConfiguration = {
+      status: "Enabled"
+    };
+    return bucket;
```
rix0rrr, line 122: Why does `applyTo()` return something? I was hoping to find the answer to that later on in the doc but didn't?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+The current CDK architecture forces customers into an "all-or-nothing" choice between sophisticated L2 abstractions and comprehensive AWS coverage.
+This creates three "treadmill" problems:
+
+1. **Coverage Treadmill**: We must provide L2s for all AWS services
+2. **Completeness Treadmill**: Each L2 must support every feature of the underlying resource
+3. **Customization Treadmill**: We must support all possible customizations
+
+These treadmills are unsustainable given AWS's pace of innovation (2,000+ features annually vs. 5 new CDK modules per year).
+
+### Why should we _not_ do this?
```
rix0rrr, line 584: Something somewhere needs to discuss that L1 mixins allow desyncing the L1 state from the L2 state. We need to acknowledge that and formulate some thoughts around it.

"Don't do that if it hurts", "we will get to that later", ... something.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+3. **Maintenance Burden**: More code to maintain and test
+
+However, these risks are mitigated by maintaining full backward compatibility and providing clear migration paths.
+
+### What is the technical solution (design) of this feature?
+
+The solution has four key components:
+
+1. **Mixin Interface**: A `.with(mixin)` method that allows composing functionality
+2. **Resource Traits**: Common interfaces (like `IEncryptable`) that enable cross-service abstractions
```
rix0rrr: Could be discussed up top where you discuss `new EncryptionAtRest`, that it is supported by this interface.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+3. **Addressable Resources**: Shared interfaces between L1s and L2s for interoperability
+4. **Automatic Generation**: Mixins generated from AWS service specifications
+
+The implementation uses TypeScript's type system to ensure type safety while maintaining runtime flexibility.
```
rix0rrr: If you say TypeScript you must say jsii 😉

text/0814-cdk-mixins.md
commiterate: Sorry I didn't make myself clear enough. I'm aware that L2s do have their L1s exposed via `IConstruct.node.children` so the question is whether `IMixin.supports`/`applyTo` should only receive L1s (i.e. have intermediary logic which automatically decomposes higher level constructs into L1s) or should higher level constructs be passed in and the CDK should only provide a utility decomposition function they can call within these mixin methods?

mrgrain: Excellent question @commiterate! Currently I have this goal split up into multiple places:

- The mixin itself should only support L1s. `IMixin.supports` will cause the the mixin to only be applied to the L1, even when given an L2.
- I envisioned the default selector for constructs to be one that picks the current construct if it's an L1 or picks `IConstruct.node.children` if that's an L1. This makes it however slightly awkward to apply mixins to L3s, so not sure.
- Constructs implementing `.with()` should either use the "magic selector" or simply pass their default resource

For mixin implementation, to recommendation would very strongly be to apply to L1s whenever possible.

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+**Cons**:
+
+* Still requires full L2 development for sophisticated features
+* Doesn't solve the composability problem
+* Limited cross-service abstraction capabilities
+
+**Why Mixins Are Better**: Mixins provide the sophisticated features of enhanced L1s
+while enabling composition and cross-service patterns that enhanced L1s cannot achieve.
+
+#### 2. Modular L2 Redesign
```
commiterate, line 645: Not too sure if there was something specific in mind for this but I'll just interpret this as any breaking L2 changes.

https://github.com/aws/aws-cdk-rfcs/pull/657#discussion_r2036706731 proposes such a change so I'll just use this thread (or the closest substitute we can get with GitHub PR comments) to discuss that proposal and how this RFC addresses it.

---

The linked proposal essentially wants L2 constructs to be split into 2 parts:

1. A "dumb/plain" dictionary/map of L1 constructs.
2. Wiring/currying/overlay functions which operate on L1 constructs.

L2 constructs today are fairly opaque since they complect both of these concerns without exposing much of the underlying details. The child L1 constructs aren't exactly that convenient to access (need to use `IConstruct.node.children` and do a bunch of if...else with `instanceof` to restore type information).

For 1, the general idea is possible but the implementation detail of using a TS interface (i.e. just a plain dictionary/map) isn't really possible since the CDK is heavily class based which is inherited from `constructs`. Passing immutable maps like in Clojure is out of the question.

For 2, it's essentially saying to extract all of the L2 wiring logic outside of the L2 construct (e.g. into functions, mixins, or whatever is decided upon).

---

If we assume something like the current class-based mixin design is accepted, that means we'd have something like this for `AWS::Lambda::Function`:

```typescript
import { Construct, IConstruct } from "constructs";
import { IMixin, Resource } from "@aws-cdk/core";
import { aws_iam } from "@aws-cdk";

// L1. Auto-generated from Cfn resource definitions.

export class CfnFunctionProps {
  // ...
}

export class CfnFunction extends Construct {
  // Resource attributes. Same as today.
  public attr*: type;

  // Constructor. Same as today.
  constructor(scope: Construct, id: string, props: CfnFunctionProps) {
    // ...
  }
}

// L2. Manually written.

export interface FunctionExecutionRoleMixinProps {
  function: CfnFunction;
  role: aws_iam.CfnRole;
}

// I don't really know why this is a class. This can just be a function.
export class FunctionExecutionRoleMixin implements IMixin<FunctionExecutionRoleMixinProps> {
  static apply(props: FunctionExecutionRoleMixinProps): FunctionExecutionRoleMixinProps {
    // Add the Lambda-managed basic execution role policy.
    props.function.managedPolicyArns = (props.function.managedPolicyArns ?? []) + [aws_iam.ManagedPolicy.fromAwsManagedPolicyName("AWSLambdaBasicExecutionRole").managedPolicyArn];
    return props;
  }
}

export interface FunctionComponents {
  function: CfnFunction;
  role: aws_iam.CfnRole;
}

// Same as today.
export interface FunctionProps {
  // ...
}

export class Function extends Resource {
  // L1 components.
  public readonly components: FunctionComponents;

  // Existing properties for backwards compatibility.

  // Constructor.
  constructor(scope: Construct, id: String, props: FunctionProps) {
    this.components = {
      function: new CfnFunction(scope, "function", {
        // Wire in values from `props: FunctionProps`.
      }),
      role: props.role ?? new aws_iam.CfnRole(scope, "role", {
        // Default function execution role if absent.
      })
    };

    // Apply the function mixin. This is doing mutation in place, may not want this for the final design.
    FunctionExecutionRoleMixin.apply({
      function: this.components.function;
      role: this.components.role;
    });
  }

  // Existing methods for backwards compatibility.
  //
  // All of their logic (e.g. adding IAM policies for event sources) are
  // extracted into mixins. The methods just apply the mixin for users.
}
```

The L2 construct basically becomes a composition of L1s and default mixin applications. If users don't like the L2 construct, they're free to apply mixins directly to L1s they created or extract the L1s from the L2 and apply them.

text/0814-cdk-mixins.md
mrgrain: This was based on some feedback. When using `EncryptionAtRest` mixin, I want to make sure the mixin _really_ is applied to _all_ constructs that I expect it to be applied to. So for this example you are right: It is pointless without selector.

text/0814-cdk-mixins.md
mrgrain: CfnFunctionProps.code is a problem as well because it is required and cannot be a mixin

text/0814-cdk-mixins.md
mrgrain: I think we want to. Downside is we cannot have optional props ever here.

text/0814-cdk-mixins.md
mrgrain: Could be constructs, but that's potentially high impact. Also need to consider type covariance.

text/0814-cdk-mixins.md
mrgrain: This was the idea:

```ts
declare const bucket: CfnBucket;
const bucketWithPolicy:  CfnBucket & HasResourcePolicy  = new TypedMixin().applyTo(bucket);
```
(Doesn't effect typing in `with()` or `apply()`)

---

Application behavior of multiple `.apply()` vs. Tree and Selectors. Can we even return a changed construct?

mrgrain: `.nowGetValue();` ?

text/0814-cdk-mixins.md
mrgrain: 3 options:
- must apply to everything
- must apply to at least one
- apply to only supported (maybe none)

mrgrain: mustApply only makes sense with selector.
other two are okay without selector.
fully optional case will warn if not applied

Is this just a naming question? What are the 3 cases called?

text/0814-cdk-mixins.md
mrgrain: Use case is that I might want some buckets to be encrypted, but not others.

mrgrain: Potential future: `ConstructSelector.css("....")` a CSS style syntax

select based on tags, or construct context values

For now provides an interfaces to turn a construct tree into an ordered list of constructs

mrgrain: What's the point if with really?

```ts
// with on whole tree, must apply once
// fun.with(new PermissionsBoundaryMixin())
// Mixins.of(func).applyAtLeastOnce(new PermissionsBoundaryMixin());
// or
// typed, on L1 resource (same or default child) only, must apply
```

text/0814-cdk-mixins.md
mrgrain:
```
resourcesOfType(res: string | { readonly CFN_RESOURCE_TYPE_NAME: string })
```

mrgrain:
```ts
ConstructSelector.resourcesOfType(s3.CfnBucket.CFN_RESOURCE_TYPE_NAME)
```

text/0814-cdk-mixins.md
mrgrain: Was thinking that `all()` has a bunch of options for people who really care.

text/0814-cdk-mixins.md
mrgrain: Could ConstructSelector be a predicate. Input probably needs to be current construct plus context (root path etc)

text/0814-cdk-mixins.md
mrgrain:
```
scope.select(...).apply(...)
scope.mixins.select(...).apply(...)
```


`Mixins.of()` and `scope.mixins` and `scope.select(...)` just return a construct tree wrapped in something else.

mrgrain: Two-way door. Adding to `constructs` could be a nice to have.

text/0814-cdk-mixins.md
mrgrain: Also planning to emit a warning for mixins that didn't apply to anything.

text/0814-cdk-mixins.md
mrgrain:
```ts
const mixin = new SomeBucketMixin({ foo: "bar", "baz": false }); // this never throws

mixin.validate(bucket); // returns list of error messages or nothing on success

mixin.applyTo(bucket); // this throws if validation fails
```

mrgrain: Mixin Composition: https://github.com/aws/aws-cdk/pull/35089/files#diff-8be8cfe22ce079dd48e511a82e07e2b11e77d4ddf40e00a03c4f18199604001bR1518-R1539

mrgrain: Probably not required and we can implement this in the mixin compositor to use try-catch

mrgrain: props should be validated in constructor

mrgrain: does supports need to return information, e.g. a not supported because reason?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+CDK Mixins are composable, reusable abstractions that can be applied to any construct (L1, L2 or custom).
+They are breaking down the traditional barriers between construct levels,
+allowing customers to mix and match sophisticated features without being locked into specific implementations.
+
+## Working Backwards
+
+### README: CDK Mixins
+
+CDK Mixins provide a new, advanced way to add functionality to through composable abstractions.
```
jiayiwang7:
```suggestion
CDK Mixins provide a new, advanced way to add functionality through composable abstractions.
```

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+const bucket = new s3.CfnBucket(scope, "EnterpriseBucket")
+  .with(new EncryptionAtRest())
+  .with(new AutoDeleteObjects())
+  .with(new ComplianceLogging())
+  .with(new CostOptimization());
+
+// Access day-one AWS features with abstractions
+const bucketWithLatestFeature = new s3.CfnBucket(scope, "LatestBucket")
+  .with(new CfnBucketPropsMixin({ 
+    // New CloudFormation property available immediately
```
jiayiwang7, line 348: ah i was wondering before reading this section that - why would user ever want to use auto-generated mixins on L1s

and here ru saying that once the feature is available in service spec, u can use the `CfnXpropsMixin` to set up new properties dynamically here, instead of waiting for the L1 property is available in CDK version after we release like [this](https://github.com/aws/aws-cdk/pull/35816)  

my understanding correct?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+Mixins apply in declaration order. Later mixins can override earlier ones:
+
+```typescript
+// Last mixin wins for conflicting properties
+const bucket = new s3.CfnBucket(scope, "Bucket")
+  .with(new EncryptionAtRest({ algorithm: "AES256" }))
+  .with(new EncryptionAtRest({ algorithm: "aws:kms" })); // KMS wins
+
+// Mixins could detect and react to conflicts, but this is generally not encouraged
```
jiayiwang7: why?

text/0814-cdk-mixins.md
```diff
@@ -0,0 +1,758 @@
...
+
+// Mix L2 properties with mixin capabilities
+const hybridBucket = new s3.Bucket(scope, "HybridBucket", {
+  versioned: true,
+  lifecycleRules: [/* ... */]
+}).with(new AdvancedMonitoring())
+  .with(new SecurityCompliance());
+```
+
+#### Mixins and Aspects
```
jiayiwang7, line 404: should we also talk abt its comparision to property injections? i know they are applied in different ways and serving mostly different purpose but isnt there overlap use cases?

text/0814-cdk-mixins.md
mrgrain: Ofc, that's the point here.

text/0814-cdk-mixins.md
mrgrain: Clarified.

text/0814-cdk-mixins.md
mrgrain: sure we can make that work

text/0814-cdk-mixins.md
```diff
@@ -312,43 +375,73 @@ const hybridBucket = new s3.Bucket(scope, "HybridBucket", {
...
+Mixins are a new fundamental feature of the Constructs Programming Model (CPM).
+As such, the Mixin interface `IMixin` and the `.with()` method are introduced to to the `constructs` package.
 
-```typescript
-// Last mixin wins for conflicting properties
-const bucket = new s3.CfnBucket(scope, "Bucket")
-  .with(new EncryptionAtRest({ algorithm: "AES256" }))
-  .with(new EncryptionAtRest({ algorithm: "aws:kms" })); // KMS wins
+Every construct is recommended to implement it's own version of `.with()`.
+The default implementation is deliberately simple and lenient:
```
mrgrain: what lenient in this context

text/0814-cdk-mixins.md
```diff
@@ -312,43 +375,73 @@ const hybridBucket = new s3.Bucket(scope, "HybridBucket", {
...
+
+```ts
+class Construct implements IConstruct {
+  
+  // ...
+
+  public with(mixin: IMixin): void {
+    if (mixin.supports(this)) {
+      mixin.applyTo(this);
+    }
```
mrgrain: needs a deep iteration as default

text/0814-cdk-mixins.md
mrgrain: Child constructs inherit the construct ID path of their parents. So yes, users can know the ID. I don't think this is the most useful of all selectors, but shows what can be done.

text/0814-cdk-mixins.md
mrgrain: Decided against this. This introduces an otherwise useless intermediate object/type. Also a similar pattern can be achieved with a wrapper mixin: `SelectiveConstruct.selectAll(new BucketMixin())`

text/0814-cdk-mixins.md
mrgrain: yes

text/0814-cdk-mixins.md
mrgrain: The " Modular L2 Redesign" alternative is roughly what you have been [proposing before](https://github.com/aws/aws-cdk-rfcs/pull/657#discussion_r2036706731). Not necessarily exactly _that_, but stands for any approach that changes the shape of L2s, how they are created and used. There's no detail to it, so it's hard to argue about it more.

> The L2 construct basically becomes a composition of L1s and default mixin applications. If users don't like the L2 construct, they're free to apply mixins directly to L1s they created or extract the L1s from the L2 and apply them.

Agreed. That nicely captures the spirit and eventual goal of the Mixins proposal.

text/0814-cdk-mixins.md
mrgrain: this has to be string, otherwise jsii. We could have a selector that takes a callback, but then what's the point? Users can just implement the IConstructSelector interface.

text/0814-cdk-mixins.md
commiterate: In that case, would the above snippet that makes L2 constructs (e.g. `Function`) expose a property with a dictionary of L1 constructs (e.g. `Function.components`) nicely expose the L1 constructs with type information (no `IConstruct.nodes.children` with verbose `instanceof` matching to restore type information) without breaking backwards compatibility?

This could be enforced through an interface all higher level constructs (L2, L3) must implement (e.g. `IResource`). Said interface would require a `components: Record<string, IInspectable>` property (where L1 constructs must implement `IInspectable`).

mrgrain: I think that's an interesting idea that is worth thinking about. It fits nicely into other improvements we made recently like `I_Ref` interfaces. That makes it independent of this RFC though. Do you agree?

Like you said `IConstruct.nodes.children` already exists, it's just not strongly typed. An interface like `FunctionComponents` would help. The downside is that it limits refactoring to some extent (cannot remove implementation details anymore), that might be okay though.

I like the idea of `IInspectable` - what would it look like?

text/0814-cdk-mixins.md
mrgrain: principle of least surprise. added a blurb.

text/0814-cdk-mixins.md
commiterate: `IInspectable` [already exists](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.IInspectable.html). I just assumed that was some interface only implemented by L1's since those seem to be the only constructs that implement it.

The purpose of proposing that in the mixins RFC is because mixins are supposed to target L1 constructs and this lets users fish out the L1's to pass to a mixin function (e.g. `functionIamExecutionRoleMixin(props: { function: aws_lambda.CfnFunction, role: aws_iam.CfnRole })`).

That's assuming mixins are intended to be manually invoked at a per-construct level (e.g. `functionIamExecutionRoleMixin({ function: myFunction.components.function, role: myFunction.components.role })`). I haven't thought too hard about how such an interface would work if we applied a mixin to an entire construct tree like what's being proposed right now (closer to CSS selectors).

mrgrain: > IInspectable [already exists](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.IInspectable.html). I just assumed that was some interface only implemented by L1's since those seem to be the only constructs that implement it.

Oh haha, how embarrassing of me. I am aware of this interface, though clearly I didn't remember the name. I'm not sure how this interface helps with the problem though. Was that just to pick a type that represents all L1s?

> The purpose of proposing that in the mixins RFC is because mixins are supposed to target L1 constructs and this lets users fish out the L1's to pass to a mixin function (e.g. functionIamExecutionRoleMixin(props: { function: aws_lambda.CfnFunction, role: aws_iam.CfnRole })).

Okay gotcha, thanks for the clarification. The intention of this proposal is that Mixins can be applied to either the L2 or the L1 or even a custom version of an L2. Implementation-wise Mixins should be at L1 level, but are applied to the tree and will determine if they are applicable ("support") a construct.

> I haven't thought too hard about how such an interface would work if we applied a mixin to an entire construct tree like what's being proposed right now (closer to CSS selectors).

We have a provision for this in the advances `Mixins.of(...)` mode. Currently not implementing a CSS Selector like syntax, but with this proposal it could be easily done in userland.

commiterate: Yup I was just using `IInspectable` to represent L1s, apologies for the confusion.

PR #826, opened 2025-10-14
RFC 789: Bedrock AgentCore Gateway L2 construct

dineshSajwan:
This is a request for Amazon Bedrock AgentCore L2 construct that simplifies the deployment and management of AI agents at scale by wrapping the AgentCore L1 constructs. It provides a high-level, object-oriented approach to creating and managing Amazon Bedrock AgentCore resources. This enables developers to deploy highly effective agents securely using any framework and model. The package will provide constructs for the following primitives:

**AgentCore Gateway**

See # https://github.com/aws/aws-cdk-rfcs/issues/785 for
additional details.

---

### DIFF
--- a/FULL_INDEX.md
+++ b/FULL_INDEX.md
@@ -201,8 +201,9 @@
 [764](https://github.com/aws/aws-cdk-rfcs/issues/764)|[L2 Constructs for Bedrock AgentCore Memory](https://github.com/aws/aws-cdk-rfcs/issues/764)||💡 proposed
 [785](https://github.com/aws/aws-cdk-rfcs/issues/785)|[Bedrock AgentCore L2 alpha package](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0785-bedrock-agentcore-runtime-l2.md)||💡 proposed
 [789](https://github.com/aws/aws-cdk-rfcs/issues/789)|[EC2 Image Builder L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0789-imagebuilder-l2.md)||👷 implementing
-[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||✍️ review
+[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||⏰ final comments
 [808](https://github.com/aws/aws-cdk-rfcs/issues/808)|[Library Quality Scoring](https://github.com/aws/aws-cdk-rfcs/issues/808)||⏰ final comments
 [814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||💡 proposed
 [835](https://github.com/aws/aws-cdk-rfcs/issues/835)|[Native EBS Snapshot Asset Handling](https://github.com/aws/aws-cdk-rfcs/issues/835)||💡 proposed
+[840](https://github.com/aws/aws-cdk-rfcs/issues/840)|[proposal+titleRFC](https://github.com/aws/aws-cdk-rfcs/issues/840)||💡 proposed
 <!--END_TABLE-->
--- a/PROPOSED.md
+++ b/PROPOSED.md
@@ -34,6 +34,7 @@
 [785](https://github.com/aws/aws-cdk-rfcs/issues/785)|[Bedrock AgentCore L2 alpha package](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0785-bedrock-agentcore-runtime-l2.md)||💡 proposed
 [814](https://github.com/aws/aws-cdk-rfcs/issues/814)|[Unified mechanism for abstractions that change constructs after construction](https://github.com/aws/aws-cdk-rfcs/issues/814)||💡 proposed
 [835](https://github.com/aws/aws-cdk-rfcs/issues/835)|[Native EBS Snapshot Asset Handling](https://github.com/aws/aws-cdk-rfcs/issues/835)||💡 proposed
-[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||✍️ review
+[840](https://github.com/aws/aws-cdk-rfcs/issues/840)|[proposal+titleRFC](https://github.com/aws/aws-cdk-rfcs/issues/840)||💡 proposed
+[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||⏰ final comments
 [808](https://github.com/aws/aws-cdk-rfcs/issues/808)|[Library Quality Scoring](https://github.com/aws/aws-cdk-rfcs/issues/808)||⏰ final comments
 <!--END_TABLE-->
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ future state of the libraries and to discover projects for contribution.
 [673](https://github.com/aws/aws-cdk-rfcs/issues/673)|[AWS CloudWatch Application Signals L2 Constructs for SLO](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0673-aws-applicationsignals-slo-l2.md)||👷 implementing
 [732](https://github.com/aws/aws-cdk-rfcs/issues/732)|[CDK CLI Telemetry](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0732-cdk-cli-telemetry.md)|[@kaizencc](https://github.com/kaizencc)|👷 implementing
 [789](https://github.com/aws/aws-cdk-rfcs/issues/789)|[EC2 Image Builder L2 Constructs](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0789-imagebuilder-l2.md)||👷 implementing
+[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||⏰ final comments
 [808](https://github.com/aws/aws-cdk-rfcs/issues/808)|[Library Quality Scoring](https://github.com/aws/aws-cdk-rfcs/issues/808)||⏰ final comments
-[792](https://github.com/aws/aws-cdk-rfcs/issues/792)|[L2 Construct for AIOps](https://github.com/aws/aws-cdk-rfcs/issues/792)||✍️ review
 <!--END_TABLE-->
 
 ## What is an RFC?


text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+  }),
+});
+```
+
+**IAM** – Authorizes through the credentials of the AWS IAM identity trying to access the gateway.
+
+```typescript fixture=default
+const gateway = new agentcore.Gateway(this, "MyGateway", {
+  gatewayName: "my-gateway",
+  authorizerConfiguration: agentcore.GatewayAuthorizer.awsIam,
```
alvazjor: Lets keep the name pattern consistent with the standard we already defined: `GatewayAuthorizer.usingIAM()`, similar to the name we defined in the `RuntimeAuthorizerConfiguration`

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+    supportedVersions: [agentcore.MCPProtocolVersion.MCP_2025_03_26],
+  }),
+});
+```
+
+### Inbound authorization
+
+Before you create your gateway, you must set up inbound authorization. Inbound authorization validates users who attempt to access targets through
+your AgentCore gateway. By default, if not provided, the construct will create and configure Cognito as the default identity provider
+(inbound Auth setup). AgentCore supports the following types of inbound authorization:
```
alvazjor: Before we go into custom specification, lets add an example of how the cognito default will look like, because the customers will still need to send some configuration parameters

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+  assumedBy: new iam.ServicePrincipal("lambda.amazonaws.com"),
+});
+
+// The Lambda needs permission to invoke the gateway
+gateway.grantInvoke(lambdaRole);
+```
+
+### Gateway with KMS Encryption
+
+You can provide a KMS key, and configure the authorizer as well as the protocol configuration.
```
alvazjor: Lets add a couple of sentences what will that allow, pretty much explain here what is the use case of using KMS, so the customers have a clearer understanding on when should they use it

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+    discoveryUrl: "https://auth.example.com/.well-known/openid-configuration",
+    allowedAudience: ["my-app"],
+    allowedClients: ["my-client-id"],
+  }),
+  kmsKey: encryptionKey,
+  exceptionLevel: agentcore.GatewayExceptionLevel.DEBUG,
+});
+```
+
+### Gateway with Custom Execution Role
```
alvazjor, line 178: Lets add a bit of explanation on this: User is allowed to provide their own custom role.... and so on

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+a more structured approach to defining APIs compared to OpenAPI, and are particularly useful for connecting to AWS services.
+AgentCore Gateway supports built-in AWS service models only. It Connects to services using Smithy model definitions
+
+- Supports OAUTH and API_KEY credential providers
+- Ideal for AWS service integrations
+- Need API schema. The construct provide 3 ways to upload a API schema to Smity target
+
+> Note: For Smithy model targets that access AWS services, your Gateway's execution role needs permissions to access those services.
+For example, for a DynamoDB target, your execution role needs permissions to perform DynamoDB operations.
+This is not managed by the construct due to the large number of options.
```
alvazjor: This is very important information. Lets be sure that is properly documented also in the README of the resource in the library

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+        properties: {
+          message: {
+            type: agentcore.SchemaDefinitionType.STRING,
+            description: "The greeting message",
+          },
+        },
+      },
+    },
+  ]),
+  credentialProviderConfigurations: [agentcore.GatewayCredentialProvider.iamRole()],
```
alvazjor: Should we not restrict already the type of cred provider configuration that can be passed? This line implines 2 things:
- The user needs to be aware only iam role is possible for lambda. This is documented, but I wonder if we can make it easier for them
- We need to validate that they wont pass anything else

How feasible would it be to already ask here directly for just IAM creds providers? And I think the same applies for the other types. Can we just accept the supported types? Was this considered?

Also, a diff question, can more thatn 1 be given in the array?

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,786 @@
...
+The Gateway Target construct provides convenient methods for granting IAM permissions:
+
+```typescript fixture=default
+// Create a gateway and target
+const gateway = new agentcore.Gateway(this, "MyGateway", {
+  gatewayName: "my-gateway",
+  protocolConfiguration: new agentcore.McpProtocolConfiguration({
+    instructions: "Use this gateway to connect to external MCP tools",
+    searchType: agentcore.McpGatewaySearchType.SEMANTIC,
+    supportedVersions: [agentcore.MCPProtocolVersion.MCP_2025_03_26],
```
alvazjor: Where is this version format coming from?

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Versions are available at https://github.com/modelcontextprotocol/modelcontextprotocol/releases
This is present in JSdoc of the code. Will add it to readme too.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: In implementation we have restricted it to `ICredentialProvider` and `GatewayCredentialProvider.iamRole() `returns `ICredentialProvider`. one can only pass `GatewayIamRoleCredentialProvider` (`GatewayCredentialProvider.iamRole()` return `GatewayIamRoleCredentialProvider`) , `ApiKeyCredentialProviderConfiguration` or `OAuthCredentialProviderConfiguration` which implements `ICredentialProvider`.

`credentialProviderConfigurations?: ICredentialProvider[];` 

where ICredentialProvider is 
```
export interface ICredentialProvider {
  /**
   * The credential provider type
   */
  readonly credentialProviderType: CredentialProviderType;

  /**
   * Renders as CFN Property
   *  @internal
   */
  _render(): CfnGatewayTarget.CredentialProviderConfigurationProperty | IResolvable;

  /**
   * Grant the role the permissions
   */
  grantNeededPermissionsToRole(role: IRole): Grant | undefined;
}
```

dineshSajwan: Screenshot 2025-10-23 at 2 51 15 PM
As per this smithy does support multiple credential providers but I am not sure if that will be the case in real world.

alvazjor: What I meant is: since `OAuthCredentialProviderConfiguration` and `ApiKeyCredentialProviderConfiguration` also return a `ICredentialProvider` type, the user could potentially try to send those when using the Lambda target. But that will fail, since is should only accept `GatewayIamRoleCredentialProvider`:
```
const target = agentcore.GatewayTarget.forLambda(this, "MyLambdaTarget", {
  ... other props,
  credentialProviderConfigurations: [agentcore.GatewayCredentialProvider.oauth({PARAMS})], 
```
This code is syntactically correct, but logically incorrect, because I am using the oath one (that is also a `ICredentialProvider`) but I should not. 
So my question is more on the fact if we could already prevent this from being constructed, rather than to verify if the param combination is correct when the user tries to synth the code. But if this is too complex, then I guess is ok to validate the correct combinations in our end, and document it clearly in the JSDoc of each target type

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,805 @@
...
+  }),
+  authorizerConfiguration: agentcore.GatewayAuthorizer.usingCustomJwt({
+    discoveryUrl: "https://auth.example.com/.well-known/openid-configuration",
+    allowedAudience: ["my-app"],
+    allowedClients: ["my-client-id"],
+  }),
+});
+
+// outbound auth (Use AWS console to create it, Once Identity L2 construct is available you can use it to create identity)
+const apiKeyIdentityArn = "your-idp-arn"
```
alvazjor: As I am understanding, this Arn will be the Identiy Arn, correct? Do we have a plan on how to migrate the code, once the identity L2 is available? 
If no, then I would suggest to name the methods that will receive this Arn param as `agentcore.GatewayCredentialProvider.fromXXXXArn` and reserve the `agentcore.GatewayCredentialProvider.apiKey` (more generic name) to the actual L2 once we have it.

so, they could be named:
`agentcore.GatewayCredentialProvider.fromApiKeyIdentityArn`
`agentcore.GatewayCredentialProvider.fromOauthIdentityArn`

And then when the Identity L2 exist, the will have:
`agentcore.GatewayCredentialProvider.fromApiKeyIdentityArn`
`agentcore.GatewayCredentialProvider.fromOauthIdentityArn`
`agentcore.GatewayCredentialProvider.apiKey`// this one will accept the L2 object rather than Arn string
`agentcore.GatewayCredentialProvider.oath` // this one will accept the L2 object rather than Arn string

What do you think?

alvazjor (dismissed):

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,805 @@
...
+
+// Grant manage permissions (Create, Update, Delete actions)
+target.grantManage(userRole);
+
+// Grant specific custom permissions
+target.grant(userRole, "bedrock-agentcore:GetGatewayTarget");
+
+
+// Grants permission to invoke this Gateway
+gateway.grantInvoke(grantee: iam.IGrantable);
```
alvazjor: Is this a typo? is the invoke permission for the gateway or for the specific target? @kumvprat fo visibility

dineshSajwan: `gateway.grantInvoke()` grants user access to the gateway, not target access. By default the construct create an execution role which handles the target invocation implicitly. And if there is any extra permission needed to be added to the role then it can be added manually using` gateway.role.addToPolicy()`

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Very good suggestion. Along with providerArn we also need secretArn (to set the permission, missed it earlier, will update the readme sample to add that too). 

I will add these methods to construct 

`agentcore.GatewayCredentialProvider.fromApiKeyIdentityArn`
`agentcore.GatewayCredentialProvider.fromOauthIdentityArn`
`agentcore.GatewayCredentialProvider.apiKey`// this one will accept the L2 object rather than Arn string
`agentcore.GatewayCredentialProvider.oath` // this one will accept the L2 object rather than Arn string

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: If a user gets invoke access to the Gateway are they able to implicitly invoke all targets?
The gateway can invoke and and all target which seems reasonable, BUT with this it would not be possible to limit user's access to certain targets right?

alvazjor: Does that mean that if I am allowlisted to assume the gateway role, I will also be allowlisted to use any of the targets?

dineshSajwan: No the user will not be invoke all the targets implicitly because we have outbound auth for all the targets. By default the permission will be set but the user still need JWT token to access the targets.  For example by default the gateway will set the permission to the role to access a lambda for the lambda target but then the user need to set the JWT token to invoke the lambda from the gateway.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Make sense. I will check with the implementation if we can strongly type that. If not then will add the validations.

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: This where couple of question of mine come : 
- Is it possible to have a target without authentication/authorization setup ?Because this property `credentialProviderConfigurations` seems to be optional, if a cdk user doesn't set it will it default to unauthenticated target invocations ? 

- If not and we default to  value as `[GatewayCredentialProvider.iamRole()]` which would make the some targeta not usable with default value. (since openapi/smithy targets don't support the default type)

dineshSajwan: If no `credentialProviderConfigurations` is set then it default to I AM and you can only invoke Lambda target with it.

dineshSajwan: For openAPI you must have `credentialProviderConfigurations`

dineshSajwan: Updated the different target type section with below information. Smithy and Lambda support IAM and for openAPI we must provide `credentialProviderConfigurations`

**Lambda Target**
When using the default IAM authentication (no `credentialProviderConfigurations` specified),
  the construct automatocally grants the gateway role permission to invoke your Lambda function (`lambda:InvokeFunction`).
  
  **OpenAPI Schema Target**
  Supports OAUTH and API_KEY credential providers (Do not support IAM, you must provide `credentialProviderConfigurations`)
  
  **Smithy Model Target** 
  When using the default IAM authentication (no `credentialProviderConfigurations` specified), The construct only
  grants permission to read the Smithy schema file from S3. You MUST manually grant permissions for the gateway
  role to invoke the actual Smithy API endpoints.

kumvprat: I see an new field called `secretArn` in the `apiKey` object. Is that change related to this ?

Also 
> You MUST manually grant permissions for the gateway role to invoke the actual Smithy API endpoints

Do we plan to provide a way to add the execution permission for these endpoints via CDK ? Like a property to provide an `arn` ? If this has already been discussed you can just point me in that direction

For [Gateway Target Table](https://github.com/aws/aws-cdk-rfcs/blob/0c9c74f013f339ea606c10d7c850dbcc2c6ceca0/text/0789-bedrock-agentcore-gateway-l2.md#gateway-target) we should add a note for `credentialProviderConfigurations` so that users know to look further in documentation for `OpenApi` targets.

dineshSajwan: Since Identity module is not available yet hence we need to use the arn (`proverarn` and `secretarn`) to set up the auth and permission accordingly. I missed to mention the `secretarn` earlier.

> grant permissions for the gateway role to invoke the actual Smithy API endpoints
The construct as of now do not support it and expect the user to set the permission. This can be a future feature request.

Gateway target table updated with a note for openapi

text/0788-bedrock-agentcore-memory-l2.md
```diff
@@ -296,7 +296,7 @@ const memory = new agentcore.Memory(this, "MyMemory", {
 
 ### Memory with self-managed Strategies
 
-A self-managed strategy in Amazon Bedrock AgentCore Memory gives you complete control over your memory extraction and consolidation pipelines. 
+A self-managed strategy in Amazon Bedrock AgentCore Memory gives you complete control over your memory extraction and consolidation pipelines.
```
kumvprat, line 299: Not sure why memory RFC changes are being reflected here but can we remove this if it's not needed

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+});
+
+// The Lambda needs permission to invoke the gateway
+gateway.grantInvoke(lambdaRole);
+```
+
+### Gateway with KMS Encryption
+
+You can provide your own AWS KMS customer-managed key (CMK) for greater control over the encryption process of your
+gateway data. If you don't specify a KMS key, AWS encrypts the gateway with an AWS-managed key. Use a customer-managed
```
kumvprat, line 157: >If you don't specify a KMS key, AWS encrypts the gateway with an AWS-managed key.

Does this mean encryption is enabled by default ?
I assume the encryption actually encrypt the communication between the Gateway and targets right ?

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+KMS key when you need to meet specific compliance requirements, implement custom key rotation policies, maintain
+audit trails through CloudTrail, or enforce fine-grained access controls on who can encrypt/decrypt gateway data. For
+more information, see Encrypt your AgentCore gateway with a customer-managed KMS key.
```
kumvprat: I think there is a documentation link missing for the following section 

> see Encrypt your AgentCore gateway with a customer-managed KMS key.

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+
+```typescript fixture=default
+// Create a KMS key for encryption
+const encryptionKey = new kms.Key(this, "GatewayEncryptionKey", {
+  enableKeyRotation: true,
+  description: "KMS key for gateway encryption",
+});
+
+// Create gateway with KMS encryption
+const gateway = new agentcore.Gateway(this, "MyGateway", {
```
kumvprat, line 170: Since `protocolConfiguration` and `authorizerConfiguration` are optional parameters can we exclude them from example where it's not needed. Here the verbose creation of the Gateway hides away the `kmsKey` property that you are supposed to set.

If these properties become requirement when `kmsKey` is enabled let's call that out as well

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+// Create a custom execution role
+const executionRole = new iam.Role(this, "GatewayExecutionRole", {
+  assumedBy: new iam.ServicePrincipal("bedrock-agentcore.amazonaws.com"),
+  managedPolicies: [
+    iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonBedrockAgentCoreGatewayExecutionRolePolicy"),
+  ],
+});
+
+// Create gateway with custom execution role
+const gateway = new agentcore.Gateway(this, "MyGateway", {
```
kumvprat, line 197: Since `protocolConfiguration` and `authorizerConfiguration` are optional parameters can we exclude them from example where it's not needed. Here the verbose creation of the Gateway hides away the `role` property that you are supposed to set.

If these properties become requirement when `role` is enabled let's call that out as well

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+
+### Gateway Target Properties
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `targetName` | `string` | Yes | The name of the gateway target. Valid characters are a-z, A-Z, 0-9, _ (underscore) and - (hyphen) |
+| `description` | `string` | No | Optional description for the gateway target. Maximum 200 characters |
+| `gateway` | `IGateway` | Yes | The gateway this target belongs to |
+| `targetConfiguration` | `ITargetConfiguration` | Yes | The target configuration (Lambda, OpenAPI, or Smithy). Use `LambdaTargetConfiguration.create()`, `OpenApiTargetConfiguration.create()`, or `SmithyTargetConfiguration.create()` |
+| `credentialProviderConfigurations` | `IGatewayCredentialProvider[]` | No | Credential providers for authentication. Defaults to `[GatewayCredentialProvider.iamRole()]`. Use `GatewayCredentialProvider.apiKey()`, `GatewayCredentialProvider.oauth()`, or `GatewayCredentialProvider.iamRole()`. **Note:** OpenAPI targets do not support IAM authentication and require explicit `credentialProviderConfigurations` (API Key or OAuth). See [Targets types](#targets-types) for details |
```
kumvprat: Thanks for adding this, makes it visible and allows user to be careful when creating such targets

> **Note:** OpenAPI targets do not support IAM authentication and require explicit `credentialProviderConfigurations` (API Key or OAuth)

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+
+- Supports GATEWAY_IAM_ROLE credential provider only
+- Ideal for custom serverless function integration
+- Need tool schema (tool schema is a blueprint that describes the functions your Lambda provides to AI agents).
+  The construct provide 3 ways to upload a tool schema to Lambda target
+- When using the default IAM authentication (no `credentialProviderConfigurations` specified),
+  the construct automatocally grants the gateway role permission to invoke your Lambda function (`lambda:InvokeFunction`).
+
+**OpenAPI Schema Target** : OpenAPI widely used standard for describing RESTful APIs. Gateway supports OpenAPI 3.0
+specifications for defining API targets. It onnects to REST APIs using OpenAPI specifications
```
kumvprat: nit: c is missing from connects

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: Adding to this : Is there any documentation link or a page where we can route users for learning more about the callout ? Adding that can help us direct users to the necessary information needed for Smithy targets creation

> For Smithy model targets that access AWS services, your Gateway's execution role needs permissions to access those services.
For example, for a DynamoDB target, your execution role needs permissions to perform DynamoDB operations.

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+
+You can create the following targets types:
+
+**Lambda Target**: Lambda targets allow you to connect your gateway to AWS Lambda functions that implement your tools. This is useful
+when you want to execute custom code in response to tool invocations.
+
+- Supports GATEWAY_IAM_ROLE credential provider only
+- Ideal for custom serverless function integration
+- Need tool schema (tool schema is a blueprint that describes the functions your Lambda provides to AI agents).
+  The construct provide 3 ways to upload a tool schema to Lambda target
```
kumvprat: I see this called out for all the targets, can we also hyperlink the section of the document that explains the possible ways to attach tool schema(I think it's described a few sections below)

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+    path.join(__dirname, "schemas", "my-tool-schema.json")
+  ),
+```
+
+- From an existing S3 file:
+
+```typescript
+
+  toolSchema: agentcore.ToolSchema.fromS3File(
+    s3.Bucket.fromBucketName(this, "SchemasBucket", "my-schemas-bucket"),
```
kumvprat, line 303: The example leverages our referenced resource concepts, can we have examples with both referenced and a S3 bucket that's created by the user in the same cdk app

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+    "tools/complex-tool-schema.json",
+    "123456789012"
+  ),
+```
+
+```typescript
+const toolSchema: agentcore.ToolSchema.fromInline( {
```
kumvprat, line 324: Is this the third way to add ToolSchema to targets ? All target allow 3 ways to add tool schema and I assume this is the third one.
Can we just call that out as a small point above this line? 

If that's not the case, what is the third way to do this.

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+- From an existing S3 file:
+
+```typescript fixture=default
+const bucket = s3.Bucket.fromBucketName(this, "ExistingBucket", "my-schema-bucket");
+const s3Schema = agentcore.ApiSchema.fromS3File(bucket, "schemas/action-group.yaml");
+```
+
+### Outbound auth
+
+Outbound authorization lets Amazon Bedrock AgentCore gateways securely access gateway targets on behalf of users authenticated
```
kumvprat, line 384: I am not sure about this 

Does this mean inbound and outbound authorization are 2 separate processes when invoking a tool via Gateway ? 

If so then I think this section lacks example, even if that's not in scope of Gateway can we link documentation for users to understand how to set this up? 

If I missed it then you can just point me in the right direction.

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+
+**API key** – Use the AgentCore service/AWS console to generate an API key to authenticate access to the gateway target.
+
+**Note > You need to set up the outbound identity before you can create a gateway target.
+
+### Basic Gateway Target Creation
+
+You can create targets in two ways: using the static factory methods on `GatewayTarget` or using the convenient `addTarget` methods on the gateway instance.
+
+#### Using addTarget methods (Recommended)
```
kumvprat, line 403: Just a nit but can we have consistent examples for both `addTarget` and static method factory ? 

If the inconsistent examples are intentional, can we call this out in the section ?

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: > Grants permission to invoke this Gateway
gateway.grantInvoke(grantee: iam.IGrantable);

😅 Looks like method definition instead of invocation is this intentional ?

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,818 @@
...
+The construct library provides interfaces for Gateway services:
+
+### Gateway Interfaces
+
+- `IGateway` - Represents a Gateway resource
+- `IGatewayTarget` - Represents a Gateway Target resource
+- `IGatewayProps` - Properties for creating a Gateway
+- `IGatewayTargetProps` - Properties for creating a Gateway Target
+- `IGatewayProtocol` - Protocol configuration interface
+- `IGatewayAuthorizer` - Authorizer configuration interface
```
kumvprat: Should add configuration/props as a suffix here ? 

If I am understanding this correctly, we are creating this interface to provide configuration for authorizers and not creating the authorizer itself.

Currently it looks like we are creating the authorizer itself.

text/0788-bedrock-agentcore-memory-l2.md
dineshSajwan: > Looks like when merging with main branch an extra space got added for memory.

Checked again and found that main branch has this extra space causing a linting  error and build failure.   `text/0788-bedrock-agentcore-memory-l2.md:299:143 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]`

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Yes. The underlying service does that and it encrypts the data at rest not sure of transit (Ideally it should but can't find it in the documentation). - https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-encryption.html

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Yes , adding the link back.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Make sense. Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Yes. We have separate sections for Inbound and Outbound auth in the RFC. 

-  **Inbound authorization** validates users who attempt to access targets through your AgentCore gateway.  validates incoming request to gateway.
-  **Outbound authorization** lets Amazon Bedrock AgentCore gateways securely access gateway targets on behalf of users that were authenticated and authorized during inbound authorization. Validates outgoing request from gateway to target.

The inbound auth is required before user can create gateway, hence it is at the top  right below protocol configuration / basic gateway creation section. 

The outbound auth is for target hence it is in gateway target section.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack. Removed optional params for all the examples to keep it simple and easy to understand.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: The naming is intentional and follows a factory pattern design. `IGatewayAuthorizer` and `IGatewayProtocol` are common interfaces that define the contract for all configuration types .  It use property names `authorizerConfiguration`, `protocolConfiguration` to  make it clear that these are configurations and not actual  instances.

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: I assumed gateway was a pass through like APIGW for non-agentcore use.

The encryption of data-at-rest does not make sense to me if this just a gateway connecting a user call to an executable target.

Also maybe I missed it but do the examples here share details around how it will tie up with AgentCoreRuntime constructs ?

dineshSajwan: No we don't have any example because that will increase the complexity, we will have to add data plane operations to invoke gateway from runtime. However we are planning to publish few  samples in sample org, we can link those samples to construct readme later.

dineshSajwan: for all transit communication TLS is implemented by default - https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/data-encryption.html

kumvprat: The documentation here : https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/data-protection.html talks about where the customer data is stored but it's not clear why do we need to store that data. 

Anyway that part is a service level question and out of scope for this RFC.


BUT we should call out that this kmsKey support encrypts data-at-rest and the TLS is enabled by default. Also attach the link you shared :  https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/data-encryption.html

This becomes particularly important if for future use-cases we plan to allow users to configure TLS version and other properties for their Agentcore Gateway

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,757 @@
...
+credential provider attached enabling you to securely access targets whether they need IAM, API Key, or OAuth credentials.
+
+### Gateway Target Properties
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| `targetName` | `string` | Yes | The name of the gateway target. Valid characters are a-z, A-Z, 0-9, _ (underscore) and - (hyphen) |
+| `description` | `string` | No | Optional description for the gateway target. Maximum 200 characters |
+| `gateway` | `IGateway` | Yes | The gateway this target belongs to |
+| `targetConfiguration` | `ITargetConfiguration` | Yes | The target configuration (Lambda, OpenAPI, or Smithy). Use `LambdaTargetConfiguration.create()`, `OpenApiTargetConfiguration.create()`, or `SmithyTargetConfiguration.create()` |
```
kumvprat: This is labelled as Required but I don't see it's use in any of the examples below.

Is this still required ? 
If no we can remove this from all the places in RFC.
If yes, we should include it in all the target creation examples we have.

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: I guess my question here is : 
- Do we create authorizers from these configurations ?
- Do we create protocol from these configurations ?

If yes, then the configuration objects should have types that tell the user that it's a configuration object.


For example here : [Api GW integ test](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-apigateway/test/authorizers/integ.cognito-authorizer.ts#L29-L44)

A cognito authorizer is created outside of the Api GW and then passed on to the Api Gateway object.

If we are doing something similar then we should rename the fields from `authorizerConfiguration` -> `authorizer` and similar for protocol

If the configuration objects are internally(inside AgentCoreGateway) used to create these authorizer and we want to abstract this away from users, the proper naming should be used : `IGatewayAuthorizer` -> `IGatewayAuthorizerConfig` and `IGatewayProtocol` -> `IGatewayProtocolConfig`


Please point out,  if I am missing something

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack. Added a link.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Typo, updated to invocation.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Yes configuration objects are internally used to create these authorizer . Using the suggested naming.  
`IGatewayAuthorizer` -> `IGatewayAuthorizerConfig` and `IGatewayProtocol` -> `IGatewayProtocolConfig`

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Good catch. `targetConfiguration` is a required by underlying service. The gateway L2 construct provide few convenience methods like `LambdaTargetConfiguration.create()`, `OpenApiTargetConfiguration.create()`, or `SmithyTargetConfiguration.create()` which automatically set the `targetConfiguration` for the user. However, if someone don't want to use these convenience methods then they can pass the `targetConfiguration` directly.  I have updated the RFC with an example of Advanced Usage: Direct Configuration for gateway target.

dineshSajwan: I realized I have the wrong method names . Updated `LambdaTargetConfiguration.create() `-> `GatewayTarget.forLambda(),` `OpenApiTargetConfiguration.create() `->`GatewayTarget.forOpenApi() `, `SmithyTargetConfiguration.create()  `-> `GatewayTarget.forSmithy()`

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: Outbound authorization : The explanation in RFC does not tell us how / by which property will we set outbound authorization. And only open api examples in target creation talk about outbound auth setup via the credentialProviderConfigurations and other examples dont. 
Can we 
1) add it(outbound auth) for all examples ? OR 
2)give  examples for outbound auth in the Outbound authorization section (although outbound auth just by itself won't make sense, then first choice would be the way to go)

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,787 @@
...
+  exceptionLevel: agentcore.GatewayExceptionLevel.DEBUG,
+});
+```
+
+### Gateway with Custom Execution Role
+
+By default, the Gateway construct creates an execution role with minimal permissions that are automatically
+expanded as you add targets. You can provide your own custom IAM role when you need to comply with organizational IAM
+policies or maintain centralized control over permissions. When providing a custom role, ensure it's assumable by the
+bedrock-agentcore.amazonaws.com service principal and includes necessary permissions for your targets (Lambda
```
kumvprat: Can we highlight the prinicpal ? Currently it's lost in the text

`bedrock-agentcore.amazonaws.com`

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: The  lambda target and Smithy target only support  GATEWAY_IAM_ROLE as outbound auth . The user don't have to set `credentialProviderConfigurations` explicitly because the construct does it by default for these two targets. I have updated this in Targets types  section and also added a commented line in the examples - 

> credentialProviderConfigurations is not required for lambda/smithy target the underlying service by default set it as GATEWAY_IAM_ROLE

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: Is it possible to strongly type the `credentialProvider` if not we can close this comment and handle it in implementation.

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: Can you link the doc in the `Protocol configuration` section ?

text/0789-bedrock-agentcore-gateway-l2.md
kumvprat: The table(`Gateway Target Properties`) talks about `GatewayCredentialProvider.apiKey()  and the example use the arn based factory methods. From what I gather this is bound to go through some small changes when Identity implementation comes into picture. Can we add explanation in the table Gateway Target Properties that explains that if a user doesnot want to use the Identity L2 they can also use the arn based factory methods, like the ones used in the examples below

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,794 @@
...
+```
+
+- From an existing S3 file:
+
+```typescript fixture=default
+const bucket = s3.Bucket.fromBucketName(this, "ExistingBucket", "my-schema-bucket");
+const s3Schema = agentcore.ApiSchema.fromS3File(bucket, "schemas/action-group.yaml");
+```
+
+### Outbound authorization
```
kumvprat, line 382: Can we call out here the exact property used to set proper outbound authorization on the gateway ?

dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Thank you for catching that, it should be other way round I guess. Since we don't have any Identity module yet hence this RFC only refer `fromXXXarn` methods.
I have updated the table to use `GatewayCredentialProvider.fromApiKeyIdentityArn()`, `GatewayCredentialProvider.fromOauthIdentityArn()`, or `GatewayCredentialProvider.fromIamRole()`.  Also mentioned in note that when Identity will be available we will add new methods like `GatewayCredentialProvider.apiKey()` ,`GatewayCredentialProvider.oauth()`

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: I have updated `credentialProvider` type to `ICredentialProvider` and updated the implementation to use specific interface for specific target type to strongly type it.   

`ILambdaCredentialProvider` extends `ICredentialProvider` - only support IAM
`IOpenApiCredentialProvider` extends `ICredentialProvider` - only support oauth and API
`ISmithyCredentialProvider` extends `ICredentialProvider` -  only support IAM
now for smithy target the `credentialProviderConfigurations` will only support `ISmithyCredentialProvider` and so on for other target. This way user cannot provide unsupported

kumvprat: Is there any reason for this to be tied to the target type ? 

I assume we can have new target types and with each new target type if we bring a new credential provider it will not be a scalable approach.

Also authorization as a concept would be agnostic to the underlying target, keeping the ICredentialProvider and it's specific implementations of Iam, Oauth and api key would be ideal and we can have validations around which credential provider works with what kind of target.

Interfaces like `ILambdaCredentialProvider` won't bring a lot of value as the AWS service in itself won't be extensible.

Happy to hear thoughts around this as this is a design question and one of the critical things before closing the RFC.

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,796 @@
...
+| `credentialProviderConfigurations` | `ICredentialProvider[]` | No | Credential providers for authentication. Defaults to `[GatewayCredentialProvider.fromIamRole()]`. Use `GatewayCredentialProvider.fromApiKeyIdentityArn()`, `GatewayCredentialProvider.fromOauthIdentityArn()`, or `GatewayCredentialProvider.fromIamRole()`. Since the Identity L2 construct is not available we can use these arn based factory methods to set the auth, Once Idenetity is available we will add new methods `GatewayCredentialProvider.apiKey()` ,`GatewayCredentialProvider.oauth()`.  **Note:** OpenAPI targets do not support IAM authentication and require explicit `credentialProviderConfigurations` (API Key or OAuth). See [Targets types](#targets-types) for details |
```
kumvprat: nit: Typo in `Idenetity`

text/0789-bedrock-agentcore-gateway-l2.md
```diff
@@ -0,0 +1,796 @@
...
+
+- Supports `OAUTH` and `API_KEY` for [outbound authorization](#outbound-authorization). (Do not support `GATEWAY_IAM_ROLE`, you must provide `credentialProviderConfigurations`)
+- Ideal for integrating with external REST services
+- Need API schema. The construct provide [3 ways to upload a API schema to OpenAPI target](#api-schema-for-openapi-and-smithy-target)
+
+**Smithy Model Target** : Smithy is a language for defining services and software development kits (SDKs). Smithy models provide
+a more structured approach to defining APIs compared to OpenAPI, and are particularly useful for connecting to AWS services.
+AgentCore Gateway supports built-in AWS service models only. It connects to services using Smithy model definitions
+
+- Supports OAUTH and API_KEY for [outbound authorization](#outbound-authorization), The construct by default set
```
kumvprat: nit: Missing quotes around OAUTH and API_KEY

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: ack.

text/0789-bedrock-agentcore-gateway-l2.md
dineshSajwan: Agreed. The validation is already there in underlying service . If you assign an OAUTH/API to smithy target the deployment fails with the validation. The reason I added these marker interface because the previous comment was to check if we can make it strongly type to get the error at compile time then at deployment time. 
The new target types will be under a generic target called integration target ( which will support for target like slack, amazon, asana etc.) so we will have just one more marker interface.
Yes , it increases the complexity but the intention was to gain the compile type check as suggested. 
I am fine to keep the validation and use `ICredentialProvider` .

kumvprat:
@mergifyio refresh

kumvprat:
@mergifyio queue

Issue #886, opened 2026-03-20
Making Hotswap Faster


ShadowCat567:
## Description

Improvements to hotswap which serve the goal of making it faster 

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @ShadowCat567 
| Author(s)           | @ShadowCat567 
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | 

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [ ] RFC pull request submitted (label: `status/review`)
- [ ] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #887, opened 2026-03-20
RFC 886: Making Hotswap Faster

ShadowCat567:
This is a request for comments about improvements to hotswap that will make it faster. See #886 for
additional details. 

---

### DIFF
--- a/text/0886-hotswap-improvements.md
+++ b/text/0886-hotswap-improvements.md
@@ -31,9 +31,9 @@ Resources currently supported by Hotswap:
 
 * feat(hotswap): hotswap now covers significantly more resource types due to using a CCAPI-based deployment engine
 * feat(hotswap): asset hotswapping has been improved
-* feat(cli): by default hotswap deployments fallback to a Cloudformation deployment
+* feat(cli): by default hotswap deployments fallback to an AWS CloudFormation deployment
 * feat(cli): the hotswap-fallback flag can now accept a deployment mode
-* feat(hotswap): hotswap deployments are now diff-ed based on the last successful hotswap deployment instead of the last full Cloudformation deployment
+* feat(hotswap): hotswap deployments are now diff-ed based on the last successful hotswap deployment instead of the last full AWS CloudFormation deployment
 
 ### BLOG POST
 
@@ -42,7 +42,7 @@ Resources currently supported by Hotswap:
 March 18, 2026 · AWS CDK Team
 
 Today, we are announcing a set of improvements to the AWS CDK hotswap deployment feature that reduce deployment times by at least 25%.
-These changes apply to both one-off hotswap deployments (cdk deploy --hotswap) and continuous deployments via cdk watch.
+These changes apply to both one-off hotswap deployments (`cdk deploy --hotswap`) and continuous deployments via `cdk watch`.
 
 #### Background
 
@@ -51,7 +51,8 @@ bypassing the full AWS CloudFormation deployment process.
 This allows developers to see their changes reflected in AWS in seconds rather than minutes.
 Hotswap has historically supported only a limited set of resource types,
 including AWS Lambda functions, AWS Step Functions, AWS AppSync, Amazon ECS Task Definitions, and AWS CodeBuild Projects.
-Changes to any other resource type required a full CloudFormation deployment, even during active development.
+Within those resource types hotswap supports, only a limited set of properties are allowed to be changed. 
+Changes to any other resource type required a full AWS CloudFormation deployment, even during active development.
 This meant that many developers could not take full advantage of hotswap's speed benefits across their entire application.
 
 #### Broader Resource Coverage with AWS Cloud Control API
@@ -60,10 +61,11 @@ The primary driver behind this improvement is the introduction of a new hotswap 
 Previously, each supported resource type required a dedicated hotswap implementation.
 The new CCAPI-based engine takes a different approach:
 for any change that is not already handled by an existing implementation, the engine attempts to perform an in-place update using Cloud Control APIs.
-If this is not possible the change is gracefully classified as non-hotswappable and routed to a fallback CloudFormation deployment.
+If this is not possible the change is gracefully classified as non-hotswappable and routed to a fallback AWS CloudFormation deployment.
 This means it will be easier for the CDK team to add hotswap support for new resource types,
 since the need to write custom implementations with SDK will largely be eliminated.
-Developers working with resources that previously fell outside of hotswap coverage will see the most dramatic improvement in their iteration speed.
+Developers working with resources that previously fell outside of hotswap coverage will see the most dramatic improvement in their iteration speed 
+compared to performing regular AWS CloudFormation deployments.
 
 #### Additional Improvements
 
@@ -71,7 +73,7 @@ Alongside the CCAPI-based engine, we have made several complementary improvement
 
 * **Optimized asset handling** — Assets are now rebuilt only when necessary, and the cdk synth step is skipped when only asset files have changed.
 This reduces pre-deployment overhead.
-* **Improved state tracking** — Successive hotswap deployments now diff against the last successful hotswap rather than the last full CloudFormation deployment,
+* **Improved state tracking** — Successive hotswap deployments now diff against the last successful hotswap rather than the last full AWS CloudFormation deployment,
 preventing redundant resource updates.
 
 ---
@@ -88,19 +90,19 @@ RFC pull request):
 
 ### What are we launching today?
 
-We are launching a set of improvements to CDK hotswap which will affect one off hotswap deployments (cdk deploy --hotswap) and
-iterative hotswap deployments via watch (cdk watch).
+We are launching a set of improvements to CDK hotswap which will affect one off hotswap deployments
+(`cdk deploy --hotswap`/`cdk deploy --hotswap-fallback`) and iterative hotswap deployments via watch (`cdk watch`).
 The more expansive improvements that will affect most cdk hotswap users are expanding the coverage of hotswap using a CCAPI-based engine and
 improvements to hotswapping assets such as incremental bundling and skipping the cdk synth phase if the hotswap change only includes asset files which
 should help speed up the pre-hotswap synthesis phase.
 We are also improving tracking of actual resource state when doing multiple subsequent hotswap deployments,
 updating the fallback behavior of hotswap (hotswap will now fallback to doing a full deployment by default) and the hotswap fallback command.
-Our primary metric for success is that hotswap deployments will be at least 25% faster.
+Our primary metric for success is that hotswap deployments will be at least 25% faster than they were previously.
 
 ### Why should I use this feature?
 
 You should use this feature while you are working on developing and testing your CDK application.
-Hotswap significantly speeds up the make a change-deploy-test loop in comparison to performing Cloudformation deployments.
+Hotswap significantly speeds up the make a change-deploy-test loop in comparison to performing AWS CloudFormation deployments.
 This will be particularly useful in speeding up the iterations your AI Agents are performing when they develop in CDK.
 
 ## Internal FAQ
@@ -113,14 +115,14 @@ testing the functionality those changes produce.
 
 ### Why should we _not_ do this?
 
-Deployment speed could be addressed somewhere else in the deployment process, such as skipping stabilization in Cloudformation for
+Deployment speed could be addressed somewhere else in the deployment process, such as skipping stabilization in AWS CloudFormation for
 resources that don’t need it, which could allow for speed gains without introducing drift.
 Hotswap and watch are still not commands that can be used in productions environments because they introduce drift,
 however this is fine for the purpose of rapid iteration.
 
 ### Will there be any additional telemetry collection to support this feature?
 
-We will collect counts of CloudFormation resource types that cause hotswap deployments to fall back to a full deployment.
+We will collect counts of AWS CloudFormation resource types that cause hotswap deployments to fall back to a full deployment.
 Hotswap fallbacks represent a performance degradation.
 Users invoking hotswap expect to be put on a faster deployment path and falling back to the standard path results in
 significantly longer deployment times.
@@ -131,16 +133,16 @@ or custom SDK logic) to eliminate the bottleneck.
 
 ### What resources will get hotswap implementations through SDK/CCAPI?
 
-The goal of this feature is not to reinvent Cloudformation and create a treadmill in the CLI.
-This means, we will not be adding hotswap support for all resources Cloudformation supports.
+The goal of this feature is not to reinvent AWS CloudFormation and create a treadmill in the CLI.
+This means, we will not be adding hotswap support for all resources AWS CloudFormation supports.
 Instead we will add hotswap support for a minimal set of resource types that are most commonly used in practical iterative deployment activities.
 Resources that will be supported by hotswap must meet the following criteria:
 
 1. Hotswappable resources are in the top 80% of resources that are included in hotswap deployments but are not currently hotswappable.
 If a resource type that is not currently hotswappable appears in often in hotswap deployments that is a signal that resource type should get hotswap support.
 2. Hotswappable resources must recover well from drift. If making hotswap changes to a resource outside of
-Cloudformation routinely causes errors while attempting to recover from drift we do not consider this resource hotswappable.
-3. Hotswappable resources, when implemented with CCAPIs, have a deployment speed through hotswap that is at least 2x faster than deploying through Cloudformation.
+AWS CloudFormation routinely causes errors while attempting to recover from drift we do not consider this resource hotswappable.
+3. Hotswappable resources, when implemented with CCAPIs, have a deployment speed through hotswap that is at least 2x faster than deploying through AWS CloudFormation.
 We want to ensure that we are only allow listing resources that will receive significant deployments speed benefits from hotswap.
 
 ### What is the technical solution (design) of this feature?
@@ -152,8 +154,8 @@ by our previously created SDK-based implementation and attempts to perform a hot
 If the resource that we are trying to hotswap is not supported by CCAPIs or we cannot update the resource in place,
 we swallow the error that would have been generated and log it as a non-hotswappable change where we use part of the error message
 in the reason why this was not a hotswappable change.
-We also have an allow list for resources that will not cause issues with subsequent Cloudformation deployments when
-hotswapped and resources where their hotswap implementation with CCAPIs is at least 2x faster than with a regular Cloudformation deployment.
+We also have an allow list for resources that will not cause issues with subsequent AWS CloudFormation deployments when
+hotswapped and resources where their hotswap implementation with CCAPIs is at least 2x faster than with a regular AWS CloudFormation deployment.
 Changes to resource types not on this allow list will be classified has non-hotswappable changes.
 [See the appendix for a prototype of the CCAPI hotswap engine](#ai-generated-implementation-of-ccapi-hotswap-engine)
 
@@ -164,44 +166,44 @@ Hotswapping assets has been improved in the following ways:
 1. We only rebuild assets when a change happens to them that requires that they are rebuilt
 2. We skip the cdk synth step of hotswap if we detect that a change only involves assets
 
-Asset bundling is expensive and we want to avoid doing it unless absolutely necessary while doing hotswap deployment using cdk watch.
-We take advantage of the watching capabilities of the watcher in cdk watch and register assets like lambda function handler files to be watched for changes.
+Asset bundling is expensive and we want to avoid doing it unless absolutely necessary while doing hotswap deployment using `cdk watch`.
+We take advantage of the watching capabilities of the watcher in `cdk watch` and register assets like lambda function handler files to be watched for changes.
 If a change happens to an asset, then we rebuild it during the next hotswap deployment.
 
 #### State Tracking during Subsequent Hotswap Deployments
 
-Currently running cdk deploy --hotswap updates all resources since the last Cloudformation deployment.
+Currently running `cdk deploy --hotswap` updates all resources since the last AWS CloudFormation deployment.
 If you are running multiple hotswap deployments in a row, this means each new template that is generated to perform the hotswap deployment
 does not know about the last hotswap deployment that happened.
 Which leads to creating diffs that includes changes that have already been hotswapped.
-This incurs a performance penalty over time since the time it takes for a --hotswap deployment to complete is proportional to the number of changed resources.
-To address this problem we are saving the Cloudformation template synthesized from the most recent successful hotswap deployment
-so we can refer back to it when new changes are made instead of referring back to the Cloudformation template from the last full deployment.
-These hotswap templates are wiped when a Cloudformation deployment happens and they do not attempt to alter
-or replace the Cloudformation template from the last successful deployment.
+This incurs a performance penalty over time since the time it takes for a `--hotswap` deployment to complete is proportional to the number of changed resources.
+To address this problem we are saving the AWS CloudFormation template synthesized from the most recent successful hotswap deployment
+so we can refer back to it when new changes are made instead of referring back to the AWS CloudFormation template from the last full deployment.
+These hotswap templates are wiped when a AWS CloudFormation deployment happens and they do not attempt to alter
+or replace the AWS CloudFormation template from the last successful deployment.
 
 #### Hotswap Fallback Mode
 
-Currently the hotswap can only configure only type of Cloudformation deployment.
-If in the future Cloudformation releases other deployment modes, they cannot be used as a fallback depployment mode
+Currently the hotswap can only configure one type of AWS CloudFormation deployment.
+If in the future AWS CloudFormation releases other deployment modes, they cannot be used as a fallback deployment mode
 for hotswap deployments with non-hotswappable changes.
-If no fallback mode is configured when using the hotswap fallback flag then we default to a full deployment.
-However if a different fallback mode is configured, we will use that as the fallback deployment method instead.
-Current behavior will remain the same:
+In response to this, we are introducing a new flag that is used in conjunction with the `cdk deploy --hotswap` command
+to configure a specific fallback mode.
+If the fallback flag is not present, we use the default fallback mode of `cdk deploy --hotswap`.
+Current behavior will remain the same (no fallback):
 
 ```
-$ cdk deploy --hotswap --hotswap-fallback
+$ cdk deploy --hotswap
 ```
 
-Can now specify a deployment mode (performs a full Cloudformation deployment):
+Can now specify a deployment mode (performs a full AWS CloudFormation deployment):
 
 ```
-$ cdk deploy --hotswap --hotswap-fallback=FULL_DEPLOY
+$ cdk deploy --hotswap --fallback=FULL_DEPLOY
 ```
 
-The default fallback mode for hotswap will change if Cloudformation releases a deployment option that is faster than
-what we currently get from Cloudformation deployments.
-If a mode like this is released it will become the default fallback deployment method for hotswap.
+The default fallback mode for hotswap will change if AWS CloudFormation releases a deployment option that is faster than
+what we currently get from AWS CloudFormation deployments.
 
 ### Is this a breaking change?
 
@@ -218,19 +220,19 @@ While some community members would write and share hotswap implementations for n
 find publicly shared ones would benefit.
 Which makes this an incomplete solution to the speed limitations that come from hotswapping non-hotswappable resources.
 
-#### Classify deployment speed as a Cloudformation Problem
+#### Classify deployment speed as problem for AWS CloudFormation to solve
 
-We also considered classifying performance concerns about deployment speed as a Cloudformation problem
-and waiting for Cloudformation to make improvements on their end.
-Any improvements Cloudformation makes to performance is a benefit to CDK customers and if Cloudformation is able to make drastic deployment speed improvements,
-then hotswap could be rendered obsolete.
-However, even if Cloudformation is able to improve the performance of their service,
-a custom hotswap implementation that uses SDK and CCAPI will still be faster because we can eliminate any overhead that comes from Cloudformation.
+We also considered classifying performance concerns about deployment speed as a problem that should be solved in AWS CloudFormation
+and wait for the AWS CloudFormation team to make improvements on their end.
+Any improvements the AWS CloudFormation team makes to performance is a benefit to CDK customers and
+if there is a drastic deployment speed improvement in AWS CloudFormation, then hotswap could be rendered obsolete.
+However, even if the AWS CloudFormation team is able to improve the performance of their service,
+a custom hotswap implementation that uses SDK and CCAPI will still be faster because we can eliminate any overhead that comes from AWS CloudFormation.
 
 ### What are the drawbacks of this solution?
 
 The main drawback of this solution is that hotswap deployments will continue to introduce drift between actual resource state and
-the state Cloudformation thinks the resource should be in.
+the state AWS CloudFormation thinks the resource should be in.
 
 ### What is the high-level project plan?
 
@@ -242,7 +244,7 @@ The second part of this project will be focused on improving asset hotswapping.
 While we are working on improving asset hotswapping, we will also be working on decreasing the amount of time we spend preparing to hotswap resources.
 The third part of this project will focus on addressing several open issues related to hotswapping,
 including altering the default fallback deployment type, updating the hotswap fallback command,
-and improving tracking of current resource state compared to the last time a full cloudformation deployment happened.
+and improving tracking of current resource state compared to the last time a full AWS CloudFormation deployment happened.
 
 ### Are there any open issues that need to be addressed later?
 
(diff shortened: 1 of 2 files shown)


text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+bypassing the full AWS CloudFormation deployment process.
+This allows developers to see their changes reflected in AWS in seconds rather than minutes.
+Hotswap has historically supported only a limited set of resource types,
+including AWS Lambda functions, AWS Step Functions, AWS AppSync, Amazon ECS Task Definitions, and AWS CodeBuild Projects.
+Changes to any other resource type required a full CloudFormation deployment, even during active development.
+This meant that many developers could not take full advantage of hotswap's speed benefits across their entire application.
+
+#### Broader Resource Coverage with AWS Cloud Control API
+
+The primary driver behind this improvement is the introduction of a new hotswap engine built on the AWS Cloud Control API (CCAPI).
```
tmokmss, line 57: CCAPI is known to be slow compared to direct service APIs ([some benchmarks](https://www.stedi.com/blog/relative-performance-tradeoffs-of-aws-native-provisioning-methods)). For hotswap where speed is the whole point, I think SDK is still a viable option for new resource types too. With coding agents these days, the maintenance cost of SDK implementations is not as high as it used to be. Have you run any benchmarks comparing the two for the target resource types?

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+      });
+      const currentProps: Record<string, any> = JSON.parse(
+        currentResource.ResourceDescription?.Properties ?? '{}',
+      );
+
+      const patchOps: Array<{ op: string; path: string; value: any }> = [];
+      for (const propName of classifiedChanges.namesOfHotswappableProps) {
+        const newValue = await evaluateCfnTemplate.evaluateCfnExpression(
+          change.propertyUpdates[propName].newValue,
+        );
```
tmokmss, line 347: A major pain point of current hotswap is that intrinsic function resolution is very limited. For example, this makes ECS Task Definitions non-hotswappable in many real-world cases. https://github.com/aws/aws-cdk/issues/26061 https://github.com/aws/aws-cdk/issues/25563
Is there any fundamental approach to this problem in this RFC? For example, fetching resolved values from already-deployed resources instead of trying to evaluate them on the CLI side.

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Is this a breaking change?
+
+No
+
+### What alternative solutions did you consider?
+
+#### Plugin system with customer-created plugins
+
+This was the proposal associated with the initial Github issue associated with hotswap anything (https://github.com/aws/aws-cdk-cli/issues/882).
```
tmokmss, line 238: One of the appealing aspects of the plugin architecture was that custom resource authors could provide their own hotswap logic. This matters for resources like BucketDeployment where the update behavior is defined by the construct library. I hope this option is still on the table.

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+get additional special treatment with custom SDK logic if needed.
+
+### What resources will get hotswap implementations through SDK/CCAPI?
+
+The goal of this feature is not to reinvent Cloudformation and create a treadmill in the CLI.
+This means, we will not be adding hotswap support for all resources Cloudformation supports.
+Instead we will add hotswap support for a minimal set of resource types that are most commonly used in practical iterative deployment activities.
+Resources that will be supported by hotswap must meet the following criteria:
+
+1. Hotswappable resources are in the top 80% of resources that are included in hotswap deployments but are not currently hotswappable.
```
tmokmss, line 151: Just out of curiosity, could you include the list of resources that would be covered under this criteria?

tmokmss (commented):
Thanks for putting this together! I use hotswap/watch regularly and I'm excited to see the coverage expanding. The asset bundling and synthesis improvements are especially great. A few comments on specific sections below.

text/0886-making-hotswap-faster.md
mrgrain: +1000

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+If a change happens to an asset, then we rebuild it during the next hotswap deployment.
+
+#### State Tracking during Subsequent Hotswap Deployments
+
+Currently running cdk deploy --hotswap updates all resources since the last Cloudformation deployment.
+If you are running multiple hotswap deployments in a row, this means each new template that is generated to perform the hotswap deployment
+does not know about the last hotswap deployment that happened.
+Which leads to creating diffs that includes changes that have already been hotswapped.
+This incurs a performance penalty over time since the time it takes for a --hotswap deployment to complete is proportional to the number of changed resources.
+To address this problem we are saving the Cloudformation template synthesized from the most recent successful hotswap deployment
```
dgandhi62: Is this saved locally or globally? What if multiple people are hotswapping at the same time? Can this lead to drift between local versions of the same code?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+so we can refer back to it when new changes are made instead of referring back to the Cloudformation template from the last full deployment.
+These hotswap templates are wiped when a Cloudformation deployment happens and they do not attempt to alter
+or replace the Cloudformation template from the last successful deployment.
```
dgandhi62: Same question here. How do we deal with multiple people using hotswap on the same environment?

text/0886-making-hotswap-faster.md
dgandhi62, line 1: Any plans for rollback? I'm curious as to how we deal with a broken state

dgandhi62 (commented):
Clarifying Questions

text/0886-making-hotswap-faster.md
ShadowCat567: No, rollback will slow down hotswap (it is part of the reason Cloudformation deployments are currently so slow), if a resource ends up in a broken state the expectation is that you perform subsequent deployments to fix the broken state. 
Additionally the current implementation of hotswap disables rollback by default, we intend to keep that implementation: https://docs.aws.amazon.com/cdk/v2/guide/cli.html#cli-deploy

text/0886-hotswap-improvements.md
ShadowCat567: This state is saved locally. 
This proposal does not solve the problem of multiple people attempting to run hotswap in the same environment at the same time. Multiple people making changes to the same stack at the same time would be complicated to resolve even if you were making changes through full Cloudformation deployments. 
I do not think hotswap is the right place to resolve issues stemming from multiple people using the same environment to make changes to the same resources. The solution to this would be to use isolated environments or coordinate to work on different stacks if you are developing in the same environment.

text/0886-hotswap-improvements.md
ShadowCat567: see answer to previous question

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+* feat(hotswap): hotswap deployments are now diff-ed based on the last successful hotswap deployment instead of the last full Cloudformation deployment
+
+### BLOG POST
+
+#### AWS CDK Hotswap Deployments Are Now 25% Faster
+
+March 18, 2026 · AWS CDK Team
+
+Today, we are announcing a set of improvements to the AWS CDK hotswap deployment feature that reduce deployment times by at least 25%.
+These changes apply to both one-off hotswap deployments (cdk deploy --hotswap) and continuous deployments via cdk watch.
```
kaizencc: nit: and `cdk deploy --hotswap-fallback`

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+
+Today, we are announcing a set of improvements to the AWS CDK hotswap deployment feature that reduce deployment times by at least 25%.
+These changes apply to both one-off hotswap deployments (cdk deploy --hotswap) and continuous deployments via cdk watch.
+
+#### Background
+
+Hotswap is a CDK CLI feature that accelerates the development loop by updating AWS resources directly through service APIs,
+bypassing the full AWS CloudFormation deployment process.
+This allows developers to see their changes reflected in AWS in seconds rather than minutes.
+Hotswap has historically supported only a limited set of resource types,
```
kaizencc, line 49: not sure how to wordsmith this but we only support a limited set of property changes under the limited set of resource types too

text/0886-making-hotswap-faster.md
kaizencc: @ShadowCat567 the proposal should be more clear about this if true: we will maintain and add bespoke SDK implementations based on quantitative benchmarks compared to CCAPI in specific use cases, but feel like the CCAPI implementation will be adequate for most use cases.

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+#### Broader Resource Coverage with AWS Cloud Control API
+
+The primary driver behind this improvement is the introduction of a new hotswap engine built on the AWS Cloud Control API (CCAPI).
+Previously, each supported resource type required a dedicated hotswap implementation.
+The new CCAPI-based engine takes a different approach:
+for any change that is not already handled by an existing implementation, the engine attempts to perform an in-place update using Cloud Control APIs.
+If this is not possible the change is gracefully classified as non-hotswappable and routed to a fallback CloudFormation deployment.
+This means it will be easier for the CDK team to add hotswap support for new resource types,
+since the need to write custom implementations with SDK will largely be eliminated.
+Developers working with resources that previously fell outside of hotswap coverage will see the most dramatic improvement in their iteration speed.
```
kaizencc: nit: compared to default CFN deployment

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### Additional Improvements
+
+Alongside the CCAPI-based engine, we have made several complementary improvements:
+
+* **Optimized asset handling** — Assets are now rebuilt only when necessary, and the cdk synth step is skipped when only asset files have changed.
+This reduces pre-deployment overhead.
```
kaizencc: question: am i correct in thinking that this improvement is not specific to hotswap and will improve CDK synth times for regular deployment as well? If so, it should be celebrated as such and not as an afterthought in the hotswap improvments

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+
+```
+[ ] Signed-off by API Bar Raiser @xxxxx
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+We are launching a set of improvements to CDK hotswap which will affect one off hotswap deployments (cdk deploy --hotswap) and
```
kaizencc: again, and `cdk deploy --hotswap-fallback`

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+iterative hotswap deployments via watch (cdk watch).
+The more expansive improvements that will affect most cdk hotswap users are expanding the coverage of hotswap using a CCAPI-based engine and
+improvements to hotswapping assets such as incremental bundling and skipping the cdk synth phase if the hotswap change only includes asset files which
+should help speed up the pre-hotswap synthesis phase.
+We are also improving tracking of actual resource state when doing multiple subsequent hotswap deployments,
+updating the fallback behavior of hotswap (hotswap will now fallback to doing a full deployment by default) and the hotswap fallback command.
+Our primary metric for success is that hotswap deployments will be at least 25% faster.
```
kaizencc: again, faster than default CFN deployment

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,390 @@
...
+3. Hotswappable resources, when implemented with CCAPIs, have a deployment speed through hotswap that is at least 2x faster than deploying through Cloudformation.
+We want to ensure that we are only allow listing resources that will receive significant deployments speed benefits from hotswap.
+
+### What is the technical solution (design) of this feature?
+
+#### CCAPI-Based Hotswap
+
+We are introducing a CCAPI-based implementation of hotswap which takes in changes that are not already covered
+by our previously created SDK-based implementation and attempts to perform a hotswap change using Cloud Control APIs.
+If the resource that we are trying to hotswap is not supported by CCAPIs or we cannot update the resource in place,
```
kaizencc: what percent of resources created by CDK would not be supported by CCAPI?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+[See the appendix for a prototype of the CCAPI hotswap engine](#ai-generated-implementation-of-ccapi-hotswap-engine)
+
+#### Hotswapping Assets
+
+Hotswapping assets has been improved in the following ways:
+
+1. We only rebuild assets when a change happens to them that requires that they are rebuilt
+2. We skip the cdk synth step of hotswap if we detect that a change only involves assets
+
+Asset bundling is expensive and we want to avoid doing it unless absolutely necessary while doing hotswap deployment using cdk watch.
```
kaizencc: this makes it seem like this feature is only available in `cdk watch`. is that true?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+We take advantage of the watching capabilities of the watcher in cdk watch and register assets like lambda function handler files to be watched for changes.
+If a change happens to an asset, then we rebuild it during the next hotswap deployment.
+
+#### State Tracking during Subsequent Hotswap Deployments
+
+Currently running cdk deploy --hotswap updates all resources since the last Cloudformation deployment.
```
kaizencc: nit: just in terms of markdown rendering it would be nice to wrap all commands in code blocks `like this`

text/0886-making-hotswap-faster.md
dgandhi62: I see! Do we have a current workaround for broken state? How does a customer get out of it?

text/0886-hotswap-improvements.md
dgandhi62: I see!

text/0886-hotswap-improvements.md
kaizencc: instead of wiping, should `deploy` just save a local file of what was last successfully deployed (hotswap or not?) that could be simpler for hotswap to never have to make a CFN `get-template` call. but also it will mean we are not protected against CFN drift that happens elsewhere.

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+Which leads to creating diffs that includes changes that have already been hotswapped.
+This incurs a performance penalty over time since the time it takes for a --hotswap deployment to complete is proportional to the number of changed resources.
+To address this problem we are saving the Cloudformation template synthesized from the most recent successful hotswap deployment
+so we can refer back to it when new changes are made instead of referring back to the Cloudformation template from the last full deployment.
+These hotswap templates are wiped when a Cloudformation deployment happens and they do not attempt to alter
+or replace the Cloudformation template from the last successful deployment.
+
+#### Hotswap Fallback Mode
+
+Currently the hotswap can only configure only type of Cloudformation deployment.
```
kaizencc:
```suggestion
Currently the hotswap can only configure one type of Cloudformation deployment.
```

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+If in the future Cloudformation releases other deployment modes, they cannot be used as a fallback depployment mode
+for hotswap deployments with non-hotswappable changes.
+If no fallback mode is configured when using the hotswap fallback flag then we default to a full deployment.
+However if a different fallback mode is configured, we will use that as the fallback deployment method instead.
+Current behavior will remain the same:
+
+```
+$ cdk deploy --hotswap --hotswap-fallback
```
kaizencc: this is not the current behavior. you cannot set `--hotswap` and `--hotswap-fallback` at the same time

kaizencc: we currently have two fallback options:

1) `--hotswap`: ignore non-hotswappable changes and report success
2) `--hotswap-fallback`: fall back to full deployment for all when non-hostwappable changes exist

if in the future we wanted to support additional fall back modes we'd likely do `--hotswap-another-fallback` as a separate `cdk deploy` option. It sounds like you are proposing a different API here, but also a bit confused as to how it works today.

text/0886-making-hotswap-faster.md
kaizencc: at least i don't see an argument right now for why we can't expose plugin architecture in addition to what else we are supporting as a use-at-your-own-risk feature.

is there a technical limitation that stops us from allowing plugins?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+In this option we would create a plugin architecture which would allow CDK customers to write their own hotswap plugins and load them into the CLI.
+This would let customers add hotswap support for their own resources without waiting for the CDK team. However, it shifts that responsibility to the community.
+While some community members would write and share hotswap implementations for new resource types, only those who write their own or
+find publicly shared ones would benefit.
+Which makes this an incomplete solution to the speed limitations that come from hotswapping non-hotswappable resources.
+
+#### Classify deployment speed as a Cloudformation Problem
+
+We also considered classifying performance concerns about deployment speed as a Cloudformation problem
+and waiting for Cloudformation to make improvements on their end.
```
kaizencc:
```suggestion
and wait for Cloudformation to make improvements on their end.
```

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,390 @@
...
+#### Plugin system with customer-created plugins
+
+This was the proposal associated with the initial Github issue associated with hotswap anything (https://github.com/aws/aws-cdk-cli/issues/882).
+In this option we would create a plugin architecture which would allow CDK customers to write their own hotswap plugins and load them into the CLI.
+This would let customers add hotswap support for their own resources without waiting for the CDK team. However, it shifts that responsibility to the community.
+While some community members would write and share hotswap implementations for new resource types, only those who write their own or
+find publicly shared ones would benefit.
+Which makes this an incomplete solution to the speed limitations that come from hotswapping non-hotswappable resources.
+
+#### Classify deployment speed as a Cloudformation Problem
```
kaizencc: huge nit, sorry: we should refer to cloudformation as "AWS CloudFormation" (AWS at front, C and F capitalized). and any other resources should have "AWS" in front when we refer to them in the public.

text/0886-making-hotswap-faster.md
ShadowCat567: Implementing hotswap plugins and implementing additional resource coverage via CCAPIs are not mutually exclusive. This RFC is more focused on how we can improve hotswap performance for most the most users. I'm not sure how many users would be willing to write their own plugins but if there is interest from people like (@tmokmss) then we can look into adding a plugin system as well.

text/0886-making-hotswap-faster.md
ShadowCat567: I was not previously aware of this, now that you bring it up, I think it is a good idea to include in the set of improvements. Thank you for highlighting this issue, will be looking into it!

text/0886-hotswap-improvements.md
ShadowCat567: it should, additionally any changes made to improve `cdk synth` would also improve the performance of hotswap

text/0886-making-hotswap-faster.md
ShadowCat567: That is a hard number to come by if we assume cdk supports the same number of resources cloudformation does mostly because Cloudformation does not have an official number for how many resources it supports. CCAPIs support around 1,290 resource types and if we assume Cloudformation/CDK support around 1,500 resource types, ~14% of resource types will not be supported by CCAPI. Conversely, CCAPIs support ~86% of the resource types Cloudformation/CDK does.

text/0886-hotswap-improvements.md
ShadowCat567: Don't we already kind of do that in `cdk.out`? `cdk.out` also saves the template for the last hotswap deployment, so we could pull the template from `cdk.out` for state tracking. Using templates in `cdk.out` was not my initial thinking here, but you comment made me check to see what gets put there when we run `cdk deploy --hotswap`.
(Also do you mean to say: '...we are _now_ protected against CFN drift...')

text/0886-hotswap-improvements.md
ShadowCat567: This is where I was getting confused. What I am suggesting is a different API where all of the fallback options are grouped together and you specify that flag alongside `cdk deploy --hotswap` to configure the fallback mode.

text/0886-hotswap-improvements.md
ShadowCat567: Actually this is in comparison to hotswap deployments before we started making synth improvements and increased resource coverage, not Cloudformation deployments.

text/0886-hotswap-improvements.md
ShadowCat567: The way I am imagining it in this RFC, yes it would be restricted to `cdk watch`. However I would like to see if it is possible to be used outside of `watch`, but that would depend on the implementation

text/0886-making-hotswap-faster.md
rix0rrr: I would like to be very clear on the use case and perceived benefits of a plugin architecture.

For non-custom resources, we are saying that we will have a CCAPI handler if that gets us at least a 2x speed improvement over CloudFormation. It will be hard to do better than CCAPI using raw SDK calls, which means that you should not have a much reason to have a custom handler for any regular resource types: you will be within 2x of the theoretical max performance regardless, which should be good enough for most development flows.

After that, there is custom resources. But for example the "BucketDeployment" custom resource doesn't seem like a strong example for a plug-in architecture yet -- that's a CDK-provided custom resource, and I think a good argument could be made that a good local implementation of that one should just be provided with the CDK CLI out of the box. It will be so commonly used in local development that it makes perfect sense for us to take on this lift, and punting the responsibility to users is the wrong decision there.

After that one common custom resource, what remains? A custom resource that gets triggered so often that it is worth specifically optimizing for, because the ~10 additional seconds it takes to wait for CloudFormation are too much to bear? I would like to see some good examples before we take that on.

mrgrain: I think this relates to another comment here: https://github.com/aws/aws-cdk-rfcs/pull/887#discussion_r2978549234

> Just out of curiosity, could you include the list of resources that would be covered under this criteria?

What are we planning to support? Only by having a proposed list we can make a judgement on what is still missing.

mrgrain: There are more custom resources, and then there will be userland custom resources. I also see a lot of value for integrating bundling with hotswap. This will be hard for us to optimize generally without knowing a specific app. Users do know there apps and can optimize for them.

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,399 @@
...
+we swallow the error that would have been generated and log it as a non-hotswappable change where we use part of the error message
+in the reason why this was not a hotswappable change.
+We also have an allow list for resources that will not cause issues with subsequent AWS CloudFormation deployments when
+hotswapped and resources where their hotswap implementation with CCAPIs is at least 2x faster than with a regular AWS CloudFormation deployment.
+Changes to resource types not on this allow list will be classified has non-hotswappable changes.
+[See the appendix for a prototype of the CCAPI hotswap engine](#ai-generated-implementation-of-ccapi-hotswap-engine)
+
+#### Hotswapping Assets
```
mrgrain, line 193: This section doesn't really describe a technical solution.

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,399 @@
...
+* Bedrock Agentcore Runtime
+* CDK Bucket Deployment custom resource
+
+## Working Backwards
+
+### CHANGELOG
+
+* feat(hotswap): hotswap now covers significantly more resource types due to using a CCAPI-based deployment engine
+* feat(hotswap): asset hotswapping has been improved
+* feat(cli): by default hotswap deployments fallback to an AWS CloudFormation deployment
```
mrgrain: This seems to be a breaking change. What is the rollout plan here?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,399 @@
...
+* Codebuild Projects
+* Bedrock Agentcore Runtime
+* CDK Bucket Deployment custom resource
+
+## Working Backwards
+
+### CHANGELOG
+
+* feat(hotswap): hotswap now covers significantly more resource types due to using a CCAPI-based deployment engine
+* feat(hotswap): asset hotswapping has been improved
```
mrgrain: how?

text/0886-hotswap-improvements.md
```diff
@@ -0,0 +1,399 @@
...
+### What alternative solutions did you consider?
+
+#### Plugin system with customer-created plugins
+
+This was the proposal associated with the initial Github issue associated with hotswap anything (https://github.com/aws/aws-cdk-cli/issues/882).
+In this option we would create a plugin architecture which would allow CDK customers to write their own hotswap plugins and load them into the CLI.
+This would let customers add hotswap support for their own resources without waiting for the CDK team. However, it shifts that responsibility to the community.
+While some community members would write and share hotswap implementations for new resource types, only those who write their own or
+find publicly shared ones would benefit.
+Which makes this an incomplete solution to the speed limitations that come from hotswapping non-hotswappable resources.
```
mrgrain: While there are good reasons against a plugin API, I don't think this is one of them. Elsewhere we are stating that not all resource are supported by CCAPI and that we will support resource that "are in the top 80% of resources that are included in hotswap deployments but are not currently hotswappable". Both explicitly exclude resource and thus make this RFC an incomplete solution as well.

text/0886-making-hotswap-faster.md
ShadowCat567: @tmokmss There currently is a hotswap implementation for `BucketDeployments` (https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/hotswap/s3-bucket-deployments.ts), what kind of improvements would you like to see from it that we could do without introducing a plugin interface? 
Also with a plugin interface in general would your main vision for its use be adding support for hotswapping custom resources? Or do you have other ideas for what you would want out of it?

text/0886-making-hotswap-faster.md
ShadowCat567: yes, that is the idea, I'll update this section

text/0886-hotswap-improvements.md
ShadowCat567: updated this by splitting it into 2 changelog annotations - 1 for asset bundling, 1 for skipping synth

text/0886-hotswap-improvements.md
ShadowCat567: Went back to do some research on the history of hotswap fallback and decided that it would be better for us to leave it alone unless we get a faster deployment mode through Cloudformation, in that case the `hotswap-fallback` flag would be getting changed to use that deployment mode. 
The current `cdk deploy --hotswap` fallback behavior is fine, this suggestion was to try to fix a problem that was based on a misunderstanding of how it worked.

text/0886-hotswap-improvements.md
rix0rrr: This is true. The goal is not "100% coverage". The goal is: fast iteration times for local development.

text/0886-making-hotswap-faster.md
tmokmss: @ShadowCat567 @rix0rrr 

> would your main vision for its use be adding support for hotswapping custom resources?

Yes. Although I know this is a small use case, I maintain a construct library named [deploy-time-build](https://constructs.dev/packages/deploy-time-build/) (recently migrated [under cdklabs](https://github.com/cdklabs/deploy-time-build)). It contains a custom resource to build certain assets like Node.js apps or container images at deploy time. With a plugin architecture, I could provide a hotswap handler that runs the build locally instead of going through CloudFormation, which would be significantly faster when making frequent changes to the source code being built.


That said, I understand we need to prioritize. If I had to pick one thing, I'd rather see the intrinsic function resolution issue (https://github.com/aws/aws-cdk-rfcs/pull/887#discussion_r2978547059) addressed first, as that's a bigger blocker in my day-to-day workflow.

text/0886-making-hotswap-faster.md
ShadowCat567: Some examples that I have been looking at would be Bedrock Agents, DynamoDB Tables, SQS Queues, Cloudwatch Alarms and Dashboards, SNS Topics and Subscriptions, and RestAPIs from ApiGateway. Let me know if there are any resource types you would be particularly interested in seeing - will be adding a list in the RFC.

text/0886-hotswap-improvements.md
ShadowCat567: Ended up removing this section after further research, see: https://github.com/aws/aws-cdk-rfcs/pull/887#discussion_r2990075252

text/0886-hotswap-improvements.md
ShadowCat567: Summary/Conclusions from an offline discussion with @rix0rrr and @mrgrain about whether plugins should be included in this RFC:
Even if this RFC does not include plugins for hotswap, we should be designing this in a way where adding hotswap plugins should be easy/straightforward in the future. The main concerns with adding a plugin in system to this project would be that we are adding a lot of work to this project without much progress towards the main success metric (make hotswap faster for most people) and the packaging/representation of common types that will cross the plugin border like types from SDK could be more complicated and offer more opportunity for breakage than we expect (especially since we have not started getting to the weeds of how exactly this will work). 
We should consider making a pseudo plugin system so we know what a full plugin system would need and can build the hotswap architecture improvements this RFC suggests around it so if/when we implement a plugin interface for hotswap it can slot in cleanly. 
This RFC should also be a bit more clear that the goal is speed rather than coverage from hotswap.

text/0886-making-hotswap-faster.md
ShadowCat567: Improvements to intrinsic function resolution is actually required work to get the CCAPI hotswap implementation to work, so it will be a part of this RFC. The plugin system probably will be implemented at some point in the future, but it will not be a part of this RFC this comment may be interesting to you: https://github.com/aws/aws-cdk-rfcs/pull/887#discussion_r3002522637

text/0886-hotswap-improvements.md
ShadowCat567: This section has evolved, part of asset bundling improvements will only be in watch, part of them will benefit `cdk synth` as a whole, see updated version of this section for clarification.

text/0886-making-hotswap-faster.md
ShadowCat567: has been updated

text/0886-making-hotswap-faster.md
ShadowCat567: This is required work to get the CCAPI engine to work (and work well), will be calling this out in the RFC.

text/0886-making-hotswap-faster.md
tmokmss: Cool, thanks!

text/0886-making-hotswap-faster.md
```diff
@@ -0,0 +1,420 @@
...
+Within those resource types hotswap supports, only a limited set of properties are allowed to be changed.
+Changes to any other resource type required a full AWS CloudFormation deployment, even during active development.
+This meant that many developers could not take full advantage of hotswap's speed benefits across their entire application.
+
+#### Increased Speed through Broader Resource Coverage with AWS Cloud Control API
+
+The primary driver behind this improvement is the introduction of a new hotswap engine built on the AWS Cloud Control API (CCAPI).
+Previously, each supported resource type required a dedicated hotswap implementation.
+The new CCAPI-based engine takes a different approach:
+for any change that is not already handled by an existing implementation, the engine attempts to perform an in-place update using Cloud Control APIs.
```
go-to-k: > not already handled by an existing implementation

Is it at the resource type level or the property level? If at the resource type level, resource types with existing SDK implementations would not benefit from CCAPI's broader property coverage, which seems worth noting in the RFC.

If at the property level, suppose a resource type has an SDK-based hotswap implementation that only covers certain properties, and a change includes both covered and uncovered properties. Is the entire change delegated to CCAPI, split between SDK and CCAPI, or classified as non-hotswappable? \
If split, both SDK and CCAPI calls are needed, making it slower than either approach alone. If the entire change is delegated to CCAPI, SDK implementations would only be used when all changed properties happen to be SDK-covered. Are SDK implementations maintained for that case?

ShadowCat567: This section should have been changed as the RFC has evolved. We are planning to have a clear split between resources that are hotswapped with SDKs and resources that are hotswapped with CCAPIs. This means, if a resources has an SDK-based hotswap implementation it will not fallback to being hotswapped by CCAPIs if a property that cannot be hotswapped by the SDK implementation is specified. Additionally, the resources that use the CCAPI implementation will be managed by an allow list, which means adding hotswap support for new resource types becomes a one-line change. 
Essentially, there are boundaries between SDK-hotswappable resources, CCAPI-hotswappable resources, and non-hotswappable resources and resources cannot cross those boundaries and be hotswappable by both SDKs and CCAPIs. I'll clarify that in this section of the RFC, thank you for asking about this!

rix0rrr (approved):

Issue #929, opened 2026-06-04
AWS S3Files L2 Constructs


Adonca2203:
## Description

Add L2 CDK Constructs for the AWS S3Files service.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @Adonca2203
| Author(s)           | @Adonca2203
| API Bar Raiser      | @leonmk-aws 
| Stakeholders        | @Pandede, @badmintoncryer

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `status/api-approved` applied to pull request)
- [x] Final comments period (label: `status/final-comments-period`)
- [x] Approved and merged (label: `status/approved`)
- [x] Execution plan submitted (label: `status/planning`)
- [x] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

Adonca2203:
^ needs `status/review` label

Adonca2203:
https://github.com/aws/aws-cdk/pull/38435

PR #932, opened 2026-06-04
RFC for S3Files L2 CDK Constructs

Adonca2203:
This is a request for comments about S3Fiels L2 Constructs. See #929 for
additional details. 

APIs are signed off by @leonmk-aws .

---


text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,609 @@
...
+   * @default - none
+   */
+  readonly fileSystemPolicy?: iam.PolicyDocument;
+
+  /**
+   * The file system's name.
+   *
+   * @default - CDK generated name
+   */
+  readonly fileSystemName?: string;
```
badmintoncryer: AWS::S3Files::FileSystem has no Name property. I think this prop is not needed.

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,609 @@
...
+    vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS },
+  },
+  prefix: 'data/',
+  synchronizationConfiguration: {
+    importDataRules: [{
+      prefix: '/',
+      sizeLessThan: 1073741824,
+      trigger: s3files.ImportDataRuleTrigger.ON_FILE_ACCESS,
+    }],
+    expirationDataRules: [{
```
badmintoncryer: sizeLessThan (CFN: Integer bytes) and the expiration value (CFN: DaysAfterLastAccess, Integer days) are raw numbers in the proposal. Per the CDK design guidelines these should be Size / Duration. Note the CFN constraints when converting: DaysAfterLastAccess is an integer 1–365 days, so a Duration must validate whole days; SizeLessThan is in bytes.

badmintoncryer: nit: CFN requires exactly one ExpirationDataRule (min=max=1) and 1–10 ImportDataRules with exactly one root-prefix rule. The proposed expirationDataRules: ExpirationDataRule[] / importDataRules[] should validate this.

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,609 @@
...
+```ts
+declare const vpc: ec2.Vpc;
+declare const bucket: s3.Bucket;
+
+const fileSystem = new s3files.FileSystem(this, 'MyFileSystem', {
+  bucket,
+  vpcConfiguration: {
+    vpc,
+    vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS },
+  },
```
badmintoncryer, line 47: The README shows vpcSubnets but the interface defines subnets: ec2.ISubnet[] (required). Suggest aligning on a single name and using ec2.SubnetSelection (consistent with EFS L2 and other VPC-attached constructs) so users can pass { subnetType } rather than resolving subnets themselves.

badmintoncryer (changes requested):
Thank you for creating a rfc.

I've added some comments. And there are two things to say.


- The proposal automates the IAM role but exposes no grant* surface for client principals (s3files:ClientMount/ClientWrite/ClientRootAccess). Worth adding grant methods (or a grants facade) so consumers can authorize Lambda/ECS/EC2 task roles idiomatically.

- The CFN resources expose AcceptBucketWarning and ClientToken on FileSystem, and IpAddressType (IPV4_ONLY|IPV6_ONLY|DUAL_STACK) on MountTarget. None appear in the proposed API — AcceptBucketWarning in particular is needed for many buckets, and there’s no way to opt into IPv6/dual-stack mount targets.

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+// Read and write access
+fileSystem.grantReadWrite(lambdaFunction);
+
+// Root access (mount + write + root)
+fileSystem.grantRootAccess(lambdaFunction);
+```
+
+Each method calls `iam.Grant.addToPrincipalOrResource` with the
+appropriate `s3files:Client*` actions and an
+`s3files:AccessedViaMountTarget` condition. A general-purpose
```
leonmk-aws: Shouldn't this condition be: [s3files:AccessPointArn](https://docs.aws.amazon.com/AmazonS3/latest/API/s3files-access-points.html) ?

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+`addToResourcePolicy`:
+
+```ts
+declare const fileSystem: s3files.FileSystem;
+
+fileSystem.addToResourcePolicy(new iam.PolicyStatement({
+  actions: ['s3files:ClientMount'],
+  principals: [new iam.AnyPrincipal()],
+  conditions: {
+    Bool: { 's3files:AccessedViaMountTarget': 'true' },
```
leonmk-aws: Shouldn't this be: [s3files:AccessPointArn](https://docs.aws.amazon.com/AmazonS3/latest/API/s3files-access-points.html)

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+    expirationDataRule: {
+      daysAfterLastAccess: Duration.days(30),
+    },
+  },
+});
+```
+
+## Granting Access
+
+Use `grant*` methods to authorize principals for NFS client operations.
```
leonmk-aws: Let's use grants facades for this: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#grants (this is a minimal change)

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+
+### What is the high-level project plan?
+
+Phase 1 - Alpha Package (current):
+
+* `FileSystem` L2 with VPC mount targets, IAM role automation,
+  resource policy support
+* `AccessPoint` L2 with POSIX user and root directory configuration
+* `FileSystem.fromFileSystemAttributes()` and
+  `AccessPoint.fromAccessPointAttributes()` import methods
```
leonmk-aws: Earlier it is mentioned that the methods name is `fromAccessPointId` not `fromAccessPointAttributes`

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+## Access Points
+
+An access point is an application-specific view into a file system
+that applies a POSIX user/group and a file system path to any request
+made through the access point:
+
+```ts
+declare const fileSystem: s3files.FileSystem;
+
+const accessPoint = fileSystem.addAccessPoint('AccessPoint', {
```
leonmk-aws, line 181: `fileSystem.addAccessPoint('AccessPoint', { path, createAcl, posixUser })`, but the only interface defined later is `AccessPointProps`, which requires `fileSystem`. Are we missing an `AccessPointOptions` interface ?

leonmk-aws (commented):
Similar to grants, we should have metrics helper functions: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#metrics

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,741 @@
...
+   * The file system to create the access point on.
+   */
+  readonly fileSystem: IFileSystemRef;
+
+  /**
+   * The name of the access point.
+   *
+   * @default - CDK generated name
+   */
+  readonly accessPointName?: string;
```
leonmk-aws: CloudFormation doesn't have support for the name of an access point https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/TemplateReference/aws-resource-s3files-accesspoint.html, I believe this is only supported in the console ?

text/0929-s3files-l2-constructs.md
```diff
@@ -109,28 +109,40 @@ const fileSystem = new s3files.FileSystem(this, 'MyFileSystem', {
...
 ```
 
-Each method calls `iam.Grant.addToPrincipalOrResource` with the
-appropriate `s3files:Client*` actions and an
-`s3files:AccessedViaMountTarget` condition. A general-purpose
-`grant(grantee, ...actions)` method is also available for custom
-action sets.
+## Metrics
+
+Use the metrics facade to create CloudWatch metric objects:
```
leonmk-aws: Unfortunately the metrics facade is not ready yet so it cannot be used, we need to write the metrics methods by hand for now. I also means that there should not be `metrics` property on the `Filesystem`.
See this section of the design guidelines: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#metrics, we should have at least a `metric()` method and additional metric methods for the ones we find relevant (e.g `metricDataReadBytes` etc)

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,750 @@
...
+```
+
+## Metrics
+
+Use the metrics facade to create CloudWatch metric objects:
+
+```ts
+declare const fileSystem: s3files.FileSystem;
+
+const readOps = fileSystem.metrics.metric('DataReadIOBytes');
```
leonmk-aws: Nit: `DataReadIOBytes` is the EFS metric, the s3files one is `DataReadBytes`

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,750 @@
...
+
+  /**
+   * The full POSIX identity, including the user ID, group
+   * ID, and any secondary group IDs, on the access point
+   * that is used for all file system operations performed
+   * by NFS clients using the access point.
+   *
+   * @default - user identity not enforced
+   */
+  readonly posixUser?: PosixUser;
```
leonmk-aws, line 613: `PosixUser` is not defined in the interfaces, it appears in above in an example, lets define it clearly:
```ts
export interface PosixUser {
  /**
   * The POSIX user ID used for all file system operations using this access point.
   */
  readonly uid: string;

  /**
   * The POSIX group ID used for all file system operations using this access point.
   */
  readonly gid: string;

  /**
   * Secondary POSIX group IDs used for all file system operations using this access point.
   *
   * @default - None
   */
  readonly secondaryGids?: string[];
}
```

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,750 @@
...
+```
+
+* `SynchronizationConfiguration` - importDataRules,
+  expirationDataRule
+
+```ts
+/**
+ * Configuration for data import and expiration behavior.
+ */
+export interface SynchronizationConfiguration {
```
leonmk-aws, line 700: Both `importDataRules` and `ImportDataRules` are required in CFN: `AWS::S3Files::FileSystem SynchronizationConfiguration` so they should not be optional in this interface, unless the L2 abstraction provides defaults configuration (`@default - service defaults` seems to imply that we don't).

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,750 @@
...
+   */
+  readonly importDataRules?: ImportDataRule[];
+
+  /**
+   * Rule controlling when cached data expires.
+   * CFN requires exactly one expiration rule.
+   *
+   * @default - service defaults
+   */
+  readonly expirationDataRule?: ExpirationDataRule;
```
leonmk-aws: Do we need to have this additional `ExpirationDataRule` type since it only holds `daysAfterLastAccess` property ?

leonmk-aws (changes requested):
Added a few more comments after the last changes

text/0929-s3files-l2-constructs.md
Adonca2203: The L1 defines it as an object but we can flatten it for the L2

text/0929-s3files-l2-constructs.md
Adonca2203: I can include one in the doc here as an example but for actual implementation is there a priority list of metrics we want to expose if we are exposing each manually? maybe the top N make it into the first impl and we can add more down the road based on demand?

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,768 @@
...
+   * rule must have a root prefix ('/').
+   */
+  readonly importDataRules: ImportDataRule[];
+
+  /**
+   * Number of days after last access before cached data
+   * expires. Must be a whole number of days between 1
+   * and 365.
+   */
+  readonly daysAfterLastAccess: Duration;
```
leonmk-aws: Now that `ExpirationDataRule` is flattened away, `daysAfterLastAccess` sitting next to `importDataRules` reads like a threshold rather than the expiration action. The S3 Files console labels this "Data expiration after last access," so I'd rename it to match — `dataExpiration: Duration` (with "measured from last access" in the JSDoc).

text/0929-s3files-l2-constructs.md
leonmk-aws: Yes, later during the implementation or based on demand, additional methods can be added without any backwards compatibility concern.

leonmk-aws (changes requested):
See below

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,768 @@
...
+    vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS },
+  },
+  prefix: 'data/',
+  synchronizationConfiguration: {
+    importDataRules: [{
+      prefix: '/',
+      sizeLessThan: Size.gibibytes(1),
+      trigger: s3files.ImportDataRuleTrigger.ON_FILE_ACCESS,
+    }],
+    daysAfterLastAccess: Duration.days(30),
```
leonmk-aws: The rename to `dataExpiration` only landed on the interface (line 713) — this example still uses `daysAfterLastAccess`, which no longer exists on `SynchronizationConfiguration`. Could you update it to `dataExpiration`?

text/0929-s3files-l2-constructs.md
```diff
@@ -0,0 +1,768 @@
...
+   * The file system associated with this access point.
+   *
+   * @default - no file system
+   */
+  readonly fileSystem?: IFileSystemRef;
+}
+```
+
+* `SynchronizationConfiguration` - importDataRules,
+  daysAfterLastAccess
```
leonmk-aws: Same rename here: `daysAfterLastAccess` → `dataExpiration` to match the interface (line 713).

leonmk-aws (changes requested):
2 minor name changes are required

leonmk-aws:
Thank you for your review @badmintoncryer and thanks to @Adonca2203 for updating the RFC.
Signing off the RFC. Will keep it open for the final comments period.

leonmk-aws:
Final comments period is over. Closing this one

leonmk-aws (approved):

leonmk-aws:
@Mergifyio queue

leonmk-aws:
@Mergifyio refresh

leonmk-aws:
@Mergifyio queue

Issue #949, opened 2026-06-19
CDK Debootstrap Command


sanjanaravikumar-az:
## Description

Removes the CDK bootstrap stack and all associated resources from an AWS environment that get created when `cdk bootstrap` is run.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @sanjanaravikumar-az 
| Author(s)           | @sanjanaravikumar-az 
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | @alias, @alias, @alias

> See [RFC Process](https://github.com/aws/aws-cdk-rfcs#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `status/proposed`)
- [x] API bar raiser assigned (ping us at [#aws-cdk-rfcs](https://cdk-dev.slack.com/archives/C025ZFGMUCD) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `status/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [ ] API signed-off (label `status/api-approved` applied to pull request)
- [ ] Final comments period (label: `status/final-comments-period`)
- [ ] Approved and merged (label: `status/approved`)
- [ ] Execution plan submitted (label: `status/planning`)
- [ ] Plan approved and merged (label: `status/implementing`)
- [ ] Implementation complete (label: `status/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #952, opened 2026-06-19
feat: cdk debootstrap command

sanjanaravikumar-az:
This is a request for comments about `cdk debootstrap` which removes the CDK bootstrap stack and all associated resources from an AWS environment that get created when `cdk bootstrap` is run.. See #949 for additional details. 

---


text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+### README
+
+#### `cdk debootstrap`
+
+Removes the CDK bootstrap stack and all associated resources from an AWS
+environment that get created when `cdk bootstrap` is run.
+
+##### Usage
+
+When users want to debootstrap a particular account, they run:
```
rix0rrr: Style: we are writing a user manual. So the person to address here is "you":

```
When YOU want to remove bootstrap resources from your account, run:
```

(I also removed the term `debootstrap` here, because understanding your sentence requires users to learn a new term "debootstrapping" is, and we can fairly easily replace it with just describing what that does)

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+```console
+$ cdk debootstrap aws://123456789012/us-east-1
+```
+
+If users want to multi-select environments interactively (discovers bootstrapped
```
rix0rrr: YOU. Etc...

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+`cdk debootstrap` performs a complete teardown of the CDKToolkit CloudFormation
+stack in a specified environment:
+
+1. Empties the S3 staging bucket (all object versions)
+2. Deletes all images in the ECR container assets repository
+3. Deletes the CDKToolkit CloudFormation stack (which cascades IAM roles, KMS
+   key, SSM parameter)
+4. Deletes the orphaned S3 bucket (left behind due to its
+   `DeletionPolicy: Retain`)
```
rix0rrr: Style: This is very mechanical of how you're going to achieve the goal. When you are addressing a user, it's usually more interesting to them WHAT the end result is, not HOW you're going to achieve it. 

So: all bootstrapped resources, including the asset bucket and ECR repo and all files and images will be deleted. Because the bucket needs to be emptied before it can be deleted, the duration of this operation depends on how many objects are in the bucket.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+```
+This will permanently delete all assets in the staging bucket and ECR repository.
+The bootstrap stack can be re-created with `cdk bootstrap`, but stored assets cannot be recovered.
+
+You are about to destroy the CDK bootstrap stack in:
+  Account : 123456789012
+  Region  : us-east-1
+  Stack   : CDKToolkit (qualifier: hnb659fds)
+
```
rix0rrr, line 81: The order of presentation here seems weird to me, and takes up a lot of space. I suggest you take a second look at the information hierarchy. 

Also, why do these need to be 2 prompts? I should think we'd be able to combine them?

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+##### Idempotency
+
+The command is safe to re-run after partial failure. Each step checks current
+state before acting:
+
+- If the stack is already gone but the bucket still exists, it skips to bucket
+  cleanup
+- If the bucket is already empty, it skips to the next step
+- If everything is already removed, it exits cleanly
```
rix0rrr: From the point of view of a user, this could be one line:

```
If the command gets interrupted, you can run it again and it will pick up where it left off.
```

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+**Dependency detection:** CDK stacks reference bootstrap resources by
+deterministic naming convention, so we can detect them by the following steps:
+
+1. Call `DescribeStacks` to list all stacks in the account/region
+2. Match each stack's `RoleARN` field against the pattern
+   `cdk-${Qualifier}-cfn-exec-role-${Account}-${Region}`
+3. Check stack parameters and template body for references to the bootstrap
+   bucket - `cdk-${Qualifier}-assets-${Account}-${Region}` and ECR repo -
+   `cdk-${Qualifier}-container-assets-${Account}-${Region}`
+4. If there is a match that means that stack depends on this bootstrap
```
rix0rrr: I wouldn't do this. 

You can get the actual resource names from the actual bootstrap stack stack.

You can then search for ACTUAL resource names in the template, parameters and configuration of all application  stacks.

Why is that better?

-  You still catch all the stacks you were planning to catch by looking up predefined names
- But you *also* catch the stacks where users customized the stack template and decided to rename the resources.
- And it's not much more work.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+No. This is a new command that does not affect existing behavior. The existing
+`cdk bootstrap` behavior is unchanged.
+
+### What alternative solutions did you consider?
+
+**`cdk bootstrap --destroy`:** Adds a flag to the existing command. This flag
+wouldn't just modify a command's behavior, but invert it. Also,
+`cdk bootstrap --destroy --force` is three flags deep and is awkward from a UX
+perspective.
```
rix0rrr, line 277: I'm fine with it. It's not like this command is being used every day.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+   and routing to toolkit-lib.
+3. **Testing:** Unit tests for each step, integration tests for the full
+   sequence (happy path, partial failure re-run, dependent stacks warning).
+4. **Documentation:** Add `cdk debootstrap` to CLI reference docs, update
+   bootstrap documentation to reference the teardown path.
+
+### Are there any open issues that need to be addressed later?
+
+- **Cross-account notification:** A mechanism to warn trusted accounts when a
+  bootstrap stack they depend on is about to be destroyed.
```
rix0rrr: Nice and aspirational, but I don't think we can do this in practice.

There isn't a good way to "send a message to an account". In this case, I think we just trust the user to understand what they're doing.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+```console
+$ cdk debootstrap aws://123456789012/us-east-1
+```
+
+If users want to multi-select environments interactively (discovers bootstrapped
+environments and presents a picker):
+
+```console
+$ cdk debootstrap
+```
```
iankhou: This is an interesting choice, as it departs from the convention we use in `cdk bootstrap` - `cdk bootstrap` run without parameters inside of a CDK project does the following:


> By default, the CDK CLI will bootstrap all environments referenced in the CDK app or will determine an environment from default sources. This could be an environment specified using the --profile option, from environment variables, or default AWS CLI sources.

ref: https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-bootstrap.html

The RFC suggests an interactive mode, which we don't have for `cdk bootstrap`. One potential consideration is the use of this command in CI, where interactive mode would create problems.

Not necessarily saying that this needs to change, but flagging as a callout.

text/0949-cdk-debootstrap.md
iankhou: Agree, what is currently written looks like a procedure for that you will implement - which is more important to something like a technical design. For an RFC, it's OK to omit this level of detail, and focus on the high level. The end user cares more about the interface and the API, rather than the specific implementation steps.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+**Prompt 1 - Dependency warning** :
+
+This is shown when there are dependent stacks found.
+
+```
+⚠️  The following stacks use this bootstrap and will no longer be deployable or destroyable via CDK:
+  - MyAppStack
+  - MyOtherStack
+
+Do you want to continue? (y/N):
```
iankhou, line 68: Will this prompt be shown if the user uses `--yes`?

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+You are about to destroy the CDK bootstrap stack in:
+  Account : 123456789012
+  Region  : us-east-1
+  Stack   : CDKToolkit (qualifier: hnb659fds)
+
+Are you sure you want to proceed? (y/N):
+```
+
+`--force` skips the dependency check entirely. `--yes` skips all interactive
```
iankhou, line 85: I see that this answers my question above. But there is inconsistency in that we indicate "This prompt will always be shown unless the user uses `--yes`" for prompt 2 and not prompt 1. I recommend consolidating this, maybe just by removing the line I quoted.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+##### Flags
+
+| Flag | Description |
+| --- | --- |
+| `--force` | Skip the dependency check entirely |
+| `--yes` | Skip all interactive prompts (global flag) |
+| `--qualifier` | Target a non-default bootstrap stack (default: `hnb659fds`) |
+| `--toolkit-stack-name` | Override the default stack name `CDKToolkit` |
+| `--profile` | AWS named profile to use |
```
iankhou, line 96: Any non-interactive mode?

iankhou: Maybe have an `-i` flag (--interactive option), or something like that, if offering interactive mode.

text/0949-cdk-debootstrap.md
```diff
@@ -0,0 +1,335 @@
...
+
+- **Risk of accidental data loss**: Users could destroy bootstrap infrastructure
+  that active CDK applications depend on, making those applications
+  unmanageable. Unlike `cdk destroy` which is application scoped,
+  `cdk debootstrap` affects every CDK deployment in that account/region/qualifier
+  combination.
+
+This risk is mitigated by the dependency detection warning and the two
+confirmation prompts.
+
```
iankhou, line 178: We can also bootstrap again, right?

sanjanaravikumar-az: yes we can bootstrap again but new resources will get created, they can't recover their original ones.

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: non-interactive mode would be specifying the account/region, but i like the idea of --i

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: makes sense, will remove

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: nope, no prompts will be shown if --yes is used

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: this is interesting, i'll take a look into this..

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: ah makes sense, i'll try to see if there's any possible way to notify them but if not then i'll just remove this

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: true but even if it isn't used every day, wouldn't it be nice for the user experience? cdk debootstrap is easier to type, makes more sense with the deploy/destroy symmetry and is pretty simple to wire up. I don't really see any hard advantages of having --destroy over debootstrap which is why it was my final pick

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: got it, will update

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: this was as per momo's suggestion - one prompt for when dependent stacks are found to notify the user, and one as a confirmation prompt. I agree with this as well, since the --force flag can be used to bypass the dependency check if necessary

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: got it, done!

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: done

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: got it, updated

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: makes sense, this is a more robust solution, I've updated this section

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: I've updated the RFC to drop the picker and match cdk bootstrap's environment selection instead:

- cdk debootstrap aws://account/region → targets that environment
- cdk debootstrap in an app directory → targets all environments referenced by the app (same as cdk bootstrap)
- cdk debootstrap with no cdk.json and no args → errors asking for an explicit environment

This keeps debootstrap consistent with the rest of the environment-level command family (bootstrap, gc) and avoids the CI problem entirely. I left a note in the open issues about supporting environment discovery outside an app context (e.g. scanning configured regions) as a possible future enhancement.

text/0949-cdk-debootstrap.md
sanjanaravikumar-az: I ended up removing the interactive picker entirely, so there's no interactive mode to flag with -i. cdk debootstrap now always runs with either an explicit aws://account/region, or in an app directory to target the app's environments.

Issue #960, opened 2026-07-03
Log Alarm L2 Construct


rodrigotuna:
## Description

Add a `LogAlarm` L2 construct to `aws-cdk-lib/aws-cloudwatch` so customers can define CloudWatch **log alarms** (alarms evaluated against the results of a scheduled CloudWatch Logs query) in CDK, without hand-writing the `AWS::CloudWatch::LogAlarm` L1 resource.

A log alarm runs a scheduled Logs query (query string + aggregation over one or more log groups) and compares the aggregated result against a threshold using an M-out-of-N evaluation. It can also surface matching log lines in the alarm notification. The construct extends `AlarmBase` (like `Alarm`, `CompositeAlarm`, and `PromQLAlarm`), reuses the existing `ComparisonOperator`/`TreatMissingData` enums, and is interchangeable anywhere an `IAlarm` is accepted.

Related: PromQL Alarm L2.

**Proposed by:** @rodrigotuna

PR #962, opened 2026-07-03
RFC: Log Alarm L2 Construct

rodrigotuna:
Tracking issue: #960

This RFC proposes a new `LogAlarm` L2 construct in `aws-cdk-lib/aws-cloudwatch`, wrapping the `AWS::CloudWatch::LogAlarm` CloudFormation resource. It follows the accepted `PromQLAlarm` precedent (extends `AlarmBase`, reuses existing enums, interchangeable via `IAlarm`).

A proof-of-concept construct with unit tests and an integration test has been implemented and deployed against a real account, confirming it creates a valid `AWS::CloudWatch::LogAlarm`.

See `text/0960-logalarm-l2.md`.

### DIFF
--- a/text/0960-logalarm-l2.md
+++ b/text/0960-logalarm-l2.md
@@ -1,0 +1,338 @@
+# RFC: Log Alarm L2 Construct
+
+* **Original Author(s):** @rtuna
+* **Tracking Issue:** https://github.com/aws/aws-cdk-rfcs/issues/960
+* **API Bar Raiser:** TBD
+
+Add a new `LogAlarm` L2 construct to `aws-cdk-lib/aws-cloudwatch` so customers can define CloudWatch log alarms — alarms
+evaluated against the results of a scheduled CloudWatch Logs query — in CDK, without hand-writing the
+`AWS::CloudWatch::LogAlarm` L1 resource.
+
+## Summary
+
+`LogAlarm` lets customers alarm directly on logs. Instead of a metric + threshold + evaluation periods, a log alarm runs
+a scheduled CloudWatch Logs Insights (CWLI) query (a query string + aggregation over one or more log groups) and compares
+the aggregated result against a threshold using an M-out-of-N evaluation. It also can surface matching log lines in the
+alarm notification.
+The construct wraps the `AWS::CloudWatch::LogAlarm` CloudFormation resource, reuses the existing
+`ComparisonOperator`/`TreatMissingData` enums, and — like `Alarm` and `PromQLAlarm` — extends `AlarmBase` so it is
+interchangeable anywhere an `IAlarm` is accepted.
+
+---
+
+## Working Backwards
+
+`CHANGELOG: feat(cloudwatch): add Log Alarm L2 construct`
+
+### README
+
+#### Log Alarm
+
+Create a CloudWatch alarm that evaluates a scheduled CloudWatch Logs query against one or more log groups and alarms on the aggregated results.
+
+```ts
+import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
+import * as iam from 'aws-cdk-lib/aws-iam';
+import * as logs from 'aws-cdk-lib/aws-logs';
+import { Duration } from 'aws-cdk-lib';
+
+declare const logGroup: logs.LogGroup;
+declare const queryRole: iam.IRole;
+
+new cloudwatch.LogAlarm(this, 'ErrorRateAlarm', {
+  threshold: 5,
+  comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_THRESHOLD,
+  queryResultsToEvaluate: 3,
+  queryResultsToAlarm: 2,
+  scheduledQueryConfiguration: {
+    queryString: 'fields @message | filter @message like /ERROR/',
+    aggregationExpression: 'count(*)',
+    logGroupIdentifiers: [logGroup.logGroupName],
+    scheduledQueryRole: queryRole,
+    schedule: {
+      rate: Duration.minutes(5),
+      startTimeOffset: Duration.minutes(5),
+    },
+  },
+});
+```
+
+##### Including Log Lines in Notifications
+
+Set `actionLogLineCount` to include matching log lines in alarm notifications. When it is greater than 0, an
+`actionLogLineRole` is required so CloudWatch can read the log lines.
+
+```ts
+new cloudwatch.LogAlarm(this, 'ErrorRateAlarm', {
+  threshold: 5,
+  comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_THRESHOLD,
+  queryResultsToEvaluate: 3,
+  queryResultsToAlarm: 2,
+  actionLogLineCount: 10,
+  actionLogLineRole: logLineRole,
+  scheduledQueryConfiguration: { /* ... */ },
+});
+```
+
+##### Adding Actions
+
+`LogAlarm` extends `AlarmBase`, so you add alarm, OK, and insufficient-data actions the same way as existing alarms:
+
+```ts
+import * as cloudwatch_actions from 'aws-cdk-lib/aws-cloudwatch-actions';
+
+declare const logAlarm: cloudwatch.LogAlarm;
+declare const topic: sns.Topic;
+
+logAlarm.addAlarmAction(new cloudwatch_actions.SnsAction(topic));
+```
+
+Log alarms support SNS notification, Lambda, and Systems Manager OpsItem actions. Other action types (for example
+CloudWatch investigation or Systems Manager Incident) are not supported and are ignored by the service; adding one
+emits a synthesis-time warning.
+
+##### Importing an Existing Log Alarm
+
+```ts
+const imported = cloudwatch.LogAlarm.fromLogAlarmArn(
+  this, 'ImportedAlarm',
+  'arn:aws:cloudwatch:us-east-1:123456789012:alarm:MyLogAlarm',
+);
+
+const importedByName = cloudwatch.LogAlarm.fromLogAlarmName(this, 'ImportedByName', 'MyLogAlarm');
+```
+
+---
+
+Ticking the box below indicates that the public API of this RFC has been
+signed-off by the API bar raiser (the `status/api-approved` label was applied to
+the RFC pull request):
+
+```
+[ ] Signed-off by API Bar Raiser @xxxxx
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+A new `LogAlarm` L2 construct in the existing `aws-cdk-lib/aws-cloudwatch` module, wrapping the `AWS::CloudWatch::LogAlarm` CloudFormation resource.
+
+### What is a log alarm?
+
+A log alarm evaluates a scheduled CloudWatch Logs query against one or more log groups on a recurring schedule and
+alarms on an aggregation of the results (for example `count(*)`). It removes the old two-step pattern (metric filter →
+metric alarm) and can attach the matching log lines directly to the alarm notification.
+
+### Why should I use this feature?
+
+Customers monitoring logs previously had to create a metric filter and then a metric alarm, and then manually go back to
+the log group to find the offending log lines. `LogAlarm` collapses that into a single construct and can surface the
+triggering log lines in the notification, reducing MTTR.
+
+### Why do I need a new construct instead of the existing `Alarm`?
+
+The `Alarm` construct is built around a metric, comparison operator, threshold, and evaluation periods. A log alarm has
+a different model: instead of a metric it runs a scheduled query (`scheduledQueryConfiguration`), and instead of
+`evaluationPeriods`/`datapointsToAlarm` it uses `queryResultsToEvaluate` (N) and `queryResultsToAlarm` (M). A dedicated
+construct gives a clean, type-safe API without overloading `Alarm` with mutually exclusive property groups.
+
+---
+
+## Internal FAQ
+
+### Why are we doing this?
+
+Customers who want to alarm on logs currently must create a metric filter and then a metric alarm, and then manually
+return to the log group to find the log lines that triggered the alarm. CloudWatch launched `AWS::CloudWatch::LogAlarm`
+to collapse this into a single resource that runs a scheduled Logs query and can surface the triggering log lines in the
+notification. Without an L2, CDK users must hand-write the L1 `CfnLogAlarm`, manually render the `rate(...)` schedule
+expression, wire the IAM role ARN, and discover the service's validation rules by trial and error. An L2 provides a
+type-safe, validated, idiomatic API consistent with the existing `Alarm` and `PromQLAlarm` constructs.
+
+### Why should we _not_ do this?
+
+There is no strong reason not to. The core shape (query + aggregation + schedule + M-of-N) is stable and matches the
+CloudFormation resource.
+
+### What is the technical solution (design) of this feature?
+
+`LogAlarm` is a new class extending `AlarmBase` (mirroring `Alarm`, `CompositeAlarm`, and `PromQLAlarm`), so it is
+interchangeable anywhere an `IAlarm` is accepted:
+
+```
+IAlarm (extends IResource)
+  └── AlarmBase (abstract)
+        ├── Alarm
+        ├── CompositeAlarm
+        ├── PromQLAlarm
+        └── LogAlarm     ← NEW
+```
+
+It maps to the `AWS::CloudWatch::LogAlarm` resource (a distinct resource type, unlike `PromQLAlarm` which reuses
+`AWS::CloudWatch::Alarm`). The full API is in the Proposed API Design section below. Key design decisions:
+
+- **`Duration` for the schedule** (not raw `number` seconds as `PromQLAlarm` used) — per the CDK Design Guidelines'
+  preference for strong types. `rate` is rendered to a `rate(...)` schedule expression and must be a whole number of
+  minutes ≥ 1 (validated at synth); `startTimeOffset`/`endTimeOffset` convert to the integer seconds the L1 expects.
+- **`startTimeOffset` is required** — the scheduled-query service rejects a null start-time offset, so requiring it fails
+  fast at synth instead of at deploy (per the guideline on synth-time validation of always-fail-at-deploy input).
+- **`logGroupIdentifiers: string[]` (not `ILogGroup[]`)** — `aws-logs` already depends on `aws-cloudwatch`, so importing
+  `ILogGroup` into `aws-cloudwatch` would create a circular module dependency and break the jsii build (which is why no
+  `aws-cloudwatch` construct references `aws-logs`). `string[]` also matches how CWLI `SOURCE`-prefix queries identify log
+  groups.
+
+A proof-of-concept construct with unit tests and an integration test has been implemented and deployed against a real
+account, confirming it creates a valid `AWS::CloudWatch::LogAlarm`.
+
+### Is this a breaking change?
+
+No. This is a new construct; it does not alter any existing API.
+
+### What alternative solutions did you consider?
+
+1. **Extend `Alarm` with optional log-alarm props** — rejected. Would require making most `AlarmProps` optional and adding
+   mutual-exclusivity validation, degrading the type system for both standard and log alarms.
+2. **A static factory on `Alarm`** (e.g. `Alarm.fromLogQuery(...)`) — rejected. Hides the different configuration model and
+   still needs to return a distinct type.
+3. **New `LogAlarm extends AlarmBase`** — chosen. Clean dedicated surface, interchangeable via `IAlarm`, consistent with
+   the accepted `PromQLAlarm` precedent.
+
+### What are the drawbacks of this solution?
+
+The construct depends on the `AWS::CloudWatch::LogAlarm` schema published in `@aws-cdk/aws-service-spec`. The field
+constraints are final (`logGroupIdentifiers` optional, `startTimeOffset` range fixed); the L2 is written against that
+final shape and requires the corresponding published service-spec version at build time.
+
+### What is the high-level project plan?
+
+- [x] Implement the `LogAlarm` construct, unit tests, integration test, and README
+- [x] Validate against `aws-cdk-lib` built from `@aws-cdk/aws-service-spec` 0.1.190
+- [x] Deploy the integration test to a real account and confirm resource creation
+- [ ] RFC review and API bar-raiser sign-off
+- [ ] Open the public `aws/aws-cdk` PR
+- [ ] Address review feedback, merge, and release
+
+### Are there any open issues that need to be addressed later?
+
+- None.
+
+---
+
+## Proposed API Design
+
+### `LogAlarmProps`
+
+```ts
+export interface LogAlarmProps {
+  readonly threshold: number;
+  readonly comparisonOperator: ComparisonOperator;       // static-threshold operators only
+  readonly queryResultsToEvaluate: number;               // N (positive integer)
+  readonly queryResultsToAlarm: number;                  // M (positive integer, ≤ N)
+  readonly scheduledQueryConfiguration: ScheduledQueryConfiguration;
+
+  readonly alarmName?: string;                           // @default - generated
+  readonly alarmDescription?: string;                    // @default - none
+  readonly actionsEnabled?: boolean;                     // @default true
+  readonly treatMissingData?: TreatMissingData;          // @default - service default
+  readonly actionLogLineCount?: number;                  // 0–50
+  readonly actionLogLineRole?: IRole;                    // auto-created (trusts cloudwatch.amazonaws.com) when count > 0
+  readonly alarmActions?: IAlarmAction[];
+  readonly okActions?: IAlarmAction[];
+  readonly insufficientDataActions?: IAlarmAction[];
+  readonly tags?: { [key: string]: string };
+}
+
+export interface ScheduledQueryConfiguration {
+  readonly queryString: string;
+  readonly aggregationExpression: string;                // e.g. count(*)
+  readonly logGroupIdentifiers?: string[];               // names or ARNs; optional for inline CWLI SOURCE queries
+  readonly scheduledQueryRole?: IRole;                   // auto-created (trusts logs.amazonaws.com) when omitted
+  readonly schedule: ScheduledQuerySchedule;
+}
+
+export interface ScheduledQuerySchedule {
+  readonly rate: Duration;                               // rendered to rate(...); whole minutes
+  readonly startTimeOffset: Duration;                    // required by the service
+  readonly endTimeOffset?: Duration;                     // @default - none
+}
+```
+
+### `LogAlarm`
+
+```ts
+@propertyInjectable
+export class LogAlarm extends AlarmBase {
+  public static readonly PROPERTY_INJECTION_ID: string;
+  public static fromLogAlarmArn(scope: Construct, id: string, alarmArn: string): IAlarm;
+  public static fromLogAlarmName(scope: Construct, id: string, alarmName: string): IAlarm;
+  public get alarmArn(): string;
+  public get alarmName(): string;
+  // The scheduled-query role (provided or auto-created); exposed so callers can grant/extend it.
+  public readonly scheduledQueryRole: IRole;
+  // The log-line role (provided or auto-created when actionLogLineCount > 0); undefined otherwise.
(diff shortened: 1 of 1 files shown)


text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+evaluated against the results of a scheduled CloudWatch Logs query — in CDK, without hand-writing the
+`AWS::CloudWatch::LogAlarm` L1 resource.
+
+## Summary
+
+`LogAlarm` lets customers alarm directly on logs. Instead of a metric + threshold + evaluation periods, a log alarm runs
+a scheduled Logs query (a query string + aggregation over one or more log groups) and compares the aggregated result
+against a threshold using an M-out-of-N evaluation. It also can surface matching log lines in the alarm notification.
+The construct wraps the `AWS::CloudWatch::LogAlarm` CloudFormation resource, reuses the existing
+`ComparisonOperator`/`TreatMissingData` enums, and — like `Alarm` and `PromQLAlarm` — extends `AlarmBase` so it is
```
chintan-vaghani2008, line 18: Please note that ComparisonOperator we use for log alarm doesn't support `LessThanLowerOrGreaterThanUpperThreshold` | `LessThanLowerThreshold` | `GreaterThanUpperThreshold` being supported under Metric Alarm. These operations are supported for anomaly detection metric alarm and it is not supported for log alarm.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+* **API Bar Raiser:** TBD
+
+Add a new `LogAlarm` L2 construct to `aws-cdk-lib/aws-cloudwatch` so customers can define CloudWatch log alarms — alarms
+evaluated against the results of a scheduled CloudWatch Logs query — in CDK, without hand-writing the
+`AWS::CloudWatch::LogAlarm` L1 resource.
+
+## Summary
+
+`LogAlarm` lets customers alarm directly on logs. Instead of a metric + threshold + evaluation periods, a log alarm runs
+a scheduled Logs query (a query string + aggregation over one or more log groups) and compares the aggregated result
```
chintan-vaghani2008: nit: Logs insight query

text/0960-logalarm-l2.md
chintan-vaghani2008: We can still use this enum and validate the parameter at synth. ( Probably, PromQL should also be doing the similar thing if they are reusing the enum ?)

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+  comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_THRESHOLD,
+  queryResultsToEvaluate: 3,
+  queryResultsToAlarm: 2,
+  scheduledQueryConfiguration: {
+    queryString: 'fields @message | filter @message like /ERROR/',
+    aggregationExpression: 'count(*)',
+    logGroupIdentifiers: [logGroup.logGroupName],
+    scheduledQueryRole: queryRole,
+    schedule: {
+      frequency: Duration.minutes(5),
```
chintan-vaghani2008: nit: Should we just name it as `rate` to align it with scheduled expression rate expression naming ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+Create a CloudWatch alarm that evaluates a scheduled CloudWatch Logs query against one or more log groups and alarms on the aggregated results.
+
+```ts
+import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
+import * as iam from 'aws-cdk-lib/aws-iam';
+import * as logs from 'aws-cdk-lib/aws-logs';
+import { Duration } from 'aws-cdk-lib';
+
+declare const logGroup: logs.LogGroup;
+declare const queryRole: iam.IRole;
```
chintan-vaghani2008, line 40: Can we see if SQ ( as in GA) supports anything to help the role creation ? If so, customers can use that to simplify the role creationrather than having their own code to create one.

If there is one, we can also replicate the similar thing for log line role as well.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+
+`LogAlarm` extends `AlarmBase`, so you add alarm, OK, and insufficient-data actions the same way as existing alarms:
+
+```ts
+import * as cloudwatch_actions from 'aws-cdk-lib/aws-cloudwatch-actions';
+
+declare const logAlarm: cloudwatch.LogAlarm;
+declare const topic: sns.Topic;
+
+logAlarm.addAlarmAction(new cloudwatch_actions.SnsAction(topic));
```
chintan-vaghani2008, line 87: Let's see if we can add validation on supported actions for log alarm as well on synth  ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+Customers who want to alarm on logs currently must create a metric filter and then a metric alarm, and then manually
+return to the log group to find the log lines that triggered the alarm. CloudWatch launched `AWS::CloudWatch::LogAlarm`
+to collapse this into a single resource that runs a scheduled Logs query and can surface the triggering log lines in the
+notification. Without an L2, CDK users must hand-write the L1 `CfnLogAlarm`, manually render the `rate(...)` schedule
+expression, wire the IAM role ARN, and discover the service's validation rules by trial and error. An L2 provides a
+type-safe, validated, idiomatic API consistent with the existing `Alarm` and `PromQLAlarm` constructs.
+
+### Why should we _not_ do this?
+
+The only reason not to is that the `ScheduledQueryConfiguration` field constraints are still being finalised upstream
```
chintan-vaghani2008: What do we mean here ? which constraints ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+   mutual-exclusivity validation, degrading the type system for both standard and log alarms.
+2. **A static factory on `Alarm`** (e.g. `Alarm.fromLogQuery(...)`) — rejected. Hides the different configuration model and
+   still needs to return a distinct type.
+3. **New `LogAlarm extends AlarmBase`** — chosen. Clean dedicated surface, interchangeable via `IAlarm`, consistent with
+   the accepted `PromQLAlarm` precedent.
+
+### What are the drawbacks of this solution?
+
+The `ScheduledQueryConfiguration` field constraints are still being finalised upstream (notably whether
+`LogGroupIdentifiers` is required and the exact `StartTimeOffset` range). The construct follows the published
```
chintan-vaghani2008: It is final right ? LogGroupIdentifiers is optional and StartTimeOffset is also final as of now.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+- [x] Validate against `aws-cdk-lib` built from `@aws-cdk/aws-service-spec` 0.1.190
+- [x] Deploy the integration test to a real account and confirm resource creation
+- [ ] RFC review and API bar-raiser sign-off
+- [ ] Open the public `aws/aws-cdk` PR
+- [ ] Address review feedback, merge, and release
+
+### Are there any open issues that need to be addressed later?
+
+- Alignment of `ScheduledQueryConfiguration` field constraints across the Coral model, CFN schema, and Hermes validation
+  (tracked in EVENTS-4979): `LogGroupIdentifiers` required→optional, `StartTimeOffset` range `[1, 2592000]`,
```
chintan-vaghani2008: Can't we just publish this with updated schema given we have already started the deployments? Or we plan to update it once we complete all deployments?

chintan-vaghani2008: Also, curious if these validations will be done at L1 construct based on CFN schema being published ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+
+  readonly alarmName?: string;                           // @default - generated
+  readonly alarmDescription?: string;                    // @default - none
+  readonly actionsEnabled?: boolean;                     // @default true
+  readonly treatMissingData?: TreatMissingData;          // @default - service default
+  readonly actionLogLineCount?: number;                  // 0–50; requires role when > 0
+  readonly actionLogLineRole?: IRole;
+  readonly alarmActions?: IAlarmAction[];
+  readonly okActions?: IAlarmAction[];
+  readonly insufficientDataActions?: IAlarmAction[];
```
chintan-vaghani2008, line 242: Tags ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+  readonly queryResultsToAlarm: number;                  // M (positive integer, ≤ N)
+  readonly scheduledQueryConfiguration: ScheduledQueryConfiguration;
+
+  readonly alarmName?: string;                           // @default - generated
+  readonly alarmDescription?: string;                    // @default - none
+  readonly actionsEnabled?: boolean;                     // @default true
+  readonly treatMissingData?: TreatMissingData;          // @default - service default
+  readonly actionLogLineCount?: number;                  // 0–50; requires role when > 0
+  readonly actionLogLineRole?: IRole;
+  readonly alarmActions?: IAlarmAction[];
```
chintan-vaghani2008, line 240: Should we have default values for these as well or we can mark default none ? We can follow what is being followed for metric alarm. Same goes for other optional parameters like `actionLogLineRole`

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+  readonly okActions?: IAlarmAction[];
+  readonly insufficientDataActions?: IAlarmAction[];
+}
+
+export interface ScheduledQueryConfiguration {
```
chintan-vaghani2008, line 246: Tag for SQ as well ?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,328 @@
...
+  readonly queryString: string;
+  readonly aggregationExpression: string;                // e.g. count(*)
+  readonly logGroupIdentifiers: string[];                // names, ARNs, or CWLI SOURCE patterns
+  readonly scheduledQueryRole: IRole;
+  readonly schedule: ScheduledQuerySchedule;
+}
+
+export interface ScheduledQuerySchedule {
+  readonly frequency: Duration;                          // rendered to rate(...); whole minutes
```
chintan-vaghani2008: as mentioned above, should we name this as `rate` ?

text/0960-logalarm-l2.md
chintan-vaghani2008: I see you already covered below:

> `comparisonOperator` must be a static-threshold operator (anomaly-detection operators rejected).

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,336 @@
...
+  readonly queryResultsToEvaluate: number;               // N (positive integer)
+  readonly queryResultsToAlarm: number;                  // M (positive integer, ≤ N)
+  readonly scheduledQueryConfiguration: ScheduledQueryConfiguration;
+
+  readonly alarmName?: string;                           // @default - generated
+  readonly alarmDescription?: string;                    // @default - none
+  readonly actionsEnabled?: boolean;                     // @default true
+  readonly treatMissingData?: TreatMissingData;          // @default - service default
+  readonly actionLogLineCount?: number;                  // 0–50
+  readonly actionLogLineRole?: IRole;                    // auto-created (trusts cloudwatch.amazonaws.com) when count > 0
```
chintan-vaghani2008, line 239: Do you mean auto created only when not passed ? Also, we plan to auto create it , would we have all required parameters in place to auto create it ? Let's also see what further reviewer has take on this.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,336 @@
...
+Customers who want to alarm on logs currently must create a metric filter and then a metric alarm, and then manually
+return to the log group to find the log lines that triggered the alarm. CloudWatch launched `AWS::CloudWatch::LogAlarm`
+to collapse this into a single resource that runs a scheduled Logs query and can surface the triggering log lines in the
+notification. Without an L2, CDK users must hand-write the L1 `CfnLogAlarm`, manually render the `rate(...)` schedule
+expression, wire the IAM role ARN, and discover the service's validation rules by trial and error. An L2 provides a
+type-safe, validated, idiomatic API consistent with the existing `Alarm` and `PromQLAlarm` constructs.
+
+### Why should we _not_ do this?
+
+There is no strong reason not to. The `ScheduledQueryConfiguration` field constraints are final (`logGroupIdentifiers`
```
chintan-vaghani2008: Do we have to mention this now given we are going with what will be the final spec?

> The `ScheduledQueryConfiguration` field constraints are final (`logGroupIdentifiers`
is optional, and the `startTimeOffset` range is fixed), and the construct tracks the published
`@aws-cdk/aws-service-spec`

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,336 @@
...
+- [x] Implement the `LogAlarm` construct, unit tests, integration test, and README
+- [x] Validate against `aws-cdk-lib` built from `@aws-cdk/aws-service-spec` 0.1.190
+- [x] Deploy the integration test to a real account and confirm resource creation
+- [ ] RFC review and API bar-raiser sign-off
+- [ ] Open the public `aws/aws-cdk` PR
+- [ ] Address review feedback, merge, and release
+
+### Are there any open issues that need to be addressed later?
+
+- None. The `ScheduledQueryConfiguration` field constraints are final; the L2 tracks the published
```
chintan-vaghani2008: Again, we may not have to mention this "The `ScheduledQueryConfiguration` field constraints are final; the L2 tracks the published `@aws-cdk/aws-service-spec` for the `AWS::CloudWatch::LogAlarm` schema." unless needed. It will confuse the reviewer further.

chintan-vaghani2008:
Changes now looks good to me.

chintan-vaghani2008 (approved):

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,338 @@
...
+
+It maps to the `AWS::CloudWatch::LogAlarm` resource (a distinct resource type, unlike `PromQLAlarm` which reuses
+`AWS::CloudWatch::Alarm`). The full API is in the Proposed API Design section below. Key design decisions:
+
+- **`Duration` for the schedule** (not raw `number` seconds as `PromQLAlarm` used) — per the CDK Design Guidelines'
+  preference for strong types. `rate` is rendered to a `rate(...)` schedule expression and must be a whole number of
+  minutes ≥ 1 (validated at synth); `startTimeOffset`/`endTimeOffset` convert to the integer seconds the L1 expects.
+- **`startTimeOffset` is required** — the scheduled-query service rejects a null start-time offset, so requiring it fails
+  fast at synth instead of at deploy (per the guideline on synth-time validation of always-fail-at-deploy input).
+- **`logGroupIdentifiers: string[]` (not `ILogGroup[]`)** — `aws-logs` already depends on `aws-cloudwatch`, so importing
```
matboros: **Design question: `ILogGroupRef[]` instead of `string[]`**

The circular-dependency reasoning here is correct for `ILogGroup` (which lives in `aws-logs`), but `ILogGroupRef` lives in the shared `interfaces/generated/` layer (`aws-cdk-lib/interfaces/generated/aws-logs-interfaces.generated.ts`) — specifically designed to be importable from any module without creating cycles.

`ILogGroupRef` is already used cross-module by `aws-ec2`, `aws-events-targets`, and `custom-resources` without circular dependency issues.

Accepting `ILogGroupRef[]` would give users type safety and let them pass LogGroup constructs directly instead of extracting names/ARNs manually. The current `string[]` works but forgoes the type safety that CDK's typed reference interfaces are designed to provide.

Would it make sense to accept `ILogGroupRef[]` (or `Array<string | ILogGroupRef>` for flexibility) here?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,338 @@
...
+```ts
+@propertyInjectable
+export class LogAlarm extends AlarmBase {
+  public static readonly PROPERTY_INJECTION_ID: string;
+  public static fromLogAlarmArn(scope: Construct, id: string, alarmArn: string): IAlarm;
+  public static fromLogAlarmName(scope: Construct, id: string, alarmName: string): IAlarm;
+  public get alarmArn(): string;
+  public get alarmName(): string;
+  // The scheduled-query role (provided or auto-created); exposed so callers can grant/extend it.
+  public readonly scheduledQueryRole: IRole;
```
matboros: **Design question: Grant helper methods for the roles**

The construct exposes `scheduledQueryRole` and `actionLogLineRole` as public properties, which is good. But users who pass in a custom role need to know:
- Which service principal to set as the trust policy (`logs.amazonaws.com` / `cloudwatch.amazonaws.com`)
- Which exact IAM actions to grant (`logs:StartQuery`, `logs:GetQueryResults`, `logs:StopQuery`, etc.)
- Which resource ARN format to scope permissions to

This is non-trivial and exactly the kind of knowledge CDK grant methods encapsulate. Consider whether the construct should provide helpers like:

```ts
// Grant the scheduled query role permission to query additional log groups
logAlarm.grantQuery(additionalLogGroup);

// Grant the action log line role permission to read from additional log groups
logAlarm.grantReadLogLines(additionalLogGroup);
```

The auto-creation path handles the 80% case, but explicit grants become important when users pass custom roles or add log groups after construction. What do you think — is the current "expose role, user figures out permissions" sufficient, or would grant helpers meaningfully reduce friction?

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,338 @@
...
+  readonly okActions?: IAlarmAction[];
+  readonly insufficientDataActions?: IAlarmAction[];
+  readonly tags?: { [key: string]: string };
+}
+
+export interface ScheduledQueryConfiguration {
+  readonly queryString: string;
+  readonly aggregationExpression: string;                // e.g. count(*)
+  readonly logGroupIdentifiers?: string[];               // names or ARNs; optional for inline CWLI SOURCE queries
+  readonly scheduledQueryRole?: IRole;                   // auto-created (trusts logs.amazonaws.com) when omitted
```
matboros: **Design question: Scheduled query observability**

The scheduled query is the core evaluation mechanism of the alarm. Does the service emit any CloudWatch metrics about the query execution health (e.g., query latency, records scanned, query errors/timeouts)? If so, it might be worth exposing convenience `metric*()` methods so users can monitor the health of their alarm's underlying query — separate from the alarm state itself.

If no such metrics exist today, this is a non-issue. Just flagging in case the service does emit them and the RFC should document how users observe query health.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,338 @@
...
+  public readonly actionLogLineRole?: IRole;
+  constructor(scope: Construct, id: string, props: LogAlarmProps);
+}
+```
+
+### Validation performed at synth
+
+- `comparisonOperator` must be a static-threshold operator (anomaly-detection operators rejected).
+- `queryResultsToEvaluate` / `queryResultsToAlarm` positive integers; `queryResultsToAlarm ≤ queryResultsToEvaluate`.
+- `actionLogLineCount` integer in `[0, 50]`; requires `actionLogLineRole` when `> 0`.
```
matboros: nit: This line says `actionLogLineCount` "requires `actionLogLineRole` when `> 0`", but line 239 and the class section (line 273) say the role is "auto-created" when count > 0. The design intent seems clear (auto-create when omitted), so this validation bullet should probably say "when `> 0` and no `actionLogLineRole` is provided, one is auto-created" to match the rest of the RFC.

text/0960-logalarm-l2.md
```diff
@@ -0,0 +1,338 @@
+# RFC: Log Alarm L2 Construct
+
+* **Original Author(s):** @rtuna
```
matboros: nit: Author listed as `@rtuna` here but the GitHub account is `@rodrigotuna`.

text/0960-logalarm-l2.md
matboros: **Design question: `ILogGroupRef[]` instead of `string[]`**

The circular-dependency reasoning here is correct for `ILogGroup` (which lives in `aws-logs`), but `ILogGroupRef` lives in the shared `interfaces/generated/` layer (`aws-cdk-lib/interfaces/generated/aws-logs-interfaces.generated.ts`) — specifically designed to be importable from any module without creating cycles.

`ILogGroupRef` is already used cross-module by `aws-ec2`, `aws-events-targets`, and `custom-resources` without circular dependency issues.

Accepting `ILogGroupRef[]` would give users type safety and let them pass LogGroup constructs directly instead of extracting names/ARNs manually. The current `string[]` works but forgoes the type safety that CDK's typed reference interfaces are designed to provide.

Would it make sense to accept `ILogGroupRef[]` (or `Array<string | ILogGroupRef>` for flexibility) here?

text/0960-logalarm-l2.md
matboros, line 279: **Design question: Role mutation behavior and `addLogGroup()`**

The RFC says these roles are "auto-created" when omitted — clear. But what happens when a user **passes in** a custom role? Does the construct auto-mutate it (add the necessary `logs:StartQuery`, `logs:GetQueryResults`, etc. permissions scoped to the provided log groups), or does it trust the user configured it correctly?

Most CDK constructs that accept optional roles (e.g., Lambda with execution role) do auto-mutate — they add the minimum required permissions to the provided role. If `LogAlarm` does the same, that should be stated explicitly in the RFC so the behavior is clear.

If it does auto-mutate, then standalone grant methods probably aren't needed. But it might be worth considering an `addLogGroup(logGroup)` method that:
1. Adds the log group to the L1's `LogGroupIdentifiers` array
2. Grants the `scheduledQueryRole` permission to query it

This would make it easier to use the construct in loops or add log groups conditionally after construction. Not a must — just something to consider for ergonomics:

```ts
const alarm = new cloudwatch.LogAlarm(this, 'Alarm', { ... });

for (const lg of additionalLogGroups) {
  alarm.addLogGroup(lg);
}
```

text/0960-logalarm-l2.md
matboros, line 257: **Design question: Scheduled query observability**

The scheduled query is the core evaluation mechanism of the alarm. Does the service emit any CloudWatch metrics about the query execution health (e.g., query latency, records scanned, query errors/timeouts)? If so, it might be worth exposing convenience `metric*()` methods so users can monitor the health of their alarm's underlying query — separate from the alarm state itself.

If no such metrics exist today, this is a non-issue. Just flagging in case the service does emit them and the RFC should document how users observe query health.

text/0960-logalarm-l2.md
matboros: nit: This line says `actionLogLineCount` "requires `actionLogLineRole` when `> 0`", but line 239 and the class section (line 273) say the role is "auto-created" when count > 0. The design intent seems clear (auto-create when omitted), so this validation bullet should probably say "when `> 0` and no `actionLogLineRole` is provided, one is auto-created" to match the rest of the RFC.

text/0960-logalarm-l2.md
matboros: nit: Author listed as `@rtuna` here but the GitHub account is `@rodrigotuna`.

matboros (changes requested):

text/0960-logalarm-l2.md
rodrigotuna: You are right, and the stated rationale was wrong. Changed to `logGroups: ILogGroupRef[]` and corrected the reasoning: the cycle only applies to `ILogGroup`, and `aws-cloudwatch` already imports `IAlarmRef` from the same generated interface layer.

Two notes: a `string | ILogGroupRef` union is not jsii-compatible, so the prop is `ILogGroupRef[]` only, with imported log groups expressible via `LogGroup.fromLogGroupName()`. It renders `logGroupRef.logGroupName`, not the ARN — the scheduled-query service rejects log group ARNs with the trailing `:*` that `logGroupArn` carries.

text/0960-logalarm-l2.md
rodrigotuna: Documented under "IAM role behaviour". A supplied role is used verbatim — no trust statements or permissions added. Rationale: silently broadening a role the customer owns can defeat a deliberately narrow scope and hides the effective permissions from the code that defines the role. Both roles are exposed as public readonly properties, so callers add permissions explicitly when they want them.

Given that, `addLogGroup()` is not offered: without mutation it could append to `LogGroupIdentifiers` without granting query access to the added log group, producing an alarm that deploys but never evaluates.

text/0960-logalarm-l2.md
rodrigotuna: Checked empirically by deploying a log alarm and diffing metrics. `AWS/Logs` does emit `QueryBytesScanned` and `ConcurrencyUsed` (both undocumented), but neither is dimensioned by scheduled query — `QueryBytesScanned` has no dimensions at all, so it aggregates every Insights query in the account and region. A `metric*()` method on a single alarm would therefore return an unrelated figure, so none are added.

No latency, error, or timeout metric exists. Per-execution health comes from `GetScheduledQueryHistory` (`Running | InvalidQuery | Complete | Failed | Timeout`) and the alarm `EvaluationState`/`StateReason`. Documented under "Scheduled query observability".

text/0960-logalarm-l2.md
rodrigotuna: Fixed. The role is auto-created when omitted; corrected here and in the two other places with the same wording.

text/0960-logalarm-l2.md
rodrigotuna: Fixed.

text/0960-logalarm-l2.md
matboros, line 279: Thanks for the clear rationale. I understand the concern about silently broadening a customer-owned role.

However, the CDK Design Guidelines' [Roles section](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#roles) explicitly expects constructs to grant permissions to provided roles — that's why the recommended type is `iam.IRoleRef & iam.IGrantable`. The `addToRolePolicy` method is described as a **must** on the construct interface.

In practice, most CDK constructs that accept a role **do** mutate it for feature-specific permissions:
- **Lambda**: adds DLQ (`sns:Publish`/`sqs:SendMessage`), X-Ray, Insights, EFS permissions to provided roles
- **Step Functions**: adds service integration permissions, KMS grants
- **CodeBuild**: adds source access, VPC, and logging permissions
- **Firehose**: adds KMS and Kinesis source read grants

The established CDK escape hatch for users who want immutable roles is `Role.fromRoleArn(scope, id, arn, { mutable: false })` — policy additions become no-ops (with a warning). This gives users explicit control without requiring every construct to offer a non-mutation path.

For `LogAlarm` specifically: the construct knows exactly which log groups need to be queried (they're in the props). Without the right permissions, the alarm will deploy but silently fail to evaluate — a worse user experience than "role was broadened." The CDK convention is to add what's needed and let users opt out via imported immutable roles.

I'd suggest following the established pattern: auto-configure the provided role with the minimum permissions for the specified log groups. This also re-opens the `addLogGroup()` conversation — if the role is mutated, `addLogGroup()` can atomically add both the log group identifier *and* the corresponding permissions.

**If you prefer to keep the pass-through approach**, then I'd ask for explicit grant helper methods (e.g. `grantQuery(logGroup)` / `grantReadLogLines(logGroup)`) so users don't have to read docs to figure out the exact IAM actions and resource ARN formats needed. Either the construct configures the role automatically, or it provides methods that do it on demand — but asking users to hand-craft the policy statements from documentation defeats the purpose of an L2.

Not blocking — but one of these two approaches should be chosen explicitly in the RFC.

matboros (changes requested):

text/0960-logalarm-l2.md
rodrigotuna: Agreed — switched to granting in `7820d2b`. The construct now grants to whichever role is in use, scoped to the `logGroups` ARNs with a region-wide fallback for inline `SOURCE` queries, and `addToRolePolicy` is added — I had missed that it is a guidelines must. The integ test's hand-written policy is redundant now and gone.

Trust is still not modified on a supplied role, so it must already trust the right service principal. Kept the prop as `IRole` following the `Function.role` precedent — `IRole` extends `IRoleRef` and, via `IIdentity → IPrincipal`, `IGrantable`, so it satisfies the guideline and also provides `addToPrincipalPolicy`. Left `addLogGroup()` out since log groups are supplied up front.

matboros (approved):
lgtm

PR #981, opened 2026-08-04
RFC 972: Structured design context in synthesized templates

satyakigh:
This is a request for comments about Structured design context in synthesized templates. See #972 for
additional details. 

---

### DIFF
--- a/text/0972-metadata-context.md
+++ b/text/0972-metadata-context.md
@@ -58,11 +58,17 @@ This renders a `Metadata.Context` block on the `AWS::SQS::Queue` resource:
   "Metadata": {
     "Context": {
       "why": "buffer order events async; 14d retention = compliance window",
-      "must": ["VisTimeout >= 6x fn timeout, else dup on retry"],
+      "must": [
+        "VisTimeout >= 6x fn timeout, else dup on retry"
+      ],
       "mutable": "change-with-constraints",
-      "mutability": { "QueueName": "must-never-change" },
+      "mutability": {
+        "QueueName": "must-never-change"
+      },
       "ops": "check ApproxAgeOfOldestMsg before cutting VisTimeout",
-      "failureModes": ["retry 3x w/ exp backoff before DLQ"]
+      "failureModes": [
+        "retry 3x w/ exp backoff before DLQ"
+      ]
     }
   }
 }
@@ -112,9 +118,10 @@ MetadataContext.of(stack).add({
 ```
 
 Record where context came from and how much to trust it with the `trust`
-field - useful when context is produced by tooling rather than authored by
-the resource owner. When omitted, `source` defaults to `AUTHORED` and
-`confidence` to `MEDIUM`:
+field. `AUTHORED` means the context was explicitly declared through this API;
+it does not assert that a human wrote it. Producers that infer context from
+comments, commits, or other artifacts set the corresponding source instead.
+When omitted, `source` defaults to `AUTHORED` and `confidence` to `MEDIUM`:
 
 ```ts
 declare const queue: sqs.Queue;
@@ -180,6 +187,13 @@ Keep free-text values terse - drop articles and use symbols (`->`, `>=`,
 template size limit. Prefer `must` for binding rules whose violation breaks
 something, and `why` for reasoning and rejected alternatives.
 
+CDK already serializes the complete template during synthesis and emits a
+warning above 80% of its conservative 1,000,000-character threshold;
+`Metadata.Context` is included in that measurement. This RFC adds no
+context-specific size validator and never silently trims declared context. The
+existing warning remains the synth-time signal; authoring tools may respond
+using the v1 tier/drop order described in appendix A.
+
 ---
 
 Ticking the box below indicates that the public API of this RFC has been
@@ -241,17 +255,25 @@ Concrete situations this feature addresses:
   one.
 * **Organizational context at scale** - template-level `ref` entries point to shared
   context files (e.g. org-wide encryption rules) without repeating them in every
-  template, and `trust` distinguishes human-authored context from tool-inferred context
+  template, and `trust` distinguishes explicitly declared context from inferred context
   so consumers can weight it appropriately.
 
-These are not just expectations. We benchmarked the alternatives against each other - no
-embedded context, context as source comments, and structured `Metadata.Context` - on the
-same CloudFormation update tasks. Structured context consistently produced the best
-outcomes, and it is the only one of those approaches that survives synthesis and is
-retrievable from a deployed stack.
+These are not just expectations. We evaluated the alternatives on the same
+CloudFormation update tasks. The results showed that supplying design context improved
+outcomes, structured `Metadata.Context` made that context durable and machine-addressable,
+and context-aware tooling used the structured fields most effectively. Structured metadata
+is the tested form designed to survive CDK synthesis and remain structurally retrievable
+from a deployed stack.
+
+Brownfield adoption does not require manually seeding every resource. A companion
+bootstrapping skill is being developed to read existing CDK/CloudFormation source,
+comments, git history, tests, and companion service code, then propose explicit
+`MetadataContext` declarations (or `Metadata.Context` blocks for raw templates) with
+provenance and declared gaps. This subsidizes discovery and authoring effort, while
+keeping heuristic inference and its review outside the core API contract.
 
 If you already maintain design context in READMEs or wikis, this feature does not replace
-them - it puts the *operationally binding* subset where every consumer of the deployed
+them - it puts the *operationally relevant* subset where every consumer of the deployed
 stack can actually find it.
 
 ## Internal FAQ
@@ -274,20 +296,20 @@ never made it into the template. The failure is not that the agent is careless -
 that the artifact it reads does not contain the constraint it needs to respect.
 
 **We benchmarked that claim rather than assuming it.** Before settling the API we compared
-four ways of carrying design context on the same CloudFormation update tasks: no embedded
-context, context as source-file comments, structured `Metadata.Context`, and structured
-context read by tooling that understands the vocabulary. Context-free templates fared worst
-on the tasks whose correct answer depended on knowledge the template did not contain;
-structured context produced the largest improvement; context-aware tooling improved on that
-again. Appendix B has the comparison.
+four conditions on the same CloudFormation update tasks: no embedded context, context as
+natural inline YAML comments, structured `Metadata.Context`, and structured context read
+by tooling that understands the vocabulary. Context-free templates fared worst on the
+tasks whose correct answer depended on knowledge the template did not contain; structured
+context produced the largest improvement; context-aware tooling improved on that again.
+Appendix B has the comparison.
 
-The benchmark also settled the obvious objection, which is that this is what comments are
-for. Comments did score well - where they exist. But they are unavailable to CDK users:
-synthesized JSON carries no comments, and no comment in CDK source survives to the
-template. They are equally unavailable to *any* consumer retrieving a deployed template
-through `GetTemplate`, regardless of how it was authored. Structured metadata is the only
-form of this that survives synthesis and deployment, which is what makes it the right
-target for CDK.
+The benchmark also tested the obvious objection, which is that this is what comments are
+for. Comments scored well where they were present in the input. But CDK synthesis does not
+preserve source comments by contract: synthesized JSON carries no comments unless a
+separate mechanism translates them into template data. Consumers retrieving a deployed
+template through `GetTemplate` therefore cannot rely on CDK source comments. Structured
+metadata is the durable carrier; automatic comment translation remains a compatible but
+heuristic producer alternative described below.
 
 **CDK is uniquely positioned to populate it.** CDK users do not author the template — they
 author constructs, and the template is generated. So the authoring surface has to exist in
@@ -406,13 +428,13 @@ no special-casing required.
 
 #### What is explicitly out of scope for this RFC
 
-This RFC covers **explicit authoring only**: context that a developer states in CDK code.
-Automatically deriving context from other sources — inferring `why` from code, or
-populating `deps` by analysing the resolved template — is not part of this API commitment.
-Any such mechanism is heuristic, and an API contract should not rest on a heuristic. If
-derivation is added later it layers on top of this declaration model without changing it,
-and the `trust` field already exists so derived context can identify itself as such
-(`src: infer`) rather than masquerading as authored.
+This RFC covers **explicit declarations only**: context supplied through the CDK API.
+During synthesis the library does not inspect source comments, git history, tests, service
+code, or deployed state. Automatically deriving `why` or resolving `deps` remains outside
+this API commitment because those mechanisms are heuristic or require a later synthesis
+phase. The companion brownfield bootstrapping skill can layer on top by emitting calls to
+this API, with `trust` and `gaps` identifying evidence and uncertainty, without changing
+the declaration or wire-format contract.
 
 ### Is this a breaking change?
 
@@ -437,32 +459,45 @@ No. The feature is purely additive and opt-in:
    for free via jsii), and integration (the `Mixins` form, `AspectPriority` defaults, and
    eventual L2 integration points all live in core). The external path remains open to
    anyone — the API proposed here does not preclude it.
-2. **Reusing existing `Description` properties.** Many L2s expose `description` props
+2. **Automatic propagation of leading source comments.** A prototype Aspect follows a
+   resource's creation stack to its source location, reads the leading comment, applies an
+   anti-fabrication gate, and writes an attributed `Metadata.Context.why` during synthesis.
+   This can reduce authoring effort for well-commented CDK code, and a compile-time
+   transformer could avoid runtime stack inspection. It is deferred as the v1 core model:
+   compiled projects need source-map handling; comment syntax and
+   quality vary across jsii languages; and weak/circular/boilerplate comments require
+   heuristic rejection. A transformer also adds build integration and is language-specific.
+   This remains a compatible producer: once reliable, it can feed derived values through
+   `MetadataContext` or directly produce the same wire shape, while the deterministic API
+   remains the persistence target.
+3. **Reusing existing `Description` properties.** Many L2s expose `description` props
    that render as first-class resource properties. These are complementary, not
    sufficient: only some resource types have them, they conflate "what it does" with
    "why it exists", and they cannot carry structure (invariants, per-property
    change-safety, provenance). The vocabulary's anti-field rules direct consumers to read
    `Description` properties in place rather than duplicating them into context.
-3. **Tags.** Tags reach the deployed resources (not just the template) but are
+4. **Tags.** Tags reach the deployed resources (not just the template) but are
    key-value-flat, tightly length-limited, count-limited, and propagate to billing and
    IAM surfaces where design prose does not belong.
-4. **Cloud-assembly metadata (out-of-band) instead of template metadata.** Writing
+5. **Cloud-assembly metadata (out-of-band) instead of template metadata.** Writing
    context into `manifest.json`/`tree.json` keeps templates untouched, but the cloud
    assembly does not travel with the deployed stack - the consumers this feature targets
    (console users, agents calling `GetTemplate` on a live stack) never see it.
-5. **A new top-level template section or CloudFormation service feature.** Strictly more
+6. **A new top-level template section or CloudFormation service feature.** Strictly more
    powerful (server-side validation, dedicated retrieval APIs) and strictly slower. `Metadata` is
    the extension point CloudFormation already provides, and it requires no service
    change to adopt (appendix C).
 
 ### What are the drawbacks of this solution?
 
-* **Template size pressure.** Context counts against the 1 MB template limit. Mitigated
-  by the terse-shorthand convention, sparse `mutability` overrides, the hoist rule
-  (cross-cutting context stated once at template level), and `ref` externalization - but
-  pathological over-annotation is possible. The vocabulary spec includes a documented
-  drop order (shed `trust`/`ops`/`failureModes` first, never drop `must`) for tooling
-  that trims under pressure.
+* **Template size pressure.** Context counts against the 1 MB template limit. CDK's
+  existing synth-time warning measures the complete serialized template, including
+  context, above 80% of its conservative 1,000,000-character threshold. This RFC does not
+  add a second limit or silently discard user declarations. Terse values, sparse
+  `mutability`, hoisting, and `ref` externalization reduce pressure; authoring tools may
+  apply the documented drop order (`trust` then `ops`, `failureModes`, `gaps`, `deps`, and
+  lower-value mutability/why detail), but never drop safety-critical `must` entries or an
+  externalization `ref`.
 * **Drift risk.** Context that is not maintained alongside the resources it describes can
   mislead. Partially mitigated by co-location in reviewed CDK source and by `trust`
   provenance; not eliminable.
@@ -510,33 +545,39 @@ because emitting no context is the default and existing synthesized output is un
 
 Resource-level (`Resources.<LogicalId>.Metadata.Context`):
 
-| Field | Type | Meaning |
-| ------- | ------ | --------- |
-| `why` | string | Rationale - purpose, notable config choices, rejected alternatives. Non-binding. |
-| `must` | string[] | Hard invariants; violating any entry breaks something (data loss, outage, security, corruption, coupling). |
-| `mutable` | enum | Resource-default change-safety: `must-never-change` \| `change-with-constraints` \| `review-required` \| `free-to-tune`. |
-| `mutability` | map\<property, enum\> | Sparse per-property overrides; only properties deviating from the default or high-stakes. |
-| `trust` | object | Provenance: `src` (`authored`\|`comment`\|`commit`\|`infer`), `conf` (`high`\|`medium`\|`low`), optional `cite`, `note`. |
-| `ops` | string | What to check before modifying this resource. |
-| `gaps` | string[] | Declared unknowns - honest beats fabricated. |
-| `deps` | string[] | Cross-stack/cross-resource producer dependencies. |
-| `failureModes` | string[] | Failure/recovery paths (retries, timeouts, DLQs, circuit breakers). |
+| Field          | Type                  | Meaning                                                                                                                  |
+|----------------|-----------------------|--------------------------------------------------------------------------------------------------------------------------|
+| `why`          | string                | Rationale - purpose, notable config choices, rejected alternatives. Non-binding.                                         |
+| `must`         | string[]              | Hard invariants; violating any entry breaks something (data loss, outage, security, corruption, coupling).               |
+| `mutable`      | enum                  | Resource-default change-safety: `must-never-change` \| `change-with-constraints` \| `review-required` \| `free-to-tune`. |
+| `mutability`   | map\<property, enum\> | Sparse per-property overrides; only properties deviating from the default or high-stakes.                                |
+| `trust`        | object                | Provenance: `src` (`authored`\|`comment`\|`commit`\|`infer`), `conf` (`high`\|`medium`\|`low`), optional `cite`, `note`. |
+| `ops`          | string                | What to check before modifying this resource.                                                                            |
+| `gaps`         | string[]              | Declared unknowns - honest beats fabricated.                                                                             |
+| `deps`         | string[]              | Cross-stack/cross-resource producer dependencies.                                                                        |
+| `failureModes` | string[]              | Failure/recovery paths (retries, timeouts, DLQs, circuit breakers).                                                      |
+
+`authored` means explicitly declared through the API; it does not imply that a human was
+the producer. A producer deriving context from comments, commits, or code structure must
+select `comment`, `commit`, or `infer` and set confidence accordingly.
 
 Template-level (top-level `Metadata.Context`):
 
-| Field | Type | Meaning |
-| ------- | ------ | --------- |
-| `arch` | string | High-level shape/pattern of the system. |
-| `must` | string[] | Cross-cutting invariants stated once (DRY). |
-| `ref` | (string \| object)[] | Pointers to external shared/overflow context: `at` (URI), optional `has` (hint), `scope`. |
-| `owner` | string | Owner/contact, if not already a tag. |
+| Field   | Type                 | Meaning                                                                                   |
+|---------|----------------------|-------------------------------------------------------------------------------------------|
+| `arch`  | string               | High-level shape/pattern of the system.                                                   |
+| `must`  | string[]             | Cross-cutting invariants stated once (DRY).                                               |
+| `ref`   | (string \| object)[] | Pointers to external shared/overflow context: `at` (URI), optional `has` (hint), `scope`. |
+| `owner` | string               | Owner/contact, if not already a tag.                                                      |
 
 Conventions carried by the companion specification: free-text values use terse
 telegraphic shorthand; the hoist rule moves context repeated on more than ~3 resources up
 to template level; anti-field rules forbid restating anything the template already
 expresses (`Type`, logical IDs, property values, `Description` properties, `aws:cdk:path`);
-a tiered drop order governs trimming near the 1 MB limit (shed `trust`, `ops`,
-`failureModes`, `gaps`, `deps` first; `must` and template `ref` are never dropped).
+a tiered drop order guides authoring tools near the 1 MB limit (shed `trust`, `ops`,
+`failureModes`, `gaps`, `deps`, then low-value `mutability`/`why` detail; `must` and an
+externalization `ref` are never dropped). CDK warns on the whole serialized template but
+does not automatically apply this drop order.
 
 ### Appendix B - Benchmark and implementation evidence
 
@@ -549,21 +590,22 @@ precedence, targeting options, nested-stack cascade, mixin precedence and valida
 errors; snapshot-verified integration tests cover both the aspect and mixin paths.
 
 **Benchmark.** The motivating claim - that embedded context changes what a template
-consumer actually does - was benchmarked rather than asserted.
-Four approaches to carrying design context were compared on the same CloudFormation update
-tasks:
+consumer actually does - was benchmarked rather than assumed under these conditions:
 
 1. **No embedded context** — the control: the template states what exists, nothing more.
-2. **Context as source-file comments** — the strongest non-structured baseline.
-3. **Structured `Metadata.Context`** — the approach this RFC proposes.
+2. **Natural inline YAML comments** — the strongest unstructured raw-template baseline.
+3. **Structured `Metadata.Context`** — the approach this RFC proposes, consumed without
+   context-specific instructions.
 4. **Structured `Metadata.Context` plus context-aware tooling** — consumers that understand
    the vocabulary rather than merely reading it as text.
 
-The control performed worst, and by the widest margin on exactly the tasks whose correct
-answer depended on knowledge absent from the template. Structured context produced the
-largest single improvement over it; context-aware tooling improved on that again. Comments
-scored well, but only in the one place they exist — the source file — which is why they are
-not a viable answer for CDK or for anything reading a deployed template.
+The comparison showed that context-free templates performed worst, especially on tasks
+whose correct answer depended on absent knowledge. Comments showed that context itself
+provides most of the improvement, while the structured form made that context durable and
+machine-addressable and vocabulary-aware tooling added a further benefit. Comments alone
+remain source-local for CDK because synthesis does not preserve them by default; the
+automatic-propagation alternative would translate them into `Metadata.Context`, which
+remains the deployed carrier.
 
 **What the benchmark taught us about the design.** Three findings shaped this proposal:
 
--- a/text/0972-metadata-context.md
+++ b/text/0972-metadata-context.md
@@ -1,0 +1,608 @@
+# Structured Design Context in Synthesized Templates (Metadata.Context)
+
+* **Original Author(s):** @satyakigh
+* **Tracking Issue**: #972
+* **API Bar Raiser**: TBD
+
+CDK apps know *why* every resource exists - the rationale, invariants, and operational
+knowledge live in source comments, construct structure, and the author's head - but none
+of it survives `cdk synth`. This RFC adds a `MetadataContext` API to `aws-cdk-lib` that
+embeds structured, advisory design context into the `Metadata.Context` sections of
+synthesized CloudFormation templates, so that humans and automated tools (consoles, CLIs,
+AI agents) operating on the deployed stack later can act on the author's intent instead
+of guessing it.
+
+## Working Backwards
+
+### CHANGELOG
+
+```text
+feat(core): embed structured design context in synthesized templates (MetadataContext)
+```
+
+### README
+
+#### Metadata Context
+
+The `MetadataContext` class embeds structured, advisory context into the
+`Metadata.Context` sections of synthesized CloudFormation templates.
+It captures the *why* behind your infrastructure - rationale, hard
+invariants, change-safety, provenance and operational hints - so that humans
+and automated tools working with the deployed template later can act on the
+author's intent instead of guessing it.
+
+Add resource-level context on any construct scope. It is rendered onto the
+scope's *primary* resources (the `defaultChild` chain of each construct),
+skipping incidental helper resources like auto-created IAM policies:
+
+```ts
+declare const queue: sqs.Queue;
+
+MetadataContext.of(queue).add({
+  why: 'buffer order events async; 14d retention = compliance window',
+  must: ['VisTimeout >= 6x fn timeout, else dup on retry'],
+  mutable: ContextMutability.CHANGE_WITH_CONSTRAINTS,
+  mutability: {
+    QueueName: ContextMutability.MUST_NEVER_CHANGE,
+  },
+  ops: 'check ApproxAgeOfOldestMsg before cutting VisTimeout',
+  failureModes: ['retry 3x w/ exp backoff before DLQ'],
+});
+```
+
+This renders a `Metadata.Context` block on the `AWS::SQS::Queue` resource:
+
+```json
+{
+  "Type": "AWS::SQS::Queue",
+  "Metadata": {
+    "Context": {
+      "why": "buffer order events async; 14d retention = compliance window",
+      "must": ["VisTimeout >= 6x fn timeout, else dup on retry"],
+      "mutable": "change-with-constraints",
+      "mutability": { "QueueName": "must-never-change" },
+      "ops": "check ApproxAgeOfOldestMsg before cutting VisTimeout",
+      "failureModes": ["retry 3x w/ exp backoff before DLQ"]
+    }
+  }
+}
+```
+
+Context added on an outer scope cascades to all primary resources beneath it
+with nearest-wins semantics: scalar fields (`why`, `mutable`, `trust`, `ops`)
+from scopes closer to a resource override outer scopes, while list fields
+(`must`, `gaps`, `deps`, `failureModes`) accumulate and de-duplicate. Like
+`Tags`, context crosses stack boundaries - adding context on a scope that
+contains a `NestedStack` also stamps the primary resources inside the nested
+stack's template:
+
+```ts
+declare const stack: Stack;
+declare const queue: sqs.Queue;
+
+// Applies to every primary resource in the stack
+MetadataContext.of(stack).add({
+  must: ['all data encrypted w/ security-team CMK'],
+});
+
+// More specific context for one resource; inherits the stack-level `must`
+MetadataContext.of(queue).add({
+  why: 'buffers webhook events for async processing',
+});
+```
+
+Use the options to widen or narrow targeting:
+
+```ts
+declare const stack: Stack;
+
+// Stamp context onto every resource, including helper resources
+MetadataContext.of(stack).add({
+  deps: ['NetworkStack'],
+}, {
+  applyToAllResources: true,
+});
+
+// Only apply to specific resource types
+MetadataContext.of(stack).add({
+  ops: 'drain queue before changing',
+}, {
+  includeResourceTypes: ['AWS::SQS::Queue'],
+});
+```
+
+Record where context came from and how much to trust it with the `trust`
+field - useful when context is produced by tooling rather than authored by
+the resource owner. When omitted, `source` defaults to `AUTHORED` and
+`confidence` to `MEDIUM`:
+
+```ts
+declare const queue: sqs.Queue;
+
+MetadataContext.of(queue).add({
+  why: 'inferred from retry wrapper in api/handler.ts',
+  trust: {
+    source: ContextTrustSource.INFERRED,
+    confidence: ContextTrustConfidence.LOW,
+    citation: 'api/handler.ts:87',
+    note: 'no explicit doc found',
+  },
+});
+```
+
+Context can also be applied as a Mixin. `MetadataContextMixin` attaches a
+context block imperatively to exactly the constructs you target - via
+`.with()` on a single L1 resource, or in bulk via `Mixins.of()`. Context
+applied by the Mixin takes precedence over context cascaded from enclosing
+scopes (scalar fields win; list fields are unioned):
+
+```ts
+declare const stack: Stack;
+declare const cfnResource: CfnResource;
+
+// Single resource via .with()
+cfnResource.with(new MetadataContextMixin({
+  why: 'append-only audit trail buffer',
+  mutable: ContextMutability.MUST_NEVER_CHANGE,
+  must: ['never shorten retention below 14d (audit requirement)'],
+}));
+
+// Bulk application to every CloudFormation resource in a scope
+Mixins.of(stack).apply(new MetadataContextMixin({
+  deps: ['NetworkStack'],
+}));
+```
+
+Template-level context holds cross-cutting facts stated once per stack: the
+architecture overview, template-wide invariants, pointers to external shared
+context, and ownership. The stack's purpose itself belongs in the native
+CloudFormation `Description` (the `description` prop of `Stack`):
+
+```ts
+declare const stack: Stack;
+
+MetadataContext.of(stack).addToTemplate({
+  arch: 'SQS buffer -> Lambda -> DynamoDB; DLQ for poison msgs',
+  must: ['all data encrypted w/ security-team CMK'],
+  refs: [
+    {
+      at: 's3://org-iac-ctx/shared/encryption.ctx.yaml',
+      has: 'org CMK + tagging rules',
+      scope: 'shared',
+    },
+  ],
+  owner: 'order-processing@example.com',
+});
+```
+
+Keep free-text values terse - drop articles and use symbols (`->`, `>=`,
+`w/`) - since context competes with resources for the CloudFormation 1 MB
+template size limit. Prefer `must` for binding rules whose violation breaks
+something, and `why` for reasoning and rejected alternatives.
+
+---
+
+Ticking the box below indicates that the public API of this RFC has been
+signed-off by the API bar raiser (the `status/api-approved` label was applied to the
+RFC pull request):
+
+```text
+[ ] Signed-off by API Bar Raiser @xxxxx
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+A new capability in `aws-cdk-lib` core: the `MetadataContext` class and the
+`MetadataContextMixin`, which embed structured, machine-readable design context into the
+`Metadata` sections of synthesized CloudFormation templates.
+
+* `MetadataContext.of(scope).add(props, options?)` - declares resource-level context
+  (rationale, hard invariants, change-safety, provenance, operational hints, known gaps,
+  dependencies, failure modes) on any construct scope. At synthesis time the context is
+  rendered as a `Metadata.Context` block on the scope's primary resources, cascading like
+  `Tags` with nearest-wins merge semantics.
+* `MetadataContext.of(scope).addToTemplate(props)` - declares template-level context
+  (architecture overview, cross-cutting invariants, external context references,
+  ownership) rendered once as a top-level `Metadata.Context` block.
+* `MetadataContextMixin` - the same resource-level context applied imperatively to
+  exactly the constructs you target, via `.with()` or `Mixins.of(scope).apply()`.
+
+The emitted wire format follows the `Metadata.Context` v1 vocabulary - a small, closed
+set of fields (`why`, `must`, `mutable`, `mutability`, `trust`, `ops`, `gaps`, `deps`,
+`failureModes` at resource level; `arch`, `must`, `ref`, `owner` at template level)
+designed to be terse, advisory, and consumable by both humans and automated tooling. The
+context is plain CloudFormation `Metadata`: it deploys with the stack, has no runtime
+effect, and is retrievable with the standard `GetTemplate` and `DescribeStackResource`
+APIs - no new service support required.
+
+### Why should I use this feature?
+
+Because the synthesized template is the only artifact that reliably reaches everyone who
+touches your infrastructure after you. The engineer (or, increasingly, the AI agent) who
+modifies your stack six months from now often has the template and the live stack - not
+your source repository, your design doc, or you.
+
+Concrete situations this feature addresses:
+
+* **Safe change review** - `must` and `mutability` tell a reviewer (human or automated)
+  which properties are load-bearing before they approve a change. Example: an operator
+  about to lower `VisibilityTimeout` sees
+  `must: ["VisTimeout >= 6x fn timeout, else dup on retry"]` in the template itself.
+* **Operational handoff and incident response** - `why`, `ops` and `failureModes` carry
+  the on-call knowledge that normally lives in tribal memory: what the resource is for,
+  what to check before touching it, and what the failure/recovery paths are.
+* **AI-assisted infrastructure operations** - agents that read templates via
+  `GetTemplate`/`DescribeStackResource` act on documented intent instead of inferring it.
+  An agent asked to "raise the Lambda timeout to 10s" cannot know that a documented SLA
+  caps it lower, or that a queue's visibility timeout is coupled to it - unless the
+  template says so. `must` and `ops` are what turn a locally-valid edit into a correct
+  one.
+* **Organizational context at scale** - template-level `ref` entries point to shared
+  context files (e.g. org-wide encryption rules) without repeating them in every
+  template, and `trust` distinguishes human-authored context from tool-inferred context
+  so consumers can weight it appropriately.
+
+These are not just expectations. We benchmarked the alternatives against each other - no
+embedded context, context as source comments, and structured `Metadata.Context` - on the
+same CloudFormation update tasks. Structured context consistently produced the best
+outcomes, and it is the only one of those approaches that survives synthesis and is
+retrievable from a deployed stack.
+
+If you already maintain design context in READMEs or wikis, this feature does not replace
+them - it puts the *operationally binding* subset where every consumer of the deployed
+stack can actually find it.
+
+## Internal FAQ
+
+### Why are we doing this?
+
+**The deployment artifact is where design context dies today.** CDK's programming model
+concentrates rich intent at authoring time: construct hierarchy, source comments, L2/L3
+prop choices, code review discussion. Synthesis flattens all of it into L1 resources.
+What survives is `aws:cdk:path` (structural, not semantic) and whatever `Description`
+properties happen to exist. Every downstream consumer of the template - CloudFormation
+console users, change-set reviewers, incident responders, drift investigators, and now AI
+agents - works from an artifact that says *what* is deployed but never *why*.
+
+**Agentic tooling makes the gap acute.** AI agents are increasingly asked to modify
+deployed infrastructure. They retrieve templates through `GetTemplate` and
+`DescribeStackResource` and make changes that are locally valid but globally wrong: they
+cannot see cross-resource coupling rules, compliance retention windows, or team SLAs that
+never made it into the template. The failure is not that the agent is careless - it is
+that the artifact it reads does not contain the constraint it needs to respect.
+
+**We benchmarked that claim rather than assuming it.** Before settling the API we compared
+four ways of carrying design context on the same CloudFormation update tasks: no embedded
+context, context as source-file comments, structured `Metadata.Context`, and structured
+context read by tooling that understands the vocabulary. Context-free templates fared worst
+on the tasks whose correct answer depended on knowledge the template did not contain;
+structured context produced the largest improvement; context-aware tooling improved on that
+again. Appendix B has the comparison.
+
+The benchmark also settled the obvious objection, which is that this is what comments are
+for. Comments did score well - where they exist. But they are unavailable to CDK users:
+synthesized JSON carries no comments, and no comment in CDK source survives to the
+template. They are equally unavailable to *any* consumer retrieving a deployed template
+through `GetTemplate`, regardless of how it was authored. Structured metadata is the only
+form of this that survives synthesis and deployment, which is what makes it the right
+target for CDK.
+
+**CDK is uniquely positioned to populate it.** CDK users do not author the template — they
+author constructs, and the template is generated. So the authoring surface has to exist in
+CDK, and the construct tree makes it a better place to put one: a single `add()` call on a
+scope covers an entire subtree of primary resources, and `defaultChild` chains give precise
+targeting so rationale lands on the resources users actually declared, not on synthesized
+plumbing. CDK is where most production templates come from, so it is where this belongs.
+
+**Precedent exists in CDK itself.** CDK already injects metadata into every resource
+(`aws:cdk:path`, analytics metadata) because structural information was judged valuable
+enough to embed by default. This RFC extends the same channel to *semantic* information,
+opt-in, under user control.
+
+### Why should we _not_ do this?
+
+* **Stale context is worse than no context.** Rationale written once and never updated
+  actively misleads the consumers it was meant to help. Co-location in reviewed CDK source
+  and the `trust`/`gaps` fields reduce the risk but do not remove it.
+* **This is new public API surface in core.** A class, three enums and five interfaces in
+  `aws-cdk-lib` are inherited by every jsii language binding, and the field vocabulary
+  becomes a contract downstream tooling pins to. An external construct library could
+  deliver the same behavior with no core commitment (see alternatives).
+* **Context consumes the template size budget.** Context competes with resources for
+  CloudFormation's template size limit. Terse conventions, sparse `mutability` overrides
+  and `ref` externalization mitigate this, but large stacks must budget consciously.
+
+### What is the technical solution (design) of this feature?
+
+The design below is implemented as written — see
+[aws/aws-cdk#38381](https://github.com/aws/aws-cdk/pull/38381) for the code it describes.
+The field vocabulary and the authoring model were settled first, by evaluating context
+embedded in templates directly (see appendix B); this section covers how CDK produces it.
+
+#### Wire format: a small, advisory vocabulary
+
+The emitted shape follows a small, versioned vocabulary (`Metadata.Context` v1), defined by
+this RFC: the field reference in appendix A and the CDK types below are its normative
+description. It is *advisory* — nothing enforces it. CloudFormation ignores it, and any
+consumer that chooses to read or validate it does so on its own. CDK
+renders the wire format (`trust.src`, `trust.conf`, `trust.cite`, bare-string `ref` entries
+when only a URI is present) from idiomatic, fully-spelled TypeScript property names
+(`trust.source`, `trust.confidence`, `trust.citation`).
+
+Resource-level fields: `why` (rationale), `must` (hard invariants), `mutable`
+(resource-default change-safety), `mutability` (sparse per-property override map),
+`trust` (provenance: source/confidence/citation/note), `ops` (pre-change operational
+hint), `gaps` (declared unknowns), `deps` (cross-stack/resource dependencies),
+`failureModes` (failure/recovery paths).
+
+Template-level fields: `arch` (system shape), `must` (cross-cutting invariants), `ref`
+(pointers to external shared/overflow context), `owner` (contact).
+
+Change-safety uses a closed four-level enum: `must-never-change`,
+`change-with-constraints`, `review-required`, `free-to-tune`. In the CDK API this is the
+`ContextMutability` enum; the schema's single-value-or-map union was deliberately split
+into two props (`mutable` for the resource default, `mutability` for the per-property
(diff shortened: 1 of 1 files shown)


text/0972-metadata-context.md
otaviomacedo, line 1: First of all, this is awesome!

I have a comment more about the schema itself, than the proposed CDK API. Suppose I have a CFN template that encodes this graph:

```mermaid
flowchart LR
  m[m: EventSourceMapping] --> q[queue: Queue]
  m --> f[function: Function]
  f --> b[bucket: Bucket]
```

I planning to build some tooling to derive "views" out of a template. The first thing a tool like that could do is infer data flows. With some clever annotations to the CFN schema (we can talk about this separately), it could infer, for example, that data flows from `queue` to `bucket`.

The next thing it could do is check the security of this data flow. Let's suppose, for simplicity, that there are three data classification categories, in order of sensitivity: public, internal and secret. The rule is: data can only flow from less to more sensitive resources. So if I classify `queue` as secret, and `bucket` as public, for example, it should flag that as a violation.

Now, I could create my own schema to encode these things. It could use a field `dataClassification` in the metadata, for example. But, if I understand it correctly, this is an opportunity to create a sort of canonical schema, to allow interoperability between tools. I believe the best place for what I'm describing here is the `must` field. But it is free form text, so it's there is no guaranantee that different tools can communicate using it.

Are you taking into account this use case at all? Even if not, it would be nice to be able to declare extension points. Something like:

```json
{
    "extension:x.y.z.SecureDataFlow": {
        "dataClassification": "secret"
    }
}
```

otaviomacedo: Forgot to mention: you're already proposing a special case of this. The `mutable` enum is a well-defined poset:

    free-to-tune ≤ review-required ≤ change-with-constraints ≤ must-never-change

I guess what I'm asking is: make this a generic mechanism.

satyakigh: Updated the context namespace to be under `com.aws.cloudformation.Context` instead

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+The `MetadataContext` class embeds structured, advisory context into the
+`Metadata["com.aws.cloudformation.Context"]` sections of synthesized CloudFormation templates.
+It captures the *why* behind your infrastructure - rationale, hard
+invariants, change-safety, provenance and operational hints - so that humans
+and automated tools working with the deployed template later can act on the
+author's intent instead of guessing it.
+
+Add resource-level context on any construct scope. It is rendered onto the
+scope's *primary* resources (the `defaultChild` chain of each construct),
+skipping incidental helper resources like auto-created IAM policies:
```
iliapolo: What if the customer does want some context on these incidental resources? For example, a `lambda.Function` may implicitly create a dead letter queue. For completes, lets add an example of that:

```ts
MetadataContext.of(func).add(...); // primary resource context

MetadataContext.of(func.deadLetterQueue).add(...) // helper resources context
```

This then begs the question of can we apply some default context. For example:

```json
{
  "Type": "AWS::SQS::Queue",
  "Metadata:" : {
    "com.aws.cloudformation.Context": {
      "why": "Created by function MyFun to store failed invocation payloads",
      "must": ["Size === 0"]
    }
  }
}
```

Now the agent knows to see it as part of the same functionality vs just a floating resource. It would require a lot of code changes to `aws-cdk-lib` - if you agree it would be worth while, lets add it as a possible follow up / future enhancement.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+      },
+      "ops": "check ApproxAgeOfOldestMsg before cutting VisTimeout",
+      "failureModes": [
+        "retry 3x w/ exp backoff before DLQ"
+      ]
+    }
+  }
+}
+```
+
```
iliapolo, line 87: As a customer reading up to here I immediately wonder what keys are available and what they mean. Put a link to the schema appendix right here for quicker access (when we publish, the link should point to a proper page on the dev guide).

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+  the `MetadataContext.of()` call is intentionally left out of the initial release to keep
+  the surface minimal while the schema settles.
+
+## Appendix
+
+### Appendix A - CloudFormation Context advisory schema field reference
+
+Resource-level (`Resources.<LogicalId>.Metadata["com.aws.cloudformation.Context"]`):
+
+| Field          | Type                  | Meaning                                                                                                                  |
```
iliapolo: Add an `Example` column please (real life).

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+|----------------|-----------------------|--------------------------------------------------------------------------------------------------------------------------|
+| `why`          | string                | Rationale - purpose, notable config choices, rejected alternatives. Non-binding.                                         |
+| `must`         | string[]              | Hard invariants; violating any entry breaks something (data loss, outage, security, corruption, coupling).               |
+| `mutable`      | enum                  | Resource-default change-safety: `must-never-change` \| `change-with-constraints` \| `review-required` \| `free-to-tune`. |
+| `mutability`   | map\<property, enum\> | Sparse per-property overrides; only properties deviating from the default or high-stakes.                                |
+| `trust`        | object                | Provenance: `src` (`authored`\|`comment`\|`commit`\|`infer`), `conf` (`high`\|`medium`\|`low`), optional `cite`, `note`. |
```
iliapolo: Elaborate on the possible values of `src`, `conf`, `cite`, `note` (with examples).

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+| `ops`          | string                | What to check before modifying this resource.                                                                            |
+| `gaps`         | string[]              | Declared unknowns - honest beats fabricated.                                                                             |
+| `deps`         | string[]              | Cross-stack/cross-resource producer dependencies.                                                                        |
+| `failureModes` | string[]              | Failure/recovery paths (retries, timeouts, DLQs, circuit breakers).                                                      |
+
+`authored` means explicitly declared through the API; it does not imply that a human was
+the producer. A producer deriving context from comments, commits, or code structure must
```
iliapolo: > it does not imply that a human was the producer

But if its an agent producing this context - isn't it always inferred from something? Give a concrete example of when each value here would be used?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+(`must`, `gaps`, `deps`, `failureModes`) accumulate and de-duplicate. Like
+`Tags`, context crosses stack boundaries - adding context on a scope that
+contains a `NestedStack` also stamps the primary resources inside the nested
+stack's template:
+
+```ts
+declare const stack: Stack;
+declare const queue: sqs.Queue;
+
+// Applies to every primary resource in the stack
```
iliapolo: I can understand what a primary resource for `sqs.Queue` is - but what are primary resources for the stack? what gets filtered out? 

I think we need a good definition of primary resources before presenting this example.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+      "ops": "check ApproxAgeOfOldestMsg before cutting VisTimeout",
+      "failureModes": [
+        "retry 3x w/ exp backoff before DLQ"
+      ]
+    }
+  }
+}
+```
+
+Context added on an outer scope cascades to all primary resources beneath it
```
iliapolo: Why? What is the alternative? 

I think this is natural when dealing with L2s because they are intentionally authored to have a "primary resource". Does this mechanism still hold for L3s? For example [ECRDeployment](https://constructs.dev/packages/cdk-ecr-deployment/v/4.2.57?lang=typescript)? I don't have an answer but I think we need to go through this exercise because this is a one way door (we cannot take context away). 

If L3s are not really applicable then we should ship with only L1/L2/Stack support for now.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+with nearest-wins semantics: scalar fields (`why`, `mutable`, `trust`, `ops`)
+from scopes closer to a resource override outer scopes, while list fields
+(`must`, `gaps`, `deps`, `failureModes`) accumulate and de-duplicate. Like
```
dgandhi62: Is this intended to propagate down the construct tree? For instance, in an L3 construct, would list fields accumulate on every resource within the construct?

A further question is - could this lead to biasing the user's AI agent? If they see a single string in a field repeated for each resource?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+```ts
+declare const stack: Stack;
+declare const queue: sqs.Queue;
+
+// Applies to every primary resource in the stack
+MetadataContext.of(stack).add({
+  must: ['all data encrypted w/ security-team CMK'],
+});
+
+// More specific context for one resource; inherits the stack-level `must`
```
iliapolo: > inherits the stack-level `must`

This fits nicely when adding specific context to a `Queue` resource, because it holds data. But what I do:

```ts
MetadataContext.of(role).add({
  why: 'Allows lambda processor to read from the queue',
});
```

Does it make sense for the "all data encrypted w/ security-team CMK" context to appear on an `AWS::IAM::Role` resource? is there an API to prevent the inheritance?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+| `trust`        | object                | Provenance: `src` (`authored`\|`comment`\|`commit`\|`infer`), `conf` (`high`\|`medium`\|`low`), optional `cite`, `note`. |
+| `ops`          | string                | What to check before modifying this resource.                                                                            |
+| `gaps`         | string[]              | Declared unknowns - honest beats fabricated.                                                                             |
+| `deps`         | string[]              | Cross-stack/cross-resource producer dependencies.                                                                        |
+| `failureModes` | string[]              | Failure/recovery paths (retries, timeouts, DLQs, circuit breakers).                                                      |
+
+`authored` means explicitly declared through the API; it does not imply that a human was
+the producer. A producer deriving context from comments, commits, or code structure must
+select `comment`, `commit`, or `infer` and set confidence accordingly.
```
iliapolo: `comment` and `commit` are also inferred, this creates confusion as to what exactly should be used. I think a better approach is to only have `infer`, and put the source of inference as a `note` (it can be many things at once).

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+`MEDIUM`. `AUTHORED` means the context was explicitly declared through this API; it
+does not assert that a human wrote it. Producers that infer context from comments,
+commits, or other artifacts override `trust` with the corresponding source and an
+explicit confidence:
+
+```ts
+declare const queue: sqs.Queue;
+
+MetadataContext.of(queue).add({
+  why: 'inferred from retry wrapper in api/handler.ts',
```
iliapolo: but, what is inferred? this doesn't read as a `why`

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+MetadataContext.of(stack).add({
+  ops: 'drain queue before changing',
+}, {
+  includeResourceTypes: ['AWS::SQS::Queue'],
+});
+```
+
+Every resource context block records where it came from and how much to trust it.
+When the caller omits `trust`, CDK emits `source: AUTHORED` and chooses confidence
+from the declaration: `HIGH` when `why` or a non-empty `must` is present, otherwise
```
dgandhi62: Could you give more clarity on _when_ a user needs to be using a specific field. For instance, guidance in the README like

```
A human's preferred option should be .... On the other hand, an agent should be using .... by default.
```

One suggestion could be keeping AUTHORED primarily for humans, and INFERRED for agents to make it simpler.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+// Only apply to specific resource types
+MetadataContext.of(stack).add({
+  ops: 'drain queue before changing',
+}, {
+  includeResourceTypes: ['AWS::SQS::Queue'],
+});
+```
+
+Every resource context block records where it came from and how much to trust it.
+When the caller omits `trust`, CDK emits `source: AUTHORED` and chooses confidence
```
iliapolo: I assume that `AUTHORED` carries more trust than `INFERRED`? What prevents agents inferring context from using `AUTHORED`? or just not providing `trust` at all and then having CDK default to `AUTHORED`?

I think there needs to be a clearer definition and guidance. What is the main use-case for using `AUTHORED`? is it not human authored? if so - lets double down on that and provide agents with instructions to always use `INFERRED` when setting context.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+  the surface minimal while the schema settles.
+
+## Appendix
+
+### Appendix A - CloudFormation Context advisory schema field reference
+
+Resource-level (`Resources.<LogicalId>.Metadata["com.aws.cloudformation.Context"]`):
+
+| Field          | Type                  | Meaning                                                                                                                  |
+|----------------|-----------------------|--------------------------------------------------------------------------------------------------------------------------|
```
iliapolo: Is everything optional here? add an `Optionality` column

text/0972-metadata-context.md
iliapolo: So both `why` and `must` are optional? I don't think we should allow for context that doesn't specify `why`. Why are we tying confidence solely to the `must` field? I'd think a `citation` is more evidence for context than some invariants the agent may have inferred incorrectly. My proposal:

- **HIGH**: Both `must` and `citation` are present.
- **MEDIUM**: Only one is present.
- **LOW**: None are present.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+Template-level context holds cross-cutting facts stated once per stack: the
+architecture overview, template-wide invariants, pointers to external shared
+context, and ownership. The stack's purpose itself belongs in the native
+CloudFormation `Description` (the `description` prop of `Stack`):
+
+```ts
+declare const stack: Stack;
+
+MetadataContext.of(stack).addToTemplate({
```
iliapolo: `addToTemplate` is an API thats going to be used far less than `add` - so lets not put them under the same object - to reduce cognitive load. Proposal:
 
- `TemplateMetadataContext.of(stack).add() // accets Stack`
- `ResourceMetadataContext.of(stack).add() // accepts Construct`

This keeps things consistent. If construct vendors want to add template metadata (though they really shouldn't), they can use:

```ts
TemplateMetadataContext.of(Stack.of(construct)).add(...)
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+`w/`) - since context competes with resources for the CloudFormation 1 MB
+template size limit. Prefer `must` for binding rules whose violation breaks
+something, and `why` for reasoning and rejected alternatives.
+
+CDK already serializes the complete template during synthesis and emits a
+warning above 80% of its conservative 1,000,000-character threshold;
+`Metadata["com.aws.cloudformation.Context"]` is included in that measurement. This RFC adds no
+context-specific size validator and never silently trims declared context. The
+existing warning remains the synth-time signal; authoring tools may respond
+using the advisory schema's tier/drop order described in appendix A.
```
iliapolo: How would agents know to consult this schema if `cdk synth` doesn't provide this guidance? we can at least print a link to the schema for the to follow.

iliapolo (commented):
@satyakigh Are you also looking for feedback on the schema itself, or just the CDK part of things?

text/0972-metadata-context.md
satyakigh: Mainly the CDK side of things, we can update the schema if required as well if you guys have suggestions

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+Concrete situations this feature addresses:
+
+* **Safe change review** - `must` and `mutability` tell a reviewer (human or automated)
+  which properties are load-bearing before they approve a change. Example: an operator
+  about to lower `VisibilityTimeout` sees
+  `must: ["VisTimeout >= 6x fn timeout, else dup on retry"]` in the template itself.
+* **Operational handoff and incident response** - `why`, `ops` and `failureModes` carry
+  the on-call knowledge that normally lives in tribal memory: what the resource is for,
+  what to check before touching it, and what the failure/recovery paths are.
```
iliapolo: What do you mean by `failure/recovery paths`? For example if failure mode is `retry 3x w/ exp backoff before DLQ` - what is the agent expected to do? look at the DLQ?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+APIs - no new service support required.
+
+### Why should I use this feature?
+
+Because the synthesized template is the only artifact that reliably reaches everyone who
+touches your infrastructure after you. The engineer (or, increasingly, the AI agent) who
+modifies your stack six months from now often has the template and the live stack - not
+your source repository, your design doc, or you.
+
+Concrete situations this feature addresses:
```
iliapolo, line 416: This answers why the context is needed, not why it should be provided as a concrete API in the CDK. 

I'd expect also something like "CDK customers will have a hard time defining the correct context without this feature" (because escape hatches suck). 

Also needs to touch on the cascading mechanism you're proposing.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+Brownfield adoption does not require manually seeding every resource. A companion
+bootstrapping skill is being developed to read existing CDK/CloudFormation source,
+comments, git history, tests, and companion service code, then propose explicit
+`MetadataContext` declarations (or `Metadata["com.aws.cloudformation.Context"]` blocks for raw templates) with
+provenance and declared gaps. This subsidizes discovery and authoring effort, while
+keeping heuristic inference and its review outside the core API contract.
+
+If you already maintain design context in READMEs or wikis, this feature does not replace
+them - it puts the *operationally relevant* subset where every consumer of the deployed
```
iliapolo, line 452: But this also creates duplication. I'm wondering why not use `ref` as a way to also point to the artifacts in the repo (design doc/runbook). Why would `ref` be suitable for org-wide rules but not for the repository URL for example?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+**Precedent exists in CDK itself.** CDK already injects metadata into every resource
+(`aws:cdk:path`, analytics metadata) because structural information was judged valuable
+enough to embed by default. This RFC extends the same channel to *semantic* information,
+opt-in, under user control.
+
+### Why should we _not_ do this?
+
+* **Stale context is worse than no context.** Rationale written once and never updated
+  actively misleads the consumers it was meant to help. Co-location in reviewed CDK source
+  and the `trust`/`gaps` fields reduce the risk but do not remove it.
```
iliapolo: How do `trust / gaps` reduce the risk?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+Template-level fields: `arch` (system shape), `must` (cross-cutting invariants), `ref`
+(pointers to external shared/overflow context), `owner` (contact).
+
+Every emitted resource block also contains `trust`. When callers omit it, CDK emits
+`src: authored` and defaults `conf` to `medium`. CDK promotes it to `high` only when
+the final merged block contains a non-blank `why` or at least one non-blank string in
+`must`. Explicit trust values always win.
+
+Change-safety uses a closed four-level enum: `must-never-change`,
+`change-with-constraints`, `review-required`, `free-to-tune`. In the CDK API this is the
```
iliapolo: What does `change-with-constraints` mean exactly? the other ones are pretty sefl explanatory.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+`ContextMutability` enum; the schema's single-value-or-map union was deliberately split
+into two props (`mutable` for the resource default, `mutability` for the per-property
```
iliapolo: `mutable` and `mutability` are too alike - they don't convey the scope (resource / property). Prefer being more explicit:

- `resourceMutation`
- `propertyMutations`

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+embedded in templates directly (see appendix B); this section covers how CDK produces it.
+
+#### Wire format: a namespaced advisory schema
+
+The emitted shape follows the advisory Context schema under the
+`com.aws.cloudformation.Context` key in `Metadata`. The field reference in appendix A and
+the CDK types below are its normative description. It is *advisory* — nothing enforces
+it. CloudFormation ignores it, and any consumer that chooses to read or validate it does
+so on its own. CDK renders the schema (`trust.src`, `trust.conf`, `trust.cite`, and
+`ref` from the TypeScript `refs` property; a `ref` entry becomes a bare string when only
```
iliapolo: What about the other schema props? are they not defined in TypeScript as well? I don't understand why `trust` and `ref` get special treatment in this paragraph.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+At visit time the aspect walks each `CfnResource`'s ancestor scopes root → leaf,
+collecting staged entries whose targeting options match, and merges them so that entries
+closer to the resource win. Because merge order derives from the construct tree rather
+than from aspect registration order, the semantics are deterministic regardless of how
+many scopes declared context or in what order `add()` was called - the same reasoning
+that led Tags to a single-visitor design. Finally the renderer writes the merged block to
+`com.aws.cloudformation.Context`.
+
+**Precedence:** A manually supplied value at that key passes through unchanged unless the
```
iliapolo: Not sure I follow - can you give an example? Is this for the case that the customer manually defined a metadata entry that conflicts with our keys? 

In that case I would expect an error, we shouldn't be dropping explicit keys the customer added.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+Merge semantics, field by field:
+
+* Scalars (`why`, `mutable`, `trust`, `ops`): nearest scope wins.
+* Lists (`must`, `gaps`, `deps`, `failureModes`): accumulate across scopes, de-duplicated.
+* `mutability` map: per-key merge; nearest scope wins per property name.
+
+#### Primary-resource targeting
+
+By default, context lands only on *primary* resources: a resource is primary relative to
```
iliapolo: I don't understand the definition - it uses the `term` designates in two different meanings I think. Can you give some examples to help clarify?

For example, our `lambda.Function` construct internally creates:
1. A `CfnFunction` L1 resource (`defaultChild`)
2. An `iam.Role` L2 resource
3. An `CfnRole` L1 resource (`defaultChild` of 2)

What makes (1) primary but (3) secondary? Also give an example for an L3.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+### What alternative solutions did you consider?
+
+1. **A standalone Aspect/construct library (no core changes).** The behavior is
+   achievable with generic metadata APIs. Rejected as the end state because the
+   interoperable contract benefits from a validated, discoverable core API with jsii
+   bindings, consistent `Mixins` behavior, `AspectPriority` defaults, and future L2
+   integration points.
+2. **Automatic propagation of leading source comments.** A prototype Aspect follows a
+   resource's creation stack to its source location, reads the leading comment, applies an
+   anti-fabrication gate, and writes an attributed `Metadata["com.aws.cloudformation.Context"].why` during synthesis.
```
iliapolo: `anti-fabrication`? ha?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+   compiled projects need source-map handling; comment syntax and
+   quality vary across jsii languages; and weak/circular/boilerplate comments require
+   heuristic rejection. A transformer also adds build integration and is language-specific.
+   This remains a compatible producer: once reliable, it can feed derived values through
+   `MetadataContext`, while the deterministic API remains the persistence target.
+3. **Reusing existing `Description` properties.** Many L2s expose `description` props
+   that render as first-class resource properties. These are complementary, not
+   sufficient: only some resource types have them, they conflate "what it does" with
+   "why it exists", and they cannot carry structure (invariants, per-property
+   change-safety, provenance). The vocabulary's anti-field rules direct consumers to read
```
iliapolo: What are `anti-field` rules?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+   transformer could avoid runtime stack inspection. It is deferred from the initial core model:
+   compiled projects need source-map handling; comment syntax and
+   quality vary across jsii languages; and weak/circular/boilerplate comments require
+   heuristic rejection. A transformer also adds build integration and is language-specific.
+   This remains a compatible producer: once reliable, it can feed derived values through
+   `MetadataContext`, while the deterministic API remains the persistence target.
+3. **Reusing existing `Description` properties.** Many L2s expose `description` props
+   that render as first-class resource properties. These are complementary, not
+   sufficient: only some resource types have them, they conflate "what it does" with
+   "why it exists", and they cannot carry structure (invariants, per-property
```
iliapolo: > they cannot carry structure 

They can if you serialize a JSON in it. It probably has length limitations that would rule that out anyway - apart from being fugly.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+* The feature emits no additional Context unless `MetadataContext.of(...).add(...)`,
+  `addToTemplate(...)`, or the mixin is called. Apps that do not use those APIs retain
+  byte-identical synthesized output, including manually supplied Context metadata.
+* `com.aws.cloudformation.Context` is an Amazon-owned advisory metadata namespace; the
+  precedence rule defined above applies only when the new API is adopted.
+* Sibling metadata namespaces remain untouched, and Context has no deployment-behavior
+  effect because CloudFormation ignores advisory metadata.
+
+### What alternative solutions did you consider?
```
iliapolo, line 722: What about having the context just be a bunch `ref`s in the that point to source code, design docs, runbooks, etc for agents to follow? This prevents the risk of drift.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+  mislead. Partially mitigated by co-location in reviewed CDK source and by `trust`
+  provenance; not eliminable.
+* **No server-side contract.** CloudFormation will not validate the blocks; garbage in,
+  garbage out. The enforcement ceiling is client-side: the synth-time checks in this
+  proposal, plus whatever validation a consumer chooses to apply.
+* **Fabrication by tooling.** As AI tools begin *writing* CDK code, they may generate
+  confident-sounding context. The `trust` field exists precisely so generated context can
+  self-identify (`src: infer`, low confidence, citation) — but the API cannot force
+  honesty, and a caller is free to claim `authored`.
+* **Merge-semantics complexity.** Nearest-wins plus accumulate-and-dedupe is more to
```
iliapolo: Its not just about learning curve - its about wether the customer actually wants the cascading effect. I can see scenarios where context is being added to an L3 with a very specific construct in mind, but then having in travel to a bunch of other resources as well might create unexpected results. 

I'd say there is a risk of out-of-place context being added. This won't happen with pure CFN because the customer has no choice but to explicitly add the right context to each resource they want. 

To evaluate this risk I think we need more examples of L3s

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+
+The RFC is published alongside the implementation so maintainers can validate the direction
+against working code. The API arrives in one increment: `MetadataContext` (facade, staged
+metadata, rendering aspect, primary-resource targeting, template-level merge) together with
+`MetadataContextMixin`, plus validation, unit and integration tests, and the `aws-cdk-lib`
+README section. The code is available for review at
+[aws/aws-cdk#38381](https://github.com/aws/aws-cdk/pull/38381).
+
+The feature ships under the standard core review bar: it is small, opt-in, and has no
+feature-flag interaction. Nothing about it needs to bake behind an experimental gate:
```
iliapolo: Why not have a bake period though? it could help hone the corresponding skill so we don't have to break the schema often.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+the feature emits nothing until its APIs are adopted, so existing output remains unchanged.
+
+### Are there any open issues that need to be addressed later?
```
iliapolo, line 789: These are future enhancements really more than open issues. Is there anything we still need to discuss to move along with the implementation - is the spirit of this question.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,668 @@
...
+Conventions carried by the companion specification: free-text values use terse
+telegraphic shorthand; the hoist rule moves context repeated on more than ~3 resources up
+to template level; anti-field rules forbid restating anything the template already
+expresses (`Type`, logical IDs, property values, `Description` properties, `aws:cdk:path`);
+a tiered drop order guides authoring tools near the 1 MB limit (shed optional trust detail, `ops`,
+`failureModes`, `gaps`, `deps`, then low-value `mutability`/`why` detail; `must` and an
+externalization `ref` are never dropped). CDK warns on the whole serialized template but
+does not automatically apply this drop order.
+
+### Appendix B - Benchmark and implementation evidence
```
iliapolo: Can you share the actual data as well?

iliapolo (changes requested):

text/0972-metadata-context.md
satyakigh: Added examples for targeting an exposed helper directly, including a function's dead-letter queue. `applyToAllResources` supports intentionally including every helper. Selecting or populating helpers automatically by relationship remains future work.

text/0972-metadata-context.md
satyakigh: Added an early link to Appendix A and the published [CloudFormation Metadata Context schema](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema). Appendix A and the CDK API docs mirror that schema; the [CloudFormation authoring skill](https://github.com/aws/agent-toolkit-for-aws/blob/main/skills/core-skills/aws-cloudformation/SKILL.md) provides non-enforced guidance.

text/0972-metadata-context.md
satyakigh: Added a dedicated trust-field table covering `src`, `conf`, `cite`, and `note`, including requiredness, allowed values, meanings, and examples.

text/0972-metadata-context.md
satyakigh: Clarified the source meanings with examples: `authored` is written or confirmed by a person, `comment` comes directly from a source comment, `commit` comes from version-control history, and `infer` is concluded from code or behavior.

text/0972-metadata-context.md
satyakigh: Defined primary resource before using it. A `Stack` has no default resource target: use `TemplateMetadataContext` for stack-wide information or request descendant selection explicitly. The RFC now requires a zero-target declaration to fail rather than silently do nothing.

text/0972-metadata-context.md
satyakigh: Changed the design to narrow default targeting: only the selected `CfnResource` or the construct's `defaultChild` resource is selected. Multi-resource constructs require explicit `applyToDescendants` and can use resource-type filters; selecting nothing is an error.

text/0972-metadata-context.md
satyakigh: Applying information to descendants is now explicit rather than automatic. The RFC also explains that repetition can make one fact appear disproportionately important and directs template-wide information to `TemplateMetadataContext`.

text/0972-metadata-context.md
satyakigh: Added `inheritAncestorContext: false`, which removes information inherited from ancestor constructs while preserving declarations made on the resource itself. Resource-type filters provide another way to prevent broad declarations from reaching unrelated resources.

text/0972-metadata-context.md
satyakigh: I kept the four source values because they distinguish direct evidence from a conclusion. `comment` and `commit` identify the evidence type; `infer` means the producer concluded the information from code or combined evidence. Combined evidence can use `infer` with `cite` and `note`.

text/0972-metadata-context.md
satyakigh: Fixed the example. `why` now contains the design reasoning, while the source location and uncertainty are recorded in `trust.citation` and `trust.note`.

text/0972-metadata-context.md
satyakigh: Added explicit guidance: people use `AUTHORED` for information they write or confirm. Automated producers use `COMMENT`, `COMMIT`, or `INFERRED` according to their evidence and must supply confidence explicitly.

text/0972-metadata-context.md
satyakigh: Removed the automatic `AUTHORED` default. `trust` is optional and may be used alone; when present, both `source` and `confidence` are required. `AUTHORED` means human-written or human-confirmed, though the API cannot prevent a caller from claiming it incorrectly.

text/0972-metadata-context.md
satyakigh: Added requiredness columns. All top-level Resource and Template Context fields are optional. When present, `trust` requires `source` and `confidence`, and an object-form `ref` requires `at`; CDK adds no other top-level requiredness.

text/0972-metadata-context.md
satyakigh: The published schema keeps `why` and `must` optional, so CDK does too. The RFC recommends `why` for significant resources and a `must` entry to explain constrained mutability, but does not enforce either. Confidence remains explicit whenever `trust` is supplied.

text/0972-metadata-context.md
satyakigh: Split the API as proposed: `ResourceMetadataContext.of(scope).add()` handles resource information, and `TemplateMetadataContext.of(stack).add()` handles template information. `MetadataContextMixin` remains resource-only.

text/0972-metadata-context.md
satyakigh: Added an early link to the published schema and documented the fields in the README and API reference. I did not add a notice to every `cdk synth`: it would be repetitive and unavailable to later `GetTemplate` readers. The stable metadata key and published documentation provide the durable reference.

text/0972-metadata-context.md
satyakigh: Removed `failureModes` because it is not in the published schema. Failure behavior that must be preserved belongs in `must`, design reasoning belongs in `why`, and external operational procedures can be linked through template-level `ref`.

text/0972-metadata-context.md
satyakigh: Added a dedicated explanation. Low-level metadata methods provide no typed fields, allowed-value checks, required trust checks, primary-resource selection, descendant selection, merge behavior, or generated documentation across CDK languages. The dedicated APIs provide those behaviors.

text/0972-metadata-context.md
satyakigh: Updated `ref` to match the published schema: references may use relative paths, `s3://`, or `https://`. CDK does not fetch or restrict those locations. Referenced content is untrusted, and inline template context remains authoritative.

text/0972-metadata-context.md
satyakigh: Removed `gaps`, which is not in the published schema. `trust` communicates provenance and confidence; it does not prevent drift. Keeping context beside reviewed CDK code helps readers assess it, but does not guarantee that it stays current.

text/0972-metadata-context.md
satyakigh: Defined `change-with-constraints` as allowing change only while the stated rules remain true. The RFC recommends documenting those rules in `must`, but the published schema and CDK do not require it.

text/0972-metadata-context.md
satyakigh: Renamed the public API properties to `defaultMutability` and `propertyMutability`, which state their scope directly. The generated template retains the documented `mutable` and `mutability` field names.

text/0972-metadata-context.md
satyakigh: Added a complete API-to-template mapping table. Most properties use the same name; the table lists the six differences, including the mutability, reference, and trust fields.

text/0972-metadata-context.md
satyakigh: Changed this to fail instead of overwrite. If directly written `com.aws.cloudformation.Context` metadata and a metadata-context API target the same location, template generation throws a scoped validation error. Resource, template, and Mixin paths are covered.

text/0972-metadata-context.md
satyakigh: Rewrote the definition and added the Lambda example: the generated `AWS::Lambda::Function` is selected, while its generated IAM role is not. The RFC also includes a filtered multi-resource construct example and explains explicit descendant selection.

text/0972-metadata-context.md
satyakigh: Removed the term. The RFC now states plainly that tools should reject comments that only repeat code or lack useful design information, and should record source and confidence for derived information.

text/0972-metadata-context.md
satyakigh: Removed the term. The RFC now directly lists information that should not be copied into Context because it already exists in the template, such as resource type, resource keys, property values, descriptions, and `aws:cdk:path`.

text/0972-metadata-context.md
satyakigh: Corrected the wording. JSON can be stored in a Description or tag, but it remains one difficult-to-query string, is constrained by length or count limits, and mixes design information with fields used for display, billing, or access policies.

text/0972-metadata-context.md
satyakigh: Documented the trade-off: `ref` entries can point to relative paths, `s3://`, or `https://` and reduce duplication or drift. They supplement rather than replace irreducible inline context, so an unavailable reference does not block safe use.

text/0972-metadata-context.md
satyakigh: Addressed this by making descendant selection explicit. The default selects only a `CfnResource` or a construct's `defaultChild` resource. `applyToDescendants`, `applyToAllResources`, and resource-type filters must be requested explicitly, and a declaration selecting zero resources is defined to fail.

text/0972-metadata-context.md
satyakigh: Added concrete stability requirements: review the public documentation, test the authoring and reading tools on real stacks, and obtain API Bar Raiser approval. A runtime feature flag is unnecessary because applications produce no additional Context unless they call the new APIs.

text/0972-metadata-context.md
satyakigh: Split the section into requirements before declaring the API stable and separate future enhancements. The future list now covers automatic dependency discovery, automatic change-safety guidance, helper-resource selection, and properties on higher-level constructs.

text/0972-metadata-context.md
satyakigh: Added the data and methodology: 33 tasks, three runs per condition. Scores were 65.20% without added information, 93.56% with template comments, 94.70% with structured metadata but no special instructions, and 98.63% with structured metadata plus instructions. The RFC also documents the scoring and task-selection limitations.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1022 @@
...
+Pair `must-never-change` or `change-with-constraints` with a `must` entry that states the
+rule behind the restriction, so a reader sees why a value is constrained; the schema does
+not require it. A `trust` block describes the source of other content, so it reads best
+alongside a `why` or `must`, but using it alone is valid. Individual declarations may omit
+`why` or `must` when another applicable declaration supplies them, and no template-level
+field is required either.
+
+##### Propagation is explicit
+
+`add()` targets only the scope's primary resource; it does not automatically apply the
```
dgandhi62, line 116: Can you expand on how this would work on an L3 construct? Or a construct which does not declare a defaultChild?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1022 @@
...
+
+`applyToDescendants` crosses a `NestedStack` boundary because a nested stack remains part of
+the same generated application. It does not cross a `Stage`, which is a separate CDK cloud
+assembly and must declare its own context.
+
+Applying information to descendants is always explicit. Repeating the same block on many
+resources can make that information appear more important than other facts and can place a
+rule on resources it does not govern. If information applies to the whole template, move it
+to `TemplateMetadataContext` instead. As a guideline, move information to template level
+when it would otherwise be repeated on more than about three resources.
```
dgandhi62, line 183: Something to consider if we are making this recommendation - templates already have a description property. Differentiating between the two might be more important.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1022 @@
...
+| Template field | API property | Type | Required | Meaning | Example |
+| -------------- | ------------ | ---- | -------- | ------- | ------- |
+| `src` | `source` | allowed value | yes, when `trust` is present | One of `authored`, `comment`, `commit`, or `infer`. | `"infer"` |
+| `conf` | `confidence` | allowed value | yes, when `trust` is present | One of `high`, `medium`, or `low`. | `"low"` |
+| `cite` | `citation` | text | no | Location of supporting evidence, such as a file and line, web address, or commit identifier. | `"service/handler.ts:87"` |
+| `note` | `note` | text | no | Additional explanation about the source or confidence. | `"no explicit design note"` |
+
+The four allowed sources are:
+
+* `authored` - a person wrote or explicitly confirmed the information.
```
dgandhi62, line 876: What if a human authored it AND any of the other three below are also true? Just something to consider

dgandhi62: I think it becomes more important to create a source for AI agents specifically, and less so for humans

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1022 @@
...
+
+```ts
+declare const deadLetterQueue: sqs.Queue;
+
+ResourceMetadataContext.of(deadLetterQueue).add({
+  why: 'stores failed order-processing invocations for later recovery',
+});
+```
+
+To include all helper resources, set `applyToAllResources: true`. This disables the
```
dgandhi62, line 200: This is not just including all helper resources, but ALL resources right? Is there a way to filter to only helper resources as this verbage suggests?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1022 @@
...
+// Only Amazon SQS queues among the descendant constructs.
+ResourceMetadataContext.of(stack).add({
+  why: 'buffers events for asynchronous processing',
+}, {
+  applyToDescendants: true,
+  includeResourceTypes: ['AWS::SQS::Queue'],
+});
+```
+
+For a multi-resource construct, `add()` with no options requires the construct's
```
dgandhi62, line 224: What if the defaultChild is another construct? Since propagation is explicit now, wouldn't this fail for that case?

text/0972-metadata-context.md
dgandhi62: I think what Eli meant here was that we are committed to the new schema once it is live. Why not have a bake period instead to test it out?

text/0972-metadata-context.md
satyakigh: Added an example

text/0972-metadata-context.md
satyakigh: Added a "Bake Period" section
The schema is owned by CloudFormation and is already published as v1 of the `Metadata Context` schema - https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema), and the `Agent Toolkit` authoring skill https://github.com/aws/agent-toolkit-for-aws/blob/main/skills/core-skills/aws-cloudformation/SKILL.md) already writes it, so a CDK preview period would not change the schema. The schema is also advisory - CloudFormation does not validate Metadata, so a future version cannot break a deployed template, and on the CDK side a new version means additive optional properties.

text/0972-metadata-context.md
satyakigh: Expanded that section. CDK follows defaultChild repeatedly until it reaches a CfnResource, so an L3 that designates one behaves like an L2 - e.g. cloudfront.experimental.EdgeFunction → lambda.Function → AWS::Lambda::Function. An L3 with no defaultChild, a grouping Construct, or a Stack has no primary resource, so `add()` with no options fails synth with an error naming the construct and the options: target a child directly, applyToDescendants (usually with a type filter), or applyToAllResources. L3 authors can opt in by setting `node.defaultChild`

text/0972-metadata-context.md
satyakigh: Added a comparison table in the Template-level context section and a note where the template-level guideline is introduced. `Description` is one string (at most 1,024 bytes) shown in the console stack list and returned by `DescribeStacks`; it answers what the stack is. `TemplateMetadataContext` holds named fields (`arch`, `must`, `ref`, `owner`) returned only in the template body via `GetTemplate`; it answers how the system is shaped and which rules apply. Guidance: avoid repeating `Description` in `arch`, and keep rules and references out of `Description`. This matches the schema doc, which also says to use `Description` for the stack purpose - https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema

text/0972-metadata-context.md
satyakigh: Added a precedence rule to Appendix A. `authored` applies whenever a person wrote or confirmed the text, even if it came from a comment, commit, or inference; the evidence goes in `cite` and `note`. Otherwise use the most direct evidence: `comment`, then `commit`. `infer` applies when the tool reasoned or combined evidence. A worked example is included.

Agreed on the second point. Three of the four values exist for automated producers, and the RFC now says so; a person writing Context directly in CDK code can omit `trust`. The enum is fixed by the published v1 schema (https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema), so a dedicated `agent` value would be a schema change. `infer` already plays that role, and the authoring skill (https://github.com/aws/agent-toolkit-for-aws/blob/main/skills/core-skills/aws-cloudformation/SKILL.md) is where agent guidance is tuned.

text/0972-metadata-context.md
satyakigh: Yup, `applyToAllResources` selects every `CfnResource` under the scope. The "all helper resources" wording was misleading and is fixed in the README, the design section, the future-enhancements bullet, and the docstring. There is no helper-only selector because CDK has no marker for a helper other than not being on the `defaultChild` chain. Helpers are reached by type, for example `applyToAllResources` with `includeResourceTypes: ['AWS::IAM::Role']`, or by targeting an exposed helper construct directly. Relationship-based selection stays a future enhancement.

text/0972-metadata-context.md
satyakigh: The wording was unclear and is rewritten. The implementation walks the chain parent by parent and requires each parent to designate the current node as its `defaultChild`, so L3 -> L2 -> L1 selects the L1 and skips the L2's helpers. `cloudfront.experimental.EdgeFunction` is the in-repo example. The chain fails only if it reaches a construct with no `defaultChild`; an ambiguous one (both `Resource` and `Default` children) counts as none.

atennak1 (approved):

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1156 @@
...
+
+* An L3 that designates one, as `EdgeFunction` does, behaves like an L2: the declaration
+  lands on the `CfnResource` at the end of the chain.
+* An L3 that does not, such as `ecs_patterns.ApplicationLoadBalancedFargateService`, a plain
+  grouping `Construct`, or a `Stack`, has no primary resource. `add()` with no options then
+  selects nothing, and template generation fails with an error that names the construct and
+  lists the alternatives: target a child construct directly, set `applyToDescendants: true`
+  (usually with a resource-type filter), or set `applyToAllResources: true`. The chain also
+  ends without a match when it reaches a construct that has no `defaultChild`, or one whose
+  `defaultChild` is ambiguous because it has both a `Resource` and a `Default` child.
```
iliapolo, line 142: This is not possible, it will produce a synth time error:

```console
throw new Error(`Cannot determine default child for ${this.path}. There is both a child with id "Resource" and id "Default"`);
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1156 @@
...
+When a helper resource is available as a construct, target it directly instead of applying
+context to every descendant. For example, a function's dead-letter queue can record why it
+exists:
+
+```ts
+declare const deadLetterQueue: sqs.Queue;
+
+ResourceMetadataContext.of(deadLetterQueue).add({
+  why: 'stores failed order-processing invocations for later recovery',
+});
```
iliapolo, line 235: To make it extra clear that helper resources can be accessed via the L2. 

```suggestion
ResourceMetadataContext.of(lambdaFunction.deadLetterQueue).add({
  why: 'stores failed order-processing invocations for later recovery',
});
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1156 @@
...
+Whether the default works for a higher-level (L3) construct depends on whether that
+construct declares a `defaultChild`:
+
+* An L3 that designates one, as `EdgeFunction` does, behaves like an L2: the declaration
+  lands on the `CfnResource` at the end of the chain.
+* An L3 that does not, such as `ecs_patterns.ApplicationLoadBalancedFargateService`, a plain
+  grouping `Construct`, or a `Stack`, has no primary resource. `add()` with no options then
+  selects nothing, and template generation fails with an error that names the construct and
+  lists the alternatives: target a child construct directly, set `applyToDescendants: true`
+  (usually with a resource-type filter), or set `applyToAllResources: true`. The chain also
```
iliapolo, line 140: > usually with a resource-type filter

Why usually and not always? What would setting `applyToDescendants: true` mean without providing a resource type filter?

We have 3 different properties that all relate to the same filtering capability, I think we can get away with just two and give different names. Also the current API doesn't lend itself well to future filters we might want to introduce.

*Suggestion:*

```ts
// 1. default - apply only to primary resource
ResourceMetadataContext.of(stack).add({
  must: ['delivery settings must preserve in-flight messages'],
});

// 2. propagate to ALL child resources
ResourceMetadataContext.of(stack).add({
  must: ['delivery settings must preserve in-flight messages'],
}, {
  propagate: true
});

// 3. propagate only to child resources of a specific type
ResourceMetadataContext.of(stack).add({
  must: ['delivery settings must preserve in-flight messages'],
}, {
  propagate: true,
  propagationFilter: PropagationFilter.includeResourceTypes(['AWS::SQS::Queue'])
});
```

We could then introduce a new filter type, for example:

```ts
// 4. propagate only to child resources NOT of a specific type
ResourceMetadataContext.of(stack).add({
  must: ['delivery settings must preserve in-flight messages'],
}, {
  propagate: true,
  propagationFilter: PropagationFilter.excludeResourceTypes(['AWS::SQS::Queue'])
});
```

text/0972-metadata-context.md
iliapolo: I was actually hoping for a change in the schema itself. Given the schema was already published with `mutable` and `mutability` I think we should remain consistent with it and have CDK properties map 1:1 with them.

text/0972-metadata-context.md
iliapolo: I see the table. There's no need for the naming differences between (`src` and `source` are the same...) - lets just stick to having CFN's schema names.

text/0972-metadata-context.md
dgandhi62: I see it. I think this information should be an instruction to authors and AI agents both, and should be part of the README (if not also the skill)

iliapolo (changes requested):
Also please provide the full public API of all classes we are introducing (in TypeScript form). Right now its kind of scattered between examples and hard to get a full picture. 

We can then approve that API and merge it as stable from the get go.

text/0972-metadata-context.md
dgandhi62: One idea - if we use the following API, we could provide an example to indirectly target all helper resources only by excluding the defaultChild resource.

```ts
// 4. propagate only to child resources NOT of a specific type
ResourceMetadataContext.of(stack).add({
  must: ['delivery settings must preserve in-flight messages'],
}, {
  propagate: true,
  propagationFilter: PropagationFilter.excludeResourceTypes(['AWS::SQS::Queue'])
});
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+The metadata-context APIs add structured design information that CloudFormation stores but
+does not enforce under the
+`com.aws.cloudformation.Context` key in generated CloudFormation templates. The information
+can include reasoning, hard rules, change-safety guidance, and source and confidence.
+People and automated tools that inspect a deployed template can
+therefore use the author's intent instead of guessing it.
+
+Two classes write the same documented template fields: `ResourceMetadataContext` writes
+information on individual resources, and `TemplateMetadataContext` writes information once
+for the whole template. `MetadataContextMixin` is a CDK Mixin, which is an API applied
```
iliapolo: For consistency with the aspect classes - it also makes sense because it can only be applied to resources. 

```suggestion
for the whole template. `ResourceMetadataContextMixin` is a CDK Mixin, which is an API applied
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+#### Metadata Context
+
+The metadata-context APIs add structured design information that CloudFormation stores but
+does not enforce under the
+`com.aws.cloudformation.Context` key in generated CloudFormation templates. The information
+can include reasoning, hard rules, change-safety guidance, and source and confidence.
+People and automated tools that inspect a deployed template can
+therefore use the author's intent instead of guessing it.
+
+Two classes write the same documented template fields: `ResourceMetadataContext` writes
```
iliapolo:
```suggestion
Two aspects write the same documented template fields: `ResourceMetadataContext` writes
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+}
+```
+
+No `trust` block appears because the caller did not provide one. The `trust` field is
+optional, and CDK never adds it automatically (see *Source and confidence* below).
+`mutable` is `change-with-constraints`, and the governing rule is recorded in
+`must`: `VisibilityTimeout` must remain at least six times the Lambda timeout.
+`change-with-constraints` means a value may change only while its stated rules remain true;
+using that value without a corresponding `must` rule gives the reader no useful guidance.
+
```
iliapolo: No need for this entire paragraph - it should all be mentioned in the schema definition we link to already. Unless you want to emphasize a specific property - but then also give a short explanation about it. I think we can just remove it.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+merely to populate the field.
+
+Pair `must-never-change` or `change-with-constraints` with a `must` entry that states the
+rule behind the restriction, so a reader sees why a value is constrained; the schema does
+not require it. A `trust` block describes the source of other content, so it reads best
+alongside a `why` or `must`, but using it alone is valid. Individual declarations may omit
+`why` or `must` when another applicable declaration supplies them, and no template-level
+field is required either.
+
+##### Propagation is explicit
```
iliapolo, line 87: This should come sooner - just after you show the example of `add`, that way the reader still has this API in their mind.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+```
+
+No `trust` block appears because the caller did not provide one. The `trust` field is
+optional, and CDK never adds it automatically (see *Source and confidence* below).
+`mutable` is `change-with-constraints`, and the governing rule is recorded in
+`must`: `VisibilityTimeout` must remain at least six times the Lambda timeout.
+`change-with-constraints` means a value may change only while its stated rules remain true;
+using that value without a corresponding `must` rule gives the reader no useful guidance.
+
+##### Resource context quality
```
iliapolo: Should go last after all other field specific explanations. Also rename to `Writing good context`

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+rule behind the restriction, so a reader sees why a value is constrained; the schema does
+not require it. A `trust` block describes the source of other content, so it reads best
+alongside a `why` or `must`, but using it alone is valid. Individual declarations may omit
+`why` or `must` when another applicable declaration supplies them, and no template-level
+field is required either.
+
+##### Propagation is explicit
+
+`add()` targets only the scope's primary resource; it does not automatically apply the
+information to descendant constructs. A declaration must match at least one resource, or
```
iliapolo:
```suggestion
information to auxillary constructs. A declaration must match at least one resource, or
```

Because a primary resource is also a descendant.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+for the whole template. `MetadataContextMixin` is a CDK Mixin, which is an API applied
+directly to selected low-level `CfnResource` objects. API property names are the field names
+of the published
+[CloudFormation Metadata Context schema](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema),
+so code and template use one vocabulary.
+
+Add resource-level information to a construct scope with `ResourceMetadataContext`. A scope
+is a node in the CDK construct hierarchy. By default, the information is written to the
+scope's *primary resource*: the CloudFormation resource reached by following CDK's
+`defaultChild` property. For example, the primary resource of an Amazon Simple Queue
```
iliapolo:
```suggestion
[`defaultChild`](https://docs.aws.amazon.com/cdk/api/v2/docs/constructs.Node.html#defaultchild) property. For example, the primary resource of an Amazon Simple Queue
```

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+
+* An L3 that designates one, as `EdgeFunction` does, behaves like an L2: the declaration
+  lands on the `CfnResource` at the end of the chain.
+* An L3 that does not, such as `ecs_patterns.ApplicationLoadBalancedFargateService`, a plain
+  grouping `Construct`, or a `Stack`, has no primary resource. `add()` with no options then
+  selects nothing, and template generation fails with an error that names the construct and
+  lists the alternatives: target a child construct, or set `propagate: true` (optionally
+  with a `propagationFilter`). The chain also ends without a match when it
+  reaches a construct that has no `defaultChild`.
+* A construct with both a `Resource` and a `Default` child has an ambiguous `defaultChild`.
```
iliapolo: A situation like this cannot exist so this becomes guidance on how to write L3s, not how to apply context. Remove it.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+construct declares a `defaultChild`:
+
+* An L3 that designates one, as `EdgeFunction` does, behaves like an L2: the declaration
+  lands on the `CfnResource` at the end of the chain.
+* An L3 that does not, such as `ecs_patterns.ApplicationLoadBalancedFargateService`, a plain
+  grouping `Construct`, or a `Stack`, has no primary resource. `add()` with no options then
+  selects nothing, and template generation fails with an error that names the construct and
+  lists the alternatives: target a child construct, or set `propagate: true` (optionally
+  with a `propagationFilter`). The chain also ends without a match when it
+  reaches a construct that has no `defaultChild`.
```
iliapolo: Lots of words that essentially boil down to "applying context to a scope without a default child will fail unless explicit filtering options are supplied (see below)"

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+}, {
+  propagate: true,
+});
+
+// 3. Propagate only to resources of a specific type. The queue above also receives
+//    this declaration.
+ResourceMetadataContext.of(stack).add({
+  must: ['delivery settings must preserve in-flight messages'],
+}, {
+  propagate: true,
```
iliapolo, line 132: Make it clear that `propagate: true` voids the `defaultChild` selection and only considers the provided filter.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+
+A `propagationFilter` requires `propagate: true`; `add()` throws otherwise, because default
+targeting already selects exactly one resource. A filter that excludes every candidate fails
+template generation like any other declaration that matches nothing.
+
+When several declarations apply to one resource, CDK combines them. For fields that hold
+one value (`why`, `mutable`, and `trust`), the declaration closest to the
+resource takes precedence. For array fields (`must` and `deps`),
+CDK combines the entries and removes duplicates. For `mutability`, CDK combines the
+maps and uses the closest declaration for each property name.
```
iliapolo: Give an example

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+
+Propagation crosses a `NestedStack` boundary because a nested stack remains part of
+the same generated application. It does not cross a `Stage`, which is a separate CDK cloud
```
iliapolo, line 189: So if you apply context on an `App` that only contains a few stages (no direct other constructs) then nothing happens? I think this merits an example of applying similar context to different stages (think dev/prod that will have different design considerations)

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+assembly and must declare its own context.
+
+Propagation is always explicit. Repeating the same block on many
+resources can make that information appear more important than other facts and can place a
+rule on resources it does not govern. If information applies to the whole template, move it
+to `TemplateMetadataContext` instead. As a guideline, move information to template level
+when it would otherwise be repeated on more than about three resources.
```
iliapolo: > on more than about three resources.

Seems completely random. Lets just say that context that applies to ALL resources in the template belongs in the template metadata.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+
+Template level here means `TemplateMetadataContext`, not the template's built-in
+`Description`. The two serve different readers: `Description` is one short, unstructured
+string (at most 1,024 bytes) that CloudFormation shows in the console stack list and
+returns from `DescribeStacks`, so it works best as a one-line statement of what the stack is.
+`TemplateMetadataContext` holds the structured fields `arch`, `must`, `ref`, and `owner`,
+which are returned only inside the template body (`GetTemplate`) and answer how the system
+is shaped, which rules apply everywhere, and where supporting material lives. Avoid repeating
+the `Description` text in `arch`, and keep rules out of `Description`. See *Template-level
+context* below for a side-by-side comparison.
```
iliapolo: Too much information about a side concept. Remove it (instead just link to the relevant section in the paragraph above).

text/0972-metadata-context.md
iliapolo: Add "See *Template-level context* below for more information"

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+   the strongest evidence; record the original evidence in `cite` (the comment's file and
+   line, or the commit identifier) and, when useful, in `note`.
+2. Otherwise, the most direct evidence: `COMMENT` when the text was copied or lightly
+   rephrased from a source comment; `COMMIT` when it came from version-control history.
+3. `INFER` when the tool combined evidence or reasoned from code structure or behavior
+   without an explicit statement, even if a comment or commit contributed. Name the
+   contributing evidence in `cite` and `note`.
+
+For example, a tool that lifts `why` from a comment writes `src: COMMENT` and
+`cite: 'lib/queue.ts:42'`; when the author reviews and accepts it, `src` becomes `AUTHORED`
```
iliapolo: You mean "`src` **should** become `AUTHORED`"? As in, the PR author should make a change after getting approval?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+and `cite` stays. Three of the four values exist for automated producers, where `src` matters
+most: a reader must be able to tell tool-derived Context from Context a person stands behind.
+A person writing Context directly in CDK code can omit `trust`, because the reviewed source
+already shows who wrote it. The Agent Toolkit's CloudFormation and CDK skills will be
```
iliapolo: > already shows who wrote it

What do you mean? how? I thought a person writing context should explicitly use `AUTHORED`...

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+updated to carry the same rule (see *Follow-ups*). See
+[Appendix A](#appendix-a---cloudformation-context-template-field-reference) for the
+`trust` object.
+
+##### Mixin form
```
iliapolo: Why do we actually need the mixin here? what does it give that the aspect doesn't? 

Im also seeing the mixin API doesn't have all the capabilities, for example `inheritAncestorContext`, is that intentional? If we don't really need the mixin, lets not have it. If we are keeping it - we should also mention how conflicts between the aspect and the mixin are handled.

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1265 @@
...
+Mixins.of(stack).apply(new MetadataContextMixin({
+  deps: ['NetworkStack'],
+}));
+```
+
+##### Conflict with manually added context
+
+`com.aws.cloudformation.Context` is a normal metadata key, so callers can also write it
+directly with `CfnResource.addMetadata()`. If manually added information and a metadata-
+context API target the same resource and key, template generation fails instead of silently
```
iliapolo:
```suggestion
directly with `CfnResource.addMetadata()`. If manually added information and a metadata-context API target the same resource and key, template generation fails instead of silently
```

iliapolo (changes requested):

text/0972-metadata-context.md
satyakigh: The Mixin is removed entirely (see the thread on the *Mixin form* section below)

text/0972-metadata-context.md
satyakigh: Changed to "Two APIs". I avoided "aspects" because `TemplateMetadataContext` isn't aspect-backed: it writes the stack's template `Metadata` directly when `add()` is called. Only `ResourceMetadataContext` registers an aspect.

text/0972-metadata-context.md
satyakigh: Removed

text/0972-metadata-context.md
satyakigh: Moved. *Propagation is explicit* now follows the first `add()` example directly, and the quality guidance moved to the end of the README as *Writing good context*.

text/0972-metadata-context.md
satyakigh: Done. It is now the last section, titled *Writing good context*, and it also absorbed the conciseness and size paragraphs that used to trail *Template-level context*.

text/0972-metadata-context.md
satyakigh: done

text/0972-metadata-context.md
satyakigh: Added

text/0972-metadata-context.md
satyakigh: Removed, together with the following paragraph about L3 authors setting `defaultChild`, which was construct-authoring guidance by the same logic

text/0972-metadata-context.md
satyakigh: Collapsed to one sentence: applying context to a scope with no `defaultChild` (most L3 patterns, a plain `Construct`, or a `Stack`) fails unless `propagate` is set.

text/0972-metadata-context.md
satyakigh: Added `Propagation replaces `defaultChild` selection entirely: the declaration applies to every resource beneath the scope, helpers included, and only a `PropagationFilter` narrows it.`

text/0972-metadata-context.md
satyakigh: Added a Stack-level plus queue-level example with the merged JSON: the closer `why` wins and the `must` entries combine, Stack entry first

text/0972-metadata-context.md
satyakigh: It fails rather than doing nothing: a declaration on an `App` whose only children are Stages matches no resource, and synthesis errors with "declare context inside each Stage". This mirrors aspects, which never cross assembly boundaries. The text now says so, and added the dev/prod example

text/0972-metadata-context.md
satyakigh: Reworded: information that applies to every resource in the template belongs in `TemplateMetadataContext`, followed by the pointer to *Template-level context* below. I fixed the same heuristic in the design section and Appendix A.

text/0972-metadata-context.md
satyakigh: Removed. The paragraph above now links to *Template-level context*

text/0972-metadata-context.md
satyakigh: Yes. It now reads: "the author (or the tool, on the author's confirmation) should change `src` to `AUTHORED` and keep `cite`"

text/0972-metadata-context.md
satyakigh: That sentence conflated source history with what a template reader can see. It now reads: a person writing Context directly uses `src: AUTHORED`; a tool uses `COMMENT`, `COMMIT`, or `INFER` according to its evidence and switches to `AUTHORED` only after a person confirms the text. `trust` stays optional per the schema, so a block without it leaves the source unstated, and the text recommends setting it wherever tool-derived and human-written Context may share a template.

text/0972-metadata-context.md
satyakigh: Agreed, and removed. It was a strict subset of the aspect: `.with()` on an L1 is `ResourceMetadataContext.of(cfnResource).add()`, and `Mixins.of(scope).apply()` is `add(..., { propagate: true })`, except that it applied eagerly, so resources created after the call were missed. It also needed an awslint `mixin-namespace` exemption. The Public API is now three classes

text/0972-metadata-context.md
satyakigh: Applied, rewrapped so `metadata-context` stays on one line.

text/0972-metadata-context.md
iliapolo: Thinking about this again - isn't every change to the template eventually human reviewed? and is therefore `AUTHORED`? when will my deployed template change without human review?

text/0972-metadata-context.md
```diff
@@ -0,0 +1,1241 @@
+# Structured Design Context in Synthesized Templates
+
+* **Original Author(s):** @satyakigh
+* **Tracking Issue**: #972
+* **API Bar Raiser**: TBD
```
iliapolo:
```suggestion
* **API Bar Raiser**: @iliapolo 
```

iliapolo:
@satyakigh Im happy with the API - this can now move onto final comments period. Can you please create a tracking issue similar to https://github.com/aws/aws-cdk-rfcs/issues/949 where we can apply the correct labels? See https://github.com/aws/aws-cdk-rfcs#1-tracking-issue for more details. 

Thanks!

satyakigh:
Updated RFC to the correct number, created https://github.com/aws/aws-cdk-rfcs/issues/994

